PR #1, opened 2015-01-25
Update gravity-adv.sh

dwin999:
This works much faster.


jacobsalmela:
Could you explain--in detail--how your code works?  It is definitely faster, but I want to understand how it works.  Explain what each piece of code does.  Thanks

dwin999:
catting a file to stdin is much faster than looping over a file in bash. 

cat /tmp/matter.txt | sort | uniq | sed '/^$/d' |  I just copied your code.
awk -v "IP=$piholeIP" '{sub(/\r$/,""); print "address=/"$0"/"IP}' just formats the code to the dnsmasq format, removing any \r using awk regex.

jacobsalmela:
I'm just not too familiar with all of the awk stuff.  I have really only used the most basic to print a column.

saltman007web (commented):
commit

saltman007web (commented):
2

PR #5, opened 2015-03-01
Change how script pulls in whitelist.txt

iamdanielv:
I noticed that when I ran the script as sudo the whitelist.txt file would not be found.
This change will use the home directory of the sudoer.
Hope it can help others who use sudo to run the script as well.

### DIFF
--- a/gravity-adv.sh
+++ b/gravity-adv.sh
@@ -7,7 +7,8 @@ piholeIP="127.0.0.1"
 
 # Config file to hold URL rules
 eventHorizion="/etc/dnsmasq.d/adList.conf"
-whitelist=$HOME/whitelist.txt
+USER_HOME=$(getent passwd $SUDO_USER | cut -d: -f6)
+whitelist=$USER_HOME/whitelist.txt
 
 echo "Getting yoyo ad list..." # Approximately 2452 domains at the time of writing
 curl -s -d mimetype=plaintext -d hostformat=unixhosts http://pgl.yoyo.org/adservers/serverlist.php? | sort > /tmp/matter.txt


jacobsalmela:
Thanks.  I'll have to test this out and see how it works.  

The script seems to run fine without sudo; is there any specific reason for doing so?

ajp2k13:
Without sudo I get a permission error when it tries to overwrite adList.conf.... (sorry about the Swedish error message but I think you understand)

pi@raspberrypi ~ $ /usr/local/bin/gravity.sh
Getting yoyo ad list...
Getting winhelp2002 ad list...
Getting adaway ad list...
Getting hosts-file ad list...
Getting malwaredomainlist ad list...
Getting adblock.gjtech ad list...
Getting someone who cares ad list...
Getting Mother of All Ad Blocks list...
100 3989k  100 3989k    0     0  1329k      0  0:00:03  0:00:03 --:--:-- 1400k
Removing duplicates, whitelisting, and formatting the list of domains...
2 domains whitelisted.
124734 ad domains blocked.
mv: försök skriva över ”/etc/dnsmasq.d/adList.conf”, åsidosätt rättigheterna 0644 (rw-r--r--)?

ajp2k13:
I manually added the pull request from iamdanielv to gravity.sh and now it works fine when running as sudo. Won't work at all if I run it without sudo, then I get access errors and failure to restart dnsmasq etc.

iamdanielv:
@ajp2k13 Great! I'm glad it's working for you.

I also had some issues running it as a normal user. Not sure what is different between our setup and @jacobsalmela 

The Pi may be small enough to be single user, but it still seems safer to run it through sudo. I guess it wouldn't hurt anything if it was left as-is either.

ajp2k13:
@iamdanielv, thanks for the commit! I thought I was doing something wrong before...
 I'm no Linux expert but I didn't think Raspian had a root user, not as default anyway...

jacobsalmela:
Thanks for the comments.  I'll try to test and verify soon before merging the request.  

Looking at the original code, I do wonder how I got it to work without sudo as the /etc directory usually needs root access.  Maybe I am crazy.

This is what's cool about open source though--it can make the project much better.

iamdanielv:
@jacobsalmela  Thanks for sharing the original!

ajp2k13:
I second that, thanks for sharing @jacobsalmela! I had something like it working directly on my router but this is more powerful and flexible. :)

jacobsalmela:
I'm wondering if it wouldn't just be better/easier to store the whitelist.txt somewhere else, like in /etc/dnsmasq.d/ but I'm not sure if having txt files in that folder will cause issues.  But it makes sense to have it in the same place as the ad list.  

This would also eliminate having to adjust the code to determine the current user.  What do you think?

iamdanielv:
According to the README:
# All files in this directory will be read by dnsmasq as
# configuration files, except if their names end in
# ".dpkg-dist",".dpkg-old" or ".dpkg-new"
#
# This can be changed by editing /etc/default/dnsmasq

So probably can't put the whitelist file in /etc/dnsmasq.d

iamdanielv:
how about something like /etc/pi-hole/hosts.allow ?

jacobsalmela:
Good find on the README--thanks!  

/etc/pihole/ is probably a good idea.  As the script improves, it will probably need a home for different files.  So having a consistent place for them would make sense.

ajp2k13:
Good thinking, sounds like a good idea not having parts of the script be user dependent.

jacobsalmela:
I opted to create /etc/pihole and users can put there whitelist there fixed in [this commit](https://github.com/jacobsalmela/pi-hole/commit/a8874fb60606a47737680b46b8c39f14b9ac8aa4).  This avoids the sudo issues and makes more sense for the files to be in their own folder.

PR #8, opened 2015-03-12
Whitelist and Blacklist improved. Reduce size of dnsmasq conf file, same blocking power.

pizzafritta:
Whitelist fix: only first element of file was actually removed from output file

blacklist added: just testing the opportunity to add a local blacklist file 

reverse domain sort: domains are sorted from right to left, 1st TLD, 2nd level domain and so on

### DIFF
--- a/gravity-adv.sh
+++ b/gravity-adv.sh
@@ -18,7 +18,7 @@ if [[ -d $piholeDir ]];then
 	:
 else
 	echo "Forming pihole directory..."
-	sudo mkdir piholeDir
+	sudo mkdir $piholeDir
 fi
 
 tmpDir='/tmp/'    
--- a/gravity-adv.sh
+++ b/gravity-adv.sh
@@ -8,6 +8,7 @@ piholeIP="127.0.0.1"
 # Config file to hold URL rules
 eventHorizion="/etc/dnsmasq.d/adList.conf"
 whitelist=/etc/pihole/whitelist.txt
+blacklist=/etc/pihole/blacklist.txt
 
 # Create the pihole resource directory if it doesn't exist.  Future files will be stored here
 if [[ -d /etc/pihole/ ]];then
@@ -38,7 +39,7 @@ curl -A 'Mozilla/5.0 (X11; Linux x86_64; rv:30.0) Gecko/20100101 Firefox/30.0' -
 # Remove entries from the whitelist file if it exists at the root of the current user's home folder
 if [[ -f $whitelist ]];then
 	echo "Removing duplicates, whitelisting, and formatting the list of domains..."
-	cat /tmp/matter.txt | sed $'s/\r$//' | sort | uniq | sed '/^$/d' | grep -v -x -f $whitelist | awk -v "IP=$piholeIP" '{sub(/\r$/,""); print "address=/"$0"/"IP}' > /tmp/andLight.txt
+	cat /tmp/matter.txt $blacklist| sed $'s/\r$//' | awk -F "." '{for(i=NF; i > 1; i--) printf "%s.", $i; print $1}' | sort | awk -F "." '{for(i=NF; i > 1; i--) printf "%s.", $i; print $1}' | uniq | sed '/^$/d' | grep -vwf $whitelist | awk -v "IP=$piholeIP" '{sub(/\r$/,""); print "address=/"$0"/"IP}' > /tmp/andLight.txt
 	numberOfSitesWhitelisted=$(cat $whitelist | wc -l | sed 's/^[ \t]*//')
 	echo "$numberOfSitesWhitelisted domains whitelisted."
 else
--- a/gravity-adv.sh
+++ b/gravity-adv.sh
@@ -1,58 +1,79 @@
 #!/bin/bash
-# The Pi-hole now blocks over 120,000 ad domains
+# The Pi-hole now blocks over 90,000 ad domains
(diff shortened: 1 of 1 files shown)


jacobsalmela:
Why do you want a blacklist?  

Also, what happened when it only removed one entry via the whitelist?  It might not remove a domain if it doesn't exist in the list in the first place.

pizzafritta:
why a blacklist: lets say you want to block an ad (or anything else) that is not listed in any of the online sources. Sorry I didn't check if the file exists first.

Regarding the whitelist, this is the content of my whitelist.txt:
```
localhost
localhost.localdomain
broadcasthost
android
test
test.localhost
local
android.localhost
```
all of them exist in matter.txt but they were not removed from the output file, only the 1st one.

jacobsalmela:
Could you send me the output of the script in debug mode by running

```bash -x gravity.sh```

Every time I test it, the whitelist works for multiple entries.  I did, however notice all those localhost entries and will be removing those within the script on my next commit.

**Comment by pizzafritta** [comment]
you made me notice there is something wrong in both cases:
- using 'grep -vxf $whitelist' only first entry of whitelist is removed from output
- using 'grep -vwf $whitelist' too many entries are removed, i.e. if 'android' is in whitelist, abc.android.com android.abc.com and abc.android.def.com are removed, abcandroiddef.com is not removed. 
I am testing this on a mac.

What I really want to do is something smarter: if domainname.com is in whitelist, also whatever.domainname.com should be whitelisted automatically, what do you think?

If I manage to do this, next step will be cleaning up the output file from redundant entries i.e. if adserver.net is already blocked is not necessary to have abc.adserver.net def.adserver.net etc.

**Comment by pizzafritta** [comment]
I've found a workaround using this as whitelist
```
localhost$
localdomain$
broadcasthost$
android$
test$
local$
```
This removes all the subdomains of localhost, localdomain etc
I am not an expert of grep and regex in general, do you know if I can put the $ sign in the regular expression in a nice way, instead of putting it at the end of every line in whitelist.txt?

**Comment by jacobsalmela** [comment]
I have always struggled with regex, but I think what might be better is to make a default whitelist.txt because all of those localhost entries are in the list and shouldn't be in case someone uses those.  

I still can't figure out why it only removes the first whitelist entry for you.  It seems to work for me.

**Comment by pizzafritta** [comment]
I just did some changes to solve the following:
- whitelist and blacklist bug fix
- subdomains of items in whitelist are whitelisted too
- subdomains of items in blacklist are blacklisted too
- comments (lines starting with #) and blank lines are allowed in whitelist and in blacklist
- definition of variables at the beginning of the code
- smarter algorithm that reduces the number of entries in dnsmaq configuration file (i.e. if adsite.com is blocked there is no need to block www.adsite.com ad1.adsite.com whatever.adsite.com etc.) this reduce the number of entries in the output from ~ 140k down to ~ 92k.
- domain list is sorted in reverse (right to left) order: TLD first, then domain name and subdomains at the end.

**Comment by jacobsalmela** [comment]
This looks impressive.  Could you change the piholeIP back to 127.0.0.1?  

Also, I see you have the yin and yang temporary white/blacklists, but never use the original ones that were already setup.

**Comment by jacobsalmela** [comment]
Also, when I ran your additions, the script still came out at around 141 thousand entries and the adList.conf was not sorted.

Here is the head of my file after running your additions:

```
address=/js.cdn.ac/127.0.0.1
address=/link.ac/127.0.0.1
address=/www.link.ac/127.0.0.1
address=/content.ad/127.0.0.1
address=/api.content.ad/127.0.0.1
address=/ads.7days.ae/127.0.0.1
address=/instagramfollowers.ae/127.0.0.1
address=/www.instagramfollowers.ae/127.0.0.1
address=/ads.emirates.net.ae/127.0.0.1
address=/aviationmedia.aero/127.0.0.1
```

**Comment by pizzafritta** [comment]
There must be something different running the code on a mac or in linux. It works fine on my mac, I get 97k lines of output, 146k testing it on the raspberrypi.

**Comment by jacobsalmela** [comment]
Interesting.  I get the same results on my Mac.  It must have to do with different versions of some of the commands, probably grep or awk.

For now, I am going to merge [this request](https://github.com/jacobsalmela/pi-hole/pull/9/files), which just adds a blacklist file.  I prefer smaller changes as opposed to many at once.  

Since your code isn't quite working consistently, it still needs some work.  You may need to rebase your pull request once I merge the other one.

**Comment by pizzafritta** [comment]
I've found the bug. Raspbian uses mawk instead of POSIX awk. It happens that the function substr (line 65) behaves in a different way (string starting index is 1 instead of 0). Testing the solution before commit.

**Comment by jacobsalmela** [comment]
Interesting.  So I wonder how many domains are actually real: 97k or 140k or 120k?  I often use ``opendiff`` to find differences, but considering the files are so large, this could be a difficult task.

**Comment by pizzafritta** [comment]
I've tested it using kdiff, if the outputs are sorted with same algorithm it's easy to spot the huge difference it makes. 
Put these bad guys in your blacklist and you get less then 90k domains
```
yupiromo.ru #3825
112.2o7.net #2315
bjk321.com #1895
intellitxt.com #1881
hwqdjzbw.eu #1565
us.intellitxt.com #1521
genericrxassist.be #1361
xn--b1aebj7c.xn--p1ai #1214
51yes.com #1064
hitbox.com #999
hpg.com.br #879
```

PR #9, opened 2015-03-19
Add a local blacklist file

mospaeda:
Add a local blacklist for saving user local blacklist 
Helpful if some sites are not in the ones downloaded from the websites

Local llacklist has less priority than whitelist

### DIFF
--- a/gravity-adv.sh
+++ b/gravity-adv.sh
@@ -2,13 +2,8 @@
 # The Pi-hole now blocks over 120,000 ad domains
 # Address to send ads to (the RPi)
 piholeIP="127.0.0.1"
-# Optionally, uncomment to automatically detect the address.  Thanks Gregg
-#piholeIP=$(ifconfig eth0 | awk '/inet addr/{print substr($2,6)}')
-# Below code allows to automatically detect the address with all locale
-langsys=$LANG
-LANG=en_US.utf8
-piholeIP=$(ifconfig eth0 | awk '/inet addr/{print substr($2,6)}')
-LANG=$langsys
+# Optionally, uncomment to automatically detect the local IP address.
+# piholeIP=$(hostname -I)
 
 # Config file to hold URL rules
 eventHorizion="/etc/dnsmasq.d/adList.conf"
--- a/gravity-adv.sh
+++ b/gravity-adv.sh
@@ -7,6 +7,10 @@ piholeIP="127.0.0.1"
 
 # Config file to hold URL rules
 eventHorizion="/etc/dnsmasq.d/adList.conf"
+
+
+
+
 whitelist=/etc/pihole/whitelist.txt
 
 # Create the pihole resource directory if it doesn't exist.  Future files will be stored here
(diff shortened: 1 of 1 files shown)


jacobsalmela:
I like your implementation.  However, will you please replace the hardcoded path within the ```if``` block to be the variable ```$blacklist```?  

I also think you could just put a colon in the else block since it may confuse the user if it notifies them that the file is not there (the file isn't required for it to run properly).

mospaeda:
Done and tested

pi@raspberrypi /usr/local/bin $ sudo ./gravity.sh
Getting yoyo ad list...
Getting winhelp2002 ad list...
Getting adaway ad list...
Getting hosts-file ad list...
Getting malwaredomainlist ad list...
Getting adblock.gjtech ad list...
Getting someone who cares ad list...
Getting Mother of All Ad Blocks list...
100 3989k  100 3989k    0     0   911k      0  0:00:04  0:00:04 --:--:--  962k
Getting the local blacklist from /etc/pihole directory
Removing duplicates and formatting the list of domains...
104836 ad domains blocked.
[ ok ] Restarting DNS forwarder and DHCP server: dnsmasq.
pi@raspberrypi /usr/local/bin $ dig ezakus.net

; <<>> DiG 9.8.4-rpz2+rl005.12-P1 <<>> ezakus.net
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 28484
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;ezakus.net.                    IN      A

;; ANSWER SECTION:
ezakus.net.             0       IN      A       192.168.1.55

;; Query time: 26 msec
;; SERVER: 192.168.1.55#53(192.168.1.55)
;; WHEN: Fri Mar 20 23:46:12 2015
;; MSG SIZE  rcvd: 44

jacobsalmela:
So you detect the ```LANG``` variable and then set it to US, and then restore it after completion?

mospaeda:
Yes, but it's a very preliminary solution as it may have side effect
Current automatic detection has a problem on my conf
Diagnostic was : 
ifconfig output is different if system is setup in english or french
-> then language setting is important to use the current solution
Preliminary solution is to fall back to a working solution (here english configuration) and revert to the user language

potential issues on the current solution with LANG
 - saving and setting again the LANG setting may be to wide and special user settings may be lost (ex LC_ message which is not set to initial lang-ex french-  but in other lang -ex german- du to user special setup)
- current is to set LANG to en-US-utf8 (may be this setting is not installed on the default rpi or have been removed by user) . may be a test with locale -a and select one with en

other potential issue (not covered by your automatic detection
- output is also not tested to be a non empty or valid IPV4 (xx.xx.xx.xx) or valid IPV6
- may be we should test the output to fit to an non empty IPV4/IPV6 IP address

mospaeda:
After some thinkings , I think the best option for automatic detection should be to detect the pattern xx.xx.xx.xx on line2 of ifconfig output.
It will be much simpler and easy to read (then to maintain). It will be also more portable as it does not take any assumption on what is available on RPI in term of lang package. It will also avoid to swich between lang context (with potential user content loss).
I will try to find the command that is doing that.

pizzafritta:
What about this one?
``hostname -I``

mospaeda:
Yes, much better solution

jacobsalmela:
I'll have to look at this in a couple of weeks. Sorry for the delay

jacobsalmela:
I am merging your request.  There is another pull request, which makes a lot of improvements, but it [isn't working consistently yet](https://github.com/jacobsalmela/pi-hole/pull/8#issuecomment-87450197).

PR #12, opened 2015-03-23
Fix Hulu ad blocking

vineyard:
1.) Grab a 1x1 Transparent GIF from Wikipedia and hook that to the 404 error handler. This appears to be faster than a URL rewrite + serving up an HTML page.
2.) In the lighttpd.conf for /block hulu ads/, change "^/(.*)" to "^/published/(.*)" in the url.redirect configuration. This allows non-video communications with the Hulu ad servers to go through unmolested, while redirecting only the ad videos to the local media server. I've tested the change and it works well for me, at least on Xbox 360 and Xbox One devices.
3.) Add black.mp4, a 1 second black screen 720p video in MP4 format which will stream in place of the Hulu ads.


vineyard:
Apologies, this is my first pull request. I had intended to separate out the 404 redirection changes into a separate pull request.

jacobsalmela:
Thanks for this.  I have been out of town for this week and next.  So I will have to look at this in a another week or so.

jacobsalmela:
Sorry.  I haven't been able to confirm this works yet, but I am still testing with it.

jacobsalmela:
I have been testing this for a while now and I cannot get it to go.  What does the `published` line in `lighttpd.conf` do?  Also, are you using `minidlna`?

vineyard:
I noticed by looking at packet captures that the ad videos were always being requested from the /published/ directory on Hulu's server. This type of redirection seems to be enough to fool Xbox 360 / Xbox One devices, however I have not had success using this method on PC, Android, or Chromecast players.

jacobsalmela:
Interesting.  I wasn't able to get it to work on my computers either, but I will try it on the Xbox One.  If it works, it is better than it is working now, so I will merge it.  From there, we can figure out how to get other devices blocked.

jacobsalmela:
I have been working on this a long time, but can never get consistent results...have you made any more progress?

Mcat12:
This will be closed unless new info appears.

PR #13, opened 2015-03-30
Output shorter deleting subdomains when possible

pizzafritta:
what's new:
- sorting not necessary on every source, it's already executed at the end
- comments are allowed in blacklist and in whitelist. Everything following the symbol '#' is considered comment, even if it's not at the beginning of the line
- blank lines are allowed in blacklist and whitelist.
- tested this with awk (mac), mawk (raspbian) and also with gawk.
- bugfix
...
- subdomains removed from output when domain already exists
- output sorted in reverse order (right to left)
- filepaths defined at the top of the code.
- subdomains of items in whitelist/blacklist are whitelisted/blacklisted too

### DIFF
--- a/gravity-adv.sh
+++ b/gravity-adv.sh
@@ -64,8 +64,9 @@ echo "Removing duplicates, whitelisting, and formatting the list of domains..."
 grep -vhE "^\s*(#|$)" $tmpAdList| 
     sed $'s/\r$//'| 
     awk -F. '{for (i=NF; i>1; --i) printf "%s.",$i;print $1}'| 
-    sort -t'.' -k1,2| uniq |  grep -vwf $tmpWhiteList |
+    sort -t'.' -k1,2| uniq |  
     awk -F. 'NR!=1&&substr($0,1,length(p))==p {next} {p=$0".";for (i=NF; i>1; --i) printf "%s.",$i;print $1}'| 
+    grep -vwf $tmpWhiteList |
     awk -v "IP=$piholeIP" '{sub(/\r$/,""); print "address=/"$0"/"IP}' > $tmpConf
 numberOfSitesWhitelisted=$(cat $tmpWhiteList | wc -l | sed 's/^[ \t]*//')
 numberOfSitesBlacklisted=$(cat $tmpBlackList | wc -l | sed 's/^[ \t]*//')
--- a/gravity-adv.sh
+++ b/gravity-adv.sh
@@ -6,11 +6,18 @@ piholeIP="127.0.0.1"
 #piholeIP=$(hostname -I)
 
 # Config file to hold URL rules
-eventHorizion="/etc/dnsmasq.d/adList.conf"
+eventHorizon="/etc/dnsmasq.d/adList.conf"
 piholeDir='/etc/pihole/'     
 
 blacklist=$piholeDir'blacklist.txt'
 whitelist=$piholeDir'whitelist.txt'
+
+tmpDir='./tmp/'    
+tmpAdPrefix=$tmpDir'matter.pihole'
+tmpAdList=$tmpAdPrefix'.txt'
+tmpConf=$tmpDir'andLight.pihole.txt'
+tmpWhiteList=$tmpDir'yang.pihole.txt'
+tmpBlackList=$tmpDir'yin.pihole.txt'
 
 # Create the pihole resource directory if it doesn't exist.  Future files will be stored here
 if [[ -d $piholeDir ]];then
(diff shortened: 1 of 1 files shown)


jacobsalmela:
So I looked at a bunch of the entries from the existing list and yours.  Yours does seem to be missing some of the ones that currently exist.  

I don't quite understand all your changes for sorting and removing duplicates.  Perhaps you could explain them in detail so we can figure out how many domains are actually supposed to be there.

pizzafritta:
Give me some examples so I can test it.
I will explain the central part line by line so everybody can understand:
```grep -vE "^\s*(#|$)" $tmpAdList|```
removes blank lines and lines starting with '#' from matter.txt
```sed $'s/\r$//'```
replaces newline LFCR (dos) with CR (unix)
```awk -F. '{for (i=NF; i>1; --i) printf "%s.",$i;print $1}'```
reverse every domain, www.domain.com becomes com.domain.www
```sort -t'.' -k1,2 | sort```
sort domains, field1 then field2, delimiter is ".", then remove duplicates
```awk -F. 'NR!=1&&substr($0,1,length(p))==p {next} {p=$0".";for (i=NF; i>1; --i) printf "%s.",$i;print $1}'```
compare current line with previous, if it is a prefix skip it, otherwise reverse back to the original www.domain.com
```grep -vwf $tmpWhiteList```
removes entries from whitelist. Since entries in temporary whitelist are terminated with $, all the line endings with domaininwhitelist.com are removed
```awk -v "IP=$piholeIP" '{sub(/\r$/,""); print "address=/"$0"/"IP}' > $tmpConf```
format the output for dnsmasq

**Comment by jacobsalmela** [comment]
Sorry for the delay.  I'll check in next week.

**Comment by jacobsalmela** [comment]
Sorry for the delays.  See the screenshots below for the original list and the one created from your PR.  So I think the 97K isn't quite right.
differences between lists 1
differences between lists 2

It also seemed to slow down a bit when sorting.  Take a look at [this suggestion](https://github.com/jacobsalmela/pi-hole/issues/3#issue-58509894), which in my testing, has worked quite well.

**Comment by pizzafritta** [comment]
You don't see the differences clearly because you need to use the reverse domain name notation before sorting http://en.wikipedia.org/wiki/Reverse_domain_name_notation.
In the original code replace the command ```sort``` with
```
awk -F. '{for (i=NF; i>1; --i) printf "%s.",$i;print $1}' | 
sort | 
awk -F. '{for (i=NF; i>1; --i) printf "%s.",$i;print $1}'
```
code explanation here http://serverfault.com/questions/364330/sort-a-list-of-domain-names-fqdns-starting-from-tld-and-working-left
I know the code is a little bit slower, I will take a look at your suggestion and will do the changes in the next commit.

**Comment by jacobsalmela** [comment]
I'll have to look again when I am less tired...Does dnsmasq understand the reverse domain notation?

**Comment by pizzafritta** [comment]
The domains in the output are in the regular form. 
What the code does is: reverse domains | sort | reverse back
So if the input is:
```
www.bar.com
foo.bar.com
foo.com
foo.net
bar.com
```
with regular sort the output would be
```
bar.com
foo.bar.com
foo.com
foo.net
www.bar.com
```
that doesn't really help understand where the subdomains actually belong.
Instead, with 'reverse' | sort | 'reverse' the result is
```
bar.com
foo.bar.com
www.bar.com
foo.com
foo.net
```
this way all the subdomains come right after the main domain so they can be eventually removed.

**Comment by jacobsalmela** [comment]
So if I sort both lists using your code and then run it through `opendiff`, it should show any differences, correct?

**Comment by pizzafritta** [comment]
exactly.

**Comment by jacobsalmela** [comment]
I'm re-doing most of the code so more lists can easily be added and better-suited for future changes.  I'll do my best to implement your suggestions, but check it out when it's ready and see what you think.

**Comment by pizzafritta** [comment]
I've also made many improvements to the code in gravity-adv.sh in order to speed-up everything. 
- curl is not executed if last execution is newer than Last-modified in http request.
- extraction of domain list from file 'hosts' standardized and shortened (adding more sources is easier)
- temporary files are not overwritten if not necessary.
- domains with no dots (i.e. localhost) removed automatically from list
- the code to reduce duplicates and create the output is now shorter
- "/etc/dnsmasq.d/adList.conf" is not overwritten and dnsmasq not restarted when is not necessary
I don't know whether pushing it here or in a new pull request.

**Comment by jacobsalmela** [comment]
Let me see what you have done.  I'm still working on mine.  I can adapt my code pretty easily to whatever you have done.

**Comment by jacobsalmela** [comment]
I would like you take a look at my re-working of the code to allow the easier addition of sources.  It makes things cleaner and easier to read.  I used your code to reverse sort the domains, which seems to work as expected.

However, I do like some of your improvements:

- domains with no dots (i.e. localhost) removed automatically from list (I was currently doing this via the whitelist)
- "/etc/dnsmasq.d/adList.conf" is not overwritten and dnsmasq not restarted when is not necessary

Is there some way to share it via GitHub without pushing it to the repo?  Otherwise, I'll just paste it in here.

**Comment by pizzafritta** [comment]
Ok but keep localhost and localdomain in whitelist, otherwise you will not remove test.localhost localhost.localdomain etc. This is my whitelist
```
localhost
#this is a comment
localdomain
broadcasthost
android
test
local
```

Let me see your code, post the changes you've done here, push it in a branch or anything you like.

jacobsalmela:
Cool.  For my first branch, I don't think I messed it up too bad!

https://github.com/jacobsalmela/pi-hole/blob/usingloops/gravity-adv.sh

pizzafritta:
I like the idea of the loop, I was thinking about putting all the sources in an external file as well. (I have a couple more sources I could add)
I have a few suggestions, remarks and things that could be improved in the future. I hope you will appreciate.

- It's possible to remove the switch-case construction if we can handle all the sources the same way:
  - pgl.yoyo.org can return the output in file 'hosts' format here http://pgl.yoyo.org/adservers/serverlist.php?hostformat=hosts&mimetype=plaintext
  - adblock.mahakala.is returns a 403 Forbidden error only if user-agent is not set. But we can set the option -A "Mozilla/5.0" for all the curl requests and we should be fine.
- I am not sure the whitelist works fine this way. The option -w in grep does the job correctly if the domains have a $ at the end of every row. Otherwise it will match words in any position (i.e. localhost.adsite.com). That's why I have a temporary whitelist file where I perform the following
  - remove empty lines (they could cause issues)
  - remove comments
  - add $ at the end of every row
- Some sources have windows style newlines (LFCR), this could be a problem sometimes, for example when using ```uniq```. In [master](https://github.com/jacobsalmela/pi-hole/blob/master/gravity-adv.sh) this was done using ```sed $'s/\r$//'```
- Every single domain is written on disk 6 times in this implementation. I believe this number can be reduced.
list.N.xyz > list.N.xyz.domains > pihole.0.matter.txt > pihole.1.andLight.txt > pihole.2.eventHorizon.txt > adList.conf

If you want I can work on this, rebase my latest commit to this new branch, in order to address the improvements I already did and the points I've just mentioned.

jacobsalmela:
I agree with everything you mentioned.  I hate to make you re-do work you have already done, but I would definitely appreciate any help you can offer.  I will look at your existing code to see if I can get it to adapt properly on this branch.

jacobsalmela:
So I finally got to comparing the lists with the subdomains removed.  Your code does indeed seem to work.  Have you been using that list?  I'm guessing most of the domains have the single purpose of serving ads, but have you run into any issues where the subdomain was serving the ads, but the regular domain was some legitimate site?

subdomains removed

does removing subdomains block legitimate content?

Also, `curl -A "Mozilla/5.0"` did not work for http://hosts-file.net/.%5Cad_servers.txt.  It gives some sort of firewall warning.  `"Mozilla/10.0"` seems fine though.

I fixed the whitelist thing.

I don't quite get how we can reduce the writes to disk based on your explanation.

jacobsalmela:
Closing this since the TLD sort did not work as well and ended up blocking legitimate services.  This is because some domains offer real sites, but only their subdomain serves ads.  I did, however, implement some of the stuff in the other branch of this project.

pizzafritta:
this is not true. subdomains are removed only if main domain is also in the list. 
If a 1st level domain is in the list, it means it has been reported as server serving ads by one of the sources. 

On the other hand if someadserver.com is in the list, there is no point of having foo.someadserver.com and bar.someadserver.com in the list, they would be blocked anyway, they are actually unnecessarily redundant.

jacobsalmela:
I'm not totally convinced.  I tried using it for a week or so and ads were getting through and I had trouble loading some sites.  Have you tried using your list?

pizzafritta:
From the [man page of dnsmasq] (http://www.thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html) (for --server, but also applies to --address):
>More specific domains take precendence over less specific domains, so: --server=/google.com/1.2.3.4 --server=/www.google.com/2.3.4.5 will send queries for *.google.com to 1.2.3.4, except *www.google.com, which will go to 2.3.4.5

>The special server address '#' means, "use the standard servers", so --server=/google.com/1.2.3.4 --server=/www.google.com/# will send queries for *.google.com to 1.2.3.4, except *www.google.com which will be forwarded as usual.

jacobsalmela:
That seems to be just for sending the queries to a different IP address.  have you actually been using your shortened list?  I tried to use it for a while but browsing was very difficult.

PR #15, opened 2015-04-20
Curl executed only if file is newer

pizzafritta:
If last-modified attribute of remote file is older than local file download is not performed.
Full original file is not written on disk, only clean domain list is.

[I will do the changes one at a time]

### DIFF
--- a/gravity-adv.sh
+++ b/gravity-adv.sh
@@ -37,6 +37,13 @@ else
 	sudo mkdir $piholeDir
 fi
 
+# if pipe is not empty write it into file $1
+function writeifne() 
+    {
+    read pipe || return 1
+    { printf "%s\n" "$pipe"; cat; }  > "$1"
+    }
+
 # Loop through domain list.  Download each one and remove commented lines (lines beginning with '# 'or '/') and blank lines
 for ((i = 0; i < "${#sources[@]}"; i++))
 do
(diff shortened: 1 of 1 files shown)


jacobsalmela:
Shouldn't it have a file to compare with?  There is a pipe that has nothing after it...

pizzafritta:
The option -z tells curl to compare the date of the specified file with the remote file (last-modified attribute of http request). 
If the remote file is older, curl returns nothing, so the pipe would be empty. So at the end of the pipe I put a function ```writeifne``` (write if not empty) that writes the output of the pipe to a file only if there is something to write. 
The function ```writeifne``` replaces the '>' output redirection operator, checking at the same time if there is something to write.

jacobsalmela:
OK.  I get it now.  I also see the piped commands are just on a new line.  I think we need to add an if block in the loop though because it will always say "Getting domain list..."  Maybe we could just check the exit status of the curl command.  Something like:
```
curl -z ...
if [[ $? = 0 ]];then
echo "Getting $domain list..."
echo "    N domains found."
else
echo "Skipping $domain list because it does not have any new entries..."
fi
```

pizzafritta:
Changes made.
```$?``` is always 0, even if there is nothing to download doesn't mean the command didn't succeed.

jacobsalmela:
No exit codes, huh?  Looks like your solution works!

PR #18, opened 2015-06-04
On CentOS 7 the $supernova is empty

rmceoin:
The `echo -n` is just creating an empty file and has no standard output.  I'm guessing that's whats getting used instead of the output from grep.
When I change the pipe to an "and", I get content in the supernova.


rmceoin:
The hostsformat branch doesn't appear to have this issue.  I see it takes a very different approach with /etc/hosts so I'll have to tinker with it.

I'm using the gravity script on CentOS 7 so there are minor changes that I've done with it and I'm trying to keep the changes to a minimum.

jacobsalmela:
I am doing something a little different on the hostsformat branch, and will probably be focusing my efforts there due to the recent [performance issue that surfaced when one of the domain lists added a ton of entries](http://jacobsalmela.com/huge-update-to-ad-list-that-the-pi-hole-uses/).

PR #20, opened 2015-06-19
If exists, import a config file to allow for overriding script variables.

rmceoin:
To allow for minor variations of use of the script, parsing a config file could update some of the data locations.

In my case my `/etc/pihole/pihole.conf` looks like this:

    origin=/var/run/pihole
    adList=/etc/dnsmasq.d/adList

This would make it so that I wouldn't need to modify future versions of the script.  I could also create my own "sources" list and possibly not need to tweak the script.

### DIFF
--- a/advanced/dnsmasq.conf
+++ b/advanced/dnsmasq.conf
@@ -7,3 +7,4 @@ interface=eth0
 listen-address=127.0.0.1
 cache-size=10000
 log-queries
+local-ttl=300
--- a/gravity.sh
+++ b/gravity.sh
@@ -16,8 +16,13 @@ sources=('https://adaway.org/hosts.txt'
 'http://winhelp2002.mvps.org/hosts.txt')
 
 # Variables for various stages of downloading and formatting the list
+adList=/etc/hosts
 origin=/etc/pihole
 piholeDir=/etc/pihole
+if [ -f $piholeDir/pihole.conf ]; then
+	. $piholeDir/pihole.conf
+fi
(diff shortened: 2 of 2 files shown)


jacobsalmela:
I like everything you have done.  Have you found that the `local-ttl` setting improves performance?

Also, it may seem minor, but I typically prefer the double bracket notation for `if` statements.  If you make that change, I will merge your request.

rmceoin:
What I did to check the local-ttl was to wireshark and look at what's on the wire.  Without the local-ttl queries that ended up using adList had a TTL of 0.  And that each time I did a `nslookup` on a test name it would issue a DNS query to the server.  Whereas if it had a TTL greater than 0, the client would issue a DNS on the first `nslookup` but additional `nslookup`s would not issue a query.  That clearly means things should be faster.

I'll check out the double bracket notation.

jacobsalmela:
Cool.  So I'm guessing that is 300 seconds?  Is that just an arbitrary number or did you try out different values?

jacobsalmela:
Looks great.  Let me know what you find out about the value to set.

rmceoin:
Yea, it's 300 seconds.  I randomly picked a value to be honest.  5 minutes seems reasonable.

Thank you for the merge.  I'm trying to gear up for production use of the script/concept at work.

jacobsalmela:
No problem.  It's a great addition.  I would love to hear how you get it going at work and the response.

Thanks!

PR #21, opened 2015-06-22
Strip carriage returns on matter so that whitelists work correctly.

rmceoin:
If a line in `matter` had a carriage return, it would not match with entries in the whitelist.  This strips carriage returns at creation of matter.

Can probably remove the other sed and awk that does the same thing, but did so after whitelist processing.

BTW, I don't have a PI, so I can't test on it.  I can only assume it has `tr`.

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -77,7 +77,7 @@ done
 
 # Find all files with the .domains extension and compile them into one file
 echo "** Aggregating list of domains..."
-find $origin/ -type f -name "*.$justDomainsExtension" -exec cat {} \; > $origin/$matter
+find $origin/ -type f -name "*.$justDomainsExtension" -exec cat {} \; | tr -d '\r' > $origin/$matter
 
 # Append blacklist entries if they exist
 if [[ -f $blacklist ]];then
--- a/gravity.sh
+++ b/gravity.sh
@@ -83,7 +83,7 @@ find $origin/ -type f -name "*.$justDomainsExtension" -exec cat {} \; | tr -d '\
 if [[ -f $blacklist ]];then
         numberOf=$(cat $blacklist | wc -l | sed 's/^[ \t]*//')
         echo "** Blacklisting $numberOf domain(s)..."
-        cat $blacklist >> /tmp/matter.txt
+        cat $blacklist >> $origin/$matter
 else
         :
 fi


rmceoin:
If you'd like to test it out, we had added `res.cloudinary.com` to the whitelist so that http://meh.com/ would show it's image.  Before adding the `tr` it would still end up in the `adList`.

rmceoin:
Was adding our first blacklist site and bumped into a mistake there.  Wasn't being added to the new location for matter.

jacobsalmela:
You are correct.  Thanks for catching the error with the variable, too.  At some point, I should run the entire list through `od` and see if there are other [invisible characters](http://jacobsalmela.com/octal-dump-od-shows-you-hidden-characters-and-ascii-values/) that are causing similar issues.

Thanks for the PR!

rmceoin:
You're welcome.  I've rolled things out to a small subnet and am tailing the log a bunch.  Also playing with a simple script to get stats.  Looks like we run at about 9-10% holed rate.

    # dnshole-report
    From Jun 21 03:50:06 to Jun 22 14:58:40
    Queries: 54844 (v4 32939/v6 19430)
    Holed: 3206 9.7%

jacobsalmela:
Haha awesome.  That is very cool.

jacobsalmela:
Just curious if you would be willing to share your script that shows how many have been "holed"?

rmceoin:
Sure thing!  It's still rather basic, but handy to get a vague notion of how much traffic is being holed.  If you think it can be helpful, you're welcome to add it into the project.

https://dl.dropboxusercontent.com/u/1029018/dnshole-report

rmceoin:
I just now updated the script.  Had an issue with the regex matching the date.  We seem to doing about 8%.

    # dnshole-report -l /var/log/messages-20150705 
    From Jun 29 03:26:19 to Jul  5 03:02:08
    Queries: 276802 (v4 192550/v6 78498)
    Holed: 15333 8.0%

jacobsalmela:
Cool.  I have made something similar based on your code, but I used bash since that's what I know.  I also added a little flair for dramatic effect. =)
pihole LCD

rmceoin:
That's awesome!

jacobsalmela:
Haha.  It's been pretty fun!  I'll clean up the script a bit and include it in my next push.

PR #23, opened 2015-07-13
Parses host only file formats now.

rmceoin:
Previously only handled hosts file format.

Specifically, it can now handle the following list:
'http://mirror1.malwaredomains.com/files/justdomains'

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -69,7 +69,7 @@ do
 		# Most of the lists downloaded are already in hosts file format but the spacing/formating is not contigious
 		# This helps with that and makes it easier to read
 		# It also helps with debugging so each stage of the script can be researched more in depth
-		echo "$data" | awk 'NF {if ($1 !~ "#") print $2}' > $saveLocation."$justDomainsExtension"
+		echo "$data" | awk 'NF {if ($1 !~ "#") { if (NF>1) {print $2} else {print $1}}}' > $saveLocation."$justDomainsExtension"
 	else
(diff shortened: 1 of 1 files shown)


jacobsalmela:
That's crazy good.  I was actually just trying to figure out how to do that so I could add the disconnect lists.  I'll test it out when I get home and then merge.

rmceoin:
Yea, initially I thought it'd be complicated.  Then I mustered up my awk skills and bingo, a very simple solution popped up.   I grabbed copies of all the lists before the change, then diff'd them before and after, so my testing showed them not changing.  Excepting of course the new list I added.  Only netted 200 additional domains, but every bit helps.

Are you referring to these lists?
'https://s3.amazonaws.com/lists.disconnect.me/simple_ad.txt'
'https://s3.amazonaws.com/lists.disconnect.me/simple_tracking.txt'

I did have them included for a bit, but they seemed to be duplicates with nothing additional.  But I might try testing them again now.

rmceoin:
Adding those two lists netted 64 additional hosts to block.  I'll leave them in on my sources list.

jacobsalmela:
Those are the ones.  Yeah, even blocking one extra domain could be worth it because that domain might serve multiple ads or who knows what else...

jacobsalmela:
Sweet.  It works as expected!  I knew the `awk` command would be simple, but I just couldn't figure out how to do it.  Nice work--very elegant.

PR #27, opened 2015-07-28
Log PiHole dns stats to /var/log/pihole

Fourdee:
Log PiHole dns stats to /var/log/pihole
- Removes the requirement for rsyslog to be installed.
- Prevents reading the shared logfile /var/log/daemon.log
- Ensures PiHole dns stats are kept in its own logfile.

### DIFF
--- a/advanced/dnsmasq.conf
+++ b/advanced/dnsmasq.conf
@@ -7,5 +7,5 @@ interface=eth0
 listen-address=127.0.0.1
 cache-size=10000
 log-queries
-log-facility=/var/log/pihole
+log-facility=/var/log/pihole.log
 local-ttl=300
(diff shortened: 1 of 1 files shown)


jacobsalmela:
I do like the idea of the Pi-hole having its own log file.  However, I would like the `.log` file extension added so it's easy to identify as a log file.  Also, the file path change would need to be made [here](https://github.com/jacobsalmela/pi-hole/blob/master/advanced/Scripts/chronometer.sh#L17-L20) for anyone that is using the LCD display. 

Will you commit those additions?

Fourdee:
Hi Jake,

As per your requests, all files have been updated to reflect /var/log/pihole.log
Web interface: https://github.com/jacobsalmela/AdminLTE/pull/1
LCD screen: https://github.com/jacobsalmela/pi-hole/pull/28

PR #29, opened 2015-07-30
Patch 3 - Dont use /etc/hosts

Fourdee:
/etc/pihole/gravity.list now stores the block list. Ensures the
/etc/hosts file is left untouched.

### DIFF
--- a/advanced/Scripts/chronometer.sh
+++ b/advanced/Scripts/chronometer.sh
@@ -17,7 +17,7 @@ do
 	todaysQueryCount=$(cat /var/log/pihole.log | grep "$today" | awk '/query/ {print $7}' | wc -l)
 	todaysQueryCountV4=$(cat /var/log/pihole.log | grep "$today" | awk '/query/ && /\[A\]/ {print $7}' | wc -l)
 	todaysQueryCountV6=$(cat /var/log/pihole.log | grep "$today" | awk '/query/ && /\[AAAA\]/ {print $7}' | wc -l)
-	todaysAdsEliminated=$(cat /var/log/pihole.log | grep "$today" | awk '/\/etc\/hosts/ {print $7}' | wc -l)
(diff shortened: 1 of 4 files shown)


jacobsalmela:
Why don't you want to use `/etc/hosts`?  I like the change, but I'm just curious.  

Also, have you tested performance using `addn-hosts`?  I know when I was using `.conf` files in `/etc/dnsmasq.d` performance took a huge hit when the list went past 120,000 entries.  

`/etc/hosts` can handle 900,000+ entries so I'm just curious if you noticed an issue.  I'll test it myself, but I want to make sure the Pi-hole makes things faster, not slower.

rmceoin:
I'll chime in from my perspective.  I made a similar change to prevent touching `/etc/hosts`.  I did it because tinkering with `/etc/hosts` makes me nervous.  I'd hate for a mistake to happen that results in either a missing or blank `/etc/hosts`.  I would expect bad things to occur should that end up happening.

As much as we might try to make an error proof script, mistakes still occur and that's such a sensitive file.

Fourdee:
Hi Jake,

I think i may have missed a few /etc/host changes across my recent pulls. I'am going to run some tests just to make sure I converted all the /etc/hosts entries. I'll reply once done.

>> Why don't you want to use /etc/hosts? I like the change, but I'm just curious.
It just seems cleaner to not interfere with system files and keep everything in its own area. Also when a user uninstalls PiHole, they will manually have to edit their /etc/hosts file to remove the block list.

>> Also, have you tested performance using addn-hosts? I know when I was using .conf files in
Not yet. I'll run some. Cant see it being much of a difference as dnsmasq loads the list into ram/cache.

>> /etc/hosts can handle 900,000+ entries so I'm just curious if you noticed an issue.
Wasnt aware the /etc/hosts size limit would be any different to other files on the same filesystem?

Fourdee:
Just tested with a fresh install, PiHole and the LTEadmin run fine. I cant see any noticible decrease in performance.

jacobsalmela:
It works nicely!

PR #33, opened 2015-08-10
Fix URL of chronometer.sh

mathiasschopmans:
Github URLs seems to be case-sensitive. Now the chronometer.sh souldn't contain "Not Found" anymore. ;)


jacobsalmela:
Thanks!

Issue #32, opened 2015-08-08
Add data validation to prevent rogue entries


jacobsalmela:
Sometimes entries from the third-party lists have errors, [such as adlog..com](http://jacobsalmela.com/raspberry-pi-block-ads-adtrap/#comment-2180476491) (two periods instead of one).

This will probably require some <code>sed</code> and <code>awk</code> skills to check for the proper formatting of:

**subdomain(s) (if applicable), a period (.), the domain, another period (.), and finally, the top level domain.**

korhadris:
Here's a see command to remove duplicated '.' characters:

`sed 's/\.\+/./g/'`

kurumushi:
The real challenge is to see if these can be combined into one, I'm not very good at simplification.
Here's a few more useful ones:
`sed -e 's/ \.//g'`: remove periods at the start and the middle of domains.
`sed -e 's/\[. ]+$/$/g'` remove domains ending in a period, and trailing white space.

korhadris:
Yeah, simplifying regexp statements to be optimized is rough. As long as I really don't require the speed I usually combine statements like this:

jacobsalmela:
It looks pretty good, but it's a bit out of my skill level.  Maybe someone could make a pull request and I can test it out.

PR #34, opened 2015-08-12
fixes corrupt lines that have two dots

CHannema:
some of the downloaded files have domains entries like 'ads..com'. These are invalid and should be removed.

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -69,6 +69,8 @@ do
 		# This helps with that and makes it easier to read
 		# It also helps with debugging so each stage of the script can be researched more in depth
 		echo "$data" | awk 'NF {if ($1 !~ "#") { if (NF>1) {print $2} else {print $1}}}' > $saveLocation."$justDomainsExtension"
+		# replace domains that have two dots (..com) with one dot and let duplicate removal sort out any duplicate domains
+		sed -i -e 's/\.\./\./g' $saveLocation."$justDomainsExtension"
 	else
 		echo "Skipping $domain list because it does not have any new entries..."
 	fi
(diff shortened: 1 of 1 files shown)


jacobsalmela:
So this works well, but I would like to see some way to integrate it into an earlier part of the script.  I recently updated my lists and they were are 1.6 million entries, so your command has to parse through the entire file a second time.  I would like to minimize that if possible.

I was thinking it could be integrated into [line 58](https://github.com/jacobsalmela/pi-hole/blob/master/gravity.sh#L58).

CHannema:
You make a valid point, so I've moved the sed command earlier in the file.
This should work for both existing installations and new ones.
Seemed better/more efficient placing it past [line 68](https://github.com/jacobsalmela/pi-hole/blob/master/gravity.sh#L65).

korhadris:
Two comments:

1. I don't know if it happens, but I'd want to handle the case of two or more '.' characters in your sed expression:
 `sed 's/\.\.\+/./g'`
(I also removed the extra escape of `.` in the replacement portion of your regexp, as it isn't needed in by sed.)

2. I would think you can add your sed as part of line 71; pipe the output of awk through sed before redirecting to the file:
`echo "$data" | awk 'NF {if ($1 !~ "#") { if (NF>1) {print $2} else {print $1}}}' |  sed 's/\.\.\+/./g' > $saveLocation."$justDomainsExtension"`

korhadris:
I expect this should fix the primary concern in #32. There is some additional discussion in #32 (some more regexps) that could be applied if there are concerns with leading or trailing `.` characters as well.

korhadris:
As I was fixing other parts of #32 and #35 today, I included this part as one of my sed expressions.

CHannema:
Alright, then I guess this one can be closed?
I see the suggested solution has already been improved upon elsewhere. ( I hadn't checked the actual issues page, whoops).
Let me know if there's anything else, otherwise I'll close this one in a few days.

jacobsalmela:
This should be fixed via #38

Issue #35, opened 2015-08-12
adblock.mahakala.is has some issues


kurumushi:
This hosts file blocks many legitimate sites, including both the download site for adobe flash and rival adblock lists (it blocks hosts-file.net which is one of your pihole sources.) It also includes lines at the top that give domain names to the hosts 'debian,' and 'android,' which may not be wanted by some people.

jacobsalmela:
The mahakala list has caused some other problems, too, but it definitely helps more than it hinders.  I show these entries at the top of my list:
```
localhost
localhost.localdomain
android
android.localhost
test
test.localhost
debian
debian.localhost
```
I agree these are probably not the best entries to include by default.  You could easily add these to `/etc/pihole/whitelist.txt` but not everyone will do that.

I will look into a way to remove these entries from this, or any list.

rmceoin:
In what situation would a single bare word be appropriate to blacklist?  Unless we want to block an entire TLD.  But do any of the downloaded lists contain a TLD?  If not, then I'd say skip any line where the host specified is a single word without a period.   

For the rest perhaps a regex to strip lines with *.localdomain and *.localhost

jacobsalmela:
I'm not sure `grep` will interpret the `*`.  Currently, we are using `awk` to append a `$` to the end of each line. I have always struggled with regex, but I agree with your logic.  Any one word lines probably isn't a domain. 

We might be able to add all of those entries at the beginning of the user's `whitelist.txt` the [way I used to do it](https://github.com/jacobsalmela/pi-hole/blob/f6ccb4b658ec67971472965491b728abdcfcc060/gravity.sh#L105-L108) for the hosts file.  But I think some sort of regex/data validation would be better than hardcoding stuff.

kurumushi:
This command will delete all domains that do not contain a dot.
```
sed -e '/^[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\} .*\..*$/!d'
```

jacobsalmela:
Awesome!  `sed` experts are the best.  Would you mind explaining how each section of the command works?

korhadris:
Hopefully I'm not stealing your thunder, or being overly descriptive below...

"/": sed delimiter
"^": beginning of line
"[0-9]\{1,3\}": any 1 to 3 digit number
"\.": a single '.' character
These are repeated 4 and 3 times to match an IPv4 address.
" ": a single space
[...]

Issue #37, opened 2015-08-17
Error fork: Cannot allocate memory


jacobsalmela:
Many users are having trouble downloading the mahakala list and running out of memory.  Giving the GPU 16MB seems to work for a lot of people, but as the list grows larger, I'm looking for a long-term solution.

enjunear:
I've commented out the mahakala for now. Works fine without it, but obviously you don't get to block all those 1.6 million domains. That's well over 900,000!!!

jacobsalmela:
Yeah this annoying error has been coming up a lot more recently.  I'm looking into increasing the swap size somehow via `gravity.sh`

jpearsall:
I was able to get around this by bypassing the assignment to the data variable and piping the curl to awk (shamelessly pulled from the if/then conditional on data further down) and directing it then redirecting it to the destination file.

jacobsalmela:
I actually used to have it [setup that way](https://github.com/jacobsalmela/pi-hole/pull/15/files), but we changed it for a few reasons:

- to limit writing to the SD card (reduces the likelihood of corruption)
- to be a respectful netizen by not downloading the entire list needlessly every time `gravity.sh` runs

so I would like to keep it as is and either find a way to increase the memory or come up with a new solution all together.

korhadris:
If I maintain 2. in your list above and decrease file read/writes, can I rearrange some of the intermediate files that are at the end of gravity.sh, in gravity_advanced()? There are some sed sections that are no longer needed, or could be moved, with some of my latest commits. If so, I'd like to store the data from the curl requests to disk, that way $data isn't holding the whole mahakala list in memory at one time. Although, I'd still assume it would still use a bit of memory when it gets to the sort | uniq section later.

jacobsalmela:
If we can increase the swap size to 500 or more, if needed,  we can still limit the writing to the SD card since the swap will only be used when needed.  I was thinking of something like this:
```
dphys-swapfile swapoff
echo -e "CONF_SWAPSIZE=500" > /etc/dphys-swapfile
dphys-swapfile setup
dphys-swapfile swapon
```

jacobsalmela:
I made a more streamline **development** branch so try it out and let me know if you still get the memory allocation error.

https://github.com/jacobsalmela/pi-hole/blob/development/gravity.sh#L47-L70

x20mar:
[...]

PR #38, opened 2015-08-23
Changes for #32 and #35.

korhadris:
Let me know if there are some individual commits that you don't like. There were a few changes I made to try and make the script more concise.


jacobsalmela:
I like a lot of what you have done, but it's a lot to look through at the moment (it's times like this I wish I could do this full time!).  My main concern at the moment is fixing #37 since it can prevent new users from getting the lists.  After that, I will look at these more in depth.

korhadris:
Yup. I thought the comment would fit with your theme. :)

jacobsalmela:
It definitely fits the theme. =)

I'll have to check these out next weekend or later, so sorry for the slow response.  I wish I could do this full time...

PR #42, opened 2015-09-12
Allow for local settings to disable swap

hawson:
Not all systems need swap files, so allow this to be locally disabled by setting the `noSwap` variable in the `pihole.conf` file

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -19,9 +19,6 @@ sources=('https://adaway.org/hosts.txt'
 adList=/etc/pihole/gravity.list
 origin=/etc/pihole
 piholeDir=/etc/pihole
-if [[ -f $piholeDir/pihole.conf ]];then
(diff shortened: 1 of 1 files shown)


jacobsalmela:
Did you mean to use `-n` instead of `-z`  on [line 63](https://github.com/jacobsalmela/pi-hole/pull/42/files#diff-9e53ae7c3c4f9235e5908d902fb93c8aR63)?

hawson:
Yes.  I'd swapped some variable names ('swap' vs 'noswap'), forgot to toggle the test.

jacobsalmela:
Make the change I and I will merge your request.

jacobsalmela:
Looking at #43, I think we will try move away from storing the lists in memory, but this pull won't hurt anything and will help those who want to use their own `.conf`

PR #43, opened 2015-09-12
Reduce mem usage

hawson:
This pull aims to reduce the RAM usage needed by pihole when parsing new/updated block lists.  

The files are downloaded locally and then operated on, instead of being stored as (very large) variables in the shell script.  Some simplistic tests (query /proc/<pid>/status for RSS size) show that the code in master uses ~345,000KB when running (that's the high water mark). When processing locally, RAM usage in this case is somewhere around 8,600KB

It's important to note that the files are downloaded only if the upstream copies are newer, which should address some of the concerns mentioned in #37.  That said, if SD card IO is really that much of a concern, there should be full dependency processing to eliminate unnecessary writes.

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -131,7 +131,7 @@ echo "** Aggregating list of domains..."
 find $origin/ -type f -name "*.$justDomainsExtension" -exec cat {} \; | tr -d '\r' > $origin/$matter
 
 # Append blacklist entries if they exist
-if [[ -f $blacklist ]];then
+if [[ -r $blacklist ]];then
 	numberOf=$(cat $blacklist | sed '/^\s*$/d' | wc -l)
 	echo "** Blacklisting $numberOf domain(s)..."
 	cat $blacklist >> $origin/$matter
@@ -159,7 +159,7 @@ function gravity_advanced()
 	}
 	
 # Whitelist (if applicable) then remove duplicates and format for dnsmasq
-if [[ -f $whitelist ]];then
+if [[ -r $whitelist ]];then
 	# Remove whitelist entries
 	numberOf=$(cat $whitelist | sed '/^\s*$/d' | wc -l)
 	plural=; [[ "$numberOf" != "1" ]] && plural=s
--- a/gravity.sh
+++ b/gravity.sh
@@ -92,7 +92,7 @@ do
 			;;
 		
 		"pgl.yoyo.org") 
-			cmd="curl -s -d mimetype=plaintext -d hostformat=hosts"
+			cmd="curl -d mimetype=plaintext -d hostformat=hosts"
 			;;
 
 		# Default is a simple curl request
--- a/gravity.sh
+++ b/gravity.sh
@@ -78,28 +78,52 @@ do
 	
 	# Save the file as list.#.domain
 	saveLocation=$origin/list.$i.$domain.$justDomainsExtension
-	
-		echo -n "Getting $domain list... "
-	# Use a case statement to download lists that need special cURL commands to complete properly
-	case "$domain" in
-		"adblock.mahakala.is") data=$(curl -s -A 'Mozilla/5.0 (X11; Linux x86_64; rv:30.0) Gecko/20100101 Firefox/30.0' -e http://forum.xda-developers.com/ -z $saveLocation $url);;
-		
-		"pgl.yoyo.org") data=$(curl -s -d mimetype=plaintext -d hostformat=hosts -z $saveLocation $url);;
 
-		*) data=$(curl -s -z $saveLocation -A "Mozilla/10.0" $url);;
-	esac
+	agent="Mozilla/10.0"
 	
-	if [[ -n "$data" ]];then
+	echo -n "Getting $domain list... "
+
+	# Use a case statement to download lists that need special cURL commands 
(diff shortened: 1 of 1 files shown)


jacobsalmela:
Initially, we switched to putting them in a variable  (`data`) to prevent writing to the SD card.  Later, the feature was added to only download the lists if there were changes.  I have personally never had a failed SD card, but I have read that it happens to people often enough.

What do you mean by _dependency processing_?

hawson:
By "dependency processing", I mean something like a makefile:  you have outputs that require various actions and other dependencies.

Storing the data all in RAM (in bash variables), thus forcing a host to use additional swapfiles, in order to save writes to an SD card seems backwards.

korhadris:
I think some of your memory usage stats may be low. The sorting step alone of supernova uses 134168 kB on my system. That said, I agree we should get away from storing `data` in memory.

Here are my stats on sorting using `time` (Not the bash built-in time, the GNU command `time`, which may need to be installed:
```
command time -v sort -u pihole.2.supernova.txt > /dev/null
        Command being timed: "sort -u pihole.2.supernova.txt"
        User time (seconds): 17.21
        System time (seconds): 0.47
        Percent of CPU this job got: 308%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:05.72
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 134168
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 33727
        Voluntary context switches: 1854
        Involuntary context switches: 555
        Swaps: 0
        File system inputs: 62520
        File system outputs: 0
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

jacobsalmela:
I haven't forgotten about this.  Just waiting for some more time that I have to test it.  It looks pretty good though from my initial scan.

jacobsalmela:
Your commits are on the right track, but there are some issues:

1. Lines [47-70](https://github.com/hawson/pi-hole/blob/reduce_mem_usage/gravity.sh#L47-L70) can be removed since we won't be storing the variables in memory any more
2. On [line 103](https://github.com/hawson/pi-hole/blob/reduce_mem_usage/gravity.sh#L103), I would prefer to have the variables named to fit the theme of the script (sci-fi, Star Trek-esque).  I think [`patternBuffer`](http://en.memory-alpha.wikia.com/wiki/Pattern_buffer) would be the ideal variable name.
3. On [line 104](https://github.com/hawson/pi-hole/blob/reduce_mem_usage/gravity.sh#L104), again, a themed variable name such as [`temporalDistortion`](http://en.memory-alpha.wikia.com/wiki/Temporal_distortion) or [`tachyonEmissions`](http://en.memory-alpha.wikia.com/wiki/Tachyon)
4. On [lines 108-109](https://github.com/hawson/pi-hole/blob/reduce_mem_usage/gravity.sh#L108-L109), you display the entire command that is running.  I prefer to keep the `echo`s simple and clean to reduce on-screen clutter
5. [Lines 143-144](https://github.com/hawson/pi-hole/blob/reduce_mem_usage/gravity.sh#L143-L144) store the filename in the variable, so it displays as `** 1679450 /etc/pihole/pihole.1.andLight.txt domains being pulled in by gravity...` instead of just `** 1679450 domains being pulled in by gravity...`
6. Same thing for [152-153](https://github.com/hawson/pi-hole/blob/reduce_mem_usage/gravity.sh#L152-L153)
7. I'm far from a `sed` expert, so could you please explain how [line 119](https://github.com/hawson/pi-hole/blob/reduce_mem_usage/gravity.sh#L119) works?

hawson:
1. The patch aims to have minimal changes, and removing swap support is outside the scope of this.
1. I can update the names, but I will note that the...unusual names (even for someone versed in Star Trek Technobabble)...made coding slower than usual.
1. ditto
4. Helps with debugging (short of running with -x, which is entirely too verbose.  Easy enough to comment out.
5. Okay
6. ditto
1.  Sure:

- -n means "don't print by default" (Perl's -n option is similar).
- -r means use extedned regexes, instead of the very limited set normally supported
- Then there are two different statements, each following the two -e options.
-- The first statement finds 2 or more literal "." characters, and replaces them with just one.  I found several cases where there are multiples periods in a row, and this fixes that problem.
-- The second looks for lines with a literal "." character, and prints it (this is needed because we "-n" earlier on to suppress printing, and also handles blank lines.

jacobsalmela:
Did you want to make the changes?

jacobsalmela:
I have some time this weekend if you wanted to make the changes.

dschaper:
I'd be happy to make a PR with the changes you requested if this issue is still open for consideration.

jacobsalmela:
Yes, it is.  Thanks!

dschaper:
Okay, PR #68 opened with @jacobsalmela changes for @hawson memory reductions.

jacobsalmela:
Cool.  I'll close this one then.  Thanks!

PR #45, opened 2015-09-18
Allow external IP as an input script parameter

jofrep:
If user add an IP as an input parameter this IP will be used when creating the host file instead of the one coming form hostname -I. This is useful for hosts with multiple IPs

Example: ./gravity.sh 123.123.123.123

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -5,8 +5,8 @@
 # This script should only be run after you have a static IP address set on the Pi
 piholeIP="$1"
 
-if [[ "$piholeIP" == "" ]]; then
-        piholeIP=$(hostname -I)    
+if [ -n "$piholeIP"]; then
+        piholeIP=$(hostname -I|xargs)        
 fi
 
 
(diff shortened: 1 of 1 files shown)


jacobsalmela:
It does, but if you use the `hostname -I` variable, it seems there is a space after the address, so if you `echo` it with quotes, there is a space (ASCII character 040):
```
pi@raspberrypi ~ $ echo "$piholeIP" | od -cb
0000000   1   9   2   .   1   6   8   .   1   .   1   3   8      \n
        061 071 062 056 061 066 070 056 061 056 061 063 070 040 012
0000017
pi@raspberrypi ~ $ echo $piholeIP | od -cb
0000000   1   9   2   .   1   6   8   .   1   .   1   3   8  \n
        061 071 062 056 061 066 070 056 061 056 061 063 070 012
0000016
```

So the additional space character would add result in two spaces in the final file.  It's probably not the biggest deal in the world, but it might as well be consistent.  

Also, the conditional on line 8 would be better suited using `if [[ -n "$piholeIP ]];then`

jofrep:
I got it now.

I've updated the change as follows:
if [ -n "$piholeIP"]; then
        piholeIP=$(hostname -I|xargs)        
fi

this trims the blank space from $hostname and ensures there is only one blank space

jacobsalmela:
Will you please use the double bracket notation?  There is also a space missing in yours.  Then I will merge it.

`if [[ -n "$piholeIP" ]];then`

jacobsalmela:
I merged and then manually [added the double bracket notation](https://github.com/jacobsalmela/pi-hole/commit/d52fdde2f8462ebca8f79610f7ee4fb2ae12342d).

PR #55, opened 2015-10-15
Setup cron job to run gravity.sh weekly

bigjay517:
This will auto setup the cron job that you mention on your blog.

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -78,5 +78,16 @@ sudo chmod 755 /usr/local/bin/chronometer.sh
 echo "Entering the event horizon..."
 sudo /usr/local/bin/gravity.sh
 
+echo "Autoupdate script installation..."
+crontab -l > /tmp/mycron
+if grep -qi '\<gravity\>' /tmp/mycron
+then
+   echo "Autoupdate already configured"
+else
+   echo "0 * * * * sudo /usr/local/bin/gravity.sh" >> /tmp/mycron
+   crontab /tmp/mycron
+fi
+rm /tmp/mycron
+
 echo "Restarting..."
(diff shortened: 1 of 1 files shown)


DonLexos:
This seems different than the method explained on the webpage here:
http://jacobsalmela.com/block-millions-ads-network-wide-with-a-raspberry-pi-hole-2-0/
Is this for the automated installer to do it during the setup?

It also puts the cron entry in a different place, it uses the crontab -e and ads
@weekly /usr/local/bin/gravity.sh

Is there a preferred (or "best") way of doing this?

jacobsalmela:
I have been trying to find a good way to [deploy a `crontab`](http://bencane.com/2011/06/28/the-safest-way-to-deploy-a-crontab/), but haven't decided on the best method.  I'll try out @bigjay517 's method and see how it works.

bigjay517:
I should have explained this better.  crontab -e opens a visual editor so it can not be used in the automated install.  Instead I readout the current crontab to a temp file, check if "gravity" exists, if not then echo the task into the file.  Then I overwrite the old crontab with this contents of this file and delete the temporary file.

I am not sure if there is a "best" way of doing this, this was my first attempt at automating this.

jacobsalmela:
I think that number 2 or 3 from the link provided previously would work best.  I want it to be able to work with the automated installer and the `-e` interactive won't work for that.

So the process would be to download a crontab file, import it, and remove the file.

bigjay517:
The changes I made are similar to number 3 from that link.  My changes take into account any existing entries on the crontab, which is why I read them out to the temp file first.

jacobsalmela:
This was added in #74.  It downloads a pre-made crontab file.

PR #58, opened 2015-10-22
add whitespace after $piholeIP

Giermann:
In my case there had been no whitespace in piholeIP (I configured it statically because 'hostname -I' returns multiple IP addresses, which ends in errornous hosts file), so it's more safe to add one between IP and hostname.

Further, I removed the obsolete latentBlacklist variable.

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -27,7 +27,6 @@ eventHorizon=pihole.3.eventHorizon.txt
 accretionDisc=pihole.4.accretionDisc.txt
 eyeOfTheNeedle=pihole.5.wormhole.txt
 blacklist=$piholeDir/blacklist.txt
-latentBlacklist=$origin/latentBlacklist.txt
 whitelist=$piholeDir/whitelist.txt
 latentWhitelist=$origin/latentWhitelist.txt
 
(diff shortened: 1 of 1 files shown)


jacobsalmela:
I will have to test this one thoroughly before merging since we had issues before.  I am also looking at finding a different way to get the IP.

Giermann:
If there had been issues (do double spaces really hurt?), would it be a solution to trim all whitespace surrounding $piholeIP?

Since hosts entries have to start with the IP address followed by a list of domain names, trimming could be done with cut extracting the first IP only: 'hostname -I | cut -d ' ' -f 1'

jacobsalmela:
Another issue is that some users have multiple IPs and/or IPv6 addresses as well.  In my experience, `dnsmasq` likes just one space between the IP and the domain otherwise it gets fussy.

jacobsalmela:
I think I would like to have two variables, one for the IPv4 address and the other for IPv6.  

I have been using this in some of the dev branches:

`piholeIPv4=$(ip -4 addr show | awk '{match($0,/[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+/); ip = substr($0,RSTART,RLENGTH); print ip}' | sed '/^\s*$/d' | grep -v "127.0.0.1" | tr -d '[[:space:]]')`

`piholeIPv6=$(ip addr show | awk '/scope\ global/ && /ff:fe/ {print $2}' | cut -d'/' -f1 | tr -d '[[:space:]]')`

Will you use those in your PR or see if you have a better way to extract them without spaces?

jacobsalmela:
If we continue to use  `hostname -I` I think we also need a little more logic if there are multiple IP addresses returned from that command.

jacobsalmela:
`dnsmasq` doesn't seem to care if there is more than one space, though.  It seems to be working fine in my test environments.

PR #66, opened 2015-11-02
Add configuration of interface during installation

cicoub13:
Facilite installation when multiple interfaces (wifi for example)
Tested


jacobsalmela:
Thanks for this.  But the new installer will have a [dialog box to select the interface you want to use](https://github.com/jacobsalmela/pi-hole/commit/e222a1d5e94ece6afa03bb751ccd37c39df0a71f).

cicoub13:
OK, great

Issue #53, opened 2015-10-13
Whitelist editor in the web interface


beukhof1:
It would be nice and easy to have a whitelist editor on the web interface!

DonLexos:
You would still need to run the gravity.sh script after you have added items to the whitelist I believe because from what I can tell it collects all lists (including the whitelist) and creates one master list that is actually used by dnsmasq.

I believe @jacobsalmela is already working on a more sophisticated admin panel that (I am guessing here) would have more features and statistics than the current one does.

beukhof1:
Maybe when u hit save on the whitelist it also automaticly runs gravity.sh?

DonLexos:
Well depending on how sophisticated you want it to be you could provide the option after the ability to save the whitelist (so box with the list, save underneath, and on the result page options to return to settings (maybe you want to do something else) or "apply all updates now and run gravity.sh").

Not sure if @jacobsalmela already has an idea of what kind of things he wants/needs/woulds/shoulds have in the interface. Right now it's all stats (and even there there are many things that can be added).

jacobsalmela:
Believe it or not, I actually have a [whitelist script](https://github.com/jacobsalmela/pi-hole/blob/master/advanced/Scripts/whitelist.sh) but it's been a while since I updated it, so it probably doesn't work all that well.

I do plan to have more stats/graphs in the interface.  But first, I want to put things like update, whitelist/blacklist, restart, check service status, all from the interface.  Progress, as usual, is slow by today's standards since I have a full time job and do this in my free time.  But there are things in the works!

jacobsalmela:
As a start, I am just adding a whitelist field on the left sidebar.  Eventually, this will turn into a full-fledged editor that lets you see what is whitelisted and sites you may want to.  But this should satisfice this highly-requested feature.
![pihole whitelist](https://cloud.githubusercontent.com/assets/3843505/11667134/92f8db40-9db5-11e5-9744-48b80dea574a.png)

jacobsalmela:
I haven't forgotten about this one; [the editor looks pretty sweet](https://github.com/jacobsalmela/AdminLTE/pull/9).  I need to fix the bugs first though as they take priority.

Mcat12:
This feature is now implemented in the devel branch of the [web interface](https://github.com/jacobsalmela/AdminLTE/tree/devel).

PromoFaux:
@Mcat12 would you be annoyed if I changed the whitelist.txt content?

[...]

PR #67, opened 2015-11-04
Update gravity.sh sudo errors as root

meliton:
When gravity.sh is run as the root user, it errors out on the sudo command.
These changes make the script check to see if it is invoked by the root and sets the sudo variable to handle root and non-root runs.


jacobsalmela:
Nice.  I was hoping someone would figure this out.  I'm behind on the PRs, but I'll do my best to catch up.

jacobsalmela:
What OS are you running the Pi-hole on?

meliton:
I am running on both Raspian Jessie and VM'd Debian Jessie. This change doesn't effect Raspian since sudo is installed and as root, with or without the sudo command, the script runs fine. As non-root, Raspian needs sudo for the script to run. So it's basically a sanity check on the script that says if you're not root, use sudo. VM'd Jessie throws sudo errors since sudo is not installed.

jacobsalmela:
Will you resolve the conflicts?  We recently revamped the installer and the gravity script.  Thanks!

meliton:
Wow. Lot's of changes. Conflicts are fixed.

dschaper:
I think a lot of this could be done as a regular user. Is there a need to have the pihole directory protected? The directory could be created as the default pi user and the files copied without su privileges. That would eliminate most of the sudo calls, except for the apt-get's. Possibly a separate pihole user could be created for this process, and the cronjobs run as the separate user without priv's. (I'm thinking about the packaging, an adduser call would be a simple thing to do...)

meliton:
The idea was to run the install and gravity scripts on the Pi and a stock Debian Jessie machine. Debian Jessie doesn't have sudo installed by default, so there is a whole mess of adding sudo, then users to sudoers for the scripts to run cleanly.

dschaper:
Debian Jessie or Raspbian Jessie? The headless (no-x) version of Raspbian has sudo and I think the full NOOBS version does also (I'm not sure on that, I haven't tried the full version yet, I either run headless or raspi-ua-netinstall if I'm running a linux distro... Doing FreeBSD on one right now and headless on my spare 2B to test the deb building...)

meliton:
Actually, I'm working with both Debian Jessie and Raspian (no-x). The script runs cleanly on the Pi and on Debian Jessie. I'm in the process of fixing some cosmetic issues (line spacing) with the installer that I missed earlier. It will probably run on FreeBSD or any other comparable distro since the fixes attempt to resolve root/sudo issues.

dschaper:
There are a couple of issues that keep it from BSD, one of them is sourcing out the IP address. Hostname -I isn't available to BSD.

jacobsalmela:
- `hostname -I` is no longer used
- @meliton is correct that since it's inception, the Pi-hole was meant to be installed on a clean version of Raspbian and run on the Pi, but people have already ported it to other OSes.
- I tried to follow the Filesystem Hierarchy Standard:
  - `/etc` *should contain static configuration files and may not contain binaries*
    -`dnamasq.conf` and `lighttpd.conf

dschaper:
Everything is according to FHS, so that's not a problem. I was just wondering if you had a specific reason to keep the files under su privilege, like keeping the user from editing the lists without being root. If not, then there is a lot that can be pulled from sudo and done as either the user or as another non-root user. I don't know how @meliton is thinking but my way of thinking is to keep as much away from root as possible since it has full control. I know it's just the pi, but in practice I try to keep sudo only for absolute needs. From other projects I've worked I know that the more seasoned linux users want to keep away from sudo and root. 

Just my thoughts, I can work on seeing how far we can push before root is needed and pull those bits of code from sudo that can be run as the default pi user. Maybe refactor out some more to functions. And at the same time get towards more portability. I'll have to see what comes stock and what tools we have at our disposal. Maybe we could integrate https://github.com/debian-pi/raspbian-ua-netinst in the future and have a one-shot build and make a distribution out of the project so a user could image an SD-Card with the distribution and pi-hole all ready to go. Mel is working on a network discovery application and that could all tie in together. Might be looking to the future a bit, but I think that would be cool for you to have your own distro, don't you think?

dschaper:
Question: It seems like the sudo's could be removed from the gravity script, but there are a couple of files that are deleted as part of the process, (latentWhiteList) and files that are created in the pihole directory. For fresh users this isn't an issue, but for users that have already run the script, taking out the sudo's blocks them from copy/delete. Would it be a good idea to check the owner of the directory and chmod it to the user that is running the script as part of the process?

meliton:
That's a good point. I was suggesting changing the script to find out what privileges the user has before continuing the script since 'root' doesn't need sudo but 'pi' does. Since I was also testing this on a stock Debian Jessie VM, I discovered that sudo is not installed and it was throwing a bunch of errors because of it. So, strictly on a Raspberry Pi, you don't need the $SUDO checks BUT adding them lets the script play nice on both the Pi and the VM. Chmodding might be a good way to go IF the user even has privileges to run all the options in the script.

dschaper:
I'll try to target the config for a stock Raspi install first. I just wanted to make sure that I wasn't duplicating efforts and doing something that you were already working on @meliton. Gravity.Sh itself really doesn't need any sudo, the install script would as it is doing apt-get pulls. This is mostly transitional work, once the development is stable enough to try to package as a distro all the files would be (in concept) installed already before the user even logged in for the first time. I'll get to work on it, and I'd like to try to refactor the code some more into functions. I like the way @jacobsalmela was headed with the function concept...

jacobsalmela:
I would also like to make some more functions out of the existing code.  

And as far as the `sudo`s go, it would be nice to not use it and run it under local privileges or run everything more like a daemon and as root.  I would like it to be as portable as possible to other *nix systems.  Root is on all of them...

I'm not sure what's best, so this is a good discussion.

dschaper:
What do you think about asking the user during install, defaulting to the UID/GID of the user running the install script? Then it's a matter of chmod'ing the applicable files to the requested user (or doing the equivalent during the install of the package). The scripts and the cronjobs would be run as the user. So it would be like a daemon with dropped priv's.

meliton:
@dschaper has a point about least privileged installation. If the only thing the Pi is doing is ad-blocking, then you can install with root/sudo and then tweak down the privileges knowing that the installation 'took hold.' If you port over to other *nix distros, the users might be doing other things in addition to ad-blocking on their boxes. So installing as root on those systems could be dangerous. Ok, maybe not 'dangerous', but definitely frowned upon.

jacobsalmela:
So we need some sort of mechanism to get the UID/GID and then `chmod` all the appropriate files?  What about multi-user systems?

I am also wondering if you can stop/start services when you are not root.  `dnsmasq` needs to reload to use any new domains added to the list.

dschaper:
I think the easiest solution would be to create a pihole user and run under that user. (I use https://github.com/mutability/dump1090 as an example, it's an SDR decoder daemon, and it runs under a dump1090 user for it's functions.) Some functions would need to be run as root, but you sudo when you apt-get, so installing as a package would take care of that, the scripts run during the installation can be run with root capabilities, and would need to in order to stop/start and to copy the scripts to the proper directories. As far as standalone installation, there is going to be a need to have su rights for some of the install script functions, there's no getting around that, but gravity.sh doesn't need to run as root, and the install script can chmod installed files to change to an unprivileged user account. I do suggest that this be a user that is added by adduser/useradd during install, but you could give the user a prompt to ask what username to run as, and then create that user if it doesn't exist already. That's how dump1090 functions...

jacobsalmela:
What about the `nobody` user?  As I understand it, it is for running things that don't need special permissions, so it can run as that user, but regular users could also modify the files.

dschaper:
I'll have to do some more research on what `nobody` can do. From what I've read, nobody does not have the ability to own files, or at least shouldn't own any files, so we may be back with the problem of who owns the files under /etc/pihole? I'll need to read up some more but that may be a good solution.

jacobsalmela:
Yeah, I haven't read much into it yet either.  I'm sure we will come up with something that will work.

dschaper:
I still think creating a `pihole` user would be the best route, if we were going to make the changes necessary for `nobody` then we probably should either create `pihole` user and group and run from that. Or ask the user for a name to run under, but there would need to be a check to make sure the username is a valid linux username.

jacobsalmela:
Yeah, I was just coming to that same conclusion.  We will still need root access to make the user, but then we could have all the files writeable by the local users but owned by the `pihole` user.

dschaper:
But that would just be one `sudo` root call, and would pull `sudo` completely out of `gravity.sh`. I don't know if the files would even need to be world writeable, the cronjob could call `gravity

jacobsalmela:
Good point.  `sudo` would then only be needed for the automated install.

dschaper:
And the packaging would eliminate that call in it's version, since the packaging runs as root when unwrapping and running the install scripting.

jacobsalmela:
Cool.  Sounds like we are on the right track!

jacobsalmela:
I would like to work on this next as I have the #53 feature almost ready to go and getting rid of `sudo` would help make the script execute a lot easier.

dschaper:
Okay, that I can put a priority on. I have the new user in place, I'm just checking for possible regression errors and handling already installed setups to make sure it doesn't create conflicts.

jacobsalmela:
Awesome!

dschaper:
Which features of the dropped root priv's would help the most, writable directories, write/read files? Some pieces are easier than others, and some pieces will still require root because of where the files are being placed, but if I knew a little more how to help you out I can spend the time there... You have my email address if you need to go off-github. (Maybe I deleted it, but it's dschaper@ganymeade.com) Getting the dang deb working per spec would help because so much of that is taken care of by the debhelper scripts...

dschaper:
Is @meliton still inolved in this PR? Would like to have his input too if he has some...

meliton:
Sorry about the late reply. I saw the discussion on `nobody` and `pihole` users. It looks like if the installer handles root issues, then `sudo` is not an issue in the gravity script. Once that happens, it should be safe to close this PR.

jacobsalmela:
I think @meliton's commit is on the right track: checking the `$EUID` and seeing is `sudo` is installed or not.  I'm going to proceed on that route.  During the install, I will make a pihole user and `chown` the necessary files.

jacobsalmela:
@joshkastang's method from #153 seems to be the most simple solution:

```
if [[ $EUID -ne 0 ]]; then
    sudo bash "$0" "$@"
    exit $?
fi
```
So I'm closing this in favor of that.

PR #68, opened 2015-11-06
Hawson reduce memory

dschaper:


jacobsalmela:
My initial test seemed to work without a problem.  I will do some more before merging.

But will you please fix something:

Lines 118 and 127 do not remove the filename from the output so when you run the script, it prints the number and filepath, which ends up not making sense.

`303447 /etc/pihole/pihole.3.eventHorizon.txt unique domains trapped in the event horizon.`

dschaper:
Fixed wc,,,

jacobsalmela:
It seems to work fine.  Thanks!

PR #73, opened 2015-11-15
Dialog-based installer

jacobsalmela:
In an effort to make the Pi-hole easier to install, I have created an installer that leads the user through the setup.  Users will no longer need to set a static IP address manually before running the installer.  This simplifies the process:
1. Install Raspbian onto your SD card
2. Run this command: `curl -L install.pi-hole.net | bash`
3. Follow the dialogs to finish the installation

This enhancement will also pave the way for IPv6 support as well as the historical stats database.

pi-hole-instal2

pi-hole-instal-ip2

pi-hole-instal-ip-verify


dschaper:
Almost looks like debconf, have you thought about making this into a .deb? I have a spare B2 I could look at setting this up for packaging and PR the whole thing...

jacobsalmela:
I haven't yet, but that would be very cool!

Also, I need to figure out how to resolve the conflicts (I really have no idea how to use Github 😅)

dschaper:
https://github.com/blog/2083-start-learning-git-and-github-today-with-self-paced-training

It's a pretty good (and free for now) overview of git and github.

jacobsalmela:
Cool.  Thanks!

jacobsalmela:
Fixed by #74

PR #74, opened 2015-11-15
Cherry pick from PR73

dschaper:
Cherry picked commits from #73, the two problem commits are 5d66f0e and a6a114a. Both are commits to gravity.sh. I'll take a look at them and see what is conflicting, but in the mean time the install script commits are good to go.

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -174,7 +174,7 @@ commands["Cleaning up the dashboard temp files"]="sudo rm /var/www/master.zip 2>
 commands["Creating a log file for the Pi-hole"]="sudo touch /var/log/pihole.log"; echoes+=( "Creating a log file for the Pi-hole" )
 commands["chmodding the log file"]="sudo chmod 644 /var/log/pihole.log"; echoes+=( "chmodding the log file" )
 commands["chowning the log file so stats can be displayed"]="sudo chown dnsmasq:root /var/log/pihole.log"; echoes+=( "chowning the log file so stats can be displayed" )
-commands["Initating sub-space transport of gravity"]="sudo curl -o /usr/local/bin/gravity.sh https://raw.githubusercontent.com/jacobsalmela/pi-hole/master/gravity.sh"; echoes+=( "Initating sub-space transport of gravity" )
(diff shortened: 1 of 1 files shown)


dschaper:
5d66f0e looks like it re-introduced a bunch of code from the old swap file process. I'll try to fix that up...

jacobsalmela:
Well at least someone knows what they are doing (and it isn't me!)

dschaper:
Just make sure you double check what I did, I might not know as much as you think I do!

jacobsalmela:
I just tested it a few times and it seemed to work as expected.

dschaper:
Good, what's next?

jacobsalmela:
Now that the installer is prettier and more adaptive.  My focus is on the Web interface (adding whitelist editor, advanced stats, domain finder, etc.)  

I would also like to parse down the ABP EasyLists to get the domains (but it might be worthwhile to check if these have domains that we already have.

dschaper:
Okay, I'll take a go at the packaging then, get this into a deb that can be pulled via apt-get. (And try to port this over to FreeBSD, I'm running that on the gateway/VPN Pi...)

DonLexos:
I checked the new files, the new cron seems to mention a file that is not there yet (to collect the daily stats). I am aware the admin interface will not show them but right now only the new daily purge script is there.

dschaper:
Hope I'm not barging in and taking over things, that's not what I meant to do...

jacobsalmela:
I appreciate any help that is offered, so no worries.  

@DonLexos You found my hint!  The script isn't done yet, but eventually, it will update the Web interface with historical stats.  I just put in there now for testing it on my end.
screen shot 2015-11-15 at 16 01 59

DonLexos:
Ah ok, since I wanted to merge some of the changes in (and not do a full reinstall from scratch) I checked the differences. I do your daily cleanup by hand once a week or so (not that much traffic) but noticed that extra line in the new cron file ;-)
New admin looks good! Can't wait :+1:

meliton:
I'm wondering if the web interface can be used to configure the Pi-hole.

jacobsalmela:
Not yet, but it's in the works!

dschaper:
Thanks for understanding! I'm more than happy to help out in any way I can...

PR #79, opened 2015-11-19
FHS Location Changes

dschaper:
Per Debian Maintainer's Guide:
"Most third-party software installs itself in the /usr/local directory hierarchy. On Debian this is reserved for private use by the system administrator, so packages must not use directories such as /usr/local/bin but should instead use system directories such as /usr/bin, obeying the Filesystem Hierarchy Standard (FHS)."

https://www.debian.org/doc/manuals/maint-guide/modify.en.html

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -224,15 +224,15 @@ sudo rm /var/www/master.zip 2>/dev/null
 sudo touch /var/log/pihole.log
 sudo chmod 644 /var/log/pihole.log
 sudo chown dnsmasq:root /var/log/pihole.log
-sudo curl -o /usr/local/bin/gravity.sh https://raw.githubusercontent.com/jacobsalmela/pi-hole/master/gravity.sh
-sudo curl -o /usr/local/bin/chronometer.sh https://raw.githubusercontent.com/jacobsalmela/pi-hole/master/advanced/Scripts/chronometer.sh
-sudo curl -o /usr/local/bin/whitelist.sh https://raw.githubusercontent.com/jacobsalmela/pi-hole/master/advanced/Scripts/whitelist.sh
-sudo curl -o /usr/local/bin/piholeLogFlush.sh https://raw.githubusercontent.com/jacobsalmela/pi-hole/master/advanced/Scripts/piholeLogFlush.sh
-sudo chmod 755 /usr/local/bin/gravity.sh
-sudo chmod 755 /usr/local/bin/chronometer.sh
-sudo chmod 755 /usr/local/bin/whitelist.sh
-sudo chmod 755 /usr/local/bin/piholeLogFlush.sh
-sudo /usr/local/bin/gravity.sh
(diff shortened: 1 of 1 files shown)


jacobsalmela:
Interesting.  I'll have to read more into that.  Also, I will have to update a lot of my documentation on pi-hole.net when this pull is merged.

dschaper:
Yeah, it's under section 3.3. It's kind of a different take than what I assumed to be cannon.

jacobsalmela:
Same here.  Good to keep learning though!

dschaper:
Do you still use TOIlet or bc? I didn't see where they were still needed?

jacobsalmela:
Yes, it is used with `chronometer.sh`, the real-time mini LCD display script.  There is probably a small subset of people actually using it, but it's nice to have anyway.

jacobsalmela:
I think I will merge #85 before this as the file path change will take me a lot longer to update all of my documentation.  Plus, having the functions will be easier to work with in the future.

jacobsalmela:
Following the links through the one you provided, it points to [here](https://www.debian.org/doc/packaging-manuals/fhs/fhs-2.3.html#USRBINMOSTUSERCOMMANDS), which reads:

_The /usr/local hierarchy is for use by the system administrator when installing software locally. It needs to be safe from being overwritten when the system software is updated. It may be used for programs and data that are shareable amongst a group of hosts, but not found in /usr._

That makes me feel like it is in the right place as-is.

dschaper:
It's kind of a catch-22, as pointed out here http://unix.stackexchange.com/questions/8656/usr-bin-vs-usr-local-bin-on-linux /usr/bin would be for the packaged version, while /usr/local/bin would be for the version that is pulled directly. The packaged version would need to go in /usr/bin per the Debian Policy. Not sure how to handle this one.

dschaper:
9.1.2 Site-specific programs

As mandated by the FHS, packages must not place any files in /usr/local, either by putting them in the file system archive to be unpacked by dpkg or by manipulating them in their maintainer scripts.

jacobsalmela:
Hmm, yeah that's weird.  So once we get the `apt-get` version going, that would go into `/usr/bin` but since user's are just downloading the files and adding them manually (semi-automatically using the automated installer), they can stay in `/usr/local/bin`?

dschaper:
That's pretty much what I get from the documentation. The packaging would take care of moving to `/usr/bin`. I don't know how that would affect the documentation though, having two different install locations. I think you can probably close this one out and leave things as is while I'm sorting out the packaging.

jacobsalmela:
Sounds good.  The documentation part will be a challenge since I have so many posts out there on it, but that will be my own fun endeavor to undertake.  =)

dschaper:
Ah, the fun of documentation! I've never used GitHub's Wiki, maybe that would be a way to get some help, write up a framework in the Wiki and get the end users to fill the details.

jacobsalmela:
That's probably a good idea.  I added the Wiki when making the project for that purpose, but I made most of the posts on my blog and pi-hole.net.  It might be a good place to centralize the documentation for people.

PR #85, opened 2015-11-23
Refactored gravity.sh

dschaper:
Gravity.sh with refactored code.

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -210,6 +210,7 @@ function gravity_advanced() {
         sudo kill -HUP $(pidof dnsmasq)
 }
 
+gravity_collapse
 gravity_spinup
 gravity_Schwarzchild
 gravity_pulsar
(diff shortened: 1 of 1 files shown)


jacobsalmela:
Haha, this is my favorite PR.  I love that you went the extra mile to fit the theme; your schwarzschild function made me laugh.

I will do some testing to verify everything before merging.

jacobsalmela:
This seems to run a lot faster now.  Even on my B+.  I just have a few more tests to run.

dschaper:
Good, fix for the persistent domain (maha) in #86 should help with resolving #89.

dschaper:
Can you tag a new release for this and create a tarball? It would help on setting a base for the packaging. Debian is kind of harsh on their requirements. (The linter throws a notification that it doesn't like having .sh because it's possible the scripting language may change in the future and the extension would be invalid?!)

PR #90, opened 2015-11-25
Pull IP info from outgoing interface

dschaper:


dschaper:
Use IP to get interface information from outgoing interface. Use CIDR notation to make things easier.

dschaper:
Using IP address replace we add new secondary IP address that is the custom IP requested by user. Pi will respond to both IP's so no reboot is needed. At the next reboot the Pi will have 1 ip address. May wan't to modify the exit screen code to alert user that a reboot would be good, but it's not necessary?

jacobsalmela:
The changes work well, and I will change the final dialog as mentioned.

Issue #86, opened 2015-11-23
Domains from saved lists still get applied even when removed from the sources array


jacobsalmela:
This only happens in a few circumstances, but as an example, say you used the mahakala list at one point and then decided you did not want it anymore so you either comment it out or remove it from the `sources` array and then run `gravity.sh` to apply the changes.  

This doesn't work because the `list.n.adblock.mahakala.is.domains` file already exists in `/etc/pihole/` from when the script ran previously.  So when the `find` command searches for any `.domains` file to use to create the aggregated list, it will still apply the mahakala list because that file is still there.  

So some sort of logic needs to exist that verifies that the domains in the `sources` array match the files that exist in `/etc/pihole/` so only the correct ones apply.

Another instance when this happened is when a user added multiple new sources and then decided not to use them, but since the files were still there, they were applied anyway.

dschaper:
Fixed, confirmed by commenting out domains and noting the decrease in number of blocked domains reported. Domains restored when uncommented.

jacobsalmela:
Wow!  Nice, thanks.  It's getting hard to keep up with you!  But that's a good thing.  I'll have some time over thanksgiving to test everything and try to merge it all without any conflicts.

dschaper:
Sounds good!

PR #94, opened 2015-11-26
Bug Fix - Cleanup remnant files

dschaper:
Fix #93 and modify #86. Extra list files now removed when domain is removed.


jacobsalmela:
Cool.  It works well!  Thanks.

PR #96, opened 2015-11-27
Feature/atomic functions

dschaper:
Misc code formatting cleanup and a few thematic changes to fit output on one line.

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -145,7 +145,7 @@ function gravity_Schwarzchild() {
 echo "** Aggregating list of domains..."
 truncate -s 0 $piholeDir/$matter
 for i in "${activeDomains[@]}"
-do
+do	
    cat $i |tr -d '\r' >> $piholeDir/$matter
 done
 
@@ -199,6 +199,16 @@ function gravity_hostFormat() {
         # Copy the file over as /etc/pihole/gravity.list so dnsmasq can use it
         cp $piholeDir/$accretionDisc $adList
 }
+function gravity_blackbody() {
+		for file in $piholeDir/*.$justDomainsExtension
+		do
+			if [[ " ${activeDomains[@]} " =~ " ${file} " ]]; then
+				:
+			else
+				rm -f $file
+			fi
+		done
+}
 function gravity_advanced() {
 
         # Remove comments and print only the domain name
@@ -221,4 +231,4 @@ gravity_Schwarzchild
 gravity_pulsar
 gravity_hostFormat
 gravity_advanced
-
+gravity_blackbody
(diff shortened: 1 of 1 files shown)


jacobsalmela:
So is #94 separate from this?  It looks similar.

dschaper:
#94 Is separate, it has a fix for the merge conflicts that was preventing the stuff from yesterday that was merged into the development branch from merging to master. This PR has some other feature updates, if you merge #94 then a lot of the commits listed here should be cleared.

jacobsalmela:
OK.  I plan to merge it today--I just want to do a bit more testing on my end.  Most of the comments I get are that the Pi-hole is very stable and I want to keep it that way.  Thanks for all this hard work!!

dschaper:
No problem, that's why I'm trying to break stuff out to bug branches or feature branches. What's the best way for you, to PR to development?

dschaper:
And now I'm seeing some problems with Maha, it returns different lists with each curl. Sometimes a small list, sometimes a huge list, sometimes no list at all. All with the same curl...?

jacobsalmela:
Yeah, for big changes PRs to the development branch is best.  If it's just small stuff that I can look at quickly, PRs to master is better for a faster merge.

jacobsalmela:
The mahakala list is a weird one...it needs to have be referred from that certain URL.  It's really called the Mother of All Ad Blocks:

http://forum.xda-developers.com/showthread.php?t=1916098

dschaper:
Apparently that's not all it blocks! ;)
\

dschaper:
I found another list, http://www.malwaredomains.com/ but I'm not sure if you just want to target ads, or would confirmed malware spouts be part of what you'd like to add? (They don't have a hosts format, so there would be some massaging the list to get it to PiHole format.)

jacobsalmela:
Is that different than the [malwaredomains that is already in the list](https://github.com/jacobsalmela/pi-hole/blob/master/gravity.sh#L22)?

I also have noticed inconsistent numbers with the mahakala list, which is weird.

dschaper:
Oh, it's the same, nevermind!

jacobsalmela:
OK, so this PR looks like it is mostly some formatting and then the `piholeIP` change, yes?

dschaper:
Yeah, and I moved the blacklisting from Scwartz to pulsar, to keep each function doing only one thing. Pulsar now just handles black and white listing (hence pulsar) and Schwartz only collapses all the individual files into a single list.

jacobsalmela:
OK, cool.  Thanks!  It looks good.

PR #137, opened 2015-12-28
Fix typo in installer

iblamefish:
IPv4 was mentioned in an IPv6 setting - changed to correctly reference IPv6.


jacobsalmela:
Thanks!

PR #141, opened 2016-01-01
Fix bug in gravity.sh where IPv6 list was not always populated

iblamefish:
Failed to be populated if /tmp/piholeIP does not exist. This may happen under
the following conditions:

- was not generated during install
- gravity.sh being run again after the original install (the file is deleted
  during the installation process)


jacobsalmela:
Good catch, thanks!

PR #143, opened 2016-01-02
fixes #142 kill -HUP erroring on gravity_reload if dnsmasq is not running

### LINKED ISSUE #142: gravity_reload() kill -HUP errors if no dnsmasq process is running.
kill -HUP errors out with `kill: invalid argument H` if dnsmasq is not running. Need to detect if `pidof dnsmasq` returns false then start dnsmasq instead of kill hup'ing it.

iblamefish:


jacobsalmela:
Nice!  Thanks!

PR #146, opened 2016-01-02
installer - only select first (probably default) interface in list instead of all

iblamefish:
For https://github.com/jacobsalmela/pi-hole/issues/138

99% sure that this works in general. Works for me, but could do with being tested on a few other devices just to be safe.


jacobsalmela:
Nice.  I knew some sort of increment variable was needed...just couldn't get the logic.

Thanks!!

PR #153, opened 2016-01-10
Automated Scripts Cleanup

JoshKastang:
The automated installation/removal scripts have some room for optimizations and improvements. This is the first wave of proposed changes in this PR:

* Install script now installs pihole crons to /etc/cron.d/pihole instead of /etc/crontab (which should never be touched by an automated process). The uninstall script takes this into account and also attempts to provide backwards comparability when uninstalling pihole.
*  Uninstall script now gracefully handles /var/www/html removal. Its not safe to assume when uninstalling that the entire /var/www/html directory should be removed.
* Removed sudo commands from install and uninstall scripts - script now requires a user with root permissions to run. Will receive "ERROR: You must run this script as a root user" message when running either as a non-root user.

### DIFF
--- a/advanced/pihole.cron
+++ b/advanced/pihole.cron
@@ -1,17 +1,3 @@
-# /etc/crontab: system-wide crontab
-# Unlike any other crontab you don't have to run the `crontab'
-# command to install the new version when you edit this file
-# and files in /etc/cron.d. These files also have username fields,
-# that none of the other crontabs do.
-
-SHELL=/bin/sh
-PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
-
-# m h dom mon dow user  command
-17 *    * * *   root    cd / && run-parts --report /etc/cron.hourly
-25 6    * * *   root    test -x /usr/sbin/anacron || ( cd / && run-parts --report /etc/cron.daily )
-47 6    * * 7   root    test -x /usr/sbin/anacron || ( cd / && run-parts --report /etc/cron.weekly )
-52 6    1 * *   root    test -x /usr/sbin/anacron || ( cd / && run-parts --report /etc/cron.monthly )
 
 # Pi-hole: Update the ad sources once a week on Sunday at 01:59
 #          Download any updates from the ad lists
@@ -24,3 +10,4 @@ PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
 # Pi-hole: Flush the log daily at 11:58 so it doesn't get out of control
 #          Stats will be viewable in the Web interface thanks to the cron job above
 58 11   * * *   root    /usr/local/bin/piholeLogFlush.sh
+
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -16,6 +16,12 @@
 #
 # curl -L install.pi-hole.net | bash
 
+# Must be root to install
+if [[ $EUID -ne 0 ]]; then
+   echo "ERROR: You must run this script as a root user"
+   exit 1
+fi
+
 ######## VARIABLES #########
 tmpLog=/tmp/pihole-install.log
 instalLogLoc=/etc/pihole/install.log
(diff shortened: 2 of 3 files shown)


jacobsalmela:
The cron.d looks good.  

I also like the better `/var/www/` removal, however, there is [a typo](https://github.com/JoshKastang/pi-hole/commit/cb401589102f226ebf316413902e9c75f3e8d01a#diff-618a200ff1dc21ccf1494ad7b86b8b9aR26)

As for the `sudo` issue, see #67.  I'm looking to provide use for both.  For most users, they will be running the command as `pi` and the install would fail immediately.  They could just as easily prepend `sudo` to the `curl` command, but I'd rather not have them have to do that.

JoshKastang:
Wow that was fast!

The type was corrected in 362fe62. 

As for the sudo issue - good point. Since sudo with the pi user doesn't require a password by defaut:

```
# cat /etc/sudoers | grep pi
pi ALL=(ALL) NOPASSWD: ALL
```
How about auto-elevation? If running as `pi`, it will be transparent to the user. If running as another user, they will be prompted for a password.

Something like this should work, ill test and update the PR
```
if [[ $EUID -ne 0 ]]; then
    sudo "$0" "$@"
    exit $?
fi

```

jacobsalmela:
[I try to be my best](http://dollhouse.wikia.com/wiki/Self-improvement).  

So that snippet means, if you are not root, use sudo to run the script (`"$0"`) with any arguments (`"$@"`)?  There shouldn't be any arguments passed to the script as is just runs as-is.

JoshKastang:
Thats correct. There are a few different ways of handling self elevation, but in my experience this is the most transparent to the user. 

Here would be the different use cases:

1. **`pi` user executes `bash basic-install.sh` (which should be equivalent to `curl -L install.pi-hole.net | bash`).**
In this case, the script self elevates and installs/uninstalls as expected.

Example:
```
$ whoami
pi

$ bash uninstall.sh 
Reading package lists... Done
Building dependency tree       
...
```

2. **User executes `bash basic-install.sh` as a user other then `pi`. Assuming the account is in the sudo group (or the user is added in `/etc/sudoers`), the following will occur:**

```
$ whoami
kastang

$ bash uninstall.sh 
[sudo] password for kastang: 
Reading package lists... Done
Building dependency tree       
```

jacobsalmela:
Cool.  I'll try it out once you have it set up.

JoshKastang:
Sounds good! I pushed the commit at basically the same time you submitted your last comment.

jacobsalmela:
Well look at that.  =)

jacobsalmela:
While trying to download your script via `curl` as `pi`, I get the error:
```
/bin/bash: /bin/bash: cannot execute binary file
```
Like it's trying to use the shell twice or something.

First Google search showed [this](https://github.com/docker/toolbox/issues/199).

JoshKastang:
It looks like there is a difference between executing `bash basic-install.sh` and `curl -L <url> | bash` which, in hindsight makes absolute sense.

I created a [test gist](https://gist.github.com/JoshKastang/54c0bb4f2538f7bb3dfc) to mimic the functionality, and you're correct:

```
$ curl -L https://gist.githubusercontent.com/JoshKastang/54c0bb4f2538f7bb3dfc/raw/349c4d22a17635f2d7de32e961da87505fcfd0e3/self-elevation-test.sh | bash
100   117  100   117    0     0    434      0 --:--:-- --:--:-- --:--:--   434
$0: bash
$@: 
/bin/bash: /bin/bash: cannot execute binary file
```

As you can see, when attempting self-elevation via `curl -L <url> | bash`, $0 is considered bash, which is no good - the filename nor the url is preserved after being piped to bash.

I think the most graceful approach would be to change the install command to:

```
curl -L <url> | sudo bash
```
but I can understand why that is not desirable as well.

Let me think on some other alternatives as well. I mean, worse-case scenario we could do something like:

```
if [[ $EUID -ne 0  ]]; then
    curl -L install.pi-hole.net | sudo bash
    exit $?
fi
```

But that has some pros and cons to it as well - users installing via the curl method would be fine, but users executing via `bash basic-install.sh` may not receive the intended version of their script (especially users who are testing new branches).

A happy medium may be something like the following (not tested, but if you agree I can modify the scripts):
```
if [[ $EUID -ne 0  ]]; then
    if [[ $0 == "bash" ]]; then
        curl -L install.pi-hole.net | sudo bash
    else
        sudo bash "$0" "$@"
    fi
    exit $?
fi
```

jacobsalmela:
One thing we looked at was putting `sudo` in a `$SUDO` variable if the user was not root.  Then, the variable would just be empty so the commands would execute either way...

I'll think on it some more, too.

`curl`ing and piping scripts into bash is typically frowned upon, so visibly showing `sudo` might turn off more users.  I totally get that.  Homebrew does it and opinions go bothways.  

@dschaper is working on a .deb so that could be an alternate method at some point, or even replace the `curl` install command.

dschaper:
Yeah, when I can get it working right, (Debian politics aside) a lot of the issues would become non-existent. Apt-get is pretty much always run as root, so the elevations issues would be gone. It's taking a whole lot longer than I anticipated (and a lot longer than I would like to be taking) but there is a lot of ground work to cover if you want the package to be a "non-native" package that would be install-able on other distros. So it's a bit of short term PITA for long term stability and gain. 

Along with the deb process, we were going over the creation of the PiHole user to run the scripts under anyways, and a few other odds and ends to tie up. If you ever want to cure insomnia, read through the Debian Policy Manual a few times!

Plus real life is always a fun thing to have suddenly take precedence, lol...

jacobsalmela:
@dschaper You are sure committed!  

I'm in a similar boat handling questions and making sure the project stays alive, while trying to keep my real full-time job and life going.

I'm sure we will figure out the solution at some point @JoshKastang

dschaper:
Lol, it's a good idea and a good project. I'm interested to see how the new 'pay for play' aspect works in, I think that's a great idea and it actually gets the revenue to the right place and not to the aggregators that eat up so much of the funds. I like the idea that I can pay to support the sites I find useful and help them out.

PR #157, opened 2016-01-10
Sub space transport fix

miketedesco:
This bit me so I wanted to fix it.

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -39,7 +39,7 @@ dhcpcdFile=/etc/dhcpcd.conf
 ####### FUCNTIONS ##########
 backupLegacyPihole()
 {
-if [[ -f /etc/dnsmasq.d/adList.conf ]];then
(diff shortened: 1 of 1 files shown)


jacobsalmela:
Sorry about that.  However, if you simply check if `/etc/pihole` is there, it doesn't mean it is the original Pi-hole.  The `adList.conf` was really the only unique file that does not exist in the new version.  

So your code would run every time for current users.

miketedesco:
I just started using this app a month ago.  As I was looking around here it seemed that you it was redesigned.   Can you give me a list of files for the new version you would like to back up?  I will fix it right up if that cool with you.

jacobsalmela:
The function you are editing already backs up all the necessary files.  Likely, there won't be many users who want or need these files.  I thought of even abandoning the function all together since the 2.0 is so much better.

miketedesco:
Cool

Issue #139, opened 2015-12-29
Use whitelist.sh instead of built-in function


jacobsalmela:
The last update to `whitelist.sh` works really well.  I would like run this script within `gravity.sh` and eliminate the function.  They work essentially the same, but it will help simplify the main script and also reduce its complexity.

Mcat12:
How easy would it be to add a blacklist script?

jacobsalmela:
Easy enough.  You basically just need to append a domain to the list and then restart `dnsmasq`.

Mcat12:
I have no real experience in bash. Would you be able to create a blacklist.sh?

jacobsalmela:
Yes.  I have a lot of other things to do first.  Bug fixes take priority and I have a few long-term issues.  Feel free to make it into another issue so we can track it.

jacobsalmela:
Fixed by #180

PR #168, opened 2016-01-14
Possible fix for gravity.sh whitelisting

PromoFaux:
Just something that was bugging me, whitelist.sh was failing on the last line due to incorrect permissions.

### DIFF
--- a/advanced/Scripts/blacklist.sh
+++ b/advanced/Scripts/blacklist.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+# (c) 2015 by Jacob Salmela
+# This file is part of Pi-hole.
+#
+# Pi-hole is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+
+blackList=/etc/pihole/blacklist.txt
+if [[ ! -f $blackList ]];then
+    touch $blackList
+fi
+
+if [[ $# = 0 ]]; then
+    echo "Immediately blacklists one or more domains."
+    echo "Usage: blacklist.sh domain1 [domain2 ...]"
+fi
+
+boolA=false
+boolB=false
+for var in "$@"
+do
+		bool=false;
+    echo "Blacklisting $var..."
+    #add to whitelist.txt if it is not already there
+    grep -Ex -q "$var" $blackList || boolB=true
+    if $boolB; then
+        echo $var >> $blackList        
+        boolA=true;
+    else
+        echo "$var Already in blacklist.txt"
+    fi
+done
+
+if $boolA; then
+    echo "New domains added to blacklist. Running gravity.sh"
+    /usr/local/bin/gravity.sh
+else
+	echo "No need to update Hosts list, given domains already in blacklist"
+fi
--- a/advanced/Scripts/whitelist.sh
+++ b/advanced/Scripts/whitelist.sh
@@ -52,6 +52,16 @@ if [[ "$combopattern" != "" ]]; then
   sed -r -i '/'"$pattern"'/d' $adList
 
   echo "** $# domain(s) whitelisted."
-  # Force dnsmasq to reload /etc/pihole/gravity.list
-  kill -HUP $(pidof dnsmasq)
+  # Reload hosts file
+	echo "** Refresh lists in dnsmasq..."
+
+	dnsmasqPid=$(pidof dnsmasq)
+
+	if [[ $dnsmasqPid ]]; then
+		# service already running - reload config
+		sudo kill -HUP $dnsmasqPid
+	else
+		# service not running, start it up
+		sudo service dnsmasq start
+	fi
 fi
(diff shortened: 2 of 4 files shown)


PromoFaux:
@jacobsalmela @dschaper 

So, I'm pretty sure this fix negates the need for whitelist.sh. Take a look and let me know what you think

dschaper:
@PromoFaux Thanks, I'll have a look and see what I can see!

PromoFaux:
ahem.

dschaper:
Branches are good... Branch early and often is the key to Git...

PromoFaux:
@dschaper Whitelist.sh updated to reflect changes to gravity.sh

PromoFaux:
Hopefully all this is good! If nothing else, I've learnt a lot about bash scripting and git tonight.

jacobsalmela:
Wow, lots of stuff to look over!  I haven't had as much time with my new job, but I think the whitelist bug needs to be squashed first as there are two issues open about it.  After that, the other bugs will be next.

I'll do some testing tonight with what you have done and try my best to get as much done as I can tonight.

PromoFaux:
@jacobsalmela There is a lot less there than it looks like.. just a large amount of commits because I kept thinking of other things/noticing a couple mistakes. 

In terms of squashing the whitelist bug, this removes entires for whitelisted domains in the gravity.list file for both ipv4 and ipv6.

for the blacklist it adds it before running gravity_advanced()...

as far as I can tell, if it exists in the whitelist.txt / blacklist.txt.. consider it white/blacklisted.

jacobsalmela:
So your changes seem to work well at removing/adding both IPv4 and IPv6.  It also creates the whitelist/blacklist text files if they don't exist.  However, I was hoping to have `whitelist.sh` and `blacklist.sh` be independent of `gravity.sh`--Partially because of [this](https://github.com/jacobsalmela/AdminLTE/pull/9), but also because to allow users to quickly white or blacklist a domain without having to run `gravity.sh`: it would just remove or add the domain and then reload `dnsmasq`.  There is no need to try to download the new lists and recreate `gravity.list` from scratch each time.

So for example, in your `gravity.sh` you are using the [cool perl command](https://github.com/PromoFaux/pi-hole/blob/master/gravity.sh#L204-L205) to whitelist domains:
```
awk -F':' '{ print $1 }' $latentWhitelist | xargs -I {} perl -i -ne'print unless /[^.]'{}'(?!.)/;' $adList
```
But then in your `whitelist.sh`, a [different approach](https://github.com/PromoFaux/pi-hole/blob/master/advanced/Scripts/whitelist.sh#L28-L36) is being used:
```
grep -Ex -q "$latentPattern" $whiteList || echo $latentPattern >> $latentWhitelist
```

So as I mentioned in #139, it would be nice to outsource the whitelisting to an external script, so it would look something like this:

```
function gravity_Whitelist() {

	# Whitelist (if applicable) domains
	if [[ -r $whitelist ]];then
              sudo /usr/local/bin/whitelist.sh $(cat /etc/pihole/whitelist.txt)
        else
              echo "* Nothing to whitelist"
        fi
```
The same for the blacklist:
```
function gravity_blacklist() {

	# Blacklist (if applicable) domains
	if [[ -r $blacklist ]];then
              sudo /usr/local/bin/blacklist.sh $(cat /etc/pihole/blacklist.txt)
        else
              echo "* Nothing to blacklist"
        fi
```
This way, users can run `whitelist.sh` and `blacklist.sh` at any time if they find a domain they want to apply.  This will also make the code more modular and easier to manage.  It also will help with the Web UI whitelist/blacklist pages that I linked to earlier.

If you are up for it, try to make those changes.  If not, we can definitely use chunks of what you have done.

I hope what I am saying makes sense.  Let me know your thoughts.

PromoFaux:
Makes sense, I can take a look tomorrow.

To explain my thinking however...
gravity.sh runs at midnight on a cron job, right? So it needs to load the whitelists anyway. It makes no sense for it to outsource to external scripts that reload the dnsmasq at this point as the master list hasn't been completed yet.

All that white/blacklist.sh do is add the domains to the existing white/latentwhite/blacklist.txt and then rerun gravity.sh. It also means that next time gravity.sh runs, the domains are re-included. The different approach is used in whitelist.sh because it doesn't need the regex required by gravity.sh to actually remove them from gravity.list, it just adds it to whitelist.txt as the user typed it, and to latentwhitelist.txt with the periods escaped. To be fair, latentwhitelist.txt could probably be scrapped.

If you like, I can change the two scripts to just change the hosts file and reload dnsmasq instead of running gravity.sh to do the same job? 

If the jobs are outsourced, then the dnsmasq is reloaded an unnecessary amount of times

jacobsalmela:
Hmmm.  Let me think it over some more.

Mcat12:
As I see it, the reasons we have whitelist.sh and now blacklist.sh are:

1. Users can easily add to the lists without waiting for gravity
2. The in-the-works white/blacklist [site](https://github.com/jacobsalmela/AdminLTE/pull/9) can quickly add (and hopefully subtract) to/from the lists.

Regarding the last part, it would be helpful to have the ability to take off items from the white/blacklist from these scripts.

PromoFaux:
@Mcat12 That makes sense.I'll take a look later on today and see what I come up with. What I'll probably end up doing is branching an submitting as a fresh pull..

PR #173, opened 2016-01-15
Whitelisting changes

PromoFaux:
Right, this is a lot more robust than my previous pull.
@jacobsalmela @dschaper @Mcat12 take a look, and see what you think.

(I even did it on a different branch this time!)

### DIFF
--- a/advanced/Scripts/blacklist.sh
+++ b/advanced/Scripts/blacklist.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+# (c) 2015 by Jacob Salmela
+# This file is part of Pi-hole.
+#
+# Pi-hole is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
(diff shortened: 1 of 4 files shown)


Mcat12:
While you did do your work on a different branch (good job with that!), you are still requesting to merge into the main branch. Also, it may be better to change the name of your branch to "whitelist", as that is the name of the branch on this repo where whitelist changes are occurring. If you do that, you should change where you're merging into from "master" to "whitelist". 

That's just from glancing at the request. I'll look into the actual changes later, when I have a bit more time. :D

advanced/Scripts/blacklist.sh
```diff
@@ -0,0 +1,180 @@
...
+	
+	if [ -z "$validDomain" ]; then
+		echo $1 is not a valid argument or domain name
+	else	  
+	  domList=("${domList[@]}" $validDomain)
+	fi
+}
+
+function PopBlacklistFile(){
+	#check whitelist file exists, and if not, create it
```
Mcat12: Looks like you forgot to change whitelist to blacklist in the comment

PromoFaux: heh, semantics. that's easily fixed.

PromoFaux:
>Also, it may be better to change the name of your branch to "whitelist", as that is the name of the branch on this repo where whitelist changes are occurring. If you do that, you should change where you're merging into from "master" to "whitelist".

Sure thing, I can look into that when I get home. I'm still fairly new to git, just finding my legs!

PromoFaux:
see #174

PR #174, opened 2016-01-15
Whitelisting changes

PromoFaux:
Note : This will break the "ugly hack"  @jacobsalmela put in yesterday...

### DIFF
--- a/advanced/Scripts/whitelist.sh
+++ b/advanced/Scripts/whitelist.sh
@@ -16,7 +16,7 @@ if [[ $# = 0 ]]; then
     echo "  -d, --delmode		Remove domains from the whitelist"
     echo "  -nr, --noreload	Update Whitelist without refreshing dnsmasq"
     echo "  -f, --force		Force updating of the hosts files, even if there are no changes"
-    
+    echo "  -q, --quiet		output is less verbose"
     exit 1
 fi
 
--- a/advanced/Scripts/whitelist.sh
+++ b/advanced/Scripts/whitelist.sh
@@ -7,83 +7,174 @@
 # the Free Software Foundation, either version 2 of the License, or
 # (at your option) any later version.
 
+if [[ $# = 0 ]]; then
+    echo "Immediately whitelists one or more domains in the hosts file"
+    echo " "
+    echo "Usage: whitelist.sh domain1 [domain2 ...]"
+    echo "  "
+    echo "Options:"
+    echo "  -d, --delmode		Remove domains from the whitelist"
+    echo "  -nr, --noreload	Update Whitelist without refreshing dnsmasq"
+    echo "  -f, --force		Force updating of the hosts files, even if there are no changes"
+    
+    exit 1
+fi
+
+#globals
 whitelist=/etc/pihole/whitelist.txt
 adList=/etc/pihole/gravity.list
-webInterfaceEchos=/tmp/whitelistEchoFile
+reload=true
+addmode=true
+force=false
+versbose=true
+domList=()
+domToRemoveList=()
 
-if [[ ! -f $whitelist ]];then
-    touch $whitelist
+piholeIPfile=/tmp/piholeIP
+piholeIPv6file=/etc/pihole/.useIPv6
+
+# Otherwise, the IP address can be taken directly from the machine, which will happen when the script is run by the user and not the installation script
+IPv4dev=$(ip route get 8.8.8.8 | awk '{for(i=1;i<=NF;i++)if($i~/dev/)print $(i+1)}')
+piholeIPCIDR=$(ip -o -f inet addr show dev $IPv4dev | awk '{print $4}' | awk 'END {print}')
+piholeIP=${piholeIPCIDR%/*}
+
+modifyHost=false
+
+
+if [[ -f $piholeIPv6file ]];then
+    # If the file exists, then the user previously chose to use IPv6 in the automated installer
+    piholeIPv6=$(ip -6 route get 2001:4860:4860::8888 | awk -F " " '{ for(i=1;i<=NF;i++) if ($i == "src") print $(i+1) }')
 fi
 
-formatEchoes()
-{
-if [[ "$(whoami)" = "www-data" ]];then
-    echo "$1" >> $webInterfaceEchos
-else
-    echo "$1"
-fi
+
+function HandleOther(){	
+  #check validity of domain
+	validDomain=$(echo $1 | perl -ne'print if /\b((?=[a-z0-9-]{1,63}\.)(xn--)?[a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,63}\b/')
+	
+	if [ -z "$validDomain" ]; then
+		echo $1 is not a valid argument or domain name
+	else	  
+	  domList=("${domList[@]}" $validDomain)
+	fi
 }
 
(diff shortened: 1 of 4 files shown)


dschaper:
Thanks!

Mcat12:
By the way, this PR will rebase whitelist with the changes from master. (I don't know if I was the only one wondering why there were so many commits :) )

Mcat12:
You should add in the commands list the quiet flag. It's implemented, but not shown with the other commands.

gravity.sh
```diff
@@ -160,48 +159,35 @@ function gravity_Schwarzchild() {
...
-function gravity_pulsar() {
 
+function gravity_Blacklist(){
 	# Append blacklist entries if they exist
-	if [[ -r $blacklist ]];then
-        numberOf=$(cat $blacklist | sed '/^\s*$/d' | wc -l)
-        echo "** Blacklisting $numberOf domain(s)..."
-        cat $blacklist >> $piholeDir/$matter
-	fi
+	blacklist.sh -f -nr -q
```
Mcat12: Do you need to add the domains you are blacklisting to this command?

Mcat12:
Once this is merged, I can change up my white/blacklist editor to use the new features.

gravity.sh
PromoFaux: All handled in blacklist.sh!

PromoFaux:
>You should add in the commands list the quiet flag. It's implemented, but not shown with the other commands.

Yeah, could do, I left it off of the list because it's only really for it being called from gravity. If a user is running it manually, they probably want some confirmation that the domains they added were blacklisted.

When run by gravity.sh, it a) makes sure the ad list domains are whitelisted, and b) just re-whitelists everything that's in whitelist.txt

PromoFaux:
>  (I don't know if I was the only one wondering why there were so many commits :) )

There weren't when I originally set the pull request to master.. but since you suggested I merge with the whitelist branch,,,

advanced/Scripts/whitelist.sh
```diff
@@ -7,83 +7,174 @@
 # the Free Software Foundation, either version 2 of the License, or
 # (at your option) any later version.
 
+if [[ $# = 0 ]]; then
+    echo "Immediately whitelists one or more domains in the hosts file"
+    echo " "
+    echo "Usage: whitelist.sh domain1 [domain2 ...]"
+    echo "  "
```
Mcat12: Slight ocd on my part, but blacklist.sh has no empty lines between the help output.

Mcat12:
No, it's good that the whitelist branch be updated (I think). We can make sure we haven't broken anything major.

gravity.sh
Mcat12: Nice. :smile:

advanced/Scripts/whitelist.sh
PromoFaux: Huh, thought i'd sorted that.. I think I might have got a bit confused with the merging earlier! Will sort it out in a bit

PromoFaux:
There was some stuff committed into the whitelist.sh with a web hook labelled as a "dirty hack" so I just wrote my changes over it, as it probably would not have worked with my new changes anyway

Mcat12:
That was @jacobsalmela trying to make a simple whitelister on the site, which would give the console output. It would redirect the console output to somewhere the script could get it and display on the site. The whitelister will soon be deprecated, if my editor is merged.

Mcat12:
Hmm, I guess the whitelist branch already got the latest from master. You just were working with an earlier version, and now git is bringing your branch up to date. Ok, so if nothing is breaking now, it should stay safe. :smile:

gravity.sh
PromoFaux: Yeah, basically all that line says is "Run the blacklist script, force it to update the hosts file, don't reload dnsmasq, and do it quietly.

advanced/Scripts/whitelist.sh
PromoFaux: fixed in f9a2ca5

dschaper:
What should we do if a user has a domain in both the blacklist and the whitelist? (Say they blacklist a domain, forget they did and a few months later try to whitelist the domain...) Just a hypothetical, not a comment on the current code. Testing out the code now and it's looking good. Thanks!

PromoFaux:
Funny you should mention that, I've been giving it some thought this evening.

What I'll probably end up doing is giving both the whitelist.sh and blacklist.sh knowledge of each other's lists. 

e.g if you add a domain to the whitelist, it will be removed from the blacklist and vice versa. We should probably decide which one takes priority.

dschaper:
I do a great impression of a forgetful user! (There is no truth to the rumor that the above scenario is based on reality in any way!)

I think whitelisting should be the winner, but that's just a biased coin flip, really.

dschaper:
And now that I think about it some more, how do we handle removing from blacklist/whitelist. Use the same process and have whitelist remove the blacklist and blacklist remove the whitelist? Just throwing things out there. I've added domains with the blacklist script and for now I have to edit `gravity.list`  by hand or run `gravity.sh` again after cleaning the `blacklist.txt`. 

I guess with the improved web interface these situations will be uncommon, it will be easy to see what's what...

PromoFaux:
To remove from the blacklist (and also remove it from gravity.list) run blacklist.sh -d <domain>

dschaper:
Ah, thanks! (I'll document that in the Wiki when the PR merges.)

PromoFaux:
I tested it as best I could earlier, but do let me know if you find any bugs, and I'll sort them out! Once it's merged, i'll start work on making the black and whitelists aware of each other.

PromoFaux:
Oh also, try passing a non domain argument to either script! Quite proud of that bit..

e.g <code>blacklist.sh kjngkdsfnshkdfhn<code>

dschaper:
Nice, lol... I'm still in a love/hate mode with regex.

Mcat12:
I think the current gravity.sh has blacklist take priority (runs whitelist first, then blacklist). Are you planning on implementing a smart listing system as described above, or just giving one list priority?

dschaper:
When I wrote `gravity_pulsar()` I didn't take into account any order, it's just happenstance that the order is like that... @jacobsalmela may have a preference though.

jacobsalmela:
Wow guys, I was busy today but my inbox was blowing up with all the chatter.  I'll have to check this stuff and all the comments on the other issues/PRs.

Thanks for your hard work.

jacobsalmela:
Impressive @PromoFaux!  

@Mcat12, should I wait to merge your changes on the Web UI?

Mcat12:
Ya, I'm going to reinstall my pihole (was running off the dietpi version using apache) and implement the new whitelist and blacklist features. In the mean time, the web interface needs sudo for white/blacklist.sh, in order to restart the dns. I think this has to be done by giving www-data no password, no tty required sudo permissions on the scripts in the sudoers file.

jacobsalmela:
Right, I was going to work on that.  

I'm going to focus on the bugs first, but go ahead and keep working so it's ready.   You are obviously better qualified for Web development then me!

Mcat12:
lol, I guess I updated my pi too soon. I'm going to have to update again with all the changes committed.

jacobsalmela:
Yeah sorry about that.; I was kind of on a spree and couldn't stop.

Issue #181, opened 2016-01-17
blacklist.sh has erroneous CRLF items, causing it to malfunction


shadowkhas:
Hi guys,

Thanks for the awesome fixes on the whitelist/blacklist stuff! It fixed my problem that I was having. :)

blacklist.sh looks like it has some troubles though - it's got CRLF items in the line breaks, which is causing problems executing the script. A "cat -v whitelist.sh" and "cat -v blacklist.sh" will show the differences - whitelist.sh doesn't  have any of the "^M" items, blacklist.sh does. Taking those out with vi seemed to get blacklist.sh working again for me.

PromoFaux:
I'll take a look at this In  the morning, might be something to do with the settings In my editor before I committed.  Although to use an old coding trope...  "it worked on my machine"  :)

PromoFaux:
This is what happens when you develop on windows, and forget to save in UNIX mode...

Should be fixed now @shadowkhas

jacobsalmela:
I believe this is fixed by #189 now.

PR #185, opened 2016-01-17
Changed line endings in blacklist.sh

PromoFaux:
FIx for #181


PromoFaux:
Why is this showing all the commits from my previous attempt at fixing the whitelist? 

The relevant commits here e4f4488 and b4b7ae7, I think.

Can somebody check out my current master and test just to make sure before this is pushed...

Issue #182, opened 2016-01-17
Lighttpd config hates custom web pages.


Mcat12:
I run a few other sites (tools and other personal stuff) from lighttpd (previously Apache). The current config redirects anything but the admin site to the ad blocking page, while at the same time using the ad blocking page as a 404 redirect. Since it redirects anything but the admin site, my pages get redirected. Does the config have to have both redirects, or can it just have the 404 redirect?

dschaper:
I was working on a config that would only need the 404 redirect. I think that probably will work, as the html page names that are going to be sent for lookup are most likely not pages that you are serving up. The downside is that your 404's are going to be blank pages and not informative 404's for legitimate page not found traffic.

dschaper:
Is there a way in 'lighttpd' to create a hierarchy of sites, like there is with 'nginx'? You can have named sites with domains configured with a stanza and then end with a 'default' site that is a catch all for non-virtual host sites. Kind of like an entire 404 site...

Mcat12:
I just commented out the redirect if not admin, reloaded, and it seems to work.

Mcat12:
Also, I added "Blocked by Pi-hole", as I was used to that from the dietpi version.

dschaper:
I think we did something like that with another issue, https://github.com/jacobsalmela/pi-hole/issues/161#issuecomment-171179325

Mcat12:
Do we need the special redirect, or can we just Add "Blocked by Pi-hole" to the blank page and only have the 404 redirect? I can make a pr if that's what we're going to do.

dschaper:
From my thinking we only need a 404. Here's how I see the process:

1. Browser tries to fetch `http://ad.domain/ThisIsAnAd.gif`
2. Pi-Hole lookup returns local address for `ad.domain`
3. Browser requests `http://local/ThisIsAnAd.gif`
4. Local lighttpd looks for `ThisIsAnAd.gif` and doesn't find it, so it signals 404.
5. Local lighttpd configuration says 404's should serve `index.html` 
6. Browser renders 'http://ad.domain/ThisIsAnAd.gif` from source 'http://local/index.html`

Mcat12:
Ok. I'll make a PR.

jacobsalmela:
I think this should work well.  There isn't much of a speed difference between getting a 404 vs. 200.

dschaper:
Resolved

PR #190, opened 2016-01-18
Allow other pages to be served from lighttpd & add text to ad blocking page

Mcat12:
Removes redirect if page is not the admin site. Adds text ("Blocked by Pi-hole") to ad blocking page to let the user know why they received a blank page.
See #182

### DIFF
--- a/advanced/lighttpd.conf
+++ b/advanced/lighttpd.conf
@@ -37,15 +37,3 @@ $HTTP["url"] =~ "^/admin/" {
 	# Create a response header for debugging using curl -I
 	setenv.add-response-header = ( "X-Pi-hole" => "The Pi-hole Web interface is working!" )
 }
-
-# If the URL does not start with /admin, then it is a query for an ad domain
-$HTTP["url"] =~ "^(?!/admin)/.*" {
-	# Create a response header for debugging using curl -I
-	setenv.add-response-header = ( "X-Pi-hole" => "A black hole for Internet advertisements." )
-
-	# Set the cache to 1 day for better performance
-	expire.url = ("" => "access plus 1 days")
-
-	# Send the query into the black hole
-	url.rewrite = (".*" => "pihole/index.html" )
-}
--- a/advanced/lighttpd.conf
+++ b/advanced/lighttpd.conf
@@ -37,3 +37,9 @@ $HTTP["url"] =~ "^/admin/" {
 	# Create a response header for debugging using curl -I
 	setenv.add-response-header = ( "X-Pi-hole" => "The Pi-hole Web interface is working!" )
 }
+
+# If the URL does not start with /admin, then it is a query for an ad domain
+$HTTP["url"] =~ "^(?!/admin)/.*" {
+	# Create a response header for debugging using curl -I
+	setenv.add-response-header = ( "X-Pi-hole" => "A black hole for Internet advertisements." )
+}
--- a/advanced/index.html
+++ b/advanced/index.html
@@ -1,4 +1,5 @@
 <html>
 <body>
+Blocked by Pi-hole
 </body>
 </html>


advanced/lighttpd.conf
```diff
@@ -37,15 +37,3 @@ $HTTP["url"] =~ "^/admin/" {
 	# Create a response header for debugging using curl -I
 	setenv.add-response-header = ( "X-Pi-hole" => "The Pi-hole Web interface is working!" )
 }
-
-# If the URL does not start with /admin, then it is a query for an ad domain
-$HTTP["url"] =~ "^(?!/admin)/.*" {
-	# Create a response header for debugging using curl -I
-	setenv.add-response-header = ( "X-Pi-hole" => "A black hole for Internet advertisements." )
```
dschaper: We still need the X-Header for debugging purposes.

Mcat12:
I added the debug header back in.

dschaper:
Thanks.

jacobsalmela:
I think this is good using the 404 handler, but do you think people would rather have the blank page, or have some indicator that it was blocked by Pi-hole?

dschaper:
I'm more of the idea that it should be blank, and we add to instructions to the Wiki (or FAQ) on how to modify the html for messages. I appreciate that DietPi has some additional configuration enhancements, but I personally prefer a blank indicator. Just my thoughts...

Mcat12:
Right now it says "Blocked by Pi-hole". Maybe we could add a link to the admin page?

Mcat12:
And maybe there could be a page to configure the pihole, like what the ad blocking page looks like or other things.

dschaper:
I wouldn't add a link to the admin page, that would give users that are not admin a link for them to access the admin page and I don't think that's a good idea. Right now there is nothing that protects the admin page from unauthorized users, and with the enhancements that you are helping with on that page (the whitelist/blacklist edit capabilities) an unauthorized user could do some mischief.

dschaper:
I do kind of like the idea of having a link in the admin page that pulls up the blocking page in an iframe that would allow for a preview of what would be displayed in place of an ad. I just don't know how much of the project is going to be altered with the addition of the pay for play aspect that will allow users to tailor their experience by paying to not see ads.

Mcat12:
I don't fully understand that concept. Do they want people to pay to not see ads?

dschaper:
That's the basic idea. Right now websites that host ads get a minuscule portion of the advertising revenue. This new model pays the content providers directly, cutting out the middle men and their markups and cuts. So if there is a website that I like to visit and I'd like to support that website, right now with an ad blocker my favorite sites don't get anything. I can donate to the sites, or whitelist the sites and give them the 1/10th of the ad revenue they should be getting. Under the new model I can pay the website directly and they turn off the ads. They get the full amount I send to them and I don't get any ads. So there's mutual benefit.

Now were talking small amounts, most websites get less than a fraction of a penny per ad impression as it is, so even if they got a full penny or a dime, that would be a massive amount in terms of what they were getting before.

Mcat12:
@jacobsalmela This is ready to be merged if you want.

dschaper:
I'll merge it to a new branch in a bit for alpha testing. Did we ever come to a consensus about the tagline in the blocking image?

jacobsalmela:
Let's leave the page blank and provide users instructions on how to replace the ads with their own text or images.  I have an [FAQ](https://pi-hole.net/faq/is-it-possible-to-change-the-blank-page-that-takes-place-of-the-ads-to-something-else/) on this, but it is very limited.

Putting in "blocked by Pi-hole" is kind of like advertising anyway, so let's leave it blank and let the user customize as they desire.

dschaper:
D'accod...

dschaper:
@Mcat12 Can you update that and I'll merge the update into a new feature branch.

dschaper:
I can try and cherry pick the two commits we'd like to add. Not sure if it works or not, we'll see!

dschaper:
Commits cherry-picked into pi-hole/development

PR #193, opened 2016-01-18
Major overhaul of chronometer script

PromoFaux:
Added -j option to output stats as JSON formatted string. If run with no option, then chronometer runs as normal.

Fixed calculations to match those on the web dashboard.

Total number of domains now takes whether or not IPv6 is used into consideration.

As this will output a JSON string, this can also be consumed by <code>index.php</code>  AND <code>api.php</code> on the web dashboard.

i.e Change <code>api.php</code> to simply read:
~~<code><?php echo exec("/usr/local/bin/chronometer.sh -j'}"); ?></code>~~
<code><?php echo exec("/usr/local/bin/chronometer.sh -j"); ?></code> (corrected)
Though I don't know anything about PHP, so someone may wish to correct that.

Similarly <code>index.php</code> can read the output of this script and parse the JSON to populate the dashboard.

Plus, think of all the other crazy things you could do with the data. 

image

### DIFF
--- a/advanced/Scripts/chronometer.sh
+++ b/advanced/Scripts/chronometer.sh
@@ -111,7 +111,7 @@ function normalChrono(){
 function displayHelp(){
  echo "Displays stats about your piHole!"
     echo " "
-    echo "Usage: whitelist.sh [optional:-j]"
+    echo "Usage: chronometer.sh [optional:-j]"
     echo "Note: If no option is passed, then stats are displayed on screen, updated every 5 seconds
     echo "  "
     echo "Options:"
--- a/advanced/Scripts/chronometer.sh
+++ b/advanced/Scripts/chronometer.sh
@@ -1,35 +1,135 @@
 #!/usr/bin/env bash
-# Displays Pi-hole stats on the Adafruit PiTFT 2.8" touch screen
-# Set the pi user to log in automatically and run this script from /etc/profile
 # (c) 2015 by Jacob Salmela
 # This file is part of Pi-hole.
 #
 # Pi-hole is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation, either version 2 of the License, or
-# (at your option) any later version.
-for (( ; ; ))
+# (at your opercentBlockedTodayion) any later version.
+
+
+#Functions##############################################################################################################
+piLog="/var/log/pihole.log"
+gravity="/etc/pihole/gravity.list"
+
+today=$(date "+%b %e")
+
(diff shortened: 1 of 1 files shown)


advanced/Scripts/chronometer.sh
```diff
@@ -1,35 +1,135 @@
...
-	clear
-	# Displays a colorful Pi-hole logo
-	toilet -f small -F gay Pi-hole
-	echo "        $(ifconfig eth0 | awk '/inet addr/ {print $2}' | cut -d':' -f2)"
-	echo ""
-	uptime | cut -d' ' -f11-
-	echo "-------------------------------"
-	# Uncomment to continually read the log file and display the current domain being blocked
-	#tail -f /var/log/pihole.log | awk '/\/etc\/pihole\/gravity.list/ {if ($7 != "address" && $7 != "name" && $7 != "/etc/pihole/gravity.list") print $7; else;}'
+# (at your opercentBlockedTodayion) any later version.
```
Mcat12: opercentBlockedTodayion? You coined a new word without even noticing! :smile:

Mcat12:
In your comment about changing api.php, I don't think you need ```}'``` in the string.

advanced/Scripts/chronometer.sh
PromoFaux: fixed

PromoFaux:
Currently api.php a JSON string like this:
<code>{"domains_being_blocked":"136774","dns_queries_today":"6114","ads_blocked_today":"764","ads_percentage_today":12.49591102388}</code>

and this script change outputs this:
<code>{"domains_being_blocked":"68387","dns_queries_today":"6095","ads_blocked_today":"764","ads_percentage_today":"12.53"}</code>

Mcat12:
What's the difference?

PromoFaux:
Oh, wait, I see you're referencing my php psuedocode!

Mcat12:
It should be ```<?php echo exec("/usr/local/bin/chronometer.sh -j"); ?>``` (Without the ```{'```)

PromoFaux:
But other than that, valid PHP script?

Mcat12:
I believe so. There's a feature that I think is standard, where you don't need the ```?>``` if there's only php in the file (and ends with *.php), so it could even be shortened to:
```php
<?php
echo exec("/usr/local/bin/chronometer.sh -j");
```

PromoFaux:
I was brave and tried the PHP changes myself!

Mcat12:
Do they work?

PromoFaux:
Of course they work, I would not commit untested code (anymore!!)

PR #196, opened 2016-01-19
cancel properly from installer

### LINKED ISSUE #154: Installer continues if user presses cancel on interface or protocol selection screens
If the user really wanted to cancel, they can hit Ctrl+C after it continues on, but it would be nice for the cancel button to actually work.

jacobsalmela:
Fixes #154


Mcat12:
Are all of those commits new?

jacobsalmela:
Closing because this was an older branch.  Made a new one off dev that will be easier.

Issue #107, opened 2015-12-03
Install config files in subdirectories


dschaper:
Instead of replacing config files, place pihole specific config files in config.d sub-directories.

dschaper:
dnsmasq seems to work with the configuration file in `/etc/dnsmasq.d/pihole.conf`. The only thing we need to watch is that if there are any duplicate configuration settings in `/etc/dnsmasq.conf` then the system throws an error. Working on an existing file check that is similar to the legacy pihole detection code.

jacobsalmela:
This was added in 2.5.1 right?

PromoFaux:
Yeah, this is live, issue can be closed...

jacobsalmela:
Cool, thanks!

PR #198, opened 2016-01-19
Upsteam DNS enhancement

ChadBHowell:

2) Allow the user to choose between Google and DynDNS upstream servers during the install.  This will later be extended to allow the user to enter custom upstream DNS servers as well.

3) Add unzip to the dependencies to avoid errors on some systems.


dschaper:
I think there was another issue open, #107 that this addressed. I'll note that this potentially fixes that issue. I agree that we should be using the `*.d` subdirectories. 

I like the choice for user selection up upstreams. And just a quick question, do you have a distro that had unzip/zip missing?

We'll test out the changes and see how things go. Can you set up a PR that merges to the development branch though? We don't like to merge to master, we'd like to keep master as pristine as possible.

Thank you very much for you PR and assistance!

jacobsalmela:
I think this will be an important PR.  I like how you followed the conventions previously introduced by used `sed` to swap in the correct values.  And to second, please make it in a branch other than `master` for stability reasons.

automated install/basic-install.sh
```diff
@@ -230,11 +251,12 @@ $SUDO chmod 755 /usr/local/bin/{gravity,chronometer,whitelist,blacklist,piholeLo
 }
 
 installConfigs(){
-$SUDO mv /etc/dnsmasq.conf /etc/dnsmasq.conf.orig
 $SUDO mv /etc/lighttpd/lighttpd.conf /etc/lighttpd/lighttpd.conf.orig
-$SUDO curl -o /etc/dnsmasq.conf https://raw.githubusercontent.com/jacobsalmela/pi-hole/master/advanced/dnsmasq.conf
+$SUDO curl -o /etc/dnsmasq.conf/01-pihole.conf https://raw.githubusercontent.com/jacobsalmela/pi-hole/master/advanced/01-pihole.conf
```
Mcat12: Did you mean to put the config into dnsmasq.d?

jacobsalmela: Yeah, I was just wondering that, too.

Mcat12:
It should go in the Devel branch. (Is it just me, or does Devel sound like Devil? :eyes:)

ChadBHowell:
The distro that was missing unzip was the latest Debian image for the Beaglebone Black.  I have fixed the typo and have resubmitted the pull request against the dev branch.

Mcat12:
This pr can be closed now.

jacobsalmela:
Closing per @Mcat12's comment.

Issue #170, opened 2016-01-15
DNSCrypt support


ajp2k13:
Support for DNSCrypt and/or DNSSec in Pi-Hole would add some extra security and prevent snooping or MITM attacks.

Maybe some pointers can be found here?
http://www.snbforums.com/threads/dnscrypt-from-opendns.11645/
##

extremeshok:
DNSSec causes allot of issues with host-rewrites.

If you want to enable dns sec.

Install bind: apt-get install bind9
/etc/bind/named.cond.options
````
options {
directory "/var/cache/bind";

dnssec-enable yes;
dnssec-validation yes;
dnssec-lookaside auto;

listen-on port 1053 { 127.0.0.1; };
auth-nxdomain no;    # conform to RFC1035
forwarders {
 		//Google
		//8.8.8.8
		//8.8.4.4
};
}
````

change dnsmasq.conf to enable bind as the upstream
````
server=127.0.0.1#1053
````

service bind9 restart
service dnsmasq restart

and there you go.

U can see bind9 working with dnssec via tail-f /var/log/syslog


##note:
dnssec-validation yes; is very slow, think 10sec per domain

dschaper:
From what I remember of the conversation before the issue was created, we were talking more about DNSCrypt, just for protection from ISP hijacking and redirecting to internals.

ajp2k13:
Maybe this can help later on...
https://github.com/simonclausen/dnscrypt-autoinstall

dschaper:
Thanks for the pointer, we'll take a look at it and see what we can do with the integration. We're focused right now on getting the existing features to be rock-solid, and then we'll take a look at adding in some more features. We have a lot planned to make the Pi-hole even better!

ajp2k13:
@dschaper No rush, I just wanted to post the link here so I don't forget it. :)

ajp2k13:
Another guide for future reference...
http://rpinow.tumblr.com/post/118887639037/dnscrypt-on-osmc

ajp2k13:
I followed the above guide (http://rpinow.tumblr.com/post/118887639037/dnscrypt-on-osmc) on a VM with Ubuntu 15.10 and it worked very well, going to try it on my Pihole next when I get the time.

dschaper:
[...]

PR #204, opened 2016-01-19
Upsteam DNS enhancement v2

ChadBHowell:

2) Allow the user to choose between Google and DynDNS upstream servers during the install. This will later be extended to allow the user to enter custom upstream DNS servers as well.

3) Add unzip to the dependencies to avoid errors on some systems.

Change Log:
1) Fixed Typo pointed out by Mcat12
2) Pull Request against Dev Branch


dschaper:
Thanks for moving the PR to development. It really does help us keep stability on Master while trying out all these awesome patches that you all are submitting. We'll test them out and see what shakes loose. (I need to go pick up another PiB+ so I have a test unit for that flavor of RaspberryPi.)

dschaper:
Merging to branch https://github.com/jacobsalmela/pi-hole/tree/feature/UpstreamResolver 

I like this, the only thing we need to take into account is that on existing installs that have `/etc/dnsmasq.conf` populated there will be `illegal repeated keywords` conflicts. So can you modify this to check for an existing `dnsmasq.conf` and replace that file with the `dnsmasq.conf.orig` file, then go ahead and drop the new config in the subdirectory?

Thanks again for the submit, looks great!

ChadBHowell:
Sure thing.  Will work on it tomorrow night.

dschaper:
@ChadBHowell You may have to pull down a new copy of the development branch, there were some changes to the `basic-install.sh` and it looks like there's a merge conflict now.

PromoFaux:
bump

extremeshok:
Rather install bind9 as per https://github.com/pi-hole/pi-hole/issues/170

Which will give you dnssec for upstream servers, bind9 is also more efficient for upsteams.

ChadBHowell:
I was just learning how to use github when I submitted this PR. I have sense done a bunch of reading and figured out branches... I think. :) I am going to resubmit this PR rebased off the latest dev branch.

PR #208, opened 2016-01-20
Allow user to be chosen during install

ChadBHowell:
Allow user to be added to the "www-data" group to be chosen during install instead of hard coding user "pi".  The current user is displayed as default.  Allows install script to run on other devices that don't have the "pi" user.  Namely every device besides the Pi. :)

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -83,6 +83,13 @@ whiptail --msgbox --backtitle "Initating network interface" --title "Static IP N
 In the next section, you can choose to use your current network settings (DHCP) or to manually edit them." $r $c
 }
 
+chooseUser()
+{
+# Choose user to be added to the www-data group
+wwwUser=$(whoami)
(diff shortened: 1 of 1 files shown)


dschaper:
Thanks for the PR! It make take a bit of time to get to this to test, we have a number of other bugs to track down before we get to enhancements but we might be able to use this further. The original plan was to add a 'pihole' user to run under so we weren't dependent upon 'Pi'. (This was going to be done as an adduser/useradd call from the debian package install.)

jacobsalmela:
Cool.  Yeah, we have been getting flooded with PRs and issues as the project has been gaining some attention.  As @dschaper mentioned, our priorities are critical bug fixes and then enhancements/features.  Please bear with as we try to work through everything.

ChadBHowell:
I was just learning how to use github when I submitted this PR.  I have sense done a bunch of reading and figured out branches... I think. :)  I am going to resubmit this PR with some enhancements.  Mainly, I ditched the user select stuff and added logic to see if system user "pihole" exists and create if not. User is created without a login or a home directory.

dschaper:
Thanks for all the help!

Issue #25, opened 2015-07-17
Blacklist count may not be correct in some cases


mospaeda:
If there is blank line in blacklist.txt, these lines are counted as blacklisted domains which is wrong.
In my case, I have 35 lines among then 4 empty lines but reported as 

** Neutrino emissions detected...
Skipping adaway.org list because it does not have any new entries...
Skipping adblock.gjtech.net list because it does not have any new entries...
Skipping adblock.mahakala.is list because it does not have any new entries...
Skipping hosts-file.net list because it does not have any new entries...
Skipping www.malwaredomainlist.com list because it does not have any new entries...
Skipping pgl.yoyo.org list because it does not have any new entries...
Skipping someonewhocares.org list because it does not have any new entries...
Skipping winhelp2002.mvps.org list because it does not have any new entries...
** Aggregating list of domains...
** Blacklisting 35 domain(s)...
** 398451 domains being pulled in by gravity...
** 377619 unique domains trapped in the event horizon.
** Formatting domains into a HOSTS file...

jacobsalmela:
I think this can be solved with something like this:

`numberOf=$(cat $blacklist | sed '/^\s*$/d' | wc -l)`

I'll test it some more

jacobsalmela:
I tested a lot on the Mac and was using this `sed 's/^[ \t]*//'` to eliminate the tab in front of the number when using `wc`.  It doesn't seem that the Raspbian version of `wc` has that, so the solution above should work for all the `numberOf` variables, while accounting for blank lines if they happen to exist.

PR #213, opened 2016-01-21
scripts to long-term stats

jacobsalmela:
These scripts will be necessary for parsing pihole.log before it is
flushed and then saving the information into a DB.  This information
can then be displayed in the Web GUI.

`mysql-server` and `php5-mysql` will need to be installed.

you can then setup the database with this command:

```
mysql -u root -p"raspberry" < /usr/local/bin/setupPiholeStats.sql
```

Once that is done, run `sudo php /usr/local/bin/lrs.php`

### DIFF
--- a/advanced/Scripts/lrs.php
+++ b/advanced/Scripts/lrs.php
@@ -1,0 +1,140 @@
+<?php
+// lrs.php (Long-rang Scanners)
+// Parses pihole.log to extract the top 10 advertisers and the long-term queries vs. ads
+date_default_timezone_set('UTC');
+
+//---------------------------------------------------------------------------------------------------
+//config
+
+$info_DEBUG = true;
+$file_name = "/var/log/pihole.log";
+$pie_chart_regexp_query = "query";
+$pie_chart_regexp_adver = ": /etc/pihole/gravity.list";
+
+$db_host = "localhost";
+$db_name = "charts";
+$db_user = "root";
+$db_pass = "hi";
+
+//---------------------------------------------------------------------------------------------------
+//logging
+
+function errorlog($error_log) {
+	global $info_DEBUG;
+	if ($info_DEBUG === true) {
+		print_r("[".date("Y-m-d H:i:s")."][".getmypid()."][error] ".$error_log."\n");
+		print_r("[".date("Y-m-d H:i:s")."][".getmypid()."][error] Script terminated under abnormal conditions\n");
+		print_r("[".date("Y-m-d H:i:s")."][".getmypid()."][error] Exiting\n");
+		exit;
+	}
+}
+
+function debuglog($debug_log) {
+	global $info_DEBUG;
+	if ($info_DEBUG === true)
+	print_r("[".date("Y-m-d H:i:s")."][".getmypid()."][debug] ".$debug_log."\n");
+}
+
+//---------------------------------------------------------------------------------------------------
+//func
+
+function parse_top_chart_advertiser($file_line) {
+	global $pie_chart_regexp_adver;
+	$adver_position = stripos($file_line, $pie_chart_regexp_adver);
+	$adver_substr = substr($file_line, $adver_position);
+	$adver_array = explode(" ", $adver_substr);
+	$adver = $adver_array[2];
+	return $adver;
+}
+
+function update_top_chart_stats($db_conn, $top_chart_array, $pie_chart_date) {
+	$top_sorted = array_count_values($top_chart_array);
+	foreach ($top_sorted as $top_id => $value) {
+		$sql_replace_top_chart_stats = "REPLACE INTO top_chart_stats VALUES ('$pie_chart_date', '$top_id', $value)";
+		if (!mysqli_query($db_conn, $sql_replace_top_chart_stats)) {
+			errorlog("SQL query failed for some reason: $sql_replace_top_chart_stats");
+		}else{
+			debuglog("SQL query successfully executed: $sql_replace_top_chart_stats");
+		}
+	}
+}
+
+function update_pie_chart_stats($db_conn, $pie_chart_date, $pie_chart_count_query, $pie_chart_count_adver) {
+	//daily insert
+	$sql_insert_pie_chart_stats = "INSERT IGNORE INTO pie_chart_stats VALUES ('$pie_chart_date', $pie_chart_count_query, $pie_chart_count_adver)";
+	if (!mysqli_query($db_conn, $sql_insert_pie_chart_stats)) {
+		errorlog("SQL query failed for some reason: $sql_insert_pie_chart_stats");
+	}else{
+		debuglog("SQL query successfully executed: $sql_insert_pie_chart_stats");
+	}
+	//total calc
+	$sql_select_pie_chart_stats  = "SELECT SUM(query_cnt) AS q_cnt, SUM(adver_cnt) AS a_cnt FROM pie_chart_stats WHERE insert_date != '0000-00-00'";
+	if ($res_select_pie_chart_stats = mysqli_query($db_conn, $sql_select_pie_chart_stats)) {
+		$arr_select_pie_chart_stats = mysqli_fetch_array($res_select_pie_chart_stats);
+		$total_query = $arr_select_pie_chart_stats['q_cnt'];
+		$total_adver = $arr_select_pie_chart_stats['a_cnt'];
+	}else{
+		errorlog("SQL query failed for some reason: $sql_select_pie_chart_stats");
+	}
+	//total update
+	$sql_update_pie_chart_stats = "UPDATE pie_chart_stats SET query_cnt = $total_query, adver_cnt = $total_adver WHERE insert_date = '0000-00-00'";
+	if (!mysqli_query($db_conn, $sql_update_pie_chart_stats)) {
+		errorlog("SQL query failed for some reason: $sql_update_pie_chart_stats");
+	}else{
+		debuglog("SQL query successfully executed: $sql_update_pie_chart_stats");
+	}
+}
+
+//---------------------------------------------------------------------------------------------------
+//main
+
+debuglog("Script is started");
+$file_name = (isset($argv[1])) ? $argv[1] : $file_name;
+if (!file_exists($file_name)) {
+	errorlog("Provided log file does not exist ".$file_name);
+}else{
+	debuglog("Trying to read and parse log file ".$file_name);
+}
+
+$pie_chart_date = date("Y-m-d");
+$pie_chart_count_query = 0;
+$pie_chart_count_adver = 0;
+debuglog("Current date == ".$pie_chart_date);
+debuglog("Initial amount of 'query' regexp for ".$pie_chart_date." == ".$pie_chart_count_query);
+debuglog("Initial amount of 'adver' regexp for ".$pie_chart_date." == ".$pie_chart_count_adver);
+$top_chart_array = array();
+
+$file_open = fopen($file_name, "r");
+while (!feof($file_open)) {
+	$file_line = fgets($file_open);
+	if (strpos($file_line, $pie_chart_regexp_query)) debuglog("\tIncremented amount of 'query' regexp for ".$pie_chart_date." == ".++$pie_chart_count_query);
+	if (strpos($file_line, $pie_chart_regexp_adver)) {
+		debuglog("\tIncremented amount of 'adver' regexp for ".$pie_chart_date." == ".++$pie_chart_count_adver);
+		array_push($top_chart_array, parse_top_chart_advertiser($file_line));
+	}
+}
+fclose($file_open);
+
+debuglog("Total amount of 'query' regexp for ".$pie_chart_date." == ".$pie_chart_count_query);
+debuglog("Total amount of 'adver' regexp for ".$pie_chart_date." == ".$pie_chart_count_adver);
+
+//---------------------------------------------------------------------------------------------------
+//mysql
+
+debuglog("** Trying to connect to database with db_host = $db_host");
+debuglog("** Trying to connect to database with db_user = $db_user");
+debuglog("** Trying to connect to database with db_pass = __skipped");
+debuglog("** Trying to connect to database with db_name = $db_name");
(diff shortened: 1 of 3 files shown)


DonLexos:
This:
mysql -u root -p"raspberry" < /usr/local/bin/setupPiholeStats.sql

Might now work is someone has changed his pi/root password away from raspberry (as you should do during initial setup imho).

DonLexos:
And I noticed that the long terms stats script has the username/password in the file, so this might need an extra step to ask for the password and then add it in the file itself?

It might be a good idea to create a database and user/pass combination somewhere in that process that is not root. 

Next to that, not to complicate things, in general the MySQL root password can (should) be a different password than the normal root password (provided it runs mysql_secure_install after the mySQL installation).

dschaper:
How about a nosql or sqlite so there wouldn't be the overhead of having to run a full MySQL package? I wish I could get my hands on a Zero so I could test the load... (May have to get a single core B+ and clock it so it behaves like a Zero instead, MicroCenter carries them rather inexpensive.)

DonLexos:
I am not sure how resource intense mySQL is if it would be only used if the admin page is displayed. You could fetch the statistics daily or every X hours, not sure if that should be updated nearly realtime (which could add some load).

dschaper:
Is there a solution that would work and not require another password to be added to the mix? The system load and having to handle another password and possible user account were my only two concerns.

extremeshok:
Why must mysql be used ?

You can use json or xml or a basic serialized array to store the values.

Update them with a php shell script.

Avoid MySQL as it eats ram.

__________________________________.    https://eXtremeSHOK.com     .__________________________________

DonLexos:
Would in the end mySQL be quicker if you want to lookup things (totals, most used domains)? Right now after just one day my pihole.log is nearly 7 MB. I understand that a lot of the data is purged when you only save the important information but would an XML not become large rather quickly?

jacobsalmela:
I'm open to suggestions.  This is just what I have and had working.

DonLexos:
Since my pi is dedicated for only running pi-hole I would not mind. I can try to set this up to see if it parses the files, is there any code available for output so I can see if it works? Otherwise I think I will get phpmyAdmin since I suck in commandline mysql to see if the database is populated.

jacobsalmela:
@DonLexos, the `lrs.php` script is set to verbose by default, so you should lots of `matter` if you have a decent log file

DonLexos:
ah ok excellent.. On my way then :-)

DonLexos:
Works like a charm! (Did install phpmyadmin to browse the database though). Right now it does not seem to purge the /var/log/pihole.log, I reckon running it again would simply add all the records again?

screenshot

extremeshok:
Whats currently stored, I doubt it. I noticed you want to add more 
metrics, so use SQLite, which is built into php5, honestly MySQL is a 
rape of resources on these little machines.

We run a 150 node/machine monitoring system with full metrics in a 
SQLite database, its light on resources.

As to the logging.. obviously you want to kill your SDcards quickly..

log to tmpfs/ram drive and do your metrics/accounting/stats generation 
ever X minutes.

This way only the stats are incremented, the logs never burn the sdcard.

__________________________________.    https://eXtremeSHOK.com     .__________________________________

DonLexos:
To retrace my steps and an attempt to help out documenting this:

First get get mysql + php5. When prompted set the root password to something (..else than your piroot AND something you can remember later)
```
apt-get install mysql-server php5-mysql
```

Secure your mySQL installation
```
mysql_secure_installation
```
- no need to change the root password (you just set it)
- disable anonymous users **Yes**
- disallow root login remotely **Yes**
- remove test database and access to it **Yes**
- reload privilege tables now **Yes**

Then add the database information, login to the mysql this way
```
mysql -u root -p
```
And then add the SQL like so (just paste it in)
```sql
CREATE DATABASE charts;
USE charts;
DROP TABLE IF EXISTS `pie_chart_stats`;
CREATE TABLE `pie_chart_stats` (
	`insert_date` DATE NOT NULL COMMENT 'date when values were captured',
	`query_cnt` INT(10) UNSIGNED NOT NULL COMMENT 'query count appreaprence',
	`adver_cnt` INT(10) UNSIGNED NOT NULL COMMENT 'advertiser count apperance',
	PRIMARY KEY (`insert_date`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COMMENT 'table to store stats for pie (queries vs advertisers) chart';

INSERT INTO `pie_chart_stats` VALUES ('0000-00-00', 0, 0);

DROP TABLE IF EXISTS `top_chart_stats`;
CREATE TABLE `top_chart_stats` (
	`insert_date` DATE NOT NULL COMMENT 'date when values were captured',
	`adver_name` VARCHAR (60) NOT NULL COMMENT 'advertiser domain name',
	`cnt` MEDIUMINT(7) NOT NULL COMMENT 'advertiser count apperance',
	PRIMARY KEY (`insert_date`, `adver_name`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```

still in mysql, add a user for the pihole admin 
```sql
GRANT ALL ON charts.* TO 'piholeadmin'@'localhost' IDENTIFIED BY 'fancypassword';
FLUSH PRIVILEGES;
```

and then quit the mysql
```sql
quit;
```
Make sure you get the latest copy of lrs.php , in the file add your user and password on line 16 + 17
```
$db_user = "piholeadmin";
$db_pass = "fancypassword";
```
and put it in /usr/local/bin, then run it
```
sudo php /usr/local/bin/lrs.php
```



sudo apt-get install phpmyadmin

extremeshok:
Replace phpmyadmin with adminer.... Think resources. 
https://www.adminer.org/

__________________________________.    https://eXtremeSHOK.com     .__________________________________

DonLexos:
@extremeshok ideas on how to make the pihole.log be in RAM instead of /var/log ? If you do log to ram a restart will flush it all or not?

extremeshok:
shutdown mysql
run: free -m -h

Start mysql
run: free -m -h

do some db queries
run: free -m -h

Decide if you want to keep mysql ;-P

__________________________________.    https://eXtremeSHOK.com     .__________________________________

extremeshok:
assuming you have udev, use /dev/shm/pihole.log

I would run the cron @ 15min intervals, does it matter if you lost 
15mins of previous logs ?

you can also extend the shutdown to run the stat generation on shutdown.

__________________________________.    https://eXtremeSHOK.com     .__________________________________

DonLexos:
```
pi@raspberrypi:/tmp $ sudo /etc/init.d/mysql stop
pi@raspberrypi:/tmp $ free -m -h
             total       used       free     shared    buffers     cached
Mem:          925M       513M       412M        14M        44M       400M
-/+ buffers/cache:        68M       856M
Swap:          99M         0B        99M

pi@raspberrypi:/tmp $ sudo /etc/init.d/mysql start
[ ok ] Starting mysql (via systemctl): mysql.service.
pi@raspberrypi:/tmp $ free -m -h
             total       used       free     shared    buffers     cached
Mem:          925M       545M       380M        14M        44M       400M
-/+ buffers/cache:       100M       825M
Swap:          99M         0B        99M
```
32 MB extra if my math is correct :-) 
I get where you are from though, I tjink @jacobsalmela is just starting up the extra stats right now, stats in /admin are all taken from pihole.log, this is tagged **help-wanted**. I am unfamiliar with SQLite  but will check it out

DonLexos:
Do I understand correctly that your idea is to add pihole here:
 /dev/shm/pihole.log
Add a SQLite version of lrs.php to cron every 15 minutes and then simply purge the log as soon as that ran?

extremeshok:
Exactly. We use this for our log collection stuff, so i know it works 
from experience.

Regarding sqlite, the syntax is almost identical to mysql.
http://www.tutorialspoint.com/sqlite/sqlite_php.htm

__________________________________.    https://eXtremeSHOK.com     .__________________________________

jacobsalmela:
Cool.  I'll probably need some help making the conversion to something better.

DonLexos:
After a reboot I was back at 750 MB free ram btw. If SQLite is less resource intense and it works almost the same there is no point in adding MySQL to the dependencies imho.

Regarding my question earlier if running lrs.php again and again would double/tripple the results, @jacobsalmela already made sure that is not happening I noticed just now, nice work!

extremeshok:
If your still not convinced ;-P

http://www.sqlite.org/whentouse.html

__________________________________. https://eXtremeSHOK.com 
.__________________________________

DonLexos:
Any chance you can see if you can get lrs.php at work in an SQLite environment? I am reading the tutorial but this involves a bit more than just fiddling with a few things I am afraid (for an SQLite challenged person like me)

DonLexos:
@jacobsalmela if you rerun the script it does update the top_chart_stats table but the pie_chart_stats are not updated. It seems to use INSERT IGNORE INTO where the others are different. Is there a specific reason for this?

```sql
[2016-01-21 21:27:57][1536][debug] SQL query successfully executed: INSERT IGNORE INTO pie_chart_stats VALUES ('2016-01-21', 16346, 1338)

[2016-01-21 21:27:57][1536][debug] SQL query successfully executed: REPLACE INTO top_chart_stats VALUES ('2016-01-21', 'data.flurry.com', 95)

```

rkallos:
Thanks for all your work on Pi-hole. I've been motivated to contribute thanks to the great service it already provides :-)

I've mostly finished the migration from mysql to sqlite using /dev/shm/pihole.log instead of /var/log/pihole.log. You can see the changes [in the dev-stats branch of my pi-hole fork](https://github.com/rkallos/pi-hole/commits/dev-stats).

The migration has brought up an interesting problem regarding permissions for the database file. From what I understand, Sqlite needs write access to the directory in which the database file is stored, not only the db file. If the database file is stored in /usr/local/bin, then the main user (assumed to be pi) needs write access to that dir.

Do you have any suggestions for getting around this? The simplest (but probably not the best) solution might be moving the db to /home/pi/.pihole or a similar directory.

dschaper:
Thanks for the words of support! It might be better to open a new issue for this, so we can track it separately. We'll definitely get back to you when the other members of the team check in!

jacobsalmela:
We have been trying to follow the FHS by keeping stuff in `/usr/local` since it is a third-party software installed by an administrator.  The permissions will be a good challenge to overcome for both the Web interface but also for supporting other platforms.  If you don't install it on a Pi, the `pi` user might not exist.

I'm far from an expert on SQL and/or databases in general, but I thought if I posted what I had someone like @rkallos would jump in and convert it to something better. 

The database and script work as-is, but definitely needs some improvements.  There have been a lot of issue and PRs coming in so we still need to sort through and prioritize them, but please keep stuff coming!

Our main focus is on stability, so fixing bugs before adding new features.

rkallos:
After doing a bit more research on SQLite, I discovered why it requires write permissions to any directory in which a database file is stored. SQLite creates and maintains a journal file in the same dir as a database file for the sake of atomic commit/rollback. A journal file is usually created when a transaction starts, and deleted when it finishes.

Fortunately, SQLite offers the journal_mode pragma for dealing with a journal file, and it is possible to keep a persistent journal file. This means that we can set access permissions to the two files and store them anywhere we like. It could make sense to store the db and journal files in `/usr/local`, or it could be stored someplace else like `/var/lib/`.

This week, I'm going to try and merge my fork with upstream, and then I'm going to play around with SQLite again, trying to get the database file working in a dir that makes sense.

jacobsalmela:
@rkallos sounds good.  Let us know what you work out.  I'm from a SQL master but I just wanted to put out what I had to spark some ideas.

rkallos:
I've made a bit more progress on this. I'm currently testing this on my home network. I've mostly moved files around manually, so nothing is reflected in the installation script (yet).

- I've moved pihole.log to /dev/shm/pihole.log
- I've moved the SQLite DB file to /var/lib/pihole/stats.db
- I've moved lrs.php to /var/www/html/admin
- I've moved setupPiholeStats.sql to /usr/local/bin. I am not sure if this is where it should live.
- I've updated piholeLogFlush.sh to create stats.db if it doesn't exist, and run lrs.php before emptying the log file.
- Currently, SQLite (and lrs.php) are run as root, since they're called via piholeLogFlush.sh by cron. As a result, permissions are not currently an issue. If you guys create a pihole user, then we might have to address this more thoroughly.

I've been keeping pihole.log in RAM for a little over a week with no issues. piholeLogFlush.sh has been writing to stats.db every 15 minutes for about an hour, so it could still use some testing.

If nothing breaks, I will write the changes into the install script in my fork sometime next week.

Thanks again for Pi-hole. It's been really great to use :-)

Mcat12:
Maybe keep the SQL file in the git directory, if it's just used to set up the database.

jacobsalmela:
`setupPiholeStats.sql` is really only needed once.  Once it is run, it can be deleted.  It just sets up the database and tables.

rkallos:
True enough. My concern is that if stats.db is deleted somehow, then lrs.php will complain every time it is run (every 15 minutes, on my system), as might index.php, if the DB is integrated with the dashboard there. I think it would be a good idea to keep the SQL file somewhere among the installed files in case the database needs to be recreated.

jacobsalmela:
That's true.  I suppose if something was corrupted it would be nice to be able to rebuild it.

jacobsalmela:
I am also trying to figure out if this PR and [this one from AdminLTE](https://github.com/pi-hole/AdminLTE/pull/25) need to work together...

Mcat12:
When this gets pulled, lots of things will need to be reworked.

DonLexos:
@jacobsalmela yes rebuilding, but maybe even flushing and starting over if you would want to. 

Next to that, I tried the MySQL version, it counts the amount of iterations of the ad.domain.com which is very nifty. This would not be compatible with the 'last ads/last request' from the AdminLTE work that you refer to since you collect and purge some information. 

Maybe if there is a working SQLite version then check what information is stored and how, maybe switch or change a few things and then work on displaying the information?

jacobsalmela:
Closing in favor of [AdminLTE #25](https://github.com/pi-hole/AdminLTE/pull/25)

PR #220, opened 2016-01-21
Use Sudo Detection from Installer in Gravity.sh Also

xd1936:
Prevents errors from being thrown when gravity.sh is run as root and the user doesn't have sudo installed (just like install.sh does)

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -14,13 +14,13 @@
 if [[ $EUID -eq 0 ]];then
 	echo "You are root."
 else
-	echo "::: sudo will be used for the install."
+	echo "::: sudo will be used."
   # Check if it is actually installed
   # If it isn't, exit because the install cannot complete
   if [[ $(dpkg-query -s sudo) ]];then
 		export SUDO="sudo"
   else
-		echo "::: Please install sudo or run this as root."
+		echo "::: Please install sudo or run this script as root."
     exit 1
   fi
 fi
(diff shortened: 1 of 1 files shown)


gravity.sh
```diff
@@ -10,8 +10,24 @@
 # the Free Software Foundation, either version 2 of the License, or
 # (at your option) any later version.
 
+# Run this script as root or under sudo
+if [[ $EUID -eq 0 ]];then
+	echo "You are root."
+else
+	echo "::: sudo will be used for the install."
```
dschaper: Thanks for the submission, can you take this one line out, I think it may have been left over from the copy from `basic-install.sh`. It's out of place since `gravity.sh` doesn't install anything.

xd1936:
Ha. Whoops. Copy-and-Paste left the *to install* verbiage in there

dschaper:
It happens, not a problem.

PromoFaux:
Gravity does not need to be run as root... The install script no longer runs gravity as root, so the files belong to pi now and not root..

dschaper:
So `gravity.sh` still needs to run as root? Is that just for the dnsmasq reload.

xd1936:
I guess I should have changed this pull request to say,
"remove sudo requirement so systems without sudo installed don't get a `command not found: sudo` error when they're running gravity.sh"

PromoFaux:
Yeah I realised that after I commented and the re-read your commits!

dschaper:
Alpha branched to xd1936-bugfix/sudo.

dschaper:
Merged into development via another PR.

Thanks @xd1936 for the fix!

PR #229, opened 2016-01-23
Changed URLs in installScripts() to match new organisation links

PromoFaux:


PromoFaux:
@dschaper @jacobsalmela Can you just sanity check the changes on this please before we merge it to the master...?

PR #234, opened 2016-01-24
Add unzip to the dependencies

ChadBHowell:
unzip is required for the install script to complete.  Some distros do
not have it installed by default.


PromoFaux:
Hey Chad, thanks for this, I've actually added it into the next iteration of the install script that i'm currently working on.

I'll pull this over to the Dev branch for now, however, in case I don't get that out for a few days.

Thanks!

ChadBHowell:
Thanks!

PR #235, opened 2016-01-24
Upsteam DNS enhancement

ChadBHowell:
The purpose of these commits is to do the following:

1) Bring the dnsmasq.conf changes to be inline with how config files are modified on current Debian type systems by using the "*.d" directory. This allows for a cleaner distro upgrade path and avoids stomping our config changes during dnsmasq upgrades.

2) Allow the user to choose between Google and DynDNS upstream servers during the install. This will later be extended to allow the user to enter custom upstream DNS servers as well.

PR #236, opened 2016-01-24
Install script rework

PromoFaux:
This install script is a lot more intelligent than the last, which makes it safer to use as an update script.

The git repos for both pi-hole and the web interface are now cloned on initial install. Subsequent runs of `basic-install.sh` perform a `git pull` on each repo to get the latest updates, and then distribute the files to where they need to be.

An additional check at the beginning of the script checks to the last date that apt-get update was run, and will skip updating apt-get if it has already been run today. The rest of the script will carry on as normal.

Other scripts outputs changed to match install script.

TODO:
- [x] `basic-install.sh`
  - [X] Grab base files from github to `/etc/.pihole` (or update)
  - [X] Grab Web Admin files from github to `/var/www/html/admin` (or update)
  - [X] Check if dependencies are already installed before trying to install
  - [X] Make output easier to read
  - [x] ~~Implement `dnsmasq.conf` naming as suggested in #235~~ going to leave this to @ChadBHowell 
  - [X] Perform date check before running `apt-get update`/`apt-get upgrade`
- [x] `gravity.sh`
  - [x] Tidy up output to match install.sh
- [x] `whitelist.sh`
  - [x] faster whitelisting (I have a plan!)
- [x] `blacklist.sh`
  - [x] Tidy up output to match all of the above

An idea of how the output looks at moment:
image

### DIFF
--- a/advanced/Scripts/blacklist.sh
+++ b/advanced/Scripts/blacklist.sh
@@ -81,13 +81,14 @@ function AddDomain(){
 	if $bool; then
 	  #domain not found in the blacklist file, add it!
 	  if $versbose; then
-	  echo "** Adding $1 to blacklist file"
+	  echo -n "::: Adding $1 to blacklist file..."
 	  fi
 		echo $1 >> $blacklist
 		modifyHost=true
+		echo " done!"
 	else
 	if $versbose; then
-		echo "** $1 already blacklisted! No need to add"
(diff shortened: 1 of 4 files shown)


jacobsalmela:
This is awesome @PromoFaux.  I thought I knew a lot about `bash` but you're teaching me things!

PromoFaux:
@jacobsalmela I'm learning as I go along! You probably still know more than me...

PR #240, opened 2016-01-25
Upsteam DNS enhancement v2

ChadBHowell:
Rebased against latest devel branch per @dschaper

2) Allow the user to choose between Google and DynDNS upstream servers during the install. This will later be extended to allow the user to enter custom upstream DNS servers as well.


dschaper:
Thanks for the PR. One thing I'd like to see is that on existing installs there is the `/etc/dnsmasq.conf` file that has the current configuration, and a copy of the original supplied config at `/etc/dnsmasq.conf.orig`. I'd like the script to check for the existence of those two files, and if they exist, move `/etc/dnsmasq.orig` back to `/etc/dnsmasq` and then drop the `01-pihole.conf` into the proper directory. This will help with preventing any breakage of existing installs. Is that something you could add to the process?

ChadBHowell:
I thought about this earlier and a problem that I see is as follows:
1) user installs pihole for the first time and the original dnsmasq.conf moves to dnsmasq.conf.orig.
2) user updats pihole to the latest version and the first installs dnsmasq.conf overwrites dnsmasq.conf.orig and the original file is lost.

I think the best we can do is pop a warning saying a default conf was installed and mv the backup to dnsmasq.conf.upgrade.  This would preserve the original dnsmasq.conf.orig and give us a backup of the current file.  This would only happen when someone upgrades pihole the first time.  After that we would be using the 01-pihole.conf.

automated install/basic-install.sh
```diff
@@ -90,7 +90,7 @@ backupLegacyPihole(){
 		echo "Original Pi-hole detected.  Initiating sub space transport"
 		$SUDO mkdir -p /etc/pihole/original/
 		$SUDO mv /etc/dnsmasq.d/adList.conf /etc/pihole/original/adList.conf.$(date "+%Y-%m-%d")
-		$SUDO mv /etc/dnsmasq.conf /etc/pihole/original/dnsmasq.conf.$(date "+%Y-%m-%d")
+		$SUDO mv /etc/dnsmasq.d/01-pihole.conf /etc/pihole/original/dnsmasq.d/01-pihole.conf.$(date "+%Y-%m-%d")
```
dschaper: This doesn't need to be modified. The backupLegacyPihole() only was to check for Pi-Hole v1, and won't be triggered unless someone is running very old code. This function is set for depreciation soon anyways.

automated install/basic-install.sh
```diff
@@ -260,6 +260,41 @@ setStaticIPv4(){
...
+	esac
+}
+
++versionCheckDNSmasq(){
+	# Check if /etc/dnsmasq.conf is from pihole.  If so replace with an original and install new in .d directory
+	dnsFile="/etc/dnsmasq.conf"
+	dnsSearch="addn-hosts=/etc/pihole/gravity.list"
+	
+	if grep -q $dnsSearch $dnsFile; then
+		$SUDO mv /etc/dnsmasq.conf /etc/dnsmasq.conf.orig
```
dschaper: Here, if dnsmasq.conf is a Pi-Hole conf then there already will be a dnsmasq.conf.orig, so we don't want to nuke the file. How about a 3 way switch? Copy dnsmasq.conf to dnsmasq.conf.pihole, then copy the dnsmasq.conf.orig back to dnsmasq.conf...

automated install/basic-install.sh
ChadBHowell: Ok.  I will revert it back.

dschaper: Thanks, it's kind of confusing and there probably should be comments in there to note what the function is really designed for...

dschaper:
I didn't see the versionCheckDNSmasq() before I made those earlier comments. I blame the bad coffee water I'm drinking. (I'll fix the coffee issue tomorrow...) I made some line notes and we can hash those out in there...

ChadBHowell:
How about something like this:

```bash
versionCheckDNSmasq(){
	# Check if /etc/dnsmasq.conf is from pihole.  If so replace with an original and install new in .d directory
	dnsFile1="/etc/dnsmasq.conf"
	dnsFile2="/etc/dnsmasq.conf"
	dnsSearch="addn-hosts=/etc/pihole/gravity.list"
	
	# Check dnsmasq.conf for pihole magic
	if grep -q $dnsSearch $dnsFile1; then
		# If true, Check dnsmasq.conf.orig for pihole magic
		if grep -q $dnsSearch $dnsFile1; then
			# If true, use advanced/dnsmasq.conf.original
			$SUDO mv /etc/dnsmasq.conf /etc/dnsmasq.conf.orig
			$SUDO cp /etc/.pihole/advanced/dnsmasq.conf.original /etc/dnsmasq.conf
		else
			# If false, mv original file back
			$SUDO mv /etc/dnsmasq.conf /etc/dnsmasq.conf.orig
	# If false, This is a fresh install
	fi
	$SUDO cp /etc/.pihole/advanced/01-pihole.conf /etc/dnsmasq.d/01-pihole.conf
	$SUDO sed -i "s/@INT@/$piholeInterface/" /etc/dnsmasq.d/01-pihole.conf
	$SUDO sed -i "s/@DNS1@/$piholeDNS1/" /etc/dnsmasq.d/01-pihole.conf
	$SUDO sed -i "s/@DNS2@/$piholeDNS2/" /etc/dnsmasq.d/01-pihole.conf
}
```

dschaper:
Looks good, just make sure your inner `if` is `fi`d.

ChadBHowell:
Good catch on the fi. This is me coding without coffee again...

dschaper:
I just took it to be very well formatted pseudo-code ;)

ChadBHowell:
Ok, that should be the last of the coffeeless mistakes... Done for the night, I will fix anything else tomorrow.

dschaper:
https://github.com/pi-hole/pi-hole/tree/CBH/UpstreamDNS

ChadBHowell:
Merged latest and fix slight merge error.

dschaper:
Okay, I'll create a new alpha branch. Thanks again.

ChadBHowell:
Updates transferred to the alpha branch #267

PR #242, opened 2016-01-25
Add user pihole if user does not exsist

ChadBHowell:
Check if user pihole exist and add if not.  This will give pihole its own system user without a login or home directory that is not system dependent.

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -431,6 +431,16 @@ runGravity(){
 	
 }
 
+setUser(){
+	# Check if user pihole exists and create if not
+	echo "::: Checking if user 'pihole' exists..."
+	if id -u pihole > /dev/null 2>&1; then
+		echo "::: User 'pihole' already exists"
+	else
+        echo "::: User 'pihole' doesn't exist.  Creating..."
+		$SUDO useradd -r -s /usr/sbin/nologin pihole
+	fi
+}
 
 installPihole(){
 	checkForDependencies # done
@@ -438,7 +448,7 @@ installPihole(){
 	
 	$SUDO chown www-data:www-data /var/www/html
 	$SUDO chmod 775 /var/www/html
-	$SUDO usermod -a -G www-data pi
+	$SUDO usermod -a -G www-data pihole
 	$SUDO lighty-enable-mod fastcgi fastcgi-php > /dev/null
 	
 	getGitFiles	
--- a/gravity.sh
+++ b/gravity.sh
@@ -228,7 +228,7 @@ function gravity_Whitelist() {
 	do
         tmp=$(echo "$url" | awk -F '/' '{print $3}')
         urls=("${urls[@]}" $tmp)
-	done & spinner $!
+	done
 	echo " done!"
 	
 	echo -n "::: Running whitelist script to update HOSTS file...."


ChadBHowell:
Rebased

ChadBHowell:
@PromoFaux This is the last thing that I have identified that stops piHole from being portable between disrtos and other hardware besides the Raspberry Pi is the dependency on the user "pi".  This PR would take care of that.  What other desting do you need done before we push this to the Development branch?

PromoFaux:
I just need myself, @Mcat12 , @dschaper , or @jacobsalmela  to test it before it's merged in...

I need to get around to installing virtualbox and debian/ubuntu or something!

dschaper:
I need to test it out, one of the things we built Pi-hole on is stability before enhancements. I'd really like the ability to run Pi-hole as a dedicated user, so your PR is high on the list, but we need to make sure that the other bugs are squashed and the code is stable before introducing another variable like a new user. We will get to this PR but we need to make sure everything is in place and ready first! I do appreciate your submissions and hope you understand our process.

dschaper:
I'm good with this, for right now it will just add the user, but not do anything with it... Yet...

PromoFaux:
^That's good enough for me. Merging in 3...2...

Mcat12:
Does the method actually get called?

Mcat12:
Fixed with latest commit.

DonLexos:
I believe this is in release 2.5.3?

PromoFaux:
Yep, it's been merged and closed... :)

DonLexos:
Doh! Apologies ... :-|

manuelgu:
Is that toggleable? I have pihole installed under 'hole' so I think the new update will now basically rip everything, right?

PromoFaux:
As far as I'm aware, you should not run into any issues.

It's (currently) only used in one place. In the [previous version](https://github.com/pi-hole/pi-hole/blob/8e289215b4f937795cfb4b1488895d0353cc7cf3/automated%20install/basic-install.sh#L554) we were usermodding `/var/www` to the user `pi`. This was fine on Rasbian systems, but not other flavours of linux.

In the [new version](https://github.com/pi-hole/pi-hole/blob/master/automated%20install/basic-install.sh#L669) we usermod that folder under to the user `pihole`, which we create if it does not exist.

As far as I'm aware, we change nothing else in terms of usernames

PromoFaux:
You could maybe try renaming `hole` to `pihole` following [this guide](http://www.cyberciti.biz/faq/howto-change-rename-user-name-id/)?

manuelgu:
Looks like it didnt override any permissions (at least `find / -group pihole` doesn't output anything).
Though I dont really want to move from `hole` to `pihole` as the user. Would really appreciate if either the name would be toggleable or the checking etc. in general :)

PromoFaux:
@dschaper - any pointers?

dschaper:
So the question is can we have the install prompt the user for the username? We were previously just glomming onto the existing `pi` user so @manuelgu if you have the install customized to run under a different user, you shouldn't see any difference. We shouldn't be stomping on an existing install. I don't think you need to move from `hole` to `pihole` as you already have enough knowledge of the scripts to be able to run under a user besides `pi`, you know what it takes to get things running. The new `pihole` user is meant to currently be transparent to current users, as it's for planned security enhancements that are in the pipeline. (And as @PromoFaux mentioned, it allows us to install on platforms that don't have an existing `pi` user...)

manuelgu:
> So the question is can we have the install prompt the user for the username

That would be really cool! 

I can see your point there, thanks :)

PromoFaux:
Prompts in the script will only work provided the script is not curl'd (I hear cheering, who's cheering?)

dschaper:
And when I can ever get the debian package working then `debconf` can handle the work...

koen84:
> So the question is can we have the install prompt the user for the username ?
Alternative is to define a variable (in the beginning of the script ?  Default it would be set as pihole, but if someone wants it differently, they can simply change it.

PromoFaux:
We could provide a whiptail dialog, offering the user a choice.

Once we have the code refactor out of the way, we can start to look at it in some more detail.

dschaper:
Just a question, but what happened to all the themeing??!! @jacobsalmela We've lost the blackhole theme :(

PR #244, opened 2016-01-25
Development

PromoFaux:
Checks for packages that needs updating, if there are any, gives user the choice to cancel installation and run apt-get update/upgrade.

Cosmetic change: Added dummy domain to the top of hosts file `Pi-Hole.IsWorking.OK` so pinging a blocked domain doesn't come back with something that looks suspect to the unaware user.

Added a new blocklist - adds in a further ~20,000 domains, if it causes problems, we can pull it.

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -293,24 +293,37 @@ stopServices(){
 }
 
 checkForDependencies(){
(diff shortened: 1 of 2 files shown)


PromoFaux:
@dschaper Confirm for merge please.

PR #255, opened 2016-01-27
Changes as in #253, but with less commits to keep the repo tidy

PromoFaux:

PR #249, opened 2016-01-26
Operation block list from gravity

PromoFaux:
Fixes/address #247

This is mainly for the advanced users. 

`Gravity.sh` now checks for the existence of `/etc/pihole/adlists.list`. 
If it finds it, then it will load it's list of adlists from the urls contained therein (one per line). 
If not, it will get the domains from `/etc/pihole/adlists.default`

PR #256, opened 2016-01-27
Install script critical bug fix. Stop dependencies from erroring when installing on a fresh pi

PromoFaux:


Mcat12:
Continuing from #255

Mcat12:
You should combine this with #249

PromoFaux:
I'm going to leave #249 til tomorrow. This is priority

PR #258, opened 2016-01-28
Update piholeLogFlush.sh

gh0stshell:
Tested 6 different codes to flush the script, this was the only one I was able to come up with that did not stop the queries from freezing on the LCD or if letting the script run before and then after the flush cron runs.


gh0stshell:
Cleaned up line 35 on the chrono code.

dschaper:
Thanks for the PR, I'll pull this to an alpha branch in a bit and test out the code tonight.

dschaper:
Branched to https://github.com/pi-hole/pi-hole/tree/gh0stshell/Bugfix/LogFlush. New PR opened.

PR #263, opened 2016-01-29
New versioning system

Mcat12:
Currently, the version number is shared between the core Pi-hole code and the web interface. Since these are on different repos and have different development paces, the versioning doesn't fully match the amount of changes in each. This pr will give the Pi-hole and web interface separate version numbers (the web interface pr is here (to be added when pr is created)). This pr is still not completely ready to be merged, but was opened for discussion (which is what Github recommends anyways).


PromoFaux:
Why not just keep it in `/etc/.pihole/`?

Mcat12:
That's fine too

PromoFaux:
This is all you need for now, yeah? Just to have the file existing?

Mcat12:
Yep, and have it updated with new versions. I'll start working on the web side soon

PromoFaux:
Cool, then I will bring it in to development!

PR #267, opened 2016-01-30
CBH/Upstream DNS

ChadBHowell:
New PR for alpha branch in Devel branch.  Continuing work from #240.


Mcat12:
I'm updating this with the latest development stuff.

ChadBHowell:
Thanks!

automated install/basic-install.sh
```diff
@@ -285,10 +330,10 @@ installConfigs() {
 	# Install the configs from /etc/.pihole to their various locations
 	$SUDO echo ":::"
 	$SUDO echo -n "::: Installing configs..."
-	$SUDO mv /etc/dnsmasq.conf /etc/dnsmasq.conf.orig
+	versionCheckDNSmasq
 	$SUDO mv /etc/lighttpd/lighttpd.conf /etc/lighttpd/lighttpd.conf.orig
+	$SUDO cp /etc/.pihole/advanced/lighttpd.conf /etc/lighttpd/lighttpd.conf 
 	$SUDO cp /etc/.pihole/advanced/dnsmasq.conf /etc/dnsmasq.conf
```
dschaper: This line causes a conflict with dnsmasq. We end up with a config file in `/etc/dnsmasq.d/01-pihole.conf` and `/etc/dnsmasq.conf`. `dnsmasq` doesn't like having keywords repeated, so it throws an error.

Mcat12: I think that was me in the merge conflict fixing...

dschaper: Okay, just need to take out line 336 and 337.

Mcat12: 337 too?

automated install/basic-install.sh
```diff
@@ -267,6 +267,51 @@ setStaticIPv4() {
...
+}
+
+versionCheckDNSmasq(){
+	# Check if /etc/dnsmasq.conf is from pihole.  If so replace with an original and install new in .d directory
+	dnsFile1="/etc/dnsmasq.conf"
+	dnsFile2="/etc/dnsmasq.conf.orig"
+	dnsSearch="addn-hosts=/etc/pihole/gravity.list"
+	
+	# Check dnsmasq.conf for pihole magic
+	if grep -q $dnsSearch $dnsFile1; then
```
dschaper: 'grep -q` throws an error when a file is non-existant. Think this needs a `grep -s`?

ChadBHowell: The function checkForDependencies() is run before versionCheckDNSmasq() in installPihole() and there for we can be certain that dnsmasq.conf that comes with the package is installed.  If not, we have other issues with the install. 

GNU had this about the -s switch:

> 
-s
--no-messages
Suppress error messages about nonexistent or unreadable files. Portability note: unlike GNU grep, 7th Edition Unix grep did not conform to POSIX, because it lacked -q and its -s option behaved like GNU grep's -q option.1 USG-style grep also lacked -q but its -s option behaved like GNU grep's. Portable shell scripts should avoid both -q and -s and should redirect standard and error output to /dev/null instead. (-s is specified by POSIX.)
>

 If you would like, we can wrap that grep with something like this:
```bash
if [ -d "/etc/dnsmasq.conf" ]; then
		grep statement
	else
		we are missing dnsmasq package
fi 
```
or pipe the errors to /dev/null

dschaper:
Well, portability would be nice, so my thoughts would be to follow GNU and remove the switches and go ahead with a conditional wrap.

ChadBHowell:
Fixed in PR #270

dschaper:
Thanks! Things look good from my end. I'll bring it up for merge with the team.

ChadBHowell:
Great, thanks for the work!

pralor:

https://discourse.pi-hole.net/t/why-is-google-the-first-dns-choice/33839/20

PR #274, opened 2016-01-31
Update Dev from master

PromoFaux:


Mcat12:
Is this for the license update?

PromoFaux:
Just so that dev is in line with anything new pushed to master. That way when we want to merge dev back into master again, it's less hastle.

Mcat12:
Ok

PR #280, opened 2016-02-02
ATV Hulu domain add

CDeeRON:
added ads-e-darwin.hulu.com which I found to be used when serving ADs on ATV.

### DIFF
--- a/block hulu ads/lighttpd.conf
+++ b/block hulu ads/lighttpd.conf
@@ -38,6 +38,6 @@ include_shell "/usr/share/lighttpd/use-ipv6.pl " + server.port
 include_shell "/usr/share/lighttpd/create-mime.assign.pl"
 include_shell "/usr/share/lighttpd/include-conf-enabled.pl"
 
-$HTTP["host"] =~ "ads.hulu.com|ads-v-darwin.hulu.com" {
+$HTTP["host"] =~ "ads.hulu.com|ads-v-darwin.hulu.com|ads-e-darwin.hulu.com" {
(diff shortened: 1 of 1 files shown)


jacobsalmela:
We haven't done much work on Hulu ads yet, at it has been quite an undertaking.  Have you been able to _skip_ the ads or just mute them?  

I don't know that too many people use the file you are editing as it has kind of fallen to the wayside.

CDeeRON:
I've been testing on AppleTV last two days. So far its a hit or miss playing the MP4 file. Appears Hulu is checking to be sure the AD is in the right video format.  Also found that the hulu ad url in lighttpd.conf should be setup to be blocked with pihole in order for the redirection done by lighttpd to work.

CDeeRON:
To answer you question it skips right over the ad. However its not consistent.

jacobsalmela:
I have also done some packet sniffing on Hulu ads and there are some timestamp issues to work out as well as some formatting and special URLs.  

We normally don't commit directly to master for stability reasons, but the file you are editing is not currently being used with Pi-hole and I _do_ know that the domain you are adding is legit and should help anyone who might be using this file.  

It's possible we might just eliminate this file all together if we can get the Hulu ad skipping to work.

I'll merge it, but **for future PRs, please make for the development branch**.

jacobsalmela:
Yep, I have had similar results.  I have spent a lot of time with the Hulu ads.  I'm close to something, but it's important to follow some of my mantras for the project:

- fix bugs before adding new features
- don't repeat yourself!

Issue #297, opened 2016-02-10
g_transport improper handle vars.


fayetted:
gravity_transport function doesn't handle the 3 variables properly.

$url, $cmd_ext, and $agent Do not get parsed properly because of the spaces in them.

fayetted:
Fixed by #298

PR #298, opened 2016-02-10
Fixes gravity_transport variable interpretation.

fayetted:
Adding (") around Variables passed to function ensures
  spaces are handled and that missing arguments are
  accounted for.


Mcat12:
Takes care of #297

fayetted:
Fixed by #298

PR #301, opened 2016-02-10
Check for 25mb of free disk space before install

### LINKED ISSUE #163: Check for available disk space
The automated installer should [check available disk space](https://github.com/jacobsalmela/pi-hole/issues/159#issuecomment-171060652), and if there is not enough, [expand the root filesystem](https://github.com/asb/raspi-config/blob/master/raspi-config#L36-L116).  

If users had to expand, a reboot would be required, but the next time they ran the install command, it would just start installing.

fayetted:
Function will check the free space of / (root)

If the free space is less than 25MB it will give directions
  on how to expand the disk.

Fixes #163

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -115,6 +115,20 @@ welcomeDialogs() {
 	In the next section, you can choose to use your current network settings (DHCP) or to manually edit them." $r $c
 }
 
+
+verifyFreeDiskSpace() {
+    # 25MB is the minimum space needed (20MB install + 5MB one day of logs.)
+    requiredFreeBytes=25600
+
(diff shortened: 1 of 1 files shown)


PromoFaux:
The logic looks sound, going to need to run some tests before we merge into dev, but thanks for your submission!

fayetted:
Your welcome, I hope it helps.

jacobsalmela:
I have tested this multiple times on a B and Pi2, both with and without enough space.  (you can generate a file to take up space if you need to test it without enough space `fallocate -l 5G gentoo_root.img`.

I would like to see this is in the next release so users don't run into issues during the install.

We can go ahead and merge this any time unless there are any objections.

PromoFaux:
Yep looks good here.

PR #305, opened 2016-02-10
Set file permisions on /etc/pihole

### LINKED ISSUE #304: pi-hole stops functioning if user has umask set to 007
Running gravity.sh as a normal, non-root user assumes sudo as needed however if the users umask is set to 007 or more restrictive the gravity.list file is regenerated with the user being the owner and 'world' not having read access to the file.

This causes dnsmask to be HUP'ed but can no longer read the gravity.list file and so no domains are blocked. 

The gravity.sh file currently changes permissions on all files in the $piholeDir but the gravity.list file is created later in the script where the umask has the ability to prevent proper access.

fayetted:
Ensures file permissions on $piholeDir before reload/restart of dnsmasq.

Fixes #304

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -321,6 +321,8 @@ function gravity_reload() {
 	echo -n "::: Refresh lists in dnsmasq..."
 	dnsmasqPid=$(pidof dnsmasq)
 
+    find "$piholeDir" -type f -exec $SUDO chmod 666 {} \; & spinner $!
+
 	if [[ $dnsmasqPid ]]; then
 		# service already running - reload config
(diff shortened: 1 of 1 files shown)


Mcat12:
Works for me.

fayetted:
@Mcat12 were you able to reproduce the error from #304 by setting your umask to 007 and running gravity.sh (not the version in this PR) as a normal user?

Mcat12:
I was running it as root (Using dietpi... I only have the root user). I'll have to try that with a new user sometime.

PR #306, opened 2016-02-11
Fixes IPv6 surviving a reinstall

Mcat12:
Checks for and removes the .useIPv6 file if it exists and the user chose to not use IPv6.


PromoFaux:
Looks good to me.

Issue #100, opened 2015-11-29
Diagnostics script


jacobsalmela:
Make a script that will verify functionality of the Pi-hole so users can see if it is working properly.

dschaper:
If you use nslookup (not installed by default, at least not in the Jessie Lite version...) you would be able to check if the resolver is using the right server and check if the address returned is correct. Maybe a `head gravity.sh

fayetted:
I just confirmed that nslookup and dig are both absent in a fresh install of Jessie-lite, however after the install of the dependencies, dnsutils is installed and includes both nslookup and dig.

What are your thoughts about testing for: 
1. Good domains - Google, openDNS, something that is relatively static
    Check that at least the 1st/2nd octet does NOT match (192.168.*, 10.*, 172.16-31)
2. Blacklisted domains - Pull a few names from the gravity.list file.
    Check that the IP resolves to the IP of the current system.
3. Whitelised domains - Pull a few names from the whitelist.txt file.
    Check that the IP matches the RFC1918 addresses from Test 1 (192.188.*, 10.*, 172.16-31.*) 

I might work on this in the next day or two.

Mcat12:
Sounds good

jacobsalmela:
Added in [v2.6](https://github.com/pi-hole/pi-hole/releases/tag/v2.6)
```
/usr/local/bin/pihole -d
```

PR #307, opened 2016-02-11
Diagnostic script

fayetted:
Script checks
 1. Google public DNS nameservers, ensure they resolve to public IP's
 2. Two random sites from the whitelist, ensure they are public IP's
 3. Two random sites from the adlist, ensure they resolve to the Pi-Hole IP
 4. Two random sites from the blacklist, ensure they resolve to the Pi-Hole IP

Addresses Feature request #100

### DIFF
--- a/advanced/Scripts/check-functionality.sh
+++ b/advanced/Scripts/check-functionality.sh
@@ -1,0 +1,115 @@
+#!/usr/bin/env bash
+
+
+IPv4dev=$(/sbin/ip route get 8.8.8.8 | awk '{for(i=1;i<=NF;i++)if($i~/dev/)print $(i+1)}')
+piholeIPCIDR=$(/sbin/ip -o -f inet addr show dev $IPv4dev | awk '{print $4}' | awk 'END {print}')
+piholeIP=${piholeIPCIDR%/*}
+
+adList=/etc/pihole/gravity.list
+blackList=/etc/pihole/blacklist.txt
+whiteList=/etc/pihole/whitelist.txt
+goodListNames="google-public-dns-a.google.com google-public-dns-b.google.com"
+
+
+# Function to resolve hostname and determine if IP is RFC1918, pi-hole, or internet address for the host.
+# Accepts 1 argument which is a host to resolve.
+verifyHostAddress() {
+    status=""
+    ip=`nslookup $1 192.168.100.122 | grep ^"Address:" | tail -1 | cut -d: -f2 | sed 's/ //g'`
+
+    if [[ $ip == $piholeIP ]]; then
+        status="pi-hole IP"
+    elif [[ $ip =~ (^127\.)|(^10\.)|(^172\.1[6-9]\.)|(^172\.2[0-9]\.)|(^172\.3[0-1]\.)|(^192\.168\.) ]]; then
+        status="RFC1918"
+    else
+        status=$ip
+    fi
+}
+
+
+if [[ -r $adList ]];then
+    adListNames=""
+    numberOf=$(cat $adList | sed '/^\s*$/d' | wc -l)
+    for n in `shuf -i 1-$numberOf -n 2`
+    do
+        adListNames="$adListNames `sed -n ${n}p $adList | cut -d\  -f2`"
+    done
+fi
+
+
+if [[ -r $blackList ]];then
+    blackListNames=""
+    numberOf=$(cat $blackList | sed '/^\s*$/d' | wc -l)
+    for n in `shuf -i 1-$numberOf -n 2`
+    do
+        blackListNames="$blackListNames `sed -n ${n}p $blackList`"
+    done
+fi
+
+
+if [[ -r $whiteList ]];then
+    whiteListNames=""
+    numberOf=$(cat $whiteList | sed '/^\s*$/d' | wc -l)
+    for n in `shuf -i 1-$numberOf -n 2`
+    do
+        whiteListNames="$whiteListNames `sed -n ${n}p $whiteList`"
+    done
+fi
+
+
+echo -e "\nTesting known good websites."
+for name in $goodListNames
+do
+    verifyHostAddress "$name"
+    if [[ "$status" == "pi-hole IP" ]]; then
+        echo -e "\tFailed - $name resolves to your pi-hole ($piholeIP)"
+    elif [[ "$status" == "RFC1918" ]]; then
+        echo -e "\tFailed - $name resolves to a non-routable address ($ip)"
+    else
+        echo -e "\tSuccess - $name resolves to a public IP address ($ip)"
+    fi
+done
+
+
+echo -e "\nTesting websites from your $whiteList file."
+for name in $whiteListNames
+do
+    verifyHostAddress "$name"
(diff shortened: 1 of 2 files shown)


PromoFaux:
Just taken a look over this script...

A couple of things:
In this screenshot, you can see that the sites listed in `gravity.list` are showing as resolving a public IP address, however when pinging them (from any device on my network) I get an internal IP address (as expected)

image

Another thing I noticed is on [this line](https://github.com/fayetted/pi-hole/blob/66e3a005d8d7c138423ebc367023735111e1e1e8/advanced/Scripts/check-functionality.sh#L18) there is a localrange IP address hard-coded... what is that, please? It appears to be the root of the `$ip` not displaying in the brackets


(as an aside, thankfully that second address is blacklisted.. wouldn't want my ISP thinking I was trying to resolve it!)

fayetted:
Good catch and thanks for looking.  I think taking the static IP out will fix the issue. But you did bring up a new use case where the name does not resolve at all because your pointed at a non existant name server.  Which could also happen if the upstream name server was not accessible or returned an NXDOMAIN response.

I'll take a look and address these and then comment on the PR.  Should be fairly quick to fix and easy to test.

As for the "aside", I'm picking at random a name from the gravity.list.  Is this ok, or would it be better to pick at random then test for certain key words and if found pick again?

fayetted:
I removed the test IP that was left in accidentally and also accounted for when a name does not resolve.

Ready for re-testing.

PromoFaux:
OK, that is actually much quicker than it was, too.. I'm guessing that was to do with the test IP you left in?

The two random addresses is fine, I was just clad I wasn't having to actually resolve it :D

Looks good to me. Will need to have a conversation with @jacobsalmela about how we include/implement this.

fayetted:
Great, I'm glad it worked this time.

jacobsalmela:
I think this is a good start, and I like how it chooses random domains to test.  

One thing I noticed though is that if you run it directly from the Pi so it can't simulate a client connection.  For example, if you turn off `dnsmasq` and try go get out to the Internet, it won't work, but the script would still say that it succeeded (because the Pi can still make it out the the Internet).

fayetted:
Would forcing the lookups to query localhost rather than relying on the resolv.conf address the problem?

dschaper:
If the queries are executed on the pi, then you can query `127.0.0.1` and get the local `dnsmasq`. Try it while you're tailing logs in another terminal window... (Or try byobu and do it from a single terminal...)

fayetted:
I'm not following your logic.  The script should be performing 4 sets of tests by querying the DNS configured on the system it is run from.  So on the RPI running Pi-Hole, mine is set to use 127.0.0.1 by default, I don't think I changed that.  So the test should perform as such.
1. Is the client able to resolve Good domains via the RPI, If an Internet IP (not the RPI, and not RFC1918) then success.
2. Is the client able to resolve white-listed hosts to an internet IP. if not, then fail
3. Is the client able to resolve adlist IP's to the RPI address. If the RPI address is returned then success, else fail.
4. Is the client able to resolve blacklisted IP's to the RIPI address. If the PI address is returned then success, else fail.

So the comment from @jacobsalmela about turning off dnsmasq is confusing.  Some of the tests would pass if you turn off dnsmasq and change your resolv.conf file to an internet DNS server but the blacklist and adlist queries would fail as they would resolve.

Maybe what is missing is a final status message that checks to make sure all 4 tests succeed and then report either success or failure based on the sum of all the tests.

jacobsalmela:
I use my Pi-hole as my sole DNS server (relying on the upstream servers set in `01-pihole.conf`.  If `dnsmasq` isn't working, my client cannot reach the Internet.  But I turn off `dnsmasq` and ping a domain from the Pi itself, it works.  So even though some of the tests pass when the script runs on the Pi, it may not work for clients if they cannot resolve any hostnames.

I think my logic is right, but I could certainly be wrong.  It starts to get confusing with so many variables!

fayetted:
@jacobsalmela I think we might be confusing each other, so I think it might be best to try to start fresh.

My understanding was that you wanted to be able to run the script from the PI to have it check and validate that the pi-hole service was functioning.  Functioning meaning that good domains resolve to their internet address and bad domains resolve to the PI's IP Address.

The way the script is currently written it relies on the /etc/resolv.conf file to be pointed only to 127.0.0.1 (Your could use the PI's real address too with the same results).  But if your resolv.conf file has your upstream DNS listed then the script will not give an accurate representation of the functioning state of Pi-Hole. (This is why I suggested that maybe hardcodeing 127.0.0.1 in the script might actually be better.  It would provide more accurate results when run from the PI, but would not work if you ran it from a client.)

If you run the script from the PI it will perform nslookups against the IP in the /etc/resolv.conf file (assumed to be 127.0.0.1).  If dnsmasq is truly turned off then all 4 tests will fail as the nslookup will not connect to any DNS service that can return a response (i.e. Every response will be a timeout.)  If you get anything except 4 sets of failures then 1 of 3 things is happening.  1. dnsmasq is not really disabled. 2. Your resolve.conf file is pointing to something other than 127.0.0.1.  3. You also have named/bind (or some other DNS service) also listening on UDP/53 and resolving addresses when DNSMASQ doesn't intercept the connection. This 3rd state would give you partially inaccurate test results as good domains will resolve correctly but bad domains will also resolve to their Internet address.  The script will say the first 2 tests passed and the 2nd 2 test failed.  Which is correct from the scripts perspective because it was able to resolve good domains but bad domains resolved in a way that we don't want and until that is fixed your pi-hole is not 100% functional (It's actually not functioning at all but the script can't differentiate that some other service responded to the DNS queries.)

When dnsmasq is turned on, then you can get any number of possible combinations.  Good address could resolve or not resolve depending on the upstream DNS server in your 01-pihole.conf file (This is independent of your resolv.conf file).  Bad domains could resolve to their actual IP or the RPI depending on if pi-hole is functioning or not ( I actually tested this by setting my umask to 007 and then running gravity before the fix I provided. You can test by changing the permission of /etc/pihole/gravity.list  to 400 and restarting the dnsmasq service. )

Pleas let me know if I'm completely misunderstanding you or the concept your trying to explain to me.  This is an awesome project and I'm glad I found it.  I wrote scripts to do this same functionality on corporate BIND servers years ago but this project gives home users some of the same benefits corporations benefit from and I think that is awesome.

manuelgu:
Any updates? :-)

Mcat12:
Does #411 do some of this?

manuelgu:
Haven't had a detailed look at it but the `piholeDebug.sh` might provide some useful information as to where the Pi-hole works or not

nate-ubiquisoft:
@Mcat12 Yes #411's piholeDebug.sh will give everything you will need, and if there is more information to be needed, I am working on an advanced debug version which will need to install additional network tools to use for advanced debugging. That will be in PiHoleController release, which can be cloned from my repo now if you want to try it out. I still havent finished testing, which after I will submit a PR.

jacobsalmela:
Can this be closed since `pihole -d` is now available?

PromoFaux:
I should say so. Unless there are any updates from @fayetted .

Mcat12:
Closing unless @fayetted says otherwise.

PR #312, opened 2016-02-14
SplitVars

jacobsalmela:
rough draft of how I was using a separate file for variables.  Don’t rely 100% on it as I was working on it a while ago before everything changed, but this might give an idea of the direction I was going…

### DIFF
--- a/advanced/pihole.vars
+++ b/advanced/pihole.vars
@@ -1,0 +1,121 @@
+#!/usr/bin/env bash
+# Pi-hole: A black hole for Internet advertisements
+# (c) 2015, 2016 by Jacob Salmela
+# Network-wide ad blocking via your Raspberry Pi
+# http://pi-hole.net
+# Exports variables for use with Pi-hole
+#
+# Pi-hole is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+
+# First check if the user running the script is root.
+# If they are, they have root access and the SUDO variable will be set to nothing
+if [[ $EUID -eq 0 ]];then
+	echo "You are root."
+else
+	# sudo will be used during install and to restart services in gravity.sh
+	export SUDO="sudo"
+fi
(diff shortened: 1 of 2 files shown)


koen84:
Cool, i can definitely see the value in this and even additional possibilities from there onward.

Looking forward to hear results of the internal discussion.

PromoFaux:
Yeah `exportvariables` was one I started on before I got a bit bogged down with work in my real job, and I believe that `splitvars` is a slightly different implementation that @jacobsalmela was starting to work on.

the `exportvariables` branch sort of works at the moment, though it was a rough implementation as a proof of concept, so it breaks a lot of things too. Needs lots of bugfixing :)

jacobsalmela:
This one won't work right now either as a lot has changed already.  It's just a copy/paste of what I had going on myself as a side project.  It works on an older Pi-hole version, but there are files in different places now and new ones.  

But that's part of the reason for the PR--if we ever need to change a file, we can just change it once in a single file and then it applies everywhere else (less work, faster scalability, and more stability).

PromoFaux:
>less work, faster scalability, and more stability

:+1:

koen84:
Sounds like this moved to top priority and will involve a major code overhaul ?
Then i guess i better wait a bit before attempting a customised install.  :)

I take either of you will work from your respective branch ?
Or should i try, possibly forking from development branch ?  (Considering there's been changes meanwhile that are creating merge conflicts.)

PromoFaux:
`exportvariables` is the WIP branch, this PR is to merge the two, though they may not be compatible...

We need to get a decent updating script together, rather than users running install script each time there is an update. To do that though, we might as well press on with the refactor.

PromoFaux:
You're morw than welcome to fork from dev and give it a go yourself! We welcome all input and look over every submission. Even if we don't accept the whole thing, we can always cherry pick commits.

koen84:
Seems like git decided to boycot me.  It's solved now, but took me hours.
And there's quite some work i need to prioritise,
so sadly i need to postpone this some time.  :(

PR #326, opened 2016-02-17
DNS changes from @slamanna212 and @fayetted

PromoFaux:
To be reviewd by @dschaper @jacobsalmela and @Mcat12


PromoFaux:
Wait.. Done goofed.

PR #343, opened 2016-02-23
CBH-Update_README md

ChadBHowell:
Add NTC CHIP and Debian based distros

Does the "Other Operating Systems" section need to be changed too?


ChadBHowell:
Disregard.. Stupid Github desktop...

PR #344, opened 2016-02-23
Update README.md

ChadBHowell:

- Add NTC CHIP and most Debian based distros

@pihole/gravity

Does the "Other Operating Systems" section need to be updated as well?


ChadBHowell:
I made the changes recommended by @PromoFaux

PR #351, opened 2016-02-28
IPv4 Choices

ryt51V:
- Ask the user if they have already configured a static IPv4 address, and if they choose yes then skip IPv4 configuration.  Default is no.
- Allow for interfaces that have multiple IPv4 addresses assigned; after the user selects an interface prompt for which IP address they wish to use.  This works with the iproute2 method with a single name for multiple IPs (e.g. "eth0" with 192.168.0.10 and 192.168.0.11).  It doesn't work with the deprecated ifconfig method with multiple aliases (e.g. "eth0" with 192.168.0.10, "eth0:0" with 192.168.0.11.).
- Simplified the main logic of the use4andor6 function to help enable the above.

@pihole/gravity

### DIFF
--- a/advanced/Scripts/blacklist.sh
+++ b/advanced/Scripts/blacklist.sh
@@ -34,13 +34,31 @@ domList=()
 domToRemoveList=()
 
 
-piholeIPfile=/tmp/piholeIP
+piholeINTfile=/etc/pihole/piholeINT
+piholeIPfile=/etc/pihole/piholeIP
 piholeIPv6file=/etc/pihole/.useIPv6
 
-# Otherwise, the IP address can be taken directly from the machine, which will happen when the script is run by the user and not the installation script
-IPv4dev=$(ip route get 8.8.8.8 | awk '{for(i=1;i<=NF;i++)if($i~/dev/)print $(i+1)}')
-piholeIPCIDR=$(ip -o -f inet addr show dev $IPv4dev | awk '{print $4}' | awk 'END {print}')
-piholeIP=${piholeIPCIDR%/*}
+# Know which interface we are using.
+if [[ -f $piholeINTfile ]]; then
+    # This file should normally exist - it was saved as part of the install.
+    IPv4dev=$(cat $piholeINTfile)
+else
+    # If it doesn't, we err on the side of working with the majority of setups and detect the most likely interface.
+    IPv4dev=$(ip route get 8.8.8.8 | awk '{for(i=1;i<=NF;i++)if($i~/dev/)print $(i+1)}')
+    echo "::: Warning: ${piholeINTfile} is missing.  Using interface ${IPv4dev}."
+fi
+
+# Know which IPv4 address we are using.
+if [[ -f $piholeIPfile ]];then
+    # This file should normally exist - it was saved as part of the install.
+    piholeIP=$(cat $piholeIPfile)
+else
+    # If it doesn't, we err on the side of working with the majority of setups and detect the most likely IPv4 address,
+    # which is the first one we find belonging to the given interface.
+    piholeIPCIDR=$(ip -o -f inet addr show dev $IPv4dev | awk '{print $4}' | head -n 1)
+    piholeIP=${piholeIPCIDR%/*}
+    echo "::: Warning: ${piholeIPfile} is missing.  Using IPv4 address ${piholeIP}."
+fi
 
 modifyHost=false
 
--- a/advanced/Scripts/blacklist.sh
+++ b/advanced/Scripts/blacklist.sh
@@ -38,14 +38,16 @@ piholeINTfile=/etc/pihole/piholeINT
 piholeIPfile=/etc/pihole/piholeIP
 piholeIPv6file=/etc/pihole/.useIPv6
 
+echo ":::"
+
 # Know which interface we are using.
 if [[ -f $piholeINTfile ]]; then
     # This file should normally exist - it was saved as part of the install.
     IPv4dev=$(cat $piholeINTfile)
 else
     # If it doesn't, we err on the side of working with the majority of setups and detect the most likely interface.
+    echo "::: Warning: ${piholeINTfile} is missing.  Auto detecting interface."
     IPv4dev=$(ip route get 8.8.8.8 | awk '{for(i=1;i<=NF;i++)if($i~/dev/)print $(i+1)}')
-    echo "::: Warning: ${piholeINTfile} is missing.  Using interface ${IPv4dev}."
 fi
 
 # Know which IPv4 address we are using.
@@ -55,10 +57,13 @@ if [[ -f $piholeIPfile ]];then
 else
     # If it doesn't, we err on the side of working with the majority of setups and detect the most likely IPv4 address,
     # which is the first one we find belonging to the given interface.
+    echo "::: Warning: ${piholeIPfile} is missing.  Auto detecting IP address."
     piholeIPCIDR=$(ip -o -f inet addr show dev $IPv4dev | awk '{print $4}' | head -n 1)
     piholeIP=${piholeIPCIDR%/*}
-    echo "::: Warning: ${piholeIPfile} is missing.  Using IPv4 address ${piholeIP}."
 fi
+
+echo "::: Large gravitational pull detected at ${piholeIP} (${IPv4dev})."
+echo ":::"
 
 modifyHost=false
 
--- a/advanced/Scripts/whitelist.sh
+++ b/advanced/Scripts/whitelist.sh
@@ -33,13 +33,31 @@ versbose=true
 domList=()
 domToRemoveList=()
 
-piholeIPfile=/tmp/piholeIP
+piholeINTfile=/etc/pihole/piholeINT
+piholeIPfile=/etc/pihole/piholeIP
 piholeIPv6file=/etc/pihole/.useIPv6
 
-# Otherwise, the IP address can be taken directly from the machine, which will happen when the script is run by the user and not the installation script
-IPv4dev=$(ip route get 8.8.8.8 | awk '{for(i=1;i<=NF;i++)if($i~/dev/)print $(i+1)}')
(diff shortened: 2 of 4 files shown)


PromoFaux:
Thanks for that. I'll review this as soon as I can get to it.

PromoFaux:
Looks good, Could you move the new IP address choice dialog to come _after_ the IPv4/IPv6 choice dialog please? And only if the user selects IPv4. I think that would make more sense from an install flow point of view.

Or if there is a reason it comes before, let me know your thoughts...

ryt51V:
I agree that makes more sense.  I've had a skim and don't see why it can't be within the `if [ $useIPv4 ]; then` block.  (I think my original reasoning was to keep it as early as possible since the `$IPv4*` declarations it was replacing were very early on, but don't see a specific reason it needs to be that way.)  I will work on it later.

I think it needs to be (immediately) before the other new dialog though - "Have you already configured a static IPv4 address".  Logistically, but also install flow wise, though perhaps alter the wording of this dialog.  Maybe something along the lines of.

> We have found the following details for the IPv4 address you have selected.  Have you already configured this as a static IPv4 address as desired?
> 
> (If you are unsure, choose No.)
> 
> IPv4 Address: \<Address\>
> Gateway: \<Gateway\>

If you wouldn't mind letting me know your thoughts

PromoFaux:
My thoughts:
:+1:

ryt51V:
All done.

PromoFaux:
Thanks. Will take a look in a bit

ryt51V:
I have just noticed a few key issues with this:

- The IP address is no longer saved to /tmp/piholeIP, as that file is only saved within getStaticIPv4Settings() which is not called.  gravity.sh thus doesn't use the correct IP when called in the install script.  Easy fix, but...
- Even if it did save, we can't rely on a file in /tmp being kept.  So long after installation when /tmp is next cleaned, gravity.sh will go ahead re-detect the IP, which may be wrong.  (And whitelist.sh, blacklist.sh...forgive me, I didn't realise how deep this rabbit hole went!)

Personally I think this ought to be changed as follows:

- The IP and interface chosen should be stored more permanently in /etc/pihole.  In an ideal world this would be a nice user-friendly config file, but to start with just the same /tmp/piholeIP and /tmp/piholeINT but replacing /tmp with /etc/pihole will work fine.
- All scripts should use these files if they are present rather than re-detect from `ip`.
- If a script finds these files missing...I'm unsure, I guess we're better erring on the side of seamlessly working for those with just one IP on an interface, so probably best to write a warning but detect IP on runtime as it does now?

Again I would welcome your thoughts if you wouldn't mind.  I'm probably going to work on this over the next few days.

(Side note: Ignoring this pull request, Pi-hole's behaviour when an interface has multiple IPv4 addresses assigned is not as expected.  A user would assume that with lack of support for multiple IP interfaces, it would default with the "main" IP address.  But in the detection below...

`ip -o -f inet addr show dev $IPv4dev | awk '{print $4}' | awk 'END {print}'`

...after `awk '{print $4}'` you have all of the IPs on the interface.  After `awk 'END {print}'` you have the _last_ IP in that list.  Notwithstanding this pull request, this oddity ought to be addressed in any case.  Especially considering that multiple IPs is one solution to pi-hole/AdminLTE#39 )

ryt51V:
Think that's all done now actually.  Saves IP and INT to an /etc/ file, and it's the first place these are read from.

PromoFaux:
Ah! Good catch. So this saves the IP selected on install in a file in `/etc/pihole`? I've been meaning to do this for a while, just not had time. If so, thanks!

ryt51V:
No problem - yes, that's what this does.  In detail:

- basic-install.sh saves the IP in /etc/pihole/piholeIP and the interface in /etc/pihole/piholeINT

  Either when the user confirms they have already configured their IP statically, or when the pi-hole configures it for them.

- gravity.sh, blacklist.sh, whitelist.sh all read from these files.

  If one is missing then it writes a warning to output and reverts back to auto detection*.  This way, in the event of a problem with the files, single-IP users are unaffected; puts the burden of troubleshooting on multi-IP users who are more likely to know their stuff.  I think this is the way to go.

  Either way, they also output the IP and interface when run, make debugging easier.

  *(Note: One change to auto detection - if it finds multiple IPs on an interface, it now settles on the FIRST rather than the LAST.)

(This pull request has ballooned quite a bit from what I expected, so appreciate this isn't a small job to review now!!)

PromoFaux:
I'll take a look over it properly on my lunch break tomorrow... :)

I bumped into a situation a while back where I had openVPN running on my pi alongside pi-hole. With `gravity.sh` working out the IP each time, `gravity.list` would get it's local IP from `tun0` instead of `eth0` (as it was the IP that had the route to `8.8.8.8`)

Good work!

koen84:
One extra taught : if the user changes his fixed IP address (neatly in /etc/dhcpcd.conf) before it would be just fine next time gravity.sh runs.  As i understand from the conversation, now it would become permanently broken unless the user also thinks to update this new /etc/pihole/piholeIP file.

Theoretically that might be caught by testing if the IP is attached to the selected interface ?

I'm also wondering, if you found the solution for the multi IP setup, can't it be duplicated to gravity.sh IP detection / handling ?  (Especially if the idea is to make that default over time.)

PromoFaux:
To save duplication of code, perhaps we could work this in once we've got the `splitvars`/`exportvariables` refactor implemented. The idea being there that we split the scripts out into  seperate files that can reference each other. That way we could have common functions (such as checking for root, testing IP address still matches etc) in a common `funcs.sh` or something...

dschaper:
(Without the .sh extension please, lol...) Question, do we really need to bind to a specific IP or can we bind to an interface instead? `dnsmasq` will happily listen and respond on all interfaces, and since we are instructing the user as to which IP to use for their DNS resolvers, could we drop that portion and requirement on `dnsmasq` and just let it run on all interfaces? Theoretically it should only be getting queried by the hosts that are configured, so what harm is there in letting it listen on all interfaces/IP's? 

Now as for which IP to use for the `gravity.list` as the source address, that obviously has to match the `lighttpd` instances address...

ryt51V:
I can only think of one scenario where dnsmasq listening on all IPs or all instances would be a bad thing - some sort of perimeter network setup with the Rpi directly assigned a public IP address as well as an internal one.  Wouldn't want the outside world to be able to resolve through it.

But surely that's way outside the scope here and not something we ought to concern ourselves about; such power user could lock down dnsmasq themselves (or indeed probably wouldn't run the auto installer).

+1 for removing `interface=@INT@` from 01-pihole.conf

dschaper:
If someone has an open to the world interface, I sure hope they have the sense to firewall the thing!

ryt51V:
You would hope ;-)

Also, forgot to say, agree with Promofaux that if there's an include / dedup refactor in progress then this should probably wait for that, as at this point it's clearly going to have an impact on all of the scripts rather than just basic-install.sh.

In any case I'm not planning on deleting this branch from my own fork.  Don't know if perhaps you want to make your own branch for me to push the work-so-far to (so you can "save it for later").

dschaper:
As long as you keep your fork, we can add it later as an upstream remote if needed...

koen84:
If there's more interfaces, it means a user added them (unless anyone has already an rpi3), which likely means they have specific intention with them, therefore making default to listen to all interfaces seems like a very bad idea to me.

dschaper:
But is it damaging to have it able to listen on the port when we know that the port isn't going to be queried? We are telling the user what interface to query on, and they are configuring either their router or their clients to query that address on that interface.

koen84:
Bad practise ?
As for firewalling, rpi default is iptables empty, you know.

dschaper:
Could you imagine the support nightmare that would be created if Raspbian came with a default firewall? I hang out a lot in the RPI forums, and the entire support system for the RPF is through volunteers in the forums.

dschaper:
NOOBS is a nightmare already on it's own...

koen84:
I meant it more : don't assume firewall is on, because it's not by default and from experience, coders running VPS don't even do so, why would a hobbyist.  (Though maybe a default iptables that only allows outgoing or related traffic, might get people to think of security.)

I think it's more compatible / smart and definitely better netiquette to keep dnsmasq bound to an interface or address.

dschaper:
Point taken... What if we were using a different package for resolving instead of `dnsmasq`?

koen84:
Principle doesn't change depending on the software used, bind to interface or address is the proper thing to do for a service.  If there's no problem with dnsmasq or there's no clear expected improvement elsewhere, i wouldn't change, because the devs now know how to work this package.

I've also seen dnsmasq being used in other tuts, which i see as another plus (larger community, less services to run, etc).  Provided pihole plays nice with others.  From seeing the install code, that seems okay.  The room for improvement i can see would be when it detects an existing dnsmasq.conf that's foreign to pihole, is to add only specific config in 01-pihole.conf (addn-hosts and log), as in my (limited) experience with dnsmasq it takes a while to tweak the parameters and this file might now override carefully crafted settings.  ((E.g. windows-netbios issues.))

dschaper:
Just a quick note on the override issue, `dnsmasq` will throw an error and refuse to start if it has duplicate configuration parameters set. So you can't override any settings, it just won't start. That's why we moved to placing our config in the subdirectory, so we wouldn't be stepping on someone's crafted configurations.

koen84:
Thanks, i didn't know that.  Though i doubt it works with the settings that take no parameters, e.g. no-resolv, if that's not used in an existing config, i'm thinking it would be "overridden" by pihole's config ?

dschaper:
If the flag isn't set in another config file, and we set the flag, then yes, we would in effect be overriding the existing config if the user did in fact want that flag to be unset. If the user is that savvy though, hopefully they will either contact us to see if we can get a setting the co=exists, or they configure their network to allow for the inconsistencies. It's kind of like the pipe to bash, if you know better you should be reading the scripts and know what you are piping in before you do it :)

jacobsalmela:
@ryt51V any input on this?  I think it's pretty far out of date with the development branch.  Esepcailly concerning the hardcoded variables as we are using `. /etc/pihole/setupVars.conf` now...

PR #353, opened 2016-02-28
Choice of web server (lighttpd vs manual configuration)

ryt51V:
- Allows the user to choose between Pi-hole using and automatically configuring lighttpd, and installing to another web server the user has manually configured.
- If the user chooses Manual, it prompts them for their web root (default /var/www/html), and it skips further lighttpd related steps.

@pihole/gravity

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -760,7 +760,10 @@ displayFinalMessage
 echo -n "::: Restarting services..."
 # Start services
 $SUDO service dnsmasq restart
-$SUDO service lighttpd start
+if [[ "$webServer" = "lighttpd" ]]
+then
+	$SUDO service lighttpd start
+fi
 echo " done."
 
 echo ":::"
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -306,7 +306,7 @@ setStaticIPv4() {
 
 function chooseWebServer() {
 	# Allow the user to choose the web server they wish to use.
-	chooseWebServerCmd=(whiptail --separate-output --radiolist "Pi-hole will automatically configure the lighttpd web server for you.  Alternatively, if you prefer, pi-hole can use a web server that you have previously manually configured yourself.\n\n(If you are unsure, choose lighttpd.)" $r $c 2)
+	chooseWebServerCmd=(whiptail --separate-output --radiolist "Pi-hole will automatically configure the lighttpd web server for you.\n\nAlternatively, if you prefer, pi-hole can use a web server that you have previously manually configured yourself.\n\n(If you are unsure, choose lighttpd.)" $r $c 2)
 	chooseWebServerOptions=(lighttpd "" on
 							Manual "" off)
 	webServer=$("${chooseWebServerCmd[@]}" "${chooseWebServerOptions[@]}" 2>&1 >/dev/tty)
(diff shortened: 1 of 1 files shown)


Mcat12:
This would go well with the nginx idea.

dschaper:
I concur, though my personal choice would be to replace lighttpd with nginx, I just think the performance is better and we'd get some more tools to play with... But that is my own personal opinion and not speaking on behalf of anyone else on the team...

dschaper:
I like this, but I think we need to hold for a bit for the code refactor first. Make sure that's stable before we introduce any more variables in to the mix. I'd love to have `nginx` but need to make sure everything is stable and settled first.

b264:
:+1: My pi already is running nginx and I don't want to add another server

PromoFaux:
@ryt51V  I think this is going to come in handy for #444

jacobsalmela:
#444 was closed due to inactivity.  Closing this as well.  Can re-open if needed.

juriansluiman:
I would like to reopen this issue and try to get this merged into master. Afaik this current branch doesn't have conflicts with the development branch, but this PR might need rebasing. My personal use case: I have one RPi3 serving as Unifi controller and Home-Assistant hub with Z-Wave controller. I'd like to join in pi-hole, but want to specify installation path and web server config (although this probably will be lighttpd, ports and hostname need to be specified).

What is required for testing? I have a spare rpi which I can use for testing. How many tests are required?

PromoFaux:
Doesn't look like it adds too much, but we'd need to see. How are you with git? You're welcome to fork from this branch and have a go at merging/rebasing to current development!

In terms of testing, there are some python tests that run to make sure nothing is broken, and the easiest way to get those running is to submit a pull request and wait for the status to update. You may need to write some extra tests, too.

As for multiple applications on one web server, take a look at [the last line of Pi-hole's lighttpd.conf](https://github.com/pi-hole/pi-hole/blob/master/advanced/lighttpd.conf.debian#L81) file. You can specify additional options in that external file

juriansluiman:
I'm pretty familiar with git; I'll fork & rebase the branch on current development. 

> As for multiple applications on one web server, take a look at the last line of Pi-hole's lighttpd.conf file. You can specify additional options in that external file

The problem is that on file level the [document root](https://github.com/pi-hole/pi-hole/blob/b0eceddcec2054273d7e1add34d1f806b0c81693/advanced/lighttpd.conf.debian#L31-L32) is fixed. In my case it's OK to provide a custom lighttpd.conf, but then you'll make "virtual" hosts by either host or port (socket) criteria. That is documented pretty well by lighttpd.

My concern is rather the `/var/www/html/` is [hard coded on much more locations](https://github.com/pi-hole/pi-hole/search?utf8=%E2%9C%93&q=%22%2Fvar%2Fwww%2Fhtml%22&type=) than only the `basic-install.sh`: 

* `automated install/uninstall.sh`
* `advanced/Scripts/update.sh`
* `advanced/Scripts/version.sh`
* `advanced/Scripts/webpage.sh`
* `advanced/Scripts/piholeDebug.sh`

For uninstall, I can imagine you just try to remove and if that doesn't work, you can always echo a message that for a differrent web root files have to be removed manually. For the other files, I'll ask a solution in the new PR. We can discuss that later, because I'm now unfamiliar with pihole it's pretty difficult to guess what the other files exactly require the webroot for.

Issue #357, opened 2016-03-01
chmod 777


koen84:
I was checking the code before running with it (you know, executing as root and for a device central to the network).

I noticed a chmod 777 in gravity.sh on line 131 ?
And as far as i understand, EVERY pihole will execute this (except on first run, likely the install or else test).

I know many websites and fora seem to promote this as a "fix all", usually followed by a comment on why this is "evil", for it being a massive security risk.  If it REALLY needs R/W access by another user, why not do chown ?  But then again, i have ZERO files or directories in /etc that are owned by anyone else than root.

I hope this can be changed.
(In checking the code + comments, i've also not found where the need for this comes from, but i'm probably missing something obvious.  If i understood, i might be able to think of a solution.)
##

kmr2319:
Based on the comments, "existing" installs are the problem from versions priror to 2.5.3 where there was no "pihole" user. 

For example on my install which was prior to 2.5.3, the /etc/cron.d/pihole still has all jobs running as root rather than pihole. In addition, (assuming RPi only installs as that was the "original" intent) the default user of "pi" would likely still need access to this directory and /var/www/html. See comments in [this pull request](https://github.com/pi-hole/pi-hole/pull/242) for some additional details (be sure to read the comments). There's more to it than just gravity.sh.

koen84:
The 777 is set on /etc/pihole directory.

The /var/www/html gets chown www-data:www-data in basic-install and then the user pihole is added to the www-data group, effectively getting both lighthttpd and pihole user write access.  (I haven't looked into the webinterface code yet.)

The cron from current master (2.5.3 i presume, looking at releases) still runs as root.

From the linked PR i understand the pihole user is currently just made, not much used ?  And if it's specifically related to prior 2.5.3 it's pity it runs also on everything more recent.  (Seeing throughout the repo security IS taken seriously.)

dschaper:
That was my massively ugly hack to fix a problem we had with permissions errors, and I just didn't get to fixing it with something not so ugly. Trying to remember what problem existed, but it was an issue with existing installs having some permission bits set that caused crashing on an update. This issue predates the introduction of the "pihole" user, this was was back a few months ago, before we fixed the IPv6 issue. It was to do with legacy installations (thing version 1.0 and prior).

dschaper:
[...]

PR #359, opened 2016-03-02
Use of sudo, pihole user, permissions.

ryt51V:
This pull request started as an attempt to simplify sudo usage in basic-install.sh.  It turns out this ties in with a LOT of things.  I've tried to keep it as small as possible, but it has become a bit of a monolith for which I do apologise.  Hope this is useful!
# Summary

This pull request:
- **Simplifies use of sudo.**  A few scripts have `$SUDO` every other line, when instead we can rerun the script as root at the start using `sudo "$0" "$@"`.  This results in faster scripts (each call to sudo adds over 0.1 seconds!), and IMO more maintainable scripts, at the cost of running the entire script as a single user.
- **Makes use of the pihole user.**  I saw from commits and comments that we have recently created a pihole user and wish to offload tasks to it, but so far not done much with it.  I am hoping that this request goes very far towards achieving that goal.  Some scripts now sudo to pihole, so we're not needlessly running as root, and don't have to worry about setting file permissions for any account other than pihole.
- **Sets less risky permissions.**  Issue #357 is about the chmod 777 that's used.  Still not perfect, but this pull request has a better way around it.
# Detail
## Sudo
- basic-install.sh, uninstall.sh, setupLCD.sh:

These now rerun themselves as **root** from the start.

These scripts heavily rely on root, with very little other than `echo` and `whiptail` not prefixed with `$SUDO`.  I'm a strong believer in least-privilege but I don't think it makes sense to maintain the error-prone complexity of using `$SUDO` with these scripts just so that a small number of relatively safe commands aren't running as root.  They're installers, it's not like they're run regularly.

updateDashboard.sh sort of comes under this list too - I haven't touched this script, but whoever pushed this in the first place already did the same `sudo "$0" "$@"` to rerun it as root.
- blacklist.sh, chronometer.sh, piholeLogFlush.sh, whitelist.sh

These now rerun themselves as **pihole** from the start.

These scripts are the opposite - they hardly ran anything as root, and we'd like to keep it that way.  They only need to do a couple of things as root: reload dnsmasq, and the chmod 777 hack.  See further below for how they elevate to run these tasks.

sudo here is run the normal way, in that it will prompt for a password if required.  This means no passwords for cron jobs, as they will either run as root (which can sudo to anything without a password) or as pihole (an if statement skips sudo if already running as the correct user).
## Sudoers

For those who don't know, we can edit the sudoers file to allow certain users (or groups) to run certain programs as root without entering a password.

Instead of modifying /etc/sudoers (dangerous!!!!), we can instead create a file in /etc/sudoers.d (much safer).  basic-install.sh now creates /etc/sudoers.d/pihole, with one entry:

`pihole  ALL=(ALL:ALL) NOPASSWD: /usr/local/bin/piholeReloadServices.sh, /usr/local/bin/piholeSetPermissions.sh`

This allows the pihole user (reminder: which now runs blacklist.sh, whitelist.sh, etc.) to non-interactively elevate to root to run two new scripts I have created.

The new scripts each run a root-requiring operation.  They are:
- piholeReloadServices.sh
  This script reloads or starts dnsmasq.
- piholeSetPermissions.sh
  This script implements an alternative to the chmod 777 hack in gravity.sh, explained further below.

Personally I don't think this is particularly elegant, but I do think it's the right choice for the project as it stands now.
## chmod 777 hack fix

On initial install, basic-install.sh now chowns /etc/pihole to root:pihole, and gives group rw permissions to the folder.

I don't know the history of the chmod 777 hack, as far as I can tell it's not needed for the current repo, so with the above we should be able to just remove it.  But comments in the script and issue #357 indicate it's for backwards compatibility with an older version.

piholeSetPermissions.sh now just runs chown and chmod again, and is called from gravity.sh instead of chmod 777.  So /etc/pihole is kept owned by root:pihole and user/group rw, other r.

(Side note: IMO we ideally ought to split off some of /etc/pihole to /var/lib/pihole, but now's not the time!)
# Known issues
- "You are root" appearing in the middle of scripts.
  
  For example: gravity.sh starts by sudoing to pihole ("You are pi hole user"), and a few lines later it runs piholeSetPermissions.sh ("You are root").
- Sudo's adding a line break messes up some formatting, particularly with multiple echos intending to span a single line (e.g. "Doing X...done")
- Permissions for /var/log/pihole.log need to be modified or piholeLogFlush.sh run as root.

@pihole/gravity

### DIFF
--- a/advanced/Scripts/blacklist.sh
+++ b/advanced/Scripts/blacklist.sh
@@ -37,8 +37,8 @@ else
 	# Check if it is actually installed
 	# If it isn't, exit because the install cannot complete
 	if [[ $(dpkg-query -s sudo) ]];then
-		echo "::: Running sudo -u pihole $@"
-		sudo -u pihole "$@"
+		echo "::: Running sudo -u pihole $0 $@"
+		sudo -u pihole "$0" "$@"
 		exit $?
 	else
 		echo "::: Please install sudo."
--- a/advanced/Scripts/blacklist.sh
+++ b/advanced/Scripts/blacklist.sh
@@ -23,6 +23,29 @@ if [[ $# = 0 ]]; then
     exit 1
 fi
 
+# Check if pihole user, and if not then rerun with sudo.
+echo ":::"
+runninguser=$(whoami)
+if [[ "$runninguser" = "pihole"  ]];then
+	echo "::: You are pihole user."
+	# Older versions of Pi-hole set $SUDO="sudo" and prefixed commands with it,
+	# rather than rerunning as sudo. Just in case it turns up by accident, 
+	# explicitly set the $SUDO variable to an empty string.
+	SUDO=""
+else
+	echo "::: sudo will be used."
+	# Check if it is actually installed
+	# If it isn't, exit because the install cannot complete
+	if [[ $(dpkg-query -s sudo) ]];then
+		echo "::: Running sudo -u pihole $@"
+		sudo -u pihole "$@"
+		exit $?
+	else
+		echo "::: Please install sudo."
+	exit 1
+	fi
+fi
+
 #globals
 blacklist=/etc/pihole/blacklist.txt
 adList=/etc/pihole/gravity.list
@@ -153,15 +176,11 @@ function Reload() {
 	echo ":::"
 	echo -n "::: Refresh lists in dnsmasq..."
 
-	dnsmasqPid=$(pidof dnsmasq)
-
-	if [[ $dnsmasqPid ]]; then
-		# service already running - reload config
-		sudo kill -HUP $dnsmasqPid
-	else
-		# service not running, start it up
-		sudo service dnsmasq start
-	fi
+	# Reloading services requires root.
+	# The installer should have created a file in sudoers.d to allow pihole user
+	# to run piholeReloadServices.sh as root with sudo without a password
+	sudo --non-interactive /usr/local/bin/piholeReloadServices.sh
+	
 	echo " done!"
 }
 
--- a/advanced/Scripts/blacklist.sh
+++ b/advanced/Scripts/blacklist.sh
@@ -23,28 +23,9 @@ if [[ $# = 0 ]]; then
     exit 1
 fi
 
-# Check if pihole user, and if not then rerun with sudo.
-echo ":::"
-runninguser=$(whoami)
-if [[ "$runninguser" = "pihole"  ]];then
-	echo "::: You are pihole user."
-	# Older versions of Pi-hole set $SUDO="sudo" and prefixed commands with it,
-	# rather than rerunning as sudo. Just in case it turns up by accident, 
-	# explicitly set the $SUDO variable to an empty string.
-	SUDO=""
-else
-	echo "::: sudo will be used."
-	# Check if it is actually installed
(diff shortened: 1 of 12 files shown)


dschaper:
Interesting, lots to look over there! (I was working on a systemD file monitor that would check for changes to white/black list and then fire off a script that would `systemctl reload dnsmasq`.) I do agree about avoiding `/etc/sudoers` and dropping into a subdir file for permissions... Might have to cherry pick some of the commits, I'm not sure where we stand with the code refactor and how this integrates, but thanks a ton for the PR and we'll be discussing it!

Mcat12:
Yes, this looks to be very interesting... Good work/ideas so far!

jacobsalmela:
This is definitely a large PR.  One thing I noticed right off the bat is a lot of the code is repeated in each scripts.  For example, checking if `pihole` user, and if not then rerun with `sudo`.  One thing I really like is the reduction in execution time by not calling `sudo` every time.

Also, since this is such a large PR, it also makes me think of the code refactor in #312.  

Since we are still just discussing this, I want to point out that we are adding a lot more complexity to Pi-hole with a PR like this, but making it difficult to maintain (i.e. having to adjust the code in each script if a change is needed).

I'm wondering if we should keep this PR in close sight, but slowly start making the code more maintainable by splitting out the variables and the functions.  Even as the codebase stands now, it seems to be a large task (especially with all the new PRs coming in).  So perhaps we should try to do this in phases:

1. Make `pihole.vars` > test > merge into master
2. Make `pihole.functions` > test > merge into master (or even make each function we need into it's own file so all the functions aren't running needlessly).
3. Merge the PRs (some updates may be needed on the contributors end)

I know `bash` isn't the most friendly language to make an entire piece of software with, but it is effective because it is very portable and it's always there.  And it almost feels like we would be making it into a OOPL by splitting everything out.

Bottom line:
I don't want the project to get so large so fast that we lose stability and the ability to scale properly, so I think we need to seriously consider revamping the codebase before trying to scale any higher.  Obviously, Pi-hole is still using a few ugly hacks and running a lot of things with elevated privileges, but it works and people like it.

ryt51V:
Point well understood from everyone that this PR is a point of discussion rather than something likely to be pulled as is, anytime soon, before the refactor, without a **lot** of testing, or even pulled at all.

@jacobsalmela Yes, the code reuse is rather poor.  I use my own fork / branches for my own use anyway, so I'm going to correct in my own functions include file for now, and once we have pihole.functions then I would look to integrating with that.

PromoFaux:
@ryt51V , you've made some quailty submissions, and I'm going to go over them. We've had a bit of a rethink as regards the refactor. Can you shoot us an email please (adblock@pi-hole.net)

dschaper:
This is a screw-up on my part, but the PR isn't based off of the Development branch.

dschaper:
The concepts are still valid, but the execution is off as the code has changed a ton since first opened.

jacobsalmela:
OK, we'll leave it open for discussion

PromoFaux:
@dschaper / @ryt51V is this PR still valid?

dschaper:
Concept it good, need to look at the user we're executing as and see if we can lower the levels of elevation, but the project has changed so much since this PR, it's good to close it and revisit the options.

PR #360, opened 2016-03-03
Store password as hash

Mcat12:
Should fix any security holes associated with storing a plaintext password and checking passwords through the web interface. See [#53](https://github.com/pi-hole/AdminLTE/pull/60) for the web side of things.

- Store hashed password using sha256 instead of plaintext.

@pi-hole/gravity

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -664,7 +664,7 @@ setPassword() {
 	
 	if [ $? = 0 ]; then
 		# Entered password
-		echo $pass > /etc/pihole/password.txt
+		echo -n "$pass" | sha256sum | awk '{print $1}' > /etc/pihole/password.txt
 	else
 		echo "::: Cancel selected, exiting...."
 		exit 1


snowchyld:
Please don't use SHA256, please use bcrypt

https://paragonie.com/blog/2016/02/how-safely-store-password-in-2016

paragonie-scott:
What version of PHP does this project use?

* **PHP 5.5 and newer**: Just use [`password_hash()` and `password_verify()`](https://secure.php.net/manual/en/book.password.php)
* **PHP 5.3.7 - 5.4**: Use [password_compat](https://github.com/ircmaxell/password_compat) by @ircmaxell (which is a compatibility layer for the 5.5 functions)
* **PHP < 5.3.7**: Your best bet is either upgrading, but your second-best is [phpass](http://openwall.com/phpass) (please choose upgrading to a [supported version](https://secure.php.net/supported-versions.php) if you find yourself here)

If you need a shell script one-liner, something like this would do the trick:

```sh
php /path/to/bcrypt_hash_generator.php # this will echo anyway
```
And the contents of bcrypt_hash_generator.php:
```php
<?php
echo password_hash($argv[1], PASSWORD_DEFAULT), "\n";
```

Mcat12:
Since the reason we are securing the php with a password is that we don't want ads whitelisting themselves, why do we need to change the encryption type? The password is just for the web interface's ability to whitelist/blacklist. I don't think that an ad company, having obtained your hash, would find it profitable to crack it. Unless I'm missing something? We will also eventually be moving the ad redirect target from port 80 to something else.

paragonie-scott:
> Since the reason we are securing the php with a password is that we don't want ads whitelisting themselves

Sure, that sounds bad.

> why do we need to change the encryption type? 

Just so we're clear, [none of this is *encryption*](https://paragonie.com/blog/2015/08/you-wouldnt-base64-a-password-cryptography-decoded).

> The password is just for the web interface's ability to whitelist/blacklist.

I think XKCD captured the risk here the best: https://xkcd.com/792/

Ideally users would use a password manager and we wouldn't worry about this, but go find the least sophisticated user you know and ask them what KeePass is. The results shouldn't be surprising.

> I don't think that an ad company, having obtained your hash, would find it profitable to crack it. Unless I'm missing something?

I'm not familiar with the cost-benefit analysis here. What if the ad company itself was hacked and the attacker wanted to crack the password in order to whitelist their own malvertising/watering-hole attack?

> We will also eventually be moving the ad redirect target from port 80 to something else.

Hopefully "something else" also means changing the protocol to HTTP over TLS. :)

Mcat12:
The ad can only guess the password, since they just link to the password protected php. They can't get anything from it except if they guessed right or not. Theoretically, they could attempt a brute force in the JavaScript by repeatably loading the php with different guesses. Still, nothing returns the hash, so the ad company can't get your hash or password.
We have been looking into TLS, but we would rather make the Pi-hole as stable as we can get it before continuing with that path. That's why on the admin repo, the whitelisting stuff got moved to it's own branch off dev. It's just not stable enough to release for awhile.

Mcat12:
I can look into using bcrypt instead, if that would be more secure.

ryt51V:
@Mcat12 - Forgive me for [repeating myself](https://github.com/pi-hole/AdminLTE/issues/39), but why do we prefer to implement password logic in PHP rather than just let mod_auth in lighttpd.conf handle it?  We could protect the whole web interface, or individual php files (e.g. add.php).  Surely it's easier to implement correctly but works just as well?

That being said, one downside is that lighttpd requires htpasswd files to be stored in MD5 format.

@paragonie-scott - Which is the lesser of two evils?  Rolling our own auth code using a good hash algorithm (say bcrypt), or using a larger project's (lighttpd's) much more scrutinized auth capabilities that requires MD5 hashes?

---

```
sudo htpasswd -cb /etc/lighttpd/passwd pihole mypassword
sudo chown root:www-data /etc/lighttpd/passwd
sudo chmod 640 /etc/lighttpd/passwd
```

lighttpd.conf:
```
server.modules = (
        # ...
        "mod_auth"
)

auth.backend = "htpasswd"
auth.backend.htpasswd.userfile = "/etc/lighttpd/passwd"

auth.require = ( "/" =>
	(
			"method" => "basic",
			"realm"  => "Authorization please",
			"require" => "user=pihole"
	)
)
```

Or, proof of concept for only protecting certain files:
```
$HTTP["url"] =~ "^/admin/queries\.php" {
	auth.require = ( "" =>
			(
					"method" => "basic",
					"realm"  => "Authorization please",
					"require" => "user=pihole"
			)
	)
}
```

dschaper:
We are looking at additional options for webserver packages at the users
choice.

Dan Schaper via GS3
Ganymeade Systems

ryt51V:
@dschaper That is a very good point.  But what's the scope of the choice of web servers - just a choice of lighttpd, apache and nginx?  IMO, even three web server config files seems both easier and better from both dev and user perspective.  Apache and nginx can use the very same htpasswd file, so it wouldn't require a lot more logic in the install script - just a few more lines in each config file.

dschaper:
Very good points to consider. @mcat12 I defer to you, you have more
expertise in this area than I do.

Dan Schaper via GS3
Ganymeade Systems

Mcat12:
We can allow for choice in web servers. I'll look into mod_auth and if it works, it'll do. The main thing is that we want a simple authentication (both for users and for us) that will block ads from trying to cause harm to the whitelist and whatever else we add.

ryt51V:
I've just implemented this in my own fork for both apache and lighttpd.  IMO it works great for both web servers.  I've passworded the entire admin interface rather than just specific files.  See the changes I had to make [here](https://github.com/ryt51V/pi-hole-alt/compare/f3f3cb7431f868b2c0d5ad4a76a3b89d0aa326f8...ryt51V:0717d4d0fe0e823135fb34d9f0853ca5c1865402).

If this is the way you want to go then I can re-fork development, reimplement it there and make a pull request.

(Though I'm still eager to hear @paragonie-scott's opinion on whether this is a step back due to MD5.)

Mcat12:
I think we will move towards mod_auth. In your implementation, is that covering every file in the web server? And with these changes, wouldn't the ajax for the php need to include the authentication headers, if the auth covers every file and folder?

Mcat12:
Here's how we can implement authentication into the JavaScript. http://stackoverflow.com/a/11960692/3267834

dschaper:
I trust you. On site so ping me if needed.

Dan Schaper via GS3
Ganymeade Systems

ryt51V:
@Mcat12  My implementation covers every file on the split port / vhost I've dedicated to the admin interface.

But on the dev branch we can do it on the combined vhost by the whole subdir (`^/admin/`) or certain files in the subdir (`^/admin/(add|remove)\.php$)`) - whatever is needed.

I'm not so sure on AJAX, but my understanding is that browsers are intelligent and reuse HTTP credentials for every connection they make to the website (my understanding is that AJAX is client side...right?).  Is this something already in the Admin-LTE?  Is there any particular link or button I need to test?  (Pretty sure I got them all!)

Mcat12:
You can test ajax by loading the whitelist branch. If the ajax requires authentication, then we can limit auth to only the php files and reuse the password prompt from this branch (adding a username box) and send the auth in the ajax headers.

ryt51V:
Tested just now with both lighttpd and apache, with the big 3 browsers.  It works exactly the same with the entire admin interface protected in lighttpd.conf / apache vhost as it does with no password whatsoever - i.e. blacklisting and whitelisting work fine*, and without any additional prompts.

*(Both with or without a web server password (so unrelated), refresh functionality seems to be broken in IE11.  Dunno if this is a known issue.)

Mcat12:
What about the case where someone wants people to be able to view the graphs and such, but not edit the whitelist? By securing everything, they need the password. By giving them the password, they can edit the interface. By securing only the php, those who want everything secured won't have their way implemented. Maybe give an option of either securing the whole thing or only the php?

Mcat12:
Refresh seems to work on my IE11. Were there errors in the console/networking tab?

ryt51V:
Having two options is a very minor config difference we should be able to offer, so long as we can lock it down at the file level (e.g. looks like it's list.php that matters).

We're spoiled for choice; there are probably fifty different ways per web server to do this.  e.g. Lighttpd can filter with `$HTTP["url"]` or in the `auth.require` line.

With Apache it could be something like
`<Location /admin/>`
versus
`<Location ~ "^/admin/list\.php$">`

With Lighttpd it could be something like
`$HTTP["url"] =~ "^/admin/" {`
versus
`$HTTP["url"] =~ "^/admin/list\.php$" {`

IMO it's a bit of feature creep, but I can see the use in it and it's not a lot of effort.

ryt51V:
No errors on IE11 console or networking tab.

Looks like some sort of cache issue.  You can open the blacklist page in IE11 (or Edge) and Chrome, blacklist a website on IE11, hit refresh in Chrome and see the website (any refresh button - browser or on-page), but even a CTRL+F5 won't show the new website in IE.  Delete temp internet files in IE11 and then CTRL+F5 shows the new website.

I'm certainly not bothered about this, so if I'm the only one then chalk it down to my setup until someone else reports it.

Mcat12:
I'm seeing the same thing. It just gets it from cache, not actually doing any more work.

Mcat12:
I say go for securing the php files and then we can modify the existing password dialog to fit it. Make a PR if you want to implement it, or we could.

ryt51V:
Just to be clear, which PHP files?  Just list.php?

Mcat12:
admin/php/add.php and admin/php/sub.php. These are the only files that can actually change anything on your device.

Mcat12:
These are the only two that need it, because if only list.php was secured, an ad website could still whitelist themselves with add.php

Mcat12:
Closing the PR, but we can continue discussing.

PR #369, opened 2016-03-07
Optionally block the BBC Breaking News banner

jpsingleton:

-  Optionally block the [BBC News website](http://www.bbc.co.uk/news) Breaking News banner by black-holing `polling.bbc.co.uk`

@pihole/gravity


dschaper:
Thanks, since the list was commented out I'll just merge it!

jpsingleton:
Cheers. I didn't think it should be blocked by default but it's annoyed enough people that it would be useful as an option. Especially for phones where using a local hosts file is difficult. Unless you think it should be uncommented?

PR #370, opened 2016-03-07
Fix URL for Windows 10 Telemetry list.

stejoo:
Fixes the URL for the Windows 10 Telemetry list.

Developer of HostsWindowsBlocker renamed that project to WindowsSpyBlocker.
Also separated the domain list for Telemetry and Windows Update into to separate files. So Windows Update should not end up being blocked anymore.

- adlist.default updated to the new URL of Windows 10 Telemetry Hosts list.

@pihole/gravity


Mcat12:
Thanks!

PR #386, opened 2016-03-15
Respond to blocked javascript requests with a valid javascript stub

### LINKED ISSUE #385: Return valid javascript instead of html stub for *.js (PR available)
##### Expected Behaviour:
When js is blocked, return a valid js file (will mostly cleanup the log).  This will not totally prevent js errors, as functions will not be present for on-page scripts, but it will clean up a large majority of console errors.

##### Actual Behaviour:
An html stub is returned regardless of the media type

```html
<html>
<body>
</body>
</html>
```

Use pi-hole, visit any site with blocked ads, open the console.

jalopnik_-_drive_free_or_die_and_untitled_text_46

rosskevin:
Fixes #385 

- add a js stub

- add a rewrite rule

- make sure it is copied in the automated install

@pihole/gravity

### DIFF
--- a/advanced/index.js
+++ b/advanced/index.js
@@ -1,0 +1,1 @@
+var x = "X-Pi-hole: A black hole for Internet advertisements."
--- a/advanced/index.js
+++ b/advanced/index.js
@@ -1,1 +1,1 @@
-var x = "X-Pi-hole: A black hole for Internet advertisements."
+var x = "Pi-hole: A black hole for Internet advertisements."
--- a/advanced/lighttpd.conf
+++ b/advanced/lighttpd.conf
@@ -53,4 +53,6 @@ $HTTP["url"] =~ "^/admin/" {
 $HTTP["url"] =~ "^(?!/admin)/.*" {
 	# Create a response header for debugging using curl -I
 	setenv.add-response-header = ( "X-Pi-hole" => "A black hole for Internet advertisements." )
+        # rewrite only js requests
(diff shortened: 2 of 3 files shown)


dschaper:
Thanks! We also have been passing around the idea of hosting some CDN content on the Pi-hole, but that's just in the discussion phase right now, along with the option to use `nginx` along with or instead of `lighttpd`.

PromoFaux:
Do you think you could maybe add something along the lines of `console.log("JS File blocked by Pi-hole!);`

Just so users have some confirmation that it's working. I'm not too hot on my javascript.. is there a way that we could also print the url that was requested, so `GET - <requested URL> - Blocked by Pi-hole!` 

Other than that, looks good, It can be merged to development for sure!

PromoFaux:
Like this :D
image

PromoFaux:
```javascript
var scripts = document.getElementsByTagName("script"),
    src = scripts[scripts.length-1].src;
var x = "Pi-hole: A black hole for Internet advertisements."
console.log("Script blocked by pi-hole!! (" + src + ")");
```

PromoFaux:
Although it doesn't work everywhere:
image

PromoFaux:
Actually, works in less places than it doesn't work. just the `console.log("Script blocked by Pi-hole!")` will suffice. We can add that in later if needs be. Thanks for the PR.

PromoFaux:
Merged into development branch (commit: 26b7e7e)

PR #389, opened 2016-03-16
Secure Pulls now with HTTPS

### LINKED ISSUE #310: Use https for install.pi-hole.net
Currently the server at install.pi-hole.net does not support https.  This means that the 301 redirect header is sent in plain text and is subject to MITM attacks.  The 301 redirect could be altered in transit because there's no way to authenticate it.  It should also be obvious that such a thing is open to a variety of DNS attacks.

In short, I'd like the following to work and the documentation to be updated accordingly.

    curl -L https://install.pi-hole.net | bash

Thank you for your consideration.

dschaper:
Fixes #310.

Add new SSL pull ability. HTTPS!!

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,11 @@
 1. Install Raspbian 
 2. Run the command below
 
-### ```curl -L install.pi-hole.net | bash```
+### ```curl -L https://install.pi-hole.net | bash```
 
 #### Alternative Semi-Automated install ####
 ```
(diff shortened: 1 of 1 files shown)


dschaper:
Ready for cross-check and merge when reviewed. (You know I hate merging my own PR's and don't like merging directly to master... I know, extra work, but I like it...)

PR #393, opened 2016-03-18
Update mechanism

### LINKED ISSUE #98: Need an update mechanism
There is no easy way for existing Pi-hole users to update their `gravity.sh` script and Pi-hole support files or the Web interface.  I would like to get some sort of mechanism installed that can query the repos and install updates.  

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28661931-need-an-update-mechanism?utm_campaign=plugin&utm_content=tracker%2F3011939&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F3011939&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jacobsalmela:
PR to add an update mechanism to close #98 and #290.

It’s pretty clean and I have been using it effectively for some time now.  Discussion or improvements are welcomed.

**_edit - promofaux_**:
todo:
- [ ] flesh out `distributeFiles()` to actually copy around new/updated files
- [ ] run gravity at the end of the script


Mcat12:
Does the script actually update the repo? And will it copy over the updated whitelist.sh, gravity.sh, etc?

PromoFaux:
It updates the repo, by the looks of things, but doesn't copy files into the correct place...?

jacobsalmela:
Yep.  I completely forgot to move them into place.  I'm obviously off my game...

jacobsalmela:
Looking at it now, maybe it's more worthwhile to do the refactor.  I feel like I would just be re-writing the same code that's already in the installer.  Just ignore my ineptitude.

Mcat12:
This would work pretty well after the refactor, because then you can simply source the install functions and call the ones that move scripts around. You did pretty good. :+1:

jacobsalmela:
So couldn't we just update the repo and then run the installer locally?

PromoFaux:
We could, but I thought the idea was that users didn't want to go through all the install steps again (interface, upstream dns etc)

jacobsalmela:
Yeah.  Maybe we can just work some logic into the installer script.  I've done a lot with the macOS port so I might be able to just add it in there.

dschaper:
See #625

PR #406, opened 2016-03-26
Pihole debug

nate-ubiquisoft:
Added script piholeDebug.sh to generate a detailed logfile `/var/log/pihole_debug.log`
It logs the following items:

- Ip adresses of the Pi
- Checks access to local Gateway
- Compares Whitelist.txt to gravity.list for any matches
- Compares Blacklist.txt to gravity.list for any matches
- Test nslookup
 - 2 nslookups, first through pihole as dns, second though 8.8.8.8
 - URL to lookup is chosen from gravity.list if there are any whitelist matches, otherwise use doubleclick.com as it is more globally available.
- Log dnsmasq.conf without whitespace or comments
- Log 01-pihole.conf without whitespace or comments
- Log lighttpd.conf without whitespace or comments
- Log gravity.list line count
- Log `hostname` of this Pi
- Log `/etc/hosts` file
- Log Whitelist.txt
- Log Blacklist.txt
- `tail -f` with interruption instructions to log pihole.log until interruption event
- Futureproof interrupt trap to execute functions after interruption before exiting script

All of this has been thoroughly tested on my end. Not much to "break" since this is just another script but will help drastically cut down time spent on support cases.

@pihole/gravity


nate-ubiquisoft:
This is also now ready for UAT. The stdout is now consistent with basic-install.sh and uninstall.sh

PR #425, opened 2016-04-03
Pi hole controller

nate-ubiquisoft:
Fixed whiptail spacing.

Only known issue is sourcing bash_completion configuration doesn't work from script. If user exits or changes shells, it reloads automatically. So this is not critical but still looking for a fix.

@pihole/gravity


PromoFaux:
Oh. I thought I merged this in.... Sorry! Next release!

PR #432, opened 2016-04-05
Hotfix 2.6.1

PromoFaux:
Bring hotfix for Pi-hole not blocking ads to master branch.

### DIFF
--- a/advanced/Scripts/blacklist.sh
+++ b/advanced/Scripts/blacklist.sh
@@ -96,7 +96,7 @@ function AddDomain(){
 		echo " done!"
 	else
 	if $versbose; then
-		echo "::: $1 already exists in blacklist.txt! No need to add"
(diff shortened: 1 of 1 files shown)


VDRrulez:
And furthermore you maybe want to update pihole.cron?

Mcat12:
Update it with what?

VDRrulez:
@Mcat12 
If you take a quick look at pihole.cron then you will recognize that it references scripts in /usr/local/bin that do not exist anymore since rel 2.6. So I do not know what the intentions of the devs are. Either /usr/local/bin/pihole has to be started or the corresponding scripts in /opt/pihole.
And this is so obvious that I really wonder why I have to explain it to you.

PromoFaux:
Ah yeah, good catch.  We'll get that updated.. 

No need to be rude @VDRrulez,  we work on this in our spare time, and we're only human after all!

PR #435, opened 2016-04-05
Fix for #432

PromoFaux:
Fixes #432

-  Update `pihole.cron` to reflect new commands

@pi-hole/gravity

### DIFF
--- a/advanced/pihole.cron
+++ b/advanced/pihole.cron
@@ -11,16 +11,16 @@
 
 # Pi-hole: Update the ad sources once a week on Sunday at 01:59
 #          Download any updates from the adlists
-59 1    * * 7   root    /usr/local/bin/gravity.sh
+59 1    * * 7   root    /usr/local/bin/pihole updateGravity
 
 # Pi-hole: Update the Web interface shortly after gravity runs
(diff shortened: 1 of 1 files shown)


advanced/pihole.cron
```diff
@@ -11,16 +11,16 @@
...
 
 # Pi-hole: Update the Web interface shortly after gravity runs
 #          This should also update the version number if it is changed in the dashboard repo
-30 2    * * 7   root    /usr/local/bin/updateDashboard.sh
+30 2    * * 7   root    /usr/local/bin/pihole updateDashboard
 
 # Pi-hole: Parse the log file before it is flushed and save the stats to a database
 #          This will be used for a historical view of your Pi-hole's performance
-#50 23  * * *   root    /usr/local/bin/dailyLog.sh
+#50 23  * * *   root    /usr/local/bin/dailyLog.sh # note: this is outdated
```
TylonHH: Why is this outdated? What will happen if I enable it?

PromoFaux: Should probably be removed all together, to be honest. the script does not exist, let alone work..!

I think the original idea was to backup the logs, rather than truncating them.

Issue #194, opened 2016-01-18
'unsafe-inline' 'unsafe-eval' directives for css and js


magarto:
Hi,

To improve security, 'unsafe-inline' 'unsafe-eval' directives for css and js must be disabled.
According to the following link which I quote at the end, in Nginx there is a header which can be added to secure it. I don't know which is the right header for lighttpd but must be similar.
If we add that security directive your application shows with errors in design. A screenshot can be found here: http://imgur.com/Yw9mP8a

https://gist.github.com/plentz/6737338

with Content Security Policy (CSP) enabled(and a browser that supports it(http://caniuse.com/#feat=contentsecuritypolicy),
you can tell the browser that it can only download content from the domains you explicitly allow
http://www.html5rocks.com/en/tutorials/security/content-security-policy/
https://www.owasp.org/index.php/Content_Security_Policy
I need to change our application code so we can increase security by disabling 'unsafe-inline' 'unsafe-eval'
directives for css and js(if you have inline css or js, you will need to keep it too).
more: http://www.html5rocks.com/en/tutorials/security/content-security-policy/#inline-code-considered-harmful
add_header Content-Security-Policy "default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval' https://ssl.google-analytics.com https://assets.zendesk.com https://connect.facebook.net; img-src 'self' https://ssl.google-analytics.com https://s-static.ak.facebook.com https://assets.zendesk.com; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com https://assets.zendesk.com; font-src 'self' https://themes.googleusercontent.com; frame-src https://assets.zendesk.com https://www.facebook.com https://s-static.ak.facebook.com https://tautt.zendesk.com; object-src 'none'";

dschaper:
I don't quite understand what this issue is. Is there something that Pi-Hole is currently doing that is incorrect?

PromoFaux:
ping

PromoFaux:
Closing this issue. Reopened a [thread](https://github.com/pi-hole/AdminLTE/issues/27) in the AdminLTE repo.

PR #444, opened 2016-04-06
Removing functionality that serves blank web page.

PromoFaux:
**DO NOT MERGE UNTIL PROMOFAUX SAYS SO. K?**

The purpose of this pull is to change the way pi-hole works, slightly. The main differences will be:

- `gravity.list` to redirect to `0.0.0.0`(IPv4) or `::` (for IPv6)
- Pi-hole will now no longer return a blank page in place of an ad, it will return nothing.
- Web Admin UI will now be moved to a port (default to be decided - but user configurable)
- when pinging a blocked domain, it will not resolve (this is a good thing!)
- It will no longer appear that we're manipulating the `hostname` (we weren't, but I can see how you might think we were)
- Hopefully some of the issues brought up [here](https://github.com/pi-hole/AdminLTE/issues/39) by @fedora-core can be bypassed using this method (as there will be no page served when naughty ad provider provides a link to `adcompany.com/admin/whitelist.php`
- with the above in mind, that means we can get on with adding neat features to the web admin page.

@pi-hole/gravity

### DIFF
--- a/advanced/Scripts/blacklist.sh
+++ b/advanced/Scripts/blacklist.sh
@@ -27,12 +27,8 @@ versbose=true
 domList=()
 domToRemoveList=()
 
-piholeIPv6file=/etc/pihole/.useIPv6
-
-# Otherwise, the IP address can be taken directly from the machine, which will happen when the script is run by the user and not the installation script
-IPv4dev=$(ip route get 8.8.8.8 | awk '{for(i=1;i<=NF;i++)if($i~/dev/)print $(i+1)}')
-piholeIPCIDR=$(ip -o -f inet addr show dev "$IPv4dev" | awk '{print $4}' | awk 'END {print}')
-piholeIP=${piholeIPCIDR%/*}
+piholeIP="0.0.0.0"
+piholeIPv6="::"
 
 modifyHost=false
 
@@ -41,13 +37,6 @@ if [[ -r $piholeDir/pihole.conf ]];then
     echo "::: Local calibration requested..."
         . $piholeDir/pihole.conf
 fi
-
-
-if [[ -f $piholeIPv6file ]];then
-    # If the file exists, then the user previously chose to use IPv6 in the automated installer
-    piholeIPv6=$(ip -6 route get 2001:4860:4860::8888 | awk -F " " '{ for(i=1;i<=NF;i++) if ($i == "src") print $(i+1) }')
-fi
-
 
 function helpFunc()
 {
--- a/advanced/Scripts/whitelist.sh
+++ b/advanced/Scripts/whitelist.sh
@@ -27,12 +27,8 @@ versbose=true
 domList=()
 domToRemoveList=()
 
-piholeIPv6file=/etc/pihole/.useIPv6
-
-# Otherwise, the IP address can be taken directly from the machine, which will happen when the script is run by the user and not the installation script
-IPv4dev=$(ip route get 8.8.8.8 | awk '{for(i=1;i<=NF;i++)if($i~/dev/)print $(i+1)}')
-piholeIPCIDR=$(ip -o -f inet addr show dev "$IPv4dev" | awk '{print $4}' | awk 'END {print}')
-piholeIP=${piholeIPCIDR%/*}
+piholeIP="0.0.0.0"
+piholeIPv6="::"
 
 modifyHost=false
 
@@ -41,12 +37,6 @@ if [[ -r $piholeDir/pihole.conf ]];then
     echo "::: Local calibration requested..."
         . $piholeDir/pihole.conf
 fi
-
-if [[ -f $piholeIPv6file ]];then
-    # If the file exists, then the user previously chose to use IPv6 in the automated installer
-    piholeIPv6=$(ip -6 route get 2001:4860:4860::8888 | awk -F " " '{ for(i=1;i<=NF;i++) if ($i == "src") print $(i+1) }')
-fi
-
 
 function helpFunc()
 {
@@ -133,20 +123,19 @@ function ModifyHostFile(){
 	 if $addmode; then
 	    #remove domains in  from hosts file
 	    if [[ -r $whitelist ]];then
-        # Remove whitelist entries
-				numberOf=$(cat $whitelist | sed '/^\s*$/d' | wc -l)
-        plural=; [[ "$numberOf" != "1" ]] && plural=s
-        echo ":::"
-        echo -n "::: Modifying HOSTS file to whitelist $numberOf domain${plural}..."
-        awk -F':' '{print $1}' $whitelist | while read -r line; do echo "$piholeIP $line"; done > /etc/pihole/whitelist.tmp
-        awk -F':' '{print $1}' $whitelist | while read -r line; do echo "$piholeIPv6 $line"; done >> /etc/pihole/whitelist.tmp
-        echo "l" >> /etc/pihole/whitelist.tmp
-        grep -F -x -v -f $piholeDir/whitelist.tmp $adList > $piholeDir/gravity.tmp
-        rm $adList
-        mv $piholeDir/gravity.tmp $adList
-        rm $piholeDir/whitelist.tmp
-        echo " done!"
-
+            # Remove whitelist entries
+			numberOf=$(cat $whitelist | sed '/^\s*$/d' | wc -l)
+	        plural=; [[ "$numberOf" != "1" ]] && plural=s
+	        echo ":::"
+	        echo -n "::: Modifying HOSTS file to whitelist $numberOf domain${plural}..."
+	        awk -F':' '{print $1}' $whitelist | while read -r line; do echo "$piholeIP $line"; done > /etc/pihole/whitelist.tmp
+	        awk -F':' '{print $1}' $whitelist | while read -r line; do echo "$piholeIPv6 $line"; done >> /etc/pihole/whitelist.tmp
+	        echo "l" >> /etc/pihole/whitelist.tmp
+	        grep -F -x -v -f $piholeDir/whitelist.tmp $adList > $piholeDir/gravity.tmp
+	        rm $adList
+	        mv $piholeDir/gravity.tmp $adList
+	        rm $piholeDir/whitelist.tmp
+	        echo " done!"
 	  	fi
 	  else
 	    #we need to add the removed domains to the hosts file
@@ -195,7 +184,7 @@ do
     "-d" | "--delmode"   ) addmode=false;;
     "-f" | "--force"     ) force=true;;
     "-q" | "--quiet"     ) versbose=false;;
-    "-h" | "--help"			 ) helpFunc;;
+    "-h" | "--help"		 ) helpFunc;;
     *                    ) HandleOther "$var";;
   esac
 done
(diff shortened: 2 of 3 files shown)


Mcat12:
If we redirect to 0.0.0.0, then would we need to move the port? And I thought I had merged that stuff fixing the hostname issue (If there's a hostname, put it on the top of the list with pi.hole below).

PromoFaux:
I've removed it, as otherwise pinging the hostname or `pi.hole` would return `0.0.0.0`

The thought behind changing the port for adminLTE was so it could be installed without interfering with a user's existing webserver (if they have one)

Mcat12:
Ah, yes... Continue :+1:

PromoFaux:
That wont work, because it will return '127.0.0.1' as the IP address for the query `pi.hole`, which means it would need to be installed on the client machine!

PromoFaux:
Though, having just tested it, setting it as the pi's actual IP _does_ work, so we could do that, also.

Fedora-Core:
Since I have been mentioned in this, allow me to comment.

"In Linux 2.2 an interface can carry multiple IP addresses,
this replaces the alias device concept in 2.0"

Linux 2.0 was published in 1996, June 9.
Linux 2.2 was published in 1999, January 25.

No longer running a WebServer for adblock would break Jacob Salmela's promise on his Blog site, where he uses the WebServer's AccessLog for the Logstalgia eye-candy.

Just my 2 cents....

Greetings, F.C.

PromoFaux:
>would break Jacob Salmela's promise on his Blog site.

I don't know. Is it a promise? I saw it as more of a "Here's some neat thing you could do if you wanted to".  `dnsmasq` also logs requests. Something else could be cobbled together, surely. 

I'm looking at this as an easy, short-term solution to provide users with something that makes things easier for them (web controls) without compromising the security too much. 

I still don't quite understand the multiple IP address angle. I'm a novice at this stuff.

Mcat12:
If we have the admin page still (we will), we will still have a web server (though possibly not running out of `/var/www/html`.

ajp2k13:
If we redirect to 0.0.0.0, won't we get a lot of broken elements on a page with just the placeholders showing or something? I know I used to when I ran adblocking on my router which redirected to 0.0.0.0, it wasn't as pretty as serving an empty page....

PromoFaux:
@ajp2k13 I haven't noticed anything so far, but then I also have uBlock origin installed. Perhaps I'll run without it for a bit.

If you want to give it a go yourself:

```bash
cd /etc/.pihole
sudo git fetch
sudo git checkout WebServerChanges
```

Which will install this branch on your pi-hole. The only change so far is that it redirects to `0.0.0.0` / `::`

PromoFaux:
The whole point of this branch is for a proof-of-concept anyway, we don't _have_ to go with it!

ajp2k13:
@PromoFaux I was just wondering because I noticed this when using 0.0.0.0 on my router. Some ads disappear but some show a broken link, looks a bit different depending on the browser used. uBlock will remove them anyway but try with an iPad or something.

Not saying I disagree or anything, just trying to understand the thinking behind it.

jacobsalmela:
I don't recall making a promise, but I do enjoy the ability to use the access log for stuff.  We also lose the ability of the X-headers for debugging.  I still need to test this out myself.

If we do go this route, maybe we should offer the option to choose this or to use the Web server.

PromoFaux:
@Fedora-Core Any helpful tips on the multiple IP angle?

Fedora-Core:
_Give a man a fish and..._

Here is your fish =>
[https://github.com/Fedora-Core/Pi-hole-adblock-multiple-IP-addresses](https://github.com/Fedora-Core/Pi-hole-adblock-multiple-IP-addresses)

Greetings,
F.C.

PromoFaux:
Top man.  I'll take a look when I get home from the pub....

PromoFaux:
@Rahel-A. Wut? Has github gone mad?

image

Rahel-A:
@PromoFaux I'm not sure. I was trying to fork and propose a bug fix, but I stopped. I don't know what this message means though.

PromoFaux:
It's weird. It's a commit for the original AdminLTE, right? I wonder how it got mixed up with this!

Rahel-A:
Yes, I think I might have mixed up the URL. That was the reason I cancelled.

Mcat12:
The whitelist/blacklist scripts will need to be updated

Mcat12:
Ok, the hostname and `pi.hole` will continue to work as they had, while everything else gets sucked into the Pi-hole. :)

Mcat12:
How about moving the port change to another PR? It would be nice to get the other changes out.

PromoFaux:
> Ok, the hostname and pi.hole will continue to work as they had, while everything else gets sucked into the Pi-hole. :)

Thought: this would enable a naughty advertiser to try and load from `http://pi.hole/admin/whitelistOrWhateverItsCalled.php`

Mcat12:
Ugh... Tricky ads. This is why we can't have nice things.

Mcat12:
Any way around that?

PromoFaux:
Maybe just not have it in there at all. If the user wants to access the pi-hole with a hostname, then they just set it up in their `hosts` file?

PromoFaux:
(which is why I removed it)

Zegnat:
Just dropping in here to comment on this:

> Thought: this would enable a naughty advertiser to try and load from `http://pi.hole/admin/whitelistOrWhateverItsCalled.php`

If the PHP back-end makes use of a proper CSRF token this should not be a problem. The advertising platforms will not be able to make a POST request to `pi.hole` without a correct CSRF token and with proper CORS headers configured on the Pi-Hole server there is no way for them to get the CSRF token in the first place.

PromoFaux:
^ Please explain like I'm a person that has no idea what CSRF is... ;)

Zegnat:
> Please explain like I'm a person that has no idea what CSRF is... ;)

I will assume you know how web forms work and what a POST request is. Else this is going to take a bit too long for my lunch break :wink:

Say the admin has a page `whitelistme.php` that users can POST to through a web form. An external party will then execute code to simulate this POST request and hope to do something. (Read more about [the CSRF attack vector](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)) on the OWASP wiki.)

To solve this we want to make absolutely sure that the POST request is coming from our actual web form on `whitelistform.php`. So we add an extra (hidden) form field with a randomly generated ID. We also store this ID somewhere (probably in `$_SESSION`). Now when `whitelistme.php` gets POST data, we check if the CSRF token send along with the data matches the one in `$_SESSION`.

Because an external party will not be able to access your session data, they will never be able to send along the CSRF token when simulating a POST request.

(Read more about [the token pattern](https://www.owasp.org/index.php/CSRF_Prevention_Cheat_Sheet#General_Recommendation:_Synchronizer_Token_Pattern) on the OWASP wiki.)

PromoFaux:
Ah-hah! I like the sound of that, so I will read about it later on (If i'm not in the pub!)

My web development understanding is quite minimal. A bit of ASP.net at work, and as much PHP as I've hacked into this project.

Zegnat:
If you have any questions regarding this, feel free to reach out to me!

My interest in Pi-Hole is purely educational for now. As I do not currently own a RPi or manage a network. Otherwise I would have gladly done some work on the PHP side of things.

PromoFaux:
Thanks, I'll be sure to bug you later :)

>As I do not currently own a RPi

I actually run mine on a full debian server I have at home. Pi-hole should work on any hardware provided the OS is based on debian (more flavours of Linux to come in future)

Zegnat:
> Pi-hole should work on any hardware provided the OS is based on debian (more flavours of Linux to come in future)

Oh, I know. But I don't like setting up VMs and as I just emigrated I do not have any excess hardware. I might be ordering a RPi soon, but 0 promises. I'll help in-proxy with any PHP security questions I see here.

Mcat12:
So with this security added, we should be fine even if we didn't move to 0.0.0.0 ... That means we can add whitelisting and host editing etc without relying on 0.0.0.0. We could even keep the ability to replace (most) ads with an image/custom webpage... If @Fedora-Core doesn't find anything wrong, this sounds like an interesting change.

jacobsalmela:
I'm liking what the 0.0.0.0 does so far, but I also think people might still like the use of the Web server.  It can be used with visualization tools.  The `access.log` could also be used to create a list of what files were attempted to be accessed, which could be a nice feature for the dashboard.

Mcat12:
And on reddit users expressed a desire to keep the ability to replace ads with a small image, in order to fill in the spot where the ad was supposed to be.

Mcat12:
With 0.0.0.0 that would not be possible afaik

jacobsalmela:
I agree, it probably would not be possible with 0.0.0.0.

If we go this route, maybe let the user choose during install and explain what each method gets them.  That way, users with an exiting Web server could install without issues, and others could still have the Web server for the added features.

Mcat12:
Sounds good

Zegnat:
> So with this security added, we should be fine even if we didn't move to 0.0.0.0 ...

Probably, but it might not make a difference.

Depending on how the admin panel is going to be set-up, things like opening the panel in a frame and tricking the user into clicking a button (possibly through [clickjacking](https://www.owasp.org/index.php/Clickjacking)) could be a path of attack. So, like I said before, you still want to have proper [CORS headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS) in place. One I would recommend is [`X-Frame-Options`](https://www.owasp.org/index.php/HTML5_Security_Cheat_Sheet#X-Frame-Options) so the pi-hole cannot be embedded on a website.

As I do not do a whole lot of server administration nor browser implementation research I am unsure if we can send `X-Frame-Options` only for `pi.hole/admin` or if the server would always send it. If the server is always going to send it, **none** of the requests to pi-hole can ever be displayed in a frame so you might as well send requests to 0.0.0.0.

(Now that I have your attention on the subject of HTTP Headers, you might also want to look into defining a strict [CSP](http://www.html5rocks.com/en/tutorials/security/content-security-policy/). [Linking to HTML5 Rocks because OWASP [mentions it](https://www.owasp.org/index.php/HTML5_Security_Cheat_Sheet#Content_Security_Policy) in their cheatsheet but does not have a proper explanatory article on the matter.])

Mcat12:
Yes, we can send headers based on what is being requested. For example, we send a custom x header based on if you're requesting something from admin or not.

Zegnat:
> Yes, we can send headers based on what is being requested. For example, we send a custom x header based on if you're requesting something from admin or not.

If browser implementations go along with that, then yes, that would be the best. That way 0.0.0.0 is indeed unnecessary for the security aspect.

Mcat12:
#194 seems to be related to this.

Mcat12:
@Zegnat thanks for the html5rocks.com article! That helps with this and #194.

Mcat12:
X-Frame-Options is in the `development` branch and CORS, and the CSRF token are now in AdminLTE's `whitelist` branch. Next up is CSP...

Mcat12:
CSP will be implemented in another branch. The whitelist branch is now open as a PR [here](https://github.com/pi-hole/AdminLTE/pull/88).

Mcat12:
CSP is now in a PR [here](https://github.com/pi-hole/AdminLTE/pull/89).

thewriteway:
In my own testing - setting a blocked host to 0.0.0.0 caused issues with DHCP discover traffic and default routing configuration. Can anyone else corroborate? This is the reason another popular blocking application actually routes the traffic to a valid external IP that rejects all http traffic. Ironically it's google own DNS servers listed for blocked hosts, used as the workaround. Thoughts?

PromoFaux:
Closing due to being VERY out of date. Can pick up again if needed.

PR #450, opened 2016-04-08
Small fixes to whitelist blacklist

PromoFaux:
Tweaks to PR submitted by @7h3ju57

- removes call to `clear` terminal when running new list function of whitelist/blacklist
- forces quiet mode to tidy up output

@pi-hole/gravity


PromoFaux:
Need to tweak something else... now it reloads `dnsmasq` :)

PromoFaux:
Yep, that works. Merging.

PR #468, opened 2016-04-17
fixed non execution of the function helpFunc

mayurkr:
This PR fixes the issue which caused the helpFunc to not execute.

Following was the error noted while executing the above scripts.

pihole -w
/opt/pihole/whitelist.sh: line 14: helpFunc: command not found

pihole -b
/opt/pihole/blacklist.sh: line 14: helpFunc: command not found


PromoFaux:
Looks good. Thanks for the submission!

PR #472, opened 2016-04-21
Save custom IP

### LINKED ISSUE #471: gravity.sh overwrites custom IP settings in gravity.list
##### Expected Behaviour:
`gravity.sh` should preserve the customized IP (for example, an external IP) if one is in place instead of detecting it.

##### Actual Behaviour:
My pi-hole is on an external network with a private IP. During setup I specified its external IP. If `pihole -g` is run afterward, all the domains in `gravity.list` point to the internal IP (detected in `gravity.sh`), instead of the external IP specified during setup.

1. Run pihole setup script and set a different IP than the one detected.
2. Domains in `gravity.list` point to the customized IP.
3. Run `pihole -g`.
4. Domains in `gravity.list` point to the detected IP.

jim-liu:
Fixes #471  .

- Make `gravity.sh` preserve custom IP setting by not deleting `/tmp/piholeIP`.

- Make `whitelist.sh` use custom IP by reading `/tmp/piholeIP` if it exists.

@pi-hole/gravity

### DIFF
--- a/advanced/Scripts/blacklist.sh
+++ b/advanced/Scripts/blacklist.sh
@@ -42,12 +42,19 @@ verbose=true
 domList=()
 domToRemoveList=()
 
+piholeIPfile=/etc/pihole/piholeIP
 piholeIPv6file=/etc/pihole/.useIPv6
 
-# Otherwise, the IP address can be taken directly from the machine, which will happen when the script is run by the user and not the installation script
-IPv4dev=$(ip route get 8.8.8.8 | awk '{for(i=1;i<=NF;i++)if($i~/dev/)print $(i+1)}')
-piholeIPCIDR=$(ip -o -f inet addr show dev "$IPv4dev" | awk '{print $4}' | awk 'END {print}')
-piholeIP=${piholeIPCIDR%/*}
+if [[ -f $piholeIPfile ]];then
+    # If the file exists, it means it was exported from the installation script and we should use that value instead of detecting it in this script
+    piholeIP=$(cat $piholeIPfile)
+    #rm $piholeIPfile
+else
+    # Otherwise, the IP address can be taken directly from the machine, which will happen when the script is run by the user and not the installation script
+    IPv4dev=$(ip route get 8.8.8.8 | awk '{for(i=1;i<=NF;i++)if($i~/dev/)print $(i+1)}')
+    piholeIPCIDR=$(ip -o -f inet addr show dev "$IPv4dev" | awk '{print $4}' | awk 'END {print}')
+    piholeIP=${piholeIPCIDR%/*}
+fi
 
 modifyHost=false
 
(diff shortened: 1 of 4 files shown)


Mcat12:
I think perhaps saving it to /etc/pihole/ would a more reliable way to store it, since tmp is a temporary place.

jim-liu:
I agree, but I was trying to minimize the changes.

Mcat12:
That's fine. We're OK with changes, since that's what a PR is for. ;)

jim-liu:
Changed basic-install.sh so it saves $piholeIP to /etc/pihole/piholeIP, and changed gravity.sh, whitelist.sh and blacklist.sh to read the file if it exists.

Mcat12:
I'm not getting any piholeIP file in my directory after installing...

jim-liu:
You don't have a `/etc/pihole/piholeIP`?

Mcat12:
After removing my pihole directory and running the install script off of your branch (with .pihole on your branch) it doesn't have piholeIP in it.

jim-liu:
I don't know how to use the automated installer to test it (since it doesn't pull the development branch), so here's what I did:

1. Using the automated installer to install everything.
2. Create `/etc/pihole/piholeIP` with the VM's IP address.
3. Modify `gravity.sh`, `whitelist.sh` and `blacklist.sh` according to my commits.
4. Run `pihole -g` and notice the custom IP set in `/etc/pihole/piholeIP` is in effect.

If you could let me know how to test using the automated installer, I would run it again =)

PromoFaux:
OK, i've pulled it into a seperate branch for you to try out. You should be able to test using the automated installed by following these steps:

```BASH
cd /etc/.pihole
sudo git fetch
sudo git checkout jim-liu-dev
```

jim-liu:
Thanks! I tested it again. If the detected value was accepted during install, `/etc/pihole/piholeIP` will not be created. If a custom IP was specified (even if it's the same as the detected), `/etc/pihole/piholeIP` will be created and used for generating `/etc/pihole/gravity.list`.

I don't think `/etc/pihole/piholeIP` is needed if the user just let the installer detect the IP, since all subsequent runs of `pihole -g` will detect IP address as well.

PromoFaux:
I think it's probably worth creating the file anyway. I faced this issue myself a while back when I had an outgoing VPN connection on my pi as well as pihole. gravity would update with the IP address of `tun0` rather than `eth0` (Rather than code a fix for this, I just moved my VPN connection to a separate machine!)

jim-liu:
Yeah that makes sense. Since the IP shouldn't change regardless of it's detected or custom set.

PromoFaux:
added that to the `jim-liu-dev` if you want to test it.

jim-liu:
Thanks! Just tested it - using all default values during installation. `piholeIP` was correctly saved.

PromoFaux:
Super, If I get a chance later I will test it myself. Bit hectic in my real job recently!

Mcat12:
Seems to work fine.

PR #489, opened 2016-05-02
Next Release

Mcat12:
DO NOT MERGE


PromoFaux:
Do you think we should merge the Fedora support PR in before we next release? Maybe if we aim for this Sunday?

Mcat12:
Ya. Then we can focus on whitelisting from the web and other stuff

PR #490, opened 2016-05-02
Fedora Support

bcambl:
Add support for Fedora/CentOS
Tested platform: [CubieBoard2](http://cubieboard.org/2013/06/19/cubieboard2-is-here/) w/ [Fedora ARM](https://fedoraproject.org/wiki/Architectures/ARM)

I have separated some of the Debian specific commands/utilities/packages into a "compatibility variables" section. This PR will not only add support for Fedora based distributions (Fedora/RHEL/CentOS,etc), but also allow compatibility to be easily added for many UNIX-like systems.

I still have a few more commits to add.. just wanted to share my effort thus far.
Mainly looking for feedback and discussion.. but a merge is of-course completely welcome.

@pi-hole/gravity

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -749,12 +749,14 @@ configureFirewall() {
 			$SUDO echo "::: Configuring firewalld for httpd and dnsmasq.."
 			$SUDO firewall-cmd --permanent --add-port=80/tcp
 			$SUDO firewall-cmd --permanent --add-port=53/tcp
+			$SUDO firewall-cmd --permanent --add-port=53/udp
 			$SUDO firewall-cmd --reload
 		fi
 	elif [ -x "$(command -v iptables)" ]; then
 		$SUDO echo "::: Configuring iptables for httpd and dnsmasq.."
 		$SUDO iptables -A INPUT -p tcp -m tcp --dport 80 -j ACCEPT
 		$SUDO iptables -A INPUT -p tcp -m tcp --dport 53 -j ACCEPT
+		$SUDO iptables -A INPUT -p tcp -m udp --dport 53 -j ACCEPT
 	else
 		$SUDO echo "::: No firewall detected.. skipping firewall configuration."
 	fi
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -68,7 +68,8 @@ if [ -x "$(command -v rpm)" ];then
 	PKG_COUNT="$PKG_MANAGER check-update | grep -v ^Last | grep -c ^[a-zA-Z0-9]"
 	INSTALLER_DEPS=( iproute procps-ng newt dhcpcd )
 	PIHOLE_DEPS=( bind-utils bc dnsmasq lighttpd lighttpd-fastcgi php-common php-cli php git curl unzip wget findutils cronie )
-	LIGHTTPD_USER="lighttpd:lighttpd"
+	LIGHTTPD_USER="lighttpd"
+	LIGHTTPD_GROUP="lighttpd"
 	LIGHTTPD_CFG="lighttpd.conf.fedora"
 	package_check() {
 		rpm -qa | grep ^$1- > /dev/null
@@ -83,7 +84,8 @@ elif [ -x "$(command -v apt-get)" ];then
 	PKG_COUNT="$PKG_MANAGER -s -o Debug::NoLocking=true upgrade | grep -c ^Inst"
 	INSTALLER_DEPS=( apt-utils whiptail )
 	PIHOLE_DEPS=( dhcpcd dnsutils bc dnsmasq lighttpd php5-common php5-cgi php5 git curl unzip wget )
-	LIGHTTPD_USER="www-data:www-data"
+	LIGHTTPD_USER="www-data"
+	LIGHTTPD_GROUP="www-data"
 	LIGHTTPD_CFG="lighttpd.conf.debian"
 	package_check() {
 		dpkg-query -W -f='${Status}' "$1" 2>/dev/null | grep -c "ok installed"
@@ -543,7 +545,7 @@ installConfigs() {
 	fi
 	$SUDO cp /etc/.pihole/advanced/$LIGHTTPD_CFG /etc/lighttpd/lighttpd.conf
 	$SUDO mkdir -p /var/cache/lighttpd/compress
-	$SUDO chown $LIGHTTPD_USER /var/cache/lighttpd/compress
+	$SUDO chown $LIGHTTPD_USER:$LIGHTTPD_GROUP /var/cache/lighttpd/compress
 }
 
 stopServices() {
@@ -746,9 +748,9 @@ installPihole() {
 	if [ ! -d "/var/www/html" ]; then
 		$SUDO mkdir -p /var/www/html
 	fi
-	$SUDO chown www-data:www-data /var/www/html
+	$SUDO chown $LIGHTTPD_USER:$LIGHTTPD_GROUP /var/www/html
 	$SUDO chmod 775 /var/www/html
-	$SUDO usermod -a -G www-data pihole
+	$SUDO usermod -a -G $LIGHTTPD_GROUP pihole
 	$SUDO lighty-enable-mod fastcgi fastcgi-php > /dev/null
 
 	getGitFiles
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -68,6 +68,7 @@ if [ -x "$(command -v rpm)" ];then
 	PKG_COUNT="$PKG_MANAGER check-update | grep -v ^Last | grep -c ^[a-zA-Z0-9]"
 	INSTALLER_DEPS=( iproute procps-ng newt dhcpcd )
 	PIHOLE_DEPS=( bind-utils bc dnsmasq lighttpd lighttpd-fastcgi php-common php-cli php git curl unzip wget findutils cronie )
+	LIGHTTPD_CFG="lighttpd.conf.fedora"
 	package_check() {
 		rpm -qa | grep ^$1- > /dev/null
 	}
@@ -81,6 +82,7 @@ elif [ -x "$(command -v apt-get)" ];then
 	PKG_COUNT="$PKG_MANAGER -s -o Debug::NoLocking=true upgrade | grep -c ^Inst"
 	INSTALLER_DEPS=( apt-utils whiptail )
 	PIHOLE_DEPS=( dhcpcd dnsutils bc dnsmasq lighttpd php5-common php5-cgi php5 git curl unzip wget )
+	LIGHTTPD_CFG="lighttpd.conf.debian"
 	package_check() {
 		dpkg-query -W -f='${Status}' "$1" 2>/dev/null | grep -c "ok installed"
 	}
@@ -536,7 +538,7 @@ installConfigs() {
 		$SUDO chown "$USER":root /etc/lighttpd
 		$SUDO mv /etc/lighttpd/lighttpd.conf /etc/lighttpd/lighttpd.conf.orig
 	fi
-	$SUDO cp /etc/.pihole/advanced/lighttpd.conf /etc/lighttpd/lighttpd.conf
+	$SUDO cp /etc/.pihole/advanced/$LIGHTTPD_CFG /etc/lighttpd/lighttpd.conf
 }
 
 stopServices() {
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -786,6 +786,7 @@ installPihole() {
 	installCron
 	runGravity
 	configureFirewall
+	configureSelinux
 }
 
 configureSelinux() {
@@ -861,9 +862,6 @@ installPihole | tee $tmpLog
 
 # Move the log file into /etc/pihole for storage
 $SUDO mv $tmpLog $instalLogLoc
-
-# Configure SELinux (if applicable)
-configureSelinux
 
 displayFinalMessage
 
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -82,8 +82,8 @@ elif [ -x "$(command -v apt-get)" ];then
 	PKG_UPDATE="$PKG_MANAGER upgrade"
 	PKG_INSTALL="$PKG_MANAGER -y -qq install"
 	PKG_COUNT="$PKG_MANAGER -s -o Debug::NoLocking=true upgrade | grep -c ^Inst"
-	INSTALLER_DEPS=( apt-utils whiptail )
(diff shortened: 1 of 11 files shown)


Mcat12:
Looks promising!

PromoFaux:
Yeah, I like this. It's something we've been thinking about for a while, just have not had time to look into it. Will have to spin up a Fedora/CentOS VM to test. Thanks for the submission!

bcambl:
If anyone wants to install on Fedora or CentOS: 
Ensure to clone this branch to /etc/.pihole prior to executing the new `basic-install.sh`. This will ensure the lighttpd configuration templates are available from this PR. (Otherwise the installer will pull master)

PromoFaux:
This PR works With Fedora AND Debian, right?

bcambl:
Yes.
Verified with the current Debian 8 DO droplet.

PromoFaux:
Excellent, The last thing we want is fragmentation! I've not had time to test this personally yet, but hopefully one of the other guys can look it over, too!

dschaper:
Okay, have this on the list for Digital Droplets, looks like Fedora, Debian and CentOS are available for spin ups. I'll give them a shot and see what goes. I'll try the debug script but I don't remember offhand which distro uses which init system.

bcambl:
I did some droplet tests:
- [x] Debian 7.10
- [x] CentOS 7.2
- [x] Debian 8.3
- [ ] Fedora 23
    - /var/log/pihole.log currently requires 2 selinux contexts (works with SELinux disabled/permissive):
      - httpd_sys_rw_content_t
      - dnsmasq_var_log_t

I am not sure how normal it is for an httpd context to be required on a dnsmasq log file.. so SELinux on Fedora 23 would need to be disabled until this is addressed.

The `toilet` package does not exist in Fedora/CentOS (only used for chronometer pi-hole logo)

fatez:
Ubuntu version? :)

PromoFaux:
Ah yes, Ubuntu is a good point. I know they do something weird with the interface names....

There was a fix for Proxmox LXC's (#497) Though that was rejected as it was against the wrong branch... Just need someone (or one of us) to open up a PR with that change against the `development` branch.

@bcambl We've not forgotten this, I promise! Real life has gotten in the way of late...

Mcat12:
The toilet issue is solved. :wink:

bcambl:
Thanks, I will rebase against development next chance I get.

PromoFaux:
OK... let's try this

image

PromoFaux:
OK, tested on Centos7. Not looked at the code yet, but it appears to loop back to the dependency check after choosing upstream DNS servers....

PromoFaux:
Looks like you may need to rework the root check (found at the beginning of various scripts ie the `pihole` command) as it currently uses `dpkg` to determine root status (not installed on centos)

PromoFaux:
Also (sorry!) `dnsmasq` does not start

`May 14 10:58:47 localhost.localdomain dnsmasq[2021]: dnsmasq: cannot open log /var/log/pihole.log: Permission denied`

bcambl:
I will ta

PromoFaux:
Ah, i'm not sure. I span up the VM on a machine at work.. so I will check later

bcambl:
No problem!
Let me push some more commits here before testing again.
Thank you for taking some time to help test.

bcambl:
To test this branch/PR..

as root:
```
setenforce permissive
```
edit /etc/selinux/config to persist reboot
```
yum install -y git
git clone https://github.com/bcambl/pi-hole.git /etc/.pihole
/etc/.pihole/automated\ install/basic-install.sh
```
I think the SELinux problems with /var/log/pihole.log should be created as a separate issue.
/var/log/pihole.log currently requires 2 SELinux contexts (works with SELinux disabled/permissive):
- httpd_sys_rw_content_t
- dnsmasq_var_log_t

dschaper:
Okay, so what needs to be checked out on this? I have an OpenVZ CentOS deployment I can test with. SE is set to disabled (yeah, I know kind of defeats the purpose of running a RedHat type distro, but it's cheap <insert DodgeBall quote here>) 

Not sure if production would allow for changing up to permissive though, is it just the logging that conflicts with going full enforced? 

Looks like you already have DO covered, I'm in Windows-land so other than duplicating with Droplets of my own I'd have to create VirtualBox instances. (Not a problem to do, unless I get some time to look at Vagrant, but that's using the same visualization engine...)

bcambl:
Yes, I believe it is just that one log file that is causing selinux to complain. It requires both dnsmasq and http contexts (not really a normal situation). I have not had a chance to look into any custom selinux rules yet to address this. 
This PR as it stands should not affect any existing functionality but will allow install on Fedora/CentOS in addition to Debian; however, the uninstall script would still need some changes for for this PR to be _more complete_.

dschaper:
Thanks for the reply. I'll have to see if I can get a setup that allows for `permissive` so I can see what complaints it's throwing. From what I understand, `disabled` will drop all errors. OpenVZ isn't the best platform for testing on. Do you know offhand what the SELinux setup is over at DO? And do you think DO is sufficient for testing this out, or should I look at a CentOS VBox spinup?

bcambl:
I just checked my DO account:
__CentOS 7.2__: defaults to selinux `disabled`. After enabling SELinux (with a reboot/relabel) the droplet came back online with SELinux enforcing as suspected.
__Fedora__: defaults to selinux `enforcing`.

dschaper:
Thanks, when I get a few minutes I'll try a VBox'ed CentOS and go from there...

dschaper:
On **CentOS 7.2.1511** the install seemed to go okay, I did get a warning about the lack of the `lighty` utility, but I don't think that affects anything. I'm running with selinux enforced so I'm getting the log error. I'll update when I go to permissive...

bcambl:
You are correct, the `lighty` utility is not required on Fedora/CentOS. I added the warning message for the sake of visibility. 

The warning could be improved to __only__ display if:
- The `lighty` utility is _not_ found
- Deployed _not_ on Fedora/CentOS.

dschaper:
I'm okay with the warning, as it's only in the install process/install logs and as long as we know that it's a RH based distro then were fine. The question that I have, and I defer to those with more experience in that realm of Linux, is how to handle the SELinux situation. My thoughts are split, it looks like DO defaults to disabled, (off hand I don't know if they are OpenVZ/KVM/In-house) so it would take user action to enable that security level. But if this is going towards an enterprise install, then a stock CentOS/Fedora has SELinux enabled. Do we want to tell the users to disable a rather useful and major feature such as SELinux, or do we want to chase that down and see if we can't play nice with SELinux?

As for now, it's looking pretty good, and if a user installs the Pi-Hole on CentOS/Fedora and runs into trouble then the `pihole -d` command is going to include the `systemctl -i status` for the daemons and it's going to be rather obvious from that when the permissions error is highlighted.

bcambl:
My vote is with SELinux `Enforcing`.

dschaper:
Seconded

bcambl:
I have a commit to address SELinux.. I will add it soon.

bcambl:
@PromoFaux https://github.com/pi-hole/pi-hole/pull/490#issuecomment-219207353

The first set of dependencies is just for the installer (ie: apt-utils whiptail)  
The second set of packages are for the pi-hole installation.

PromoFaux:
I'd quite like to get this merged sooner rather than later before it there's too much disparity between the dev branch and this!

@dschaper have you had a chance to test the latest commits RE: SELinux?

dschaper:
I'll take a look, not sure where the merge conflicts came from? I thought it was green the last time I checked. Unless that last change Jacob made to the `sudo` detection is conflicting with 639bc6d.

dschaper:
Found it, it's in `automated installer/basic-install.sh`

```
<<<<<<< HEAD
    fi
    echo ":::"
    echo "::: Checking dependencies:"

  dependencies=( dnsutils bc dnsmasq lighttpd php5-common php5-cgi php5 git curl unzip wget sudo)
        for i in "${dependencies[@]}"; do
=======
        fi
        echo ":::"
        echo "::: Checking installer dependencies..."
        for i in "${INSTALLER_DEPS[@]}"; do
>>>>>>> 72498001d0ca0917847ec38a5848e2c70e5e9f8f
```

dschaper:
@bcambl Did you want to fix that up? Looks like you've created a new function for the dependency checks to go agnostic, and there's a new dependency array? My Bash abilities are not all that good so I don't want to go in and chunk up your work for a small merge conflict resolution.

bcambl:
No problem, I will fix this up shortly. Do you prefer a commit or a rebase?

Mcat12:
Commit

Mcat12:
You should just be able to add sudo to the dependencies (or whatever else is used in Fedora).

dschaper:
What ever is easier for you, there really isn't that much commit spam so I don't think a rebase would be necessary, just a new commit.

dschaper:
Other platforms use other package managers, @bcambl's solution works that angle, can't dpkg on Fedora...

bcambl:
So I looked at 1194e48. Adding sudo to the deps will never get hit due to [Line 62](https://github.com/pi-hole/pi-hole/blob/development/automated%20install/basic-install.sh#L62). Either way, I will fix this conflict.

dschaper:
Thank you! Going to give this a whirl on CentOS 7.

bcambl:
I would like to fix/rebase that last commit. Please give me a heads up if this PR is close to merge.

dschaper:
I'm running an install test now. Using a user that is not in sudoers but `su -` to root appears to complete successfully, but `journalctl` is full of red. `sealert` shows 2 hits, for `dnsmasq` for write to log, and `php-cgi` for read from log. I just cloned your fork, checked to make sure I was on the right branch and then fired off the install. Trying again with a more sane environment.

TLDR; SELinux looks to be blocking good, and I'm daft...

dschaper:
Would you like to rebase the last commit now and have me test it afterwards?

bcambl:
@dschaper, I am curious if those warnings are false-positive due to the SELinux configuration being one of the last tasks of the install.
After install has completed, please try `semodule -l | grep pihole`

dschaper:
Okay, grabbed restored to a fresh snapshot. Installing via `su -` shell from a non-sudo user. 

Looks like my CentOS 7 doesn't have the dhcpcd package available, but that just quickly flashes on the screen as the installer dependency check looks like it calls `yum install dhcpcd` and the `basic` script parses the `yum` no package found as a good response. 

Get an error `x_tables: ip_tables: udp match: only valid for protocol 17` and a `iptables -S` shows only the two tcp ports on the INPUT chain, not the udp port.

`journalctl` looks way cleaner this time and looks like the SELinux module compiled and installed correctly. 

Testing web interface from Win10 box looks good, PHP seems to be working just fine.

Ran `pihole -d` for debug and it didn't like the lack of `ifconfig`, so it was unable to detect the pihole address to test with. The `lighttpd` log file is loaded with errors though, so it looks like `lighttpd` is trying to run prior to the SEL modules being loaded and that fills the error logs, but once the module is loaded then everything quiets down pretty well. 

Just those small things, but I think it's good to merge in.

bcambl:
Thanks for testing, I will see what I can do to address these items quickly.

automated install/basic-install.sh
```diff
@@ -859,6 +858,9 @@ use4andor6
 # Decide what upstream DNS Servers to use
 setDNS
 
+# Configure SELinux (if applicable)
```
dschaper: I think this may be causing an issue where the install script is trying to copy files from `/etc/,pihole` to build the SEModules, but since `installPihole` hasn't been run yet, that directory hasn't been created?

bcambl: Should we move it between 786 and 787?

dschaper: Yeah, that looks like it should be okay.

dschaper:
Let me know when you're comfortable with the state of the code. Running your latest looks pretty much good to go from what I can see.

bcambl:
The installer now works with Fedora.. so I am good with it :+1: 

The SELinux policies are interesting and make me question security. I assume most people deploy this application behind a firewall and take additional measures to protect their servers.

Regardless, Debian has SELinux disabled by default and should be unaffected. If there is a Debian user out there that has security enabled, they will likely know how to create a policy.

I also notice effort on database support as well as a re-write, Perhaps some of these security issues will be addressed there.

dschaper:
Okay, let me do one last build on it and I think we should be good to merge, I don't see any changes since I last ran it, but just want to check.

As for security, well, we get a ton of flack for the curl to bash, but it's my hope that people aren't blindly doing that, hoping they look at the code first! 

Database will come with the Python. Hopeing to have a few other surprises with that re-write as well. 

One other question, from your branch name, are you running on the Pi hardware itself? Do you have any guides or a link to getting the hardfloat setup?

bcambl:
Fedora ARM installation and supported platforms:
[https://fedoraproject.org/wiki/Architectures/ARM](https://fedoraproject.org/wiki/Architectures/ARM)

dschaper:
Let's check with the Judges... Chuck, what do you think? :+1: Thank you Chuck Norris!

I say its good to go!

PromoFaux:
I'm good with it, as long as it's been tested and it doesn't break things...!

dschaper:
I haven't been able to find any breakage, the code is pretty much segmented out so that any changes are only made if a Fedora-like installation is detected. But if you guys want to test it on Debian just to check me that would be good.

dschaper:
If there's no objections in the next little while, I'll go ahead and merge this, because I'm going to be needing it on a CentOS 7 box in a bit.

PromoFaux:
No objections here. Go ahead and merge!

dschaper:
@bcambl I have a small CentOS7 KVM that I'm trying to set some sane access restrictions to possibly configure a public demo of the Pi-Hole for users to try out. If you are amenable. would you be able to get me a rsa pub key and I'll get you a shell on that box? I may need some of your knowledge on the RH/CentOS environment, if you'd be open to pointing out any major errors I'm making prior to releasing the address for the test setup to the world at large. (I have all password logins blocked, keys only to try and harden the box as much as possible.)

Thanks!

bcambl:
Thank you for the merge. My hopes are that this PR is useful and/or educational in regards to the subtle differences between Fedora/CentOS/RHEL and Debian based distributions.

I would be happy to take a quick look at your implementation. 
Perhaps we continue this discussion via email.. can I assume your email in the git logs is valid?

Cheers

dschaper:
You can always get to me at dschaper@ganymeade.com (I'm pretty sure that's what I have listed for my global...) https://pgp.mit.edu/pks/lookup?op=vindex&search=0x785A8B12CF053FDA

PR #491, opened 2016-05-03
Make sure web interface can't be loaded into a frame

Mcat12:

- Add `X-Frame-Options` to make sure the interface doesn't get loaded as an ad (protects the admin functions)

@pi-hole/gravity


Zegnat:
> * Remove `X-Pi-hole`. Unless it's going to be used, it might present a problem when using CORS (I think)

Unknown headers should not have any impact, other than the few extra bytes you send over the wire.

Mcat12:
OK. I wasn't sure, since the MDN articles seemed to say that you could only send certain headers.

Mcat12:
Tested. It works:
Test

PR #498, opened 2016-05-07
replace toilet with echo in chronometer.sh

yoosi:
Related to issue #490

- Replace `toilet` command with `echo` to remove dependency on `toilet` and `figlet`

- Remove `toilet` and `figlet` from install and uninstall scripts

@pi-hole/gravity


Mcat12:
Thanks, but you will need to reopen against the development branch as per the [contributing guidelines](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md). It looks like it might work though! :)

PR #535, opened 2016-06-20
Fix link for spam404scamlist.txt

fhemberger:
The redirect `http://spam404bl.com/spam404scamlist.txt` resolves to `https://raw.githubusercontent.com/spam404scamlist.txt/Dawsey21/Lists/master/main-blacklist.txt`, which is wrong.

Instead, use `https://raw.githubusercontent.com/Dawsey21/Lists/master/main-blacklist.txt`directly.

@pi-hole/gravity

### DIFF
--- a/adlists.default
+++ b/adlists.default
@@ -46,7 +46,7 @@ https://raw.githubusercontent.com/quidsup/notrack/master/trackers.txt
 
 # Untested Lists:
 #https://raw.githubusercontent.com/reek/anti-adblock-killer/master/anti-adblock-killer-filters.txt
-#http://spam404bl.com/spam404scamlist.txt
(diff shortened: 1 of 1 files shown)


dschaper:
Thanks for the update.

fhemberger:
You're welcome. Wasn't sure if it should go into `development` or `master`, decided for the latter as it's just a blocklist.

dschaper:
If it's on Master, then it will need to wait until we are ready for another release, so it may be a bit. If you submit another PR off of Development, I can merge that in quickly and then it will be included in the next Master release.

fhemberger:
Created a new PR #536 against `development`.

PR #537, opened 2016-06-21
Missing permissions when writing to piholeIP

Mcat12:

- Fix `bash: line 239: /etc/pihole/piholeIP: Permission denied` in the install script, caused by the echo command not having permission to write to that directory.

@pi-hole/gravity


jacobsalmela:
Those were the exact changes I was thinking.  I'll test it out anyway.

dschaper:
:+1:

PR #540, opened 2016-06-24
Have `curl` auto-redirect

dschaper:
Fixes # Non-Listed

Add -L location redirection detection flag to curl

@pi-hole/gravity

This should solve the redirection issue if a list is moved.

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -152,7 +152,7 @@ function gravity_transport() {
 	fi
 
 	# Silently curl url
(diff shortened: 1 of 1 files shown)


PromoFaux:
This will stop lists failing if they move to https?

diginc:
If they move in general.  wget follows redirect by default for reference so `-L` is a pretty common flag for curl for reliability since you can't control external webservers.

PR #548, opened 2016-06-26
Debug online with sprunge.us

dan019:
Simplifies posting the log file online, instead of using pastebin. Automatically sends the output to a sprung.us link


Mcat12:
This sounds really useful! Just one thing... can you rebase your branch off of development? Thanks for pointing the PR at development though. ;)

PR #549, opened 2016-06-27
Debug with sprunge

dan019:
Replaced pihole_debug.log with a sprunge.us link for easier collaboration.

I have never contributed with git, however I did rebase from the development branch. Should I be seeing all those commits above might at the end of the list?

### DIFF
--- a/advanced/Scripts/piholeDebug.sh
+++ b/advanced/Scripts/piholeDebug.sh
@@ -330,8 +330,7 @@ function dumpPiHoleLog {
 
 # Anything to be done after capturing of pihole.log terminates
 function finalWork {
-	echo "::: Finshed debugging!" 
-	echo "::: Debug log can be found at : /var/log/pihole_debug.log"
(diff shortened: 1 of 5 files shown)


advanced/Scripts/piholeDebug.sh
```diff
@@ -330,8 +330,7 @@ function dumpPiHoleLog {
 
 # Anything to be done after capturing of pihole.log terminates
 function finalWork {
-	echo "::: Finshed debugging!" 
-	echo "::: Debug log can be found at : /var/log/pihole_debug.log"
+	echo "::: Debug log can be found at : $(cat /var/log/pihole_debug.log | curl --silent -F 'sprunge=<-' http://sprunge.us)"
```
diginc: What about some sort of check for upload success prior and printing this message.  If sprunge.us were down or DNS were broke then fallback to the old behavior would be preferable so the user is not left confused.

dschaper: Agreed, quite often we have the situation where the resolver isn't happy, and if the resolver isn't happy then uploads aren't going to work. Right now I get a `curl: (6) Could not resolve host: sprunge.us` response if the DNS is down.

PR #554, opened 2016-07-03
Easy log

dan019:
Automatically upload log file to sprunge.us
=================================

When the user is online, the log file will be uploaded to sprunge.us for easy sharing.
If their are connectivity issues and the user can not connect to sprunge, then they will
be presented the log in their log folder.


Mcat12:
Looks pretty good! One thing though. Can you target the development branch instead of master? Thanks. :+1:

PR #558, opened 2016-07-03
Upload log to sprunge.us

dan019:
Automatically upload log file to sprunge.us
=================================

When the user is online, the log file will be uploaded to sprunge.us for easy sharing.
If their are connectivity issues and the user can not connect to sprunge, then they will
be presented the log in their log folder.

Hopefully, I think I got it right this time. Thanks for pointing out I was pulling to the master branch. I didn't know I could change it on github's website.


jacobsalmela:
Looks good to me.  Works as expected.

PR #563, opened 2016-07-08
Development next release

PromoFaux:
Ready for merge.

@pi-hole/gravity


PromoFaux:
@pi-hole/gravity Is there anything that needs to be added to this/tested more before we release?

Mcat12:
I'd like to do some testing on it, but it seems to be pretty full-featured.

dschaper:
:+1:

dschaper:
The only thing I found that needed checking was Sprunge, and that fall back check seems to work okay.

PromoFaux:
Did you wanna try and get something in for Ubuntu as per #567 ?

dschaper:
Let's target the next release after this for Ubu checking. From the chat @brantje and I had last night on Gitter with another user troubleshooting, it may not be limited to Ubu, but straight Debian may have it as well.

dschaper:
http://pastebin.com/d43mmxmZ

dschaper:
Note to self, let's add a check like /etc/lsb-release or the equivalent to the debug. Find what OS is running.

Mcat12:
We should do that before we release

dschaper:
It would only work for lsb distributions, we could just add that for now and then flesh it out for next release, RH and BSD use other indicators.

dschaper:
I need to get some time to make the IPv6 debugging a little more robust.

bitterypaul:
what about adding functionality to enable or disable pi hole via the web interface. the web interface can send a command to rename the gravity.list to gravity.list.off as in [http://thetimmy.silvernight.org/pages/endisbutton/endisbutton.html](http://thetimmy.silvernight.org/pages/endisbutton/endisbutton.html) . this may be easy to implement since the web interface boasts a feature to whitelist and blacklist domains. thank you.

PromoFaux:
@jacobsalmela did actually start [work on the feature](https://github.com/pi-hole/pi-hole/tree/pause-resume-pihole) a little while back.. not sure if it's up to date! I know there is also a bounty (not by us) out there to get the feature added. It was at $40 last I checked....

jacobsalmela:
Yep.  It works fine in my testing, but I think the logic can be improved a bit:

https://github.com/pi-hole/pi-hole/blob/pause-resume-pihole/advanced/Scripts/pauseResume.sh

bitterypaul:
The code is clean and nice. And I have a doubt. If pi-hole is paused or resumed in the middle of an ongoing HTTP non-resumable download on any device using the network, will it be interrupted?
By the way,
http://sprunge.us/ is down

brantje:
@bitterypaul Once the download has started the domain name has already been resolved. 
So the download won't be interrupted, need to test this to be sure

- [ ] Test if an download is interrupted when pi-hole restarts

dschaper:
Things to consider, the resolver cache is not busted on the client machine. A pause button is only going to pause the resolver on the Pi-Hole, a Windows client is still going to have the stale data for at least 300 seconds. And as @brantje mentioned, HTTP sessions are untouched by the Pi-Hole in any way, anyways.

bitterypaul:
I already have a hotspot set up using my pi using hostapd and dnsmasq. I will start the http download and restart dnsmasq in the middle of that download. will check it in an hour.

bitterypaul:
and why there are exit codes 1 in the pihole main shell script?
I am not a Linux programmer but I think non-zero exit code indicates errors

dschaper:
http://tldp.org/LDP/abs/html/exitcodes.html We need some way to signal a pause function executed and so there is the return. I agree that 1 is probably not the best and we need to change that to a non-magic number though.

And I would be surprised if anything happened to the stream, we don't touch anything but port 53. (Port 80 of course for the built in Admin webserver, but that isn't changing any status or capabilities.)

PromoFaux:
>and why there are exit codes 1 in the pihole main shell script?

What's the line number of that @bitterypaul ?

bitterypaul:
Thanks for the info @dschaper . And I checked restarting the dnsmasq service and it doesn't interrupt the download in between.

dschaper:
This should be good to go. @bitterypaul When we release this, it's going to close out this thread, so please feel free to start a new issue thread and we can continue the discussion. I'm trying to track down the URI for the Chrome plug in and see if we need to do anything with that exit status to play nice with the extention.

bitterypaul:
guys, I have got an error in the dhcpcd.conf when using the automated pi hole installation
take a look at here
https://imgur.com/QQXsxao
the three colons are added to the file too

 
286 setDHCPCD() { 
287 	# Append these lines to dhcpcd.conf to enable a static IP 
288 	echo "::: interface $piholeInterface 
289 	static ip_address=$IPv4addr 
290 	static routers=$IPv4gw 
291 	static domain_name_servers=$IPv4gw" | $SUDO tee -a $dhcpcdFile >/dev/null 
292 }

bitterypaul:
- [x] Test if an download is interrupted when pi-hole restarts

the above problem might be easy to solve. just need to remove the three colons. firstly I ain't sure the problem exists only for me or it persists for all.

Whenever I reach the last page of any transaction in my netbanking I get everywhere var x = pi hole..

dschaper:
When you remove the comment line from dhcpcd.conf does that solve the `var x` issue on your netbanking?

bitterypaul:
sorry i wasn't clear. var x issue occured on my previous install after modifying all of the ip addresses since i use the pi as a hotspot and so auto configure wont work.

bitterypaul:
i found this issue of the three colons just now.

dschaper:
Is it causing any errors on your install? I'm just seeing `dhcpcd` ignoring that line and continuing to parse. Probably should be cleaned up for readability but I don't see it throwing any errors. Is there anything in the current list of commits that is causing errors?

bitterypaul:
it doesn't show any errors.
Another bug. the auto install script just blindly says the installation is complete even if it has encountered errors.
https://imgur.com/oqCYMy2

dschaper:
Okay, can you open a new issue for that for us so that we can track it? It's going to get lost when we close this out in a bit.

bitterypaul:
ok. i think we might have to remove the comment too because the line itself could specify the interface on which the pi settles itself on. helpful in situations when the user might have to manually look and modify the dhcpcd.conf to define multiple interfaces. i might be wrong.

dschaper:
Freezing for final release prep. Please open new issues for new items.

PR #572, opened 2016-07-15
Dan forgot this : `}`

PromoFaux:

@pi-hole/gravity


stonedbovines:
Line 75 shouldn't have the trailing slash after release
TMP=$(cat /etc/*release || echo "Failed to find release")
or it will always fail to find the file

dschaper:
Thanks for the catch, I'll update that with the IPv6 update in a bit...

PR #575, opened 2016-07-16
Remove trailing slash on os-release check

dschaper:
Fixes #CLOSED .

- Remove trailing slash that would cause an os-check to always fail out.

@pi-hole/gravity


brantje:
Looks good, no missing }

PR #600, opened 2016-07-30
Added a couple of options available by editing gravity.sh file

DelFireman:
Fixes #[no issue] .

@pi-hole/gravity

Add capability to specify add a number to the pi-hole host name to indicate the number of pi-hole boxes on the network

ie pi1.hole

Added capability to specify a different IP address in hosts file to allow for sending to a different server (a honey pot, etc.)

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -27,6 +27,16 @@ else
 	fi
 fi
 
+
+# If you are running more than one name server then
+# Uncomment next then change number.
+#piholeNSnumber=1
+
+# If you want to send your response to another server then
+# Uncomment next line to send query to a different ip address than pihole's ip 
+#piholeDumpAddress="0.0.0.0"
+
+
 piholeIPfile=/etc/pihole/piholeIP
 piholeIPv6file=/etc/pihole/.useIPv6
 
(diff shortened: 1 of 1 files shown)


PromoFaux:
Hey, thanks for the Pull Request, I like the idea, however I think it may be best to add the variable declarations into `pihole.conf` which [gravity will source if it exists](https://github.com/pi-hole/pi-hole/blob/master/gravity.sh#L72).

That way it's easier for less advanced users to update, they simply edit or create `/etc/pihole/pihole.conf`

DelFireman:
Thanks.  That is a good idea.  From what I can see that no one has started to use pihole.conf

Chris

PromoFaux:
Documentation is the key! So keep the checks in `gravity.sh` for those variables, and if they exist, use them, otherwise carry on as normal.

PromoFaux:
It's not a very often spoken about feature, but I know of a few people that use it.

PromoFaux:
@DelFireman Can you please merge lastest changes from development branch? Thanks.

Mcat12:
@DelFireman Pinging for status

CodingGoat:
`pihole.conf` is a good idea. There we can store hardcoded paths also ....

Fedora-Core:
@PromoFaux 
> Hey, thanks for the Pull Request, I like the idea, however I think it may be best to add the variable declarations into pihole.conf which gravity will source if it exists.
That way it's easier for less advanced users to update, they simply edit or create /etc/pihole/pihole.conf

@CodingGoat 
> pihole.conf is a good idea. There we can store hardcoded paths also ....


**warning**
pihole.conf is no longer "sourced" by gravity.sh

_( Yes, I know, because I was one of the two piholers that used it... )_

Grtz, Fedora-Core

PromoFaux:
>( Yes, I know, because I was one of the two piholers that used it... )

And perhaps @CodingGoat was the other ;)

The file has essentially been superseded by `setupVars.conf` (though, at the moment with less customisation, although I'm sure an advanced user like yourself has realised that you can just add in any custom paths into that file... 😄) in our on-going quest to make methods more solid. This in turn will probably eventually replaced by, or perhaps complemented by `/etc/default/pihole`. Who knows, anything can happen in this crazy, crazy world.

Mcat12:
Closing from lack of response. You can open a new PR with updated code (preferably using setupVars.conf) and we'll look at it again.

PR #624, opened 2016-08-19
Added `window.close()` to index.html

victmo:
- Adding `window.close()` will automatically close any pop-ups that might get loaded with a blocked ad.

@pi-hole/gravity Thought this might be a nice and easy addition.

### DIFF
--- a/advanced/index.html
+++ b/advanced/index.html
@@ -1,4 +1,7 @@
 <html>
+<head>
+<script>window.close();</script>
+</head>
 <body>
 </body>
 </html>


Mcat12:
Interesting idea. One thing to think about is if there's a popup that is not an ad. Though we could deal with those cases as it comes up.

victmo:
If the popup is not an add then the alternative (our) html file wouldn't load in the first place.

If the add is loaded inside a popup as an iframe, then the popup will stay open because the iframe's `window` object is pointing to the iframe's context, not the parent window 😉 

I've been using this hack for a while now and it works like a charm.

And BTW, thanks for all the effort you've put into making pi-hole. I've known about it for a while but I decided to give it a try now that I have a spare pi.

Mcat12:
Oh, just noticed that this is pointed at master. Can you make your changes based off of `development` and open a new PR? Thanks.

victmo:
Oops, you're right, sorry about that. Will do.

Mcat12:
No problem. :smiley:

PR #625, opened 2016-08-19
Forget all existing attempts at an update method... this is the one.

PromoFaux:
Introduces update method, still a few more things to add but that's mainly just spit and polish.

@pi-hole/gravity **REQUIRES TESTING** - I'm only one man, and I can only do so much, so a few sets of extra eyes on this would be great to make sure I haven't broken anything major!

- Exports options selected by user in setup whiptails to `/etc/pihole/setupVars.conf`. If this file is detected when running the install script, then an initial whiptail dialog is presented to the user to allow them to choose whether they want to update, skipping the setup whiptails, or if they want to continue (in case they want to change any of their install settings)

- Fixes `pihole -g -f`. Arguments were not being passed through.

- in the files I've updated, I've wrapped the variables in `${}`. Mainly because Idea gave me a squiggly line, and I don't like to see them.

TODO:
- [x] remove setupVars.conf if user cancels out of install
- [x] finalise update message in whiptail
- [x] less confusing switches on the `pihole` command to seperate the main update and web update
- [x] Move creation of `setupVars.conf` to the end to avoid having references to it all over the place
@pi-hole/gravity

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -920,7 +920,7 @@ View the web interface at http://pi.hole/admin or http://${IPv4addr%/*}/admin" $
 
 updateDialogs(){
 
-    if (whiptail --title "Existing Install detected" --yesno "We have detected you are installing over the top of an existing install.\n\n\nWould you like update?"  ${r} ${c}); then
+    if (whiptail --title "Existing Install detected" --yesno "We have detected you are installing over the top of an existing install.\n\nIf you would like to update, select yes.\n\n If you would like to run the complete setup again, select no."  ${r} ${c}); then
         echo "::: Updating existing install selected"
         useUpdateVars=true
     fi
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -924,7 +924,13 @@ updateDialogs(){
 ######## SCRIPT ############
 if [[ -f ${setupVars} ]];then
     . ${setupVars}
-    updateDialogs
+
+    if [ "$1" == "pihole" ]; then
+        useUpdateVars=true
+    else
+        updateDialogs
+    fi
+
 fi
 
 # Start the installer
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -929,10 +929,26 @@ View the web interface at http://pi.hole/admin or http://${IPv4addr%/*}/admin" $
 
 updateDialogs(){
 
-    if (whiptail --title "Existing Install detected" --yesno "We have detected you are installing over the top of an existing install.\n\nIf you would like to update, select yes.\n\n If you would like to run the complete setup again, select no."  ${r} ${c}); then
-        echo "::: Updating existing install selected"
-        useUpdateVars=true
-    fi
+  UpdateCmd=(whiptail --separate-output --radiolist "We have detected an existing install.\n\n    Selecting Update will retain settings from the existing install.\n\n    Selecting Install will allow you to enter new settings.\n\n(Highlight desired option, and press space to select!)" ${r} ${c} 2)
+  UpdateChoices=(Update "" on
+                 Install "" off)
+  UpdateChoice=$("${UpdateCmd[@]}" "${UpdateChoices[@]}" 2>&1 >/dev/tty)
+
+  if [[ $? = 0 ]];then
+		case ${UpdateChoice} in
+            Update)
+                echo "::: Updating existing install"
+                useUpdateVars=true
+                ;;
+            Install)
+                echo "::: Running complete install script"
+                useUpdateVars=false
+                ;;
+	    esac
+	else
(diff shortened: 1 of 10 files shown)


Mcat12:
Seems to work, although it should remove the setup vars file if the user cancels setup.

dschaper:
Wow, I think we need to retire that other PR / Roadmap. This I like!

pihole
```diff
@@ -52,6 +52,14 @@ function updateDashboardFunc {
     exit 1
 }
 
+function updateMainFunc {
+    echo "::: Fetching latest changes from Github..."
+    cd /etc/.pihole
+    ${SUDO} git pull
+    ${SUDO} /etc/.pihole/automated install/basic-install.sh
```
dschaper: Space in path fails.

PromoFaux: Ah, OK. I didn't actually test that before I committed Just needs to be wrapped in quotes?

dschaper: Escaped

automated install/basic-install.sh
```diff
@@ -938,11 +947,11 @@ fi
 installerDependencies
 
 if [[ ${useUpdateVars} == false ]]; then
-    ${SUDO} echo "" > /etc/pihole/setupVars.conf
-    ${SUDO} mkdir -p /etc/pihole/
     welcomeDialogs
     # Verify there is enough disk space for the install
     verifyFreeDiskSpace
+    ${SUDO} echo "" > ${setupVars}
+    ${SUDO} mkdir -p /etc/pihole/
```
Mcat12: Should it create the directory first?

automated install/basic-install.sh
```diff
@@ -784,52 +782,99 @@ setUser(){
...
+	fi
+}
+
+installPihole() {
+	# Install base files and web interface
+	checkForDependencies # done
+	stopServices
+	setUser
+	${SUDO} mkdir -p /etc/pihole/
+	if [ ! -d "/var/www/html" ]; then
```
diginc: Needed conditional?  mkdir -p is safe to run repeatedly.

diginc:
You could potentially get rid of all the `rm ${setupVars}` and `echo ... >> ${setupVar` lines (I counted 14 lines) if all the setup vars were `export`ed and then at the very end of a successful setup you ran `env | grep -E '(all|of|the|setup|vars|in|a|list|)' >> ${setupVars}`.  

This would give you a single central point of saving off user choices rather than scattering them about, and handles the 'aborted, remove partial setup vars' condition by only writing setupVars if the user makes it to the end which might make maintenance a tad easier.  `env | grep '^pihole' >> ${setupVars}` would work if all vars were prefixed the same too.

Side note, a separate style only PR would have been nice, but once I found the core logic changes I managed to ctrl+f to read the logic flow fairly easily.

PromoFaux:
This whole PR was sort of by accident, I was working on a completely different approach when suddenly this all clicked into place, so it may be a bit all over the place...

>`env | grep -E '(all|of|the|setup|vars|in|a|list|)' >> ${setupVars}`

Will that save the value _and_ the var name into the file? (e.g `ipv4=10.10.10.10`)

>Side note, a separate style only PR would have been nice, but once I found the core logic changes I managed to ctrl+f to read the logic flow fairly easily.

I realised that after I pushed... I'll bear it in mind for making future code reviews a bit easier!

diginc:
Yes, I ran a quick test just now and it has no problems with the IP.  Spaces/weird characters in the value of the var were a bit of a problem due to `env` not having quotes around values by default.  This SO detailed some techniques needed to make env work better for importing/exporting variables and I got it working with complex values: http://stackoverflow.com/questions/15235729/sourcing-env-output

I modified the first answer a little and added the 2nd anser's quoting solution and it works pretty well:

```
$ export ph_var1=yes
$ export ph_var2="I think"
$ export ph_var3="10.10.10.10"
$ env | grep '^ph_'
ph_var2=I think
ph_var3=10.10.10.10
ph_var1=yes
$ env | sed 's/=\(.*\)/="\1"/' | grep '^ph_' > saved
$ cat saved
ph_var2="I think"
ph_var3="10.10.10.10"
ph_var1="yes"
$ unset ph_var1 ph_var2 ph_var3
$ env | grep '^ph_'
$ while read line; do eval "export $line"; done < saved
$ env | grep '^ph_'
ph_var2=I think
ph_var3=10.10.10.10
ph_var1=yes
```

dschaper:
Syntactically, is it better form to have environment variables as all caps? Are they considered a form like a Pythonic Global?

PromoFaux:
@diginc I believe what I've done is combined both of our approaches to writing `setupVars.conf`

PromoFaux:
Approved

jacobsalmela:
Approved.  Excited about this one!

dschaper:
Approved

PR #626, opened 2016-08-20
Added window.close() to index.html

victmo:
Adding window.close() will automatically close any pop-ups that might get loaded with a blocked ad. Discussion ref #624


Mcat12:
Approved

brantje:
Approved

dschaper:
Approved

PR #636, opened 2016-08-26
Query command

PromoFaux:
Fixes [Internal]

- Adds a new function to the `pihole` command.

- Running `pihole -q <domainToQuery>` will tell you which lists contain a specified domain
@pi-hole/gravity

### DIFF
--- a/pihole
+++ b/pihole
@@ -62,6 +62,11 @@ function setupLCDFunction {
     exit 1
 }
 
+function queryFunc {
+    for list in /etc/pihole/list.*;do echo $list;grep '$@' $list;done
+    exit 1
+}
+
 function chronometerFunc {
     shift
     $SUDO /opt/pihole/chronometer.sh "$@"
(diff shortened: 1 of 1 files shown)


dschaper:
Thanks! Can we source the `pihole.conf` file and pull the directory of the lists from that, just in case there are one or two people that actually use that function? I've got it as a test case to try and use again...

dschaper:
Approved

PromoFaux:
Approved

[Approved with PullApprove](https://pullapprove.com/pi-hole/pi-hole/pull-request/636/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=pi-hole/pi-hole)

jacobsalmela:
Approved

pihole
```diff
@@ -62,6 +62,12 @@ function setupLCDFunction {
     exit 1
 }
 
+function queryFunc {
+    domain=$2
+    for list in /etc/pihole/list.*;do echo $list;grep ${domain} $list;done
+    exit 1
```
diginc: huh

dschaper: Just one of the debug commands we run a lot to find what list contains the domain in question. Is there something you see off in there?

diginc: exit 1 is traditionally a failure in bash, are you using it for something else?

dschaper: All the function exit with a return code of 1, they aren't exposed to the user, just internal. Not sure where the precedent was set for using 1 for all the functions, but we're just sticking with our convention for now.

dschaper: I'm guessing the convention was started by copying the return val for the `sudo` check, where it's properly used, and just continued on from there when the code was refactored. I don't remember offhand if we trap the return values or make any use of them directly. But yes, I do agree that it's not standard convention.

PR #679, opened 2016-09-12
Make `pihole -up` a bit more intelligent

PromoFaux:
Fixes #[internal] .

- Runs version check before updating files and makes the output a bit clearer

- points user at the changelog at the end

- tells user if nothing has been updated

@pi-hole/gravity


automated install/basic-install.sh
```diff
@@ -543,11 +543,10 @@ installScripts() {
 	${SUDO} cp /etc/.pihole/advanced/Scripts/blacklist.sh /opt/pihole/blacklist.sh
```
dschaper: This all probably should be redone with `install` instead, I think it would flow better.

dschaper:
:+1:

dschaper:
:+1:

PromoFaux:
:shipit:

jacobsalmela:
:shipit:

PR #703, opened 2016-09-19
Add additional instruction.

### LINKED ISSUE #686: Initial install script requires GUI environment
##### Expected Behaviour:
Installer runs correctly from an SSH session with no GUI, or if this isn't possible, gives a warning to that effect. Since a shell command is publicized as the install method, the need for the graphical environment is _not_ obvious. 

##### Actual Behaviour:
Installer runs, but user isn't given a chance to make any local configurations. No warning(s) are given. Especially noticeable if the local pi-hole interface needs to be something other than the default eth0. 

Install on a headless Pi using only ssh to connect. 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/37846768-initial-install-script-requires-gui-environment?utm_campaign=plugin&utm_content=tracker%2F3011939&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F3011939&utm_medium=issues&utm_source=github).
</bountysource-plugin>

PromoFaux:
Fixes #686.

-  Adds "(press space to select)" to choose interface dialog, as it may not be as obvious to users unfamiliar with keyboard-only dialogs. 

@pi-hole/gravity

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -188,7 +188,7 @@ chooseInterface() {
 
 	# Find out how many interfaces are available to choose from
 	interfaceCount=$(echo "$availableInterfaces" | wc -l)
-	chooseInterfaceCmd=(whiptail --separate-output --radiolist "Choose An Interface" ${r} ${c} ${interfaceCount})
+	chooseInterfaceCmd=(whiptail --separate-output --radiolist "Choose An Interface (press space to select)" ${r} ${c} ${interfaceCount})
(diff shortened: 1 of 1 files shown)


PromoFaux:
Approved

dschaper:
:+1:

jacobsalmela:
Approved

PR #704, opened 2016-09-20
lighttpd.conf templates change

### LINKED ISSUE #689: Backup old conf files before overwriting during upgrade
I have update pi-hole a few times now, and every time any modifications to some of the related conf files (E.g., lighttpd) get overwritten. Supporting modified configurations can be tricky, but the update software should backup then overwrite or the user be given an option to. 
##### Expected

To backup or integrate changes to conf files an update overwrites. 
##### Actual Behaviour:

Simply overwrites the altered conf file. 

Run' pihole -up 'after changing several of the related config files.  For example, I password protect the /admin dir in lighttpd.conf
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/37868582-backup-old-conf-files-before-overwriting-during-upgrade?utm_campaign=plugin&utm_content=tracker%2F3011939&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F3011939&utm_medium=issues&utm_source=github).
</bountysource-plugin>

stonedbovines:
Fixes #689 

- Provide ability to password protect the admin page by adding the required module to the lighttpd template files.

- Preserve any user-initiated alterations to lighttpd.conf by providing for the use of an external file of additional user requirements.

@pi-hole/gravity

### DIFF
--- a/advanced/lighttpd.conf.debian
+++ b/advanced/lighttpd.conf.debian
@@ -12,6 +12,7 @@
 server.modules = (
 	"mod_access",
 	"mod_accesslog",
+	"mod_auth",
 	"mod_expire",
 	"mod_compress",
 	"mod_redirect",
@@ -59,3 +60,6 @@ $HTTP["url"] =~ "^(?!/admin)/.*" {
     # rewrite only js requests
     url.rewrite = ("(.*).js" => "pihole/index.js")
 }
+
+# Add user chosen options held in external file
+include_shell "cat external.conf 2>/dev/null"
--- a/advanced/lighttpd.conf.fedora
+++ b/advanced/lighttpd.conf.fedora
@@ -11,6 +11,7 @@
 
 server.modules = (
 	"mod_access",
+	"mod_auth",
 	"mod_fastcgi",
 	"mod_accesslog",
 	"mod_expire",
@@ -75,3 +76,6 @@ $HTTP["url"] =~ "^(?!/admin)/.*" {
         # rewrite only js requests
         url.rewrite = ("(.*).js" => "pihole/index.js")
 }
+
+# Add user chosen options held in external file
+include_shell "cat external.conf 2>/dev/null"


PromoFaux:
Where does `external.conf` live?

stonedbovines:
In the same directory as the lighttpd.conf being used so (likely) /etc/lighttpd/

The danger of putting it anywhere else is that you can guarantee people will want to use different places. We _know_ there has to be a directory for lighttpd.conf, so that is where it will look.

advanced/lighttpd.conf.debian
```diff
@@ -12,6 +12,7 @@
 server.modules = (
 	"mod_access",
 	"mod_accesslog",
+	"mod_auth",
```
diginc: would `server.modules += ("mod_auth")` inside of an include work and be more modular?

advanced/lighttpd.conf.debian
```diff
@@ -59,3 +60,6 @@ $HTTP["url"] =~ "^(?!/admin)/.*" {
     # rewrite only js requests
     url.rewrite = ("(.*).js" => "pihole/index.js")
 }
+
+# Add user chosen options held in external file
+include_shell "cat external.conf 2>/dev/null"
```
diginc: why not a normal include?  what is the benefit of include_shell?

diginc:
what about making this bigger scope than just auth by making it support any lighttpd customizations a user needs via a conf.d folder?  I just saw an example using `include_shell "cat conf.d/*.conf"` to accomplish this behavior.

advanced/lighttpd.conf.debian
stonedbovines: `include external.conf` fails the syntax check if external.conf doesn't exist

stonedbovines:
@diginc 
I didn't use the folder approach from a personal desire to keep all the `conf` files in one common location, but it would certainly work. I would add the error trapping to it though. Either approach allows for 'freeform' customisations to lighttpd.

I suspect the general case is that the vast majority of users won't even be using password protection let alone multiple customisations. Using the "one additional file that we don't even care if it exists" approach leaves most systems untouched and minimises the imposition of a particular file layout. Using a specific, named file means everything else can live wherever the user wants them to. If they want to put them all in a folder, they can and _then_ `include_shell "cat myfolder/*.conf" 2>/dev/null` from external.conf.

advanced/lighttpd.conf.debian
stonedbovines: It would work, but then there wouldn't be an already written guide (Jacob's blog post) to setting up password protection that people can be pointed to.

PromoFaux:
Approved.

I think this idea has a lot of merit, and it will enable users to make customisations to their setup such as password protection without worrying about us barging in and overwriting it with every update.

Mcat12:
Approved

jacobsalmela:
Looks pretty cool.  Thanks!

jacobsalmela:
Approved

eNTi:
how is external.conf supposed to be used?

adding:
`server.port = 8081`

results in:
> Duplicate config variable in conditional 0 global: server.port
2017-10-09 21:56:03: (configfile.c.957) source: cat external.conf 2>/dev/null line: 2 pos: 1 parser failed somehow near here: (EOL)
2017-10-09 21:56:03: (configfile.c.957) source: /etc/lighttpd/lighttpd.conf line: 82 pos: 1 parser failed somehow near here: (EOL)

...and i guess that's about the most common usage? did anyone ever test that?

guygma:
Same issue here. Is this being addressed upstream?

Mcat12:
Changing the port is currently not supported as it breaks the block page and removes a major reason why we have the web server (besides the web interface).

YouveGotMeowxy:
I'm using docker, if I wanted to comment out this line in the lighttpd.conf file:

`"X-Frame-Options" => "DENY"`

but, in the external.conf instead, how would I do that? Do I just copy/paste this whole handler into the external.conf:

```
$HTTP["url"] =~ "^/admin/" {
    setenv.add-response-header = (
        "X-Pi-hole" => "The Pi-hole Web interface is working!",
        #"X-Frame-Options" => "DENY"
    )

    $HTTP["url"] =~ ".ttf$" {
        # Allow Block Page access to local fonts
        setenv.add-response-header = ( "Access-Control-Allow-Origin" => "*" )
    }
}
```

and then mount the external.conf in the docker compose file?

PR #722, opened 2016-09-26
Fix #718

### LINKED ISSUE #718: When a period is in a comment in a source, the comment is used as the URL, instead of intended URL [$5]
##### Expected Behaviour:
Comments in a source should be ignored

##### Actual Behaviour:
Periods in the comments are treated like URLs to be blocked, and actual URl is not blocked

Try putting these lines in your source:

**rcsmetrics.it #Corriere.it
esm1.net #Dealer.com**

It seems when the comment looks like a URL, PiHole is taking the comment as the domain, rather than the domain itself!

Because this is what I see in my actual gravity.list:
**192.168.1.186 #Corriere.it
192.168.1.186 #Dealer.com**

It is blocking the comment instead of the intended URL. **And it keeps the # in there too.**

<bountysource-plugin>
---
There is a **[$5 open bounty](https://www.bountysource.com/issues/37986538-when-a-period-is-in-a-comment-in-a-source-the-comment-is-used-as-the-url-instead-of-intended-url?utm_campaign=plugin&utm_content=tracker%2F3011939&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F3011939&utm_medium=issues&utm_source=github).
</bountysource-plugin>

PromoFaux:
Fixes #718 

Gravity can now handle host source files that for some reason break convention and add a comment to every single line. It's only one list out of all the default source lists... Just saying.

@pi-hole/gravity

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -289,7 +289,10 @@ function gravity_advanced() {
 	# This helps with that and makes it easier to read
 	# It also helps with debugging so each stage of the script can be researched more in depth
 	echo -n "::: Formatting list of domains to remove comments...."
-	awk '($1 !~ /^#/) { if (NF>1) {print $2} else {print $1}}' ${piholeDir}/${matterAndLight} | sed -nr -e 's/\.{2,}/./g' -e '/\./p' >  ${piholeDir}/${supernova}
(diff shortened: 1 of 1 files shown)


PromoFaux:
Approved

dschaper:
:+1:

jacobsalmela:
Approved

PR #726, opened 2016-09-27
DebugUpdate

dschaper:
Update debug script. Verbosity and IPv6

### DIFF
--- a/advanced/Scripts/piholeDebug.sh
+++ b/advanced/Scripts/piholeDebug.sh
@@ -37,7 +37,7 @@ echo "::: "
 
 ######## FIRST CHECK ########
 # Must be root to debug
-if [[ $EUID -eq 0 ]]; then
+if [[ "$EUID" -eq 0 ]]; then
 	echo "::: Script is executing as root user..."
 else
 	echo "::: Non-root user detected..."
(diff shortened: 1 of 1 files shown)


advanced/Scripts/piholeDebug.sh
```diff
@@ -330,8 +402,8 @@ function dumpPiHoleLog {
 			echo >> ${DEBUG_LOG}
 		done
 	else
-		echo "No pihole.log file found!" >> ${DEBUG_LOG}
-		printf ":::\tNo pihole.log file found!\n"
+		echo "No Pi-hole.log file found!" >> ${DEBUG_LOG}
```
PromoFaux: Should be `pihole.log`

advanced/Scripts/piholeDebug.sh
```diff
@@ -330,8 +402,8 @@ function dumpPiHoleLog {
...
+		printf ":::\tNo Pi-hole.log file found!\n"
```
PromoFaux: Should also be `pihole.log`

advanced/Scripts/piholeDebug.sh
```diff
@@ -151,6 +160,17 @@ function hostnameCheck {
...
 }
 
+function portCheck {
+    echo "############################################################" >> ${DEBUG_LOG}
+	echo "########           Open Port Information           #########" >> ${DEBUG_LOG}
+	echo "############################################################" >> ${DEBUG_LOG}
+
+    echo "::: Writing local server ports to logfile"
+
+    ${SUDO} netstat -tulpn >> ${DEBUG_LOG}
```
dschaper: All ports or just 80/53?

PromoFaux: Probably just 80/53. I can't see a use for any others?

diginc: does `netstat` require installing netstat on any of the officially supported OSs?

I know you don't support alpine but it lacked netstat by default in my docker container and I didn't want to add time to my tests so decline installing it and instead just `curl`'d 80 and ran `dig` for port 53

Dig test: https://github.com/diginc/docker-pi-hole/blob/master/test/test_start.py#L36
Curl test: https://github.com/diginc/docker-pi-hole/blob/master/test/test_start.py#L46

dschaper: We're looking for the owners of the processes at port 80 and 53 to see if they are who we expect, or if they are owned by another daemon. I guess I could change it around to run pidof or some other attempt with systemctl, but just getting a response back isn't what were looking for in this section.

There's a nslookup series of tests and I'll be adding in a couple of dig commands specific to dnsmasq to pull some metrics out of the process

dschaper: Do you have `lsof`?

diginc: correction, it wasn't that netstat wasn't installed it was that it was that alpine's busybox version which had quite different functionality and flags than the debian version.  LSOF seems to be same between alpine/debian.

dschaper:
Ready for check and merge. Further updates to be applied...

PromoFaux:
:shipit:

dschaper:
:+1:

dschaper:
Frozen for release.

Mcat12:
Approved

PR #738, opened 2016-10-01
Add check at the beginning of update script to check for existence of…

PromoFaux:
Fixes #[internal] .

- Adds an additional check for the local repos before updating. I came across it because I deleted `/var/www/html/admin` and tried `pihole -up`, which didn't like it!

@pi-hole/gravity

… local repos. If not, redirect the user to run install command again.

### DIFF
--- a/pihole
+++ b/pihole
@@ -49,6 +49,18 @@ function flushFunc {
 
 
 function updatePiholeFunc {
+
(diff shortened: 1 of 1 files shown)


PromoFaux:
Admin crimes

PR #742, opened 2016-10-02
gravityDB

dschaper:
See commit comments. Runs in about 20 seconds on a Zero, but eliminates some locked file contentions. You still can't delete a table while this is running, but that should not be an issue. Uses WAL and SHM files for transactions. Around 0.77 utilization, so the process is diskbound and not processor bound.

### DIFF
--- a/advanced/Scripts/gravity_parse.py
+++ b/advanced/Scripts/gravity_parse.py
@@ -43,31 +43,7 @@
 
 import sqlite3
 
-#-----------------------------------------------------------------------------
-# Functions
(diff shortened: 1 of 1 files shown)


dschaper:
WTH? Where did all the commits come from?!

PromoFaux:
Is 20 seconds faster or slower than it was running on a zero before?

PromoFaux:
Thanks! Tested this as working OK. Merging (No need for approval, it's not a protected branch)

PR #774, opened 2016-10-10
InstallFixes2

dschaper:
Continuation of Install Fixes - Branch 2

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -70,9 +70,9 @@ if [ -x "$(command -v apt-get)" ];then
 	#############################################
 	PKG_MANAGER="apt-get"
 	PKG_CACHE="/var/lib/apt/lists/"
-	UPDATE_PKG_CACHE="$PKG_MANAGER -qq update"
+	UPDATE_PKG_CACHE="$PKG_MANAGER update"
 	PKG_UPDATE="$PKG_MANAGER upgrade"
-	PKG_INSTALL="$PKG_MANAGER --yes --quiet install"
+	PKG_INSTALL="$PKG_MANAGER --yes --fix-missing install"
 	# grep -c will return 1 retVal on 0 matches, block this throwing the set -e with an OR TRUE
 	PKG_COUNT="$PKG_MANAGER -s -o Debug::NoLocking=true upgrade | grep -c ^Inst || true"
 	INSTALLER_DEPS=( apt-utils whiptail git dhcpcd5)
@@ -91,7 +91,7 @@ elif [ -x "$(command -v rpm)" ];then
 		PKG_MANAGER="yum"
 	fi
 	PKG_CACHE="/var/cache/$PKG_MANAGER"
-	UPDATE_PKG_CACHE="$PKG_MANAGER check-update -q"
+	UPDATE_PKG_CACHE="$PKG_MANAGER check-update"
 	PKG_UPDATE="$PKG_MANAGER update -y"
 	PKG_INSTALL="$PKG_MANAGER install -y"
 	PKG_COUNT="$PKG_MANAGER check-update | grep -v ^Last | grep -c ^[a-zA-Z0-9]"
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -102,7 +102,7 @@ elif [ -x "$(command -v rpm)" ];then
 	LIGHTTPD_GROUP="lighttpd"
 	LIGHTTPD_CFG="lighttpd.conf.fedora"
 	package_check_install() {
-		rpm -qa | grep ^"$1"- > /dev/null
+		rpm -qa | grep ^"$1"- > /dev/null || ${PKG_INSTALL} "$1"
 	}
 else
 	echo "OS distribution not supported"
@@ -922,6 +922,9 @@ notify_package_updates_available
 # Install packages used by this installation script
 install_dependent_packages INSTALLER_DEPS[@]
 
+# Install packages used by the Pi-hole
+install_dependent_packages PIHOLE_DEPS[@]
+
 if [[ ${useUpdateVars} == false ]]; then
     # Display welcome dialogs
     welcomeDialogs
@@ -943,8 +946,6 @@ if [[ ${useUpdateVars} == false ]]; then
     use4andor6
     # Decide what upstream DNS Servers to use
     setDNS
-    # Install packages used by the Pi-hole
-    install_dependent_packages PIHOLE_DEPS[@]
     # Install and log everything to a file
     installPihole | tee ${tmpLog}
 else
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -18,8 +18,6 @@
 
 set -e
 ######## VARIABLES #########
-
-
 tmpLog=/tmp/pihole-install.log
 instalLogLoc=/etc/pihole/install.log
 setupVars=/etc/pihole/setupVars.conf
@@ -31,8 +29,8 @@ piholeFilesDir="/etc/.pihole"
 
 useUpdateVars=false
 
-IPv4_address="Not Used"
-IPv6_address="Not Used"
+IPv4_address=""
+IPv6_address=""
 
 # Find the rows and columns
 rows=$(tput lines)
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -101,7 +101,7 @@ elif [ -x "$(command -v rpm)" ];then
 	LIGHTTPD_GROUP="lighttpd"
 	LIGHTTPD_CFG="lighttpd.conf.fedora"
 	package_check_install() {
-		rpm -qa | grep ^$1- > /dev/null
+		rpm -qa | grep ^"$1"- > /dev/null
 	}
 else
 	echo "OS distribution not supported"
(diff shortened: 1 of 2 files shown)


automated install/basic-install.sh
```diff
@@ -319,53 +320,57 @@ It is also possible to use a DHCP reservation, but if you are going to do that,
...
+	static ip_address=$IPv4_address
 	static routers=$IPv4gw
 	static domain_name_servers=$IPv4gw" | tee -a /etc/dhcpcd.conf >/dev/null
 }
 
 setStaticIPv4() {
+  local IFCFG_FILE
+  local IPADDR
+  local CIDR
 	if [[ -f /etc/dhcpcd.conf ]];then
```
dschaper: This will always exist. We are installing `dhcpcd5` as an installer dependency, and that file is automatically generated. So every Debian and derivative will have this file no matter what based on current install. Need ideas on how to handle this situation. Take `dhcpcd5` out of the install dependencies, but then how should we proceed?

dschaper:
```
pi@raspberrypi:/etc/pihole $ cat /etc/network/interfaces
# interfaces(5) file used by ifup(8) and ifdown(8)

# Please note that this file is written to be used with dhcpcd
# For static IP, consult /etc/dhcpcd.conf and 'man dhcpcd.conf'
```
This is the header for `/etc/network/interfaces` on Raspbian, it does not exist in other distributions in that file.

dschaper: And lastly, this is os-release
```
pi@raspberrypi:/etc/pihole $ cat /etc/os-release
PRETTY_NAME="Raspbian GNU/Linux 8 (jessie)"
NAME="Raspbian GNU/Linux"
VERSION_ID="8"
VERSION="8 (jessie)"
ID=raspbian
ID_LIKE=debian
```

automated install/basic-install.sh
```diff
@@ -319,53 +320,57 @@ It is also possible to use a DHCP reservation, but if you are going to do that,
...
+			echo "DEVICE=$piholeInterface"
+			echo "BOOTPROTO=none"
+			echo "ONBOOT=yes"
+			echo "IPADDR=$IPADDR"
+			echo "PREFIX=$CIDR"
+			echo "GATEWAY=$IPv4gw"
+			echo "DNS1=$piholeDNS1"
+			echo "DNS2=$piholeDNS2"
+			echo "USERCTL=no"
+			}>> "${IFCFG_FILE}"
```
diginc: HEREDOC this block? Just did this my self and almost used the style you implemented but switched to HEREDOC https://github.com/diginc/docker-pi-hole/pull/58/files#diff-f4647fb585ba39c152977803e97b4520R55

Like this?  `cat <<- EOF > $FILE` the dash after << gets rid of the whitespace but it only supports tabs...I personally was willing to trade echo and quotes for all of the tabs?

https://github.com/diginc/docker-pi-hole/pull/58/files#diff-f4647fb585ba39c152977803e97b4520R55

dschaper: Yeah, it was either heredoc or grouping and output. Didn't see the benefit of one over the other, actually.

automated install/basic-install.sh
```diff
@@ -968,7 +979,7 @@ fi
 
 echo ":::"
 echo "::: The install log is located at: /etc/pihole/install.log"
-echo "::: View the web interface at http://pi.hole/admin or http://${IPv4addr%/*}/admin"
+echo "::: View the web interface at http://pi.hole/admin or http://${IPv4_address%/*}/admin"
 }
 
 main "$@"
```
diginc: Could we get a dry-run that can prevent execution of this to allow just sourcing the functions and running them one by one?

This would allow unit tests to be written for the functions.  Though the test framework could also just comment this line out it may help with manual testing too, to not run everything at once.

PromoFaux:
Approved.

Tested on:
- [x] Ubuntu 16.05 DO Droplet
- [ ] Debian 8.6  DO Droplet (Some issues, but we know about them)
- [x] Rasbian Jessie on Pi2
- [x] Debian 8.6 on a laptop

dschaper:
Approved

Will add trap to next branch that will explain the Debian 8.6 DO issue, just need to instruct the user to try an install again, right now it drops to prompt without explanation.

dschaper:
And `dhcpcd5` will always make that sentinel file exist, so that's on the list of things to solve in the next round.

PromoFaux:
It's probably edge-case at this stage, mind. I wonder how many people are running it on a DO droplet anyway.

dschaper:
Very few, but it's easier to test there than in a VirtualBox VM :) Need to set them up with snapshots.

automated install/basic-install.sh
diginc: when dry run is set, do not run the main / anything. `if [[ -z "$DRYRUN" ]] ; main "$@ ; fi;` so it would always run main by default.

then if I wanted to write a test for one function I could do something like:
```
DRYRUN=TRUE . automated install/basic-install.sh;
useIPv6dialog
assert $IPv6addr == 1.2.3.4
```

dschaper: Warner had something he was working on with an undocumented flag that would pull the variables from a file and run non-interactive. But I understand what you mean now. Instead of triggering `main()` on load with the flag set it would allow for everything to load up and prep for main but then you could unit/nosetest the functions... I'll look at that and see how to get that for you.

PromoFaux:
have reset my pullApprove to Pending as this branch breaks Centos installs!

dschaper:
Can you do that? I'll just respect that it should be in hold status until all agree.

gravity.sh
```diff
@@ -254,10 +254,10 @@ function gravity_hostFormat() {
 		echo "::: Error: Unable to determine fully qualified domain name of host"
 	fi
 	# If there is a value in the $piholeIPv6, then IPv6 will be used, so the awk command modified to create a line for both protocols
-	if [[ -n ${piholeIPv6} ]];then
```
diginc: quotes around variable?

dschaper: I'm working mostly on `basic-install.sh` haven't gotten to much into `gravity.sh`

diginc: I think -n and -z require quotes to work properly:

```
[ -z $asdf ] && echo hit
hit
[ -z "$asdf" ] && echo hit
hit
[ -n $asdf ] && echo hit
hit
[ -n "$asdf" ] && echo hit
```

The 2nd to last one shouldn't have echo'd hit.

gravity.sh
```diff
@@ -254,10 +254,10 @@ function gravity_hostFormat() {
...
+	if [[ -n $"{IPv6_address}" ]];then
```
diginc: `"${IPv6_address}"`

PromoFaux: I'm ashamed to say it took me a full 30 seconds of staring at that trying to work out what you were pointing out. Good catch!

dschaper:
@bcambl Can you run this branch through your CentOS and see how it works for you? It's not `set -e` so any errors will ditch out the script and I'm trying to track down all the logic places where we rely on return values for flow control. Also, with DO, I'm finding that you can't use the automagically selected IPv4 address because it grabs the 10 net address and that makes Network Manager really mad, you have to type in the public IP address in the choices box for it to pass that section. (Along with the fact that SEL is disabled in DO so had to code around that and not throw a shoe when the firewall and SEL tried to function. SEL will install the policies, but they are not in effect. Firewall just bypasses any entries since it's non-op)... This is a major change in the install and we don't want to break CentOS.

bcambl:
```
[root@test automated install]# ./basic-install.sh 
:::
::: You are root.
::: Verifying free disk space...
:::
::: yum update has not been run today. Running now... done!
:::
::: Checking yum for upgraded packages.... done!
:::
::: There are 2 updates available for your system!
::: We recommend you run 'yum update -y' after installing Pi-Hole! 
:::
:::    Checking for iproute... installed!
:::    Checking for net-tools... installed!
:::    Checking for procps-ng... installed!
:::    Checking for newt... installed!
:::    Checking for git... installed!
:::    Checking for epel-release... installed!
:::    Checking for bind-utils... installed!
:::    Checking for bc... installed!
:::    Checking for dnsmasq... installed!
:::    Checking for lighttpd... installed!
:::    Checking for lighttpd-fastcgi... installed!
:::    Checking for php-common... installed!
:::    Checking for php-cli... installed!
:::    Checking for php... installed!
:::    Checking for curl... installed!
:::    Checking for unzip... installed!
:::    Checking for wget... installed!
:::    Checking for findutils... installed!
:::    Checking for cronie... installed!
:::    Checking for sudo... installed!
:::    Checking for nmap-ncat... installed!
:::
::: Checking for existing repository...
:::    Checking /etc/.pihole is a repo... OK!
:::     Updating repo in /etc/.pihole... done!
:::
::: Checking for existing repository...
:::    Checking /var/www/html/admin is a repo...:::    Cloning https://github.com/pi-hole/AdminLTE.git into /var/www/html/admin... done!
:::
::: Stopping dnsmasq service... done.
:::
::: Stopping lighttpd service... done.
::: Using interface: ens33
::: Warning: Unable to locate configuration file to set static IPv4 address!
[root@test automated install]#
```

dschaper:
You're going to hate me, but can you `bash -x basic-install.sh`?

dschaper:
There's something throwing the `set -e` and I don't have the trap set up yet to report to the user exactly what is going wrong...

bcambl:
Ignore the previous comment. Turns out my test server network configuration is an edge case. My `ifcfg-` file name does not match my device name. This caused `-f /etc/sysconfig/network-scripts/ifcfg-<device>` to fail. I touched the file and then everything went well.

I have a few things to add here if possible.. mind if I send a couple pull requests?

dschaper:
You are definitely more than welcome to PR what ever you'd like to contribute. We wouldn't have CentOS without your help!

bcambl:
Ok, I will test my changes within the next days and submit a couple PR's to this branch.

`updatesToInstall=$(eval "${PKG_COUNT}")` seems to be failing now.

```
+ echo -n '::: Checking dnf for upgraded packages....'
::: Checking dnf for upgraded packages....++ eval dnf check-update '|' grep -v '^Last' '|' grep -c '^[a-zA-Z0-9]'
+++ dnf check-update
+++ grep -v '^Last'
+++ grep -c '^[a-zA-Z0-9]'
+ updatesToInstall=0
[root@localhost .pihole]# 
```

This works for me:
```
if [[ $(eval "${PKG_COUNT}") -eq "0" ]]; then
```

But then you will lose your package count.  
Also, This is also the only `eval` in the installer. We should try and get rid of it anyways.

dschaper:
It may be the `grep -c`. If that evaluates to 0, then it will return an exit code of 1 and that will trip set -e. Check the package count for Debian and I think I left a note in there how to code around that one gotcha. It only happens when you have 0 packages to upgrade, any other value will pass without issue.

PromoFaux:
Approved, in it's current state, having re-tested on the following distros
- [x] Ubuntu 16.05 DO Droplet
- [x] Debian 8.6  DO Droplet (Some issues, but we know about them, requires a second run of the script)
- [x] Rasbian Jessie on Pi2
- [x] Debian 8.6 on a laptop
- [x] Centos 7.2 DO
- [x] Centos 6.8 DO

dschaper:
Okay, can we get this to merge to development and I'll continue to clean things up. I'd like to get some users to beta and doing that off devel may be easier.

dschaper:
:+1:

jacobsalmela:
Engage

dschaper:
Going to try an experiment and see if we can keep this branch open and still push to it, and keep it so we can merge to development again without losing the history and conversation.

iannao:
Hi Dschaper, i'm not able to make question in the other thread, can you please tell me how did you fix the issue? we are having the same problem in our app.

Thank you
Best Regards
Riccardo

dschaper:
Please open a new thread, do not re-use existing threads for other purposes. In 30 minutes I will delete both of these comments from this thread as we need it for tracking other issues. Thank you for your understanding.

PR #787, opened 2016-10-15
add install whiptail before update_dialogs

shadiakiki1986:

- [X] 1 (very unfamiliar)
- [ ] 2
- [ ] 3
- [ ] 4
- [ ] 5
- [ ] 6
- [ ] 7
- [ ] 8
- [ ] 9
- [ ] 10 (very familiar)

---
The update_dialogs function depends on whiptail.
whiptail is already in the INSTALLER_DEPS array,
but it gets installed at a later point than here,
so we need to install it for the sake of update_dialogs first.

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -892,6 +892,7 @@ if [[ -f ${setupVars} ]];then
   if [ "$1" == "pihole" ]; then
     useUpdateVars=true
   else
+    package_check_install "whiptail"
     update_dialogs
   fi
 fi


PromoFaux:
Hey, thanks for the submission.. I like where you're going with it, but I think for the sake of flow it would be best to move that entire if block after the call to check for packages.

I've created a new PR (fully credited to yourself :)) with this in mind. That said, it should be safe to assume that if it is an update, then `whiptail` is already installed....

PR #788, opened 2016-10-15
Move update_dialogs so that we are sure whiptail is installed! 

PromoFaux:

- [ ] 1 (very unfamiliar)
- [ ] 2
- [ ] 3
- [ ] 4
- [ ] 5
- [ ] 6
- [ ] 7
- [ ] 8
- [ ] 9
- [x] 10 (very familiar)

---
Ensures `whiptail` is installed before trying to use it.

Thanks to @shadiakiki1986 (see #787)

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -884,14 +884,6 @@ update_dialogs(){
 }
 
 main() {
-if [[ -f ${setupVars} ]];then
-  if [ "$1" == "pihole" ]; then
-    useUpdateVars=true
-  else
-    update_dialogs
-  fi
-fi
-
 # Start the installer
 # Verify there is enough disk space for the install
 if [[ $1 = "--i_do_not_follow_recommendations" ]]; then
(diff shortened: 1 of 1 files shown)


PromoFaux:
Approved

automated install/basic-install.sh
```diff
@@ -913,6 +905,15 @@ install_dependent_packages INSTALLER_DEPS[@]
 # Install packages used by the Pi-hole
 install_dependent_packages PIHOLE_DEPS[@]
```
Mcat12: The update dialog should show between the installer deps and the Pi-hole deps, because at that point we know that the installer can run, but we don't know if the user actually wants to install/update. If we add more deps, then the user would be forced to install those before they get the choice of not updating.

PromoFaux: Yeah, that's a fair comment. 

To be honest I think the pull is moot, but I submitted anyway just in case. As I said in #787, to even display the update dialogs in the first place, an existing install  needs to be detected. If an existing install is detected, then `whiptail` is already installed!

shadiakiki1986:
I was using the docker image when I noticed this. It had pihole but not
whiptail,  but I know now that the docker image (as discussed in PR 733) is
a special case.

dschaper:
@shadiakiki1986 If you'd like to fix the merge conflicts we'll take a look at getting this pulled in. Thanks

PromoFaux:
I mean, he could try, but it's my PR! (Based off of his idea!)

PromoFaux:
And even then, I'm not so sure about it any more

dschaper:
If it's a Docker edge/corner-case then might be something Bruce could help out with instead.

PromoFaux:
It's extreme edge, I think. As I mentioned above somewhere, it's only going to be a problem if `whiptail` is not installed on the system, and as this only affects the update dialogs, one would say it was safe to assume that `whiptail` is still installed from the original install!

dschaper:
Oh, then I'd say good to close it.

PR #811, opened 2016-10-19
remove direct pathanames in gravity blacklist and whitelist scripts and store them in setupVars

tuplink:

- [x ] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [ x] I have checked that [another pull request](https://github.com/pi-hole/pi-hole/pulls) for this purpose does not exist.
- [x ] I have considered, and confirmed that this submission will be valuable to others.
- [x ] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [x ] I give this submission freely, and claim no ownership to its content.

- [ ] 1 (very unfamiliar)
- [ ] 2
- [ ] 3
- [ ] 4
- [x ] 5
- [ ] 6
- [ ] 7
- [ ] 8
- [ ] 9
- [ ] 10 (very familiar)

---

_{replace this line with your pull request content}_

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -24,7 +24,7 @@ setupVars=/etc/pihole/setupVars.conf
 
 webInterfaceGitUrl="https://github.com/pi-hole/AdminLTE.git"
 webInterfaceDir="/var/www/html/admin"
-piholeGitUrl="https://github.com/tuplink/pi-hole.git"
+piholeGitUrl="https://github.com/pi-hole/pi-hole.git"
 piholeFilesDir="/etc/.pihole"
 
 useUpdateVars=false
@@ -58,7 +58,7 @@ else
   echo "::: Detecting the presence of the sudo utility for continuation of this install..."
 	if [ -x "$(command -v sudo)" ];then
 		echo "::: Utility sudo located."
-	  exec sudo bash "$0" "$@"
+	  exec curl -sSL https://install.pi-hole.net | sudo bash "$@"
     exit $?
 	else
 		echo "::: sudo is needed for the Web interface to run pihole commands.  Please run this script as root and it will be automatically installed."
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -24,7 +24,7 @@ setupVars=/etc/pihole/setupVars.conf
 
 webInterfaceGitUrl="https://github.com/pi-hole/AdminLTE.git"
 webInterfaceDir="/var/www/html/admin"
-piholeGitUrl="https://github.com/pi-hole/pi-hole.git"
+piholeGitUrl="https://github.com/tuplink/pi-hole.git"
 piholeFilesDir="/etc/.pihole"
 
 useUpdateVars=false
@@ -58,7 +58,7 @@ else
   echo "::: Detecting the presence of the sudo utility for continuation of this install..."
 	if [ -x "$(command -v sudo)" ];then
 		echo "::: Utility sudo located."
-	  exec curl -sSL https://install.pi-hole.net | sudo bash "$@"
+	  exec sudo bash "$0" "$@"
     exit $?
 	else
 		echo "::: sudo is needed for the Web interface to run pihole commands.  Please run this script as root and it will be automatically installed."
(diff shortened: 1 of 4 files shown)


tuplink:
i have not tested this with and IPv6 setup.

automated install/basic-install.sh
```diff
@@ -24,7 +24,7 @@ setupVars=/etc/pihole/setupVars.conf
 
 webInterfaceGitUrl="https://github.com/pi-hole/AdminLTE.git"
 webInterfaceDir="/var/www/html/admin"
-piholeGitUrl="https://github.com/pi-hole/pi-hole.git"
+piholeGitUrl="https://github.com/tuplink/pi-hole.git"
```
dschaper: Please check the repo line.

automated install/basic-install.sh
```diff
@@ -58,7 +58,7 @@ else
   echo "::: Detecting the presence of the sudo utility for continuation of this install..."
 	if [ -x "$(command -v sudo)" ];then
 		echo "::: Utility sudo located."
-	  exec curl -sSL https://install.pi-hole.net | sudo bash "$@"
+	  exec sudo bash "$0" "$@"
```
dschaper: This will fail on a `curl` refire.

automated install/basic-install.sh
tuplink: i fixed these how do i update?

dschaper: You already did, those were outdated commits, the latest version you have @ [2d7e316] takes care of that part.

diginc:
You should quote all variable assignments and `[ -n ]` or `[ -z  ]` conditionals.  Some cases you can get away without quotes but the cases that do require it makes it so you should just always quote bash variables as a standard convention.

See my previous comment showing how an empty variable can trigger -n if you don't quote it: https://github.com/pi-hole/pi-hole/pull/774#discussion_r82660085

tuplink:
@dschaper are you saying gravity.list shoould be in format        IPv4 IPv6 hostname?   it should return ipv6 host   if IPv6_address is set in setupVars.conf  if thats not the case it uses IPv4_address       but this should only apply to whitelist and blacklist

tuplink:
going to test as we speak

tuplink:
i need some direction.   should there be an entry for both ipv4 and ipv6? also the stable has a bug if you only select ipv6 durring installl you end up with 
 zz.zeroredirect1.com 
fe80::a00:27ff:fe54:df4e zz.zeroredirect1.com

but seeing that there is 2 entries im asuming we want 2 in the finised list file

PromoFaux:
If both IPv4 and IPv6 are selected on install, then yes there should be an entry for each. Else there should only be an entry for whichever one was selected. Nice catch on the bug... I guess not many people run on IPv6 only! If nothing else that should be easy to patch!

tuplink:
if we git this alll worked out i have it fixed going to test again

PromoFaux:
@tuplink [this is my proposed fix for current bug in stable](https://github.com/pi-hole/pi-hole/blob/60d17b97f521595562cc1dba6cd13104e4f9eea1/gravity.sh#L250-L272)

tuplink:
so i got it all working i do not have any VAR check but i can add that in way fewer lines

tuplink:
so as of right now all apears to be workig with all records reporting ipv4 / ipv6 or both and a check if none are set

PromoFaux:
The reason I took the approach I did, was so that the list remained in the order of :
```
ipv4 domain
ipv6 domain
ipv4 domain2
ipv6 domain2
```

rather than

```
ipv4 domain
ipv4 domain2
ipv6 domain
ipv6 domain2
```

tuplink:
i agree but you are probably never going to nano a file with 10000 lines you are going to use grep to check what is in there. and as the file will be recreated every time gravity runs you wouldnt change anything in it. i wonder if there is any performance gains to be had with dnsmasq eather way.  to each his own. but the ipv4 and ipv6 check should happen before any files are downloaded imo.

dschaper:
Closing as outdated.

PR #813, opened 2016-10-19
Create missing directory for lighttpd

das7pad:
The directory mentioned in the lighttpd config for uploads is not created in the current installation process.

`server.upload-dirs          = ( "/var/cache/lighttpd/uploads" )`


dschaper:
I'm not sure if we ever upload anything to the `lighttpd` process? May be a security feature, or we may need to take that directive out of the configuration. The users shouldn't be uploading to the admin interface.

PromoFaux:
Are there any errors that complain the directory does not exist?

brontide:
FYI, If you do not enable this for Fedora 24 ( and possibly other platforms ) lighttpd/fastcgi will be limited to 65k responses which kills the query log pretty quick.

before
```
1477620749|10.195.42.144|GET /admin/api.php?getAllQueries HTTP/1.1|200|65528
```

after creating directory
```
1477621277|10.195.42.144|GET /admin/api.php?getAllQueries&_=1477621276496 HTTP/1.1|200|762698
```

The clients get a ajax error because the json is truncated.

PromoFaux:
That's really odd behaviour!

Approved

I can't see adding the directory doing any harm.

dschaper:
65K? Like 65535?

brontide:
Yeah, you can see in the logs it's a return code of 200 but only 65k is returned ( well 8 bytes short 65528 ) so the query is fully executed or, I hope, there would be an error.  I believe fcgi is attempting to mmap a temp file in that directory to hold the buffer.

dschaper:
Because that's right around the max for a 16bit buffer, I'm wondering if there's any relation, if the upload directory is really being used as a cache directory and for maxing out the buffer.

dschaper:
Or if we have a hidden bug somewhere in our implementation...

das7pad:
@PromoFaux Yes, at the startup of lighttpd I can find the following in the log:
Oct 25 04:51:18 ... lighttpd[33203]: 2016-10-25 04:51:18: (configfile.c.1390) server.upload-dirs doesn't exist: /var/cache/lighttpd/uploads

brontide:
On fedora24 it throws this into the log when the response is over 64k.

```
2016-10-28 17:02:41: (chunk.c.544) opening temp-file failed: No such file or directory 
```

dschaper:
I swear it's got to be using that directory as a cache for a buffer or something. But in any case if it solves the problem and doesn't introduce anything...

dschaper:
Approved

PR #814, opened 2016-10-19
Various tweaks and fixes: Part 2 (Plus Pause/Resume button functionality!)

PromoFaux:

- [x] 10 (very familiar)

---
Various tweaks and fixes. 
Main objectives:
- [x] Fix bug in gravity that would mess up gravity.list if IPv4_address is not set


tuplink:
i have played with this for a bit and it apears to work!!!! one step in the right direction.    i did notice that if you unwhitelist a domain it isnt added back to the gravity.list

also see #816

Mcat12:
Was the domain that you took off the whitelist included in a list? If it wasn't, then it shouldn't be added back to gravity.list

tuplink:
it was

tuplink:
@ 101 in gravity.sh should be able to add 
pihole -b $dom

gravity.sh
```diff
@@ -92,20 +97,6 @@ gravity_collapse() {
 		done < ${adListDefault}
 		echo " done!"
 	fi
-
-	# Create the pihole resource directory if it doesn't exist.  Future files will be stored here
-	if [[ -d ${piholeDir} ]];then
-        # Temporary hack to allow non-root access to pihole directory
-        # Will update later, needed for existing installs, new installs should
-        # create this directory as non-root
-        chmod 777 ${piholeDir}
```
dschaper: @tuplink Are you referrring to this line 101?

tuplink: promoTweaksTheSecond branch

dschaper: Okay, if you click on "Files Changed" at the top of the screen, it should bring up a list of file differences, from there click on the line number of the file you are commenting on and leave a line note so we know exactly which line you are referring to. Thanks!

gravity.sh
```diff
@@ -92,20 +97,6 @@ gravity_collapse() {
...
-        echo ":::"
-        echo "::: Existing pihole directory found"
-	else
-        echo "::: Creating pihole directory..."
-        mkdir ${piholeDir}
-        chmod 777 ${piholeDir}
-	fi
 }
```
dschaper: Like this...

dschaper: But in the file that you are referring to. This is gravity.sh for an example.

advanced/Scripts/blacklist.sh
```diff
@@ -124,68 +93,16 @@ RemoveDomain(){
...
   else
-    #Domain is in the blacklist file, add to a temporary array
+    #Domain is in the blacklist file,remove it
     if ${verbose}; then
     echo "::: Un-blacklisting $dom..."
     fi
-    domToRemoveList=("${domToRemoveList[@]}" $1)
-    modifyHost=true
+   echo "$1" | sed 's/\./\\./g' | xargs -I {} perl -i -ne'print unless /'{}'(?!.)/;' ${blacklist}
   fi
```
tuplink: ${blacklistScript} $dom

advanced/Scripts/blacklist.sh
```diff
@@ -37,40 +36,11 @@ adList=${piholeDir}/gravity.list
 blacklist=${piholeDir}/blacklist.txt
```
dschaper: Okay, PR your changes and we'll take a look!

gravity.sh
dschaper: It's how we all learn, so it's not a problem, thanks for the code modifications and the Pull Requests.

advanced/Scripts/whitelist.sh
```diff
@@ -128,81 +98,13 @@ RemoveDomain(){
...
   	fi
   else
-    #Domain is in the whitelist file, add to a temporary array and remove from whitelist file
-    #if $verbose; then
-    #echo "::: Un-whitelisting $dom..."
-    #fi
-    domToRemoveList=("${domToRemoveList[@]}" $1)
-    modifyHost=true
+    echo "$1" | sed 's/\./\\./g' | xargs -I {} perl -i -ne'print unless /'{}'(?!.)/;' ${whitelist}
   fi
```
dschaper: But if you could write that up and submit it as a PR instead of as line item comments. We can't pull in line comments, only commits and PR's. So feel free to add that to either a new branch and PR or as another push to your existing branch that you have PR'd. (If you have a PR open, I don't remember offhand.)

gravity.sh
```diff
@@ -43,7 +45,7 @@ else
 fi
 
 #Remove the /* from the end of the IPv4addr.
-IPv4addr=${IPv4_address%/*}
+IPv4_address=${IPv4_address%/*}
```
tuplink: dose dhcpd call this file for this var tho?  all of the scripts in this branch that i have looked at that use this var remove the cidr so if it was stored in setupVars.conf without a cidr less compute used all together

dschaper: The value should be stored in `setupVars` with the CIDR notation, as that file will be sourced for headless installation and we need the subnet information. I don't believe it's used in CIDR form within `gravity.sh`, just plain as the source address for the hosts files. If you want to break out `setupVars` and store the CIDR value as a separate variable, then there's that option, but we do need to have it stored for access.

dschaper: We have a lot of users that do not use the natural masks for their addressing schemes.

tuplink: asuming yoasuming you are refering to https://github.com/pi-hole/pi-hole/blob/promoTweaksTheSecond/automated%20install/basic-install.sh#L338

tuplink: on install: should not break installer script as the info for setDHCPCD() is gatherd from find_IPv4_information() 
on update/reconfigure:even if i could find where functions were called from these scripts. ${IPv4_address} (from setupVars.conf) would be overwriten by find_IPv4_information()

dschaper: Future plans and things that are in work. Would prefer not having to undo all your hard work. I'll let the other members of the team comment as I may be misunderstanding. (Premature Optimization and all that...)

dschaper: https://github.com/pi-hole/pi-hole/blob/promoTweaksTheSecond/automated%20install/basic-install.sh#L333

dschaper: [Linux-File-System-Hierarchy-blackMORE-Ops.pdf](https://github.com/pi-hole/pi-hole/files/546177/Linux-File-System-Hierarchy-blackMORE-Ops.pdf)

This is the guide that we try to go by with the location of files, just something for the insomnia reading shelf. :)

PromoFaux: I'd say we might as well leave the CIDR in there, it's not much of an overhead to remove it in when we need it!

PromoFaux:
Approved

tuplink:
I have a feeling https://github.com/pi-hole/pi-hole/commit/82e78fb6518b47e81f22ee8022677288b70513f5 will add extra unexpected output to the web part.   Maybe on the exec in header add > /dev/null

PromoFaux:
Doesn't appear to....

image

tuplink:
I see that now on php.net thanks for verification

PromoFaux:
Approved

(I'll stop updating it now unless anyone finds anything wrong with it!!)

PromoFaux:
APPROVED

(Seriously, somebody needs to code review this before I make MORE changes)

advanced/Scripts/piholeDebug.sh
```diff
@@ -271,12 +262,11 @@ checkProcesses() {
 	echo ":::"
 	echo "::: Logging status of lighttpd and dnsmasq..."
 	PROCESSES=( lighttpd dnsmasq )
-	for i in "${PROCESSES[@]}"
-	do
+	for i in "${PROCESSES[@]}"; do
```
dschaper: Note: This is a systemD only function. Dan you fool, you need to support `service` calls as well!

advanced/Scripts/setupLCD.sh
```diff
@@ -15,28 +15,28 @@
 # Borrowed from adafruit-pitft-helper < borrowed from raspi-config
```
dschaper: Totally need to re-do this script, Adafruit is bad on this, we can do better. @dschaper

advanced/Scripts/piholeDebug.sh
dschaper: @dschaper

dschaper:
:+1:

PromoFaux:
Approved.

Apparently PullApprove doesn't like all caps!

PromoFaux:
Ping! (Before this gets stale)

dschaper:
:+1:

brantje:
Approved

Fedora-Core:
Hmm, I can't wait to see the issues and complaints about this Disable/Enable thingy...

Found in branch Development:

1) script __basic-install.sh__
line 872 and line 1010
2) script __gravity.sh__
line 247 and line 253 and line 259

Found in pi-hole/AdminLTE:

1) script __data.php__
line 203

The bad news is, that Disable will work. But Enable maybe not.

_The good news is, that there is a solution._

Greetings,
Fedora-Core

dschaper:
That's why we have a development branch, so, you know, people can develop? PR's are always welcome...

PromoFaux:
Actually, @Fedora-Core I'm not sure what you're pointing out... in my testing of enable/disable.. it works both ways? Care to clarify what it is you're pointing out? :)

Fedora-Core:
Dear, you have tried - or you have tested ?

When the additional hostfile is disabled, the domain pi.hole is disabled also.
When your try-out machine's dns becomes stale ( dns cache expires ) the dashboard web has to resolve the domain pi.hole anew. But DNSMasq will not find pi.hole (its file is disabled).

When you __try__ something, you do your utmost best to prove that something works according to the specifications.

When you __test__ something, you do your utmost best to prove that something does __not__ work according to the specifications.

That's why software developers can't test their own code:

In the morning they must produce good code, 

and in the afternoon, they must destroy their code.

Enough to make them _schizophrenic._

Greetings,
Fedora-Core

tuplink:
solution to the problem> put pi.hole in pi-holes /etc/hosts

dschaper:
Well, applause for you, 1800+ stars and almost 250 forks, and you were the first to find it! (Isn't Open Source great, you can actually see the source, and test it yourself!)

PromoFaux:
> pi-holes /etc/hosts

I think we'd like to avoid that where possible. directly editing folk's /etc/hosts file is probably a bad idea. That said, I have a solution in mind.

PromoFaux:
@Fedora-Core Thank you for the clarification. Though why you need to be so cryptic is beyond me! Keep fighting the good fight.

Fedora-Core:
The solution is: review the rules of DNSmasq's configuration.

1) DNSmasq has two kinds of configuration keywords:
a) unique configuration keywords
     example:
```
log-facility=/var/log/pihole.log
```
b) repetitive configuration keywords
     example:
```
server=8.8.8.8
server=8.8.4.4
```
2) DNSMasq - as we have learned thru issues - is sometimes used for DHCP server, and even PXEboot server

Proposal:

/etc/dnsmasq/config.d/
members:
__01-pihole.conf__
    contains only the unique keywords, that the installer/updater knows
__11-pihole.conf__
   contains the repetitive keywords, that the installer/updater knows
   ° the upstream dnsservers is an example
   ° the additional hostfile with the pi.hole domain is another example,
   ° it does __not__ belong to gravity.sh
__21-pihole.conf__
   contains the repetitive keywords, that the installer/updater knows
   ° the additional hostfile list identifier for the gravity list is it !
   ° This is the one that may be Disabled/Enabled

And with this setup, there is space for DNSMasq configurations for DHCP server en PXEboot server 

As long as we document therationale of the setp for the advanced users.

Greetings,
Fedora-Core

PR #817, opened 2016-10-20
Blacklist the unwhitelisted

tuplink:

- [ x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [x ] I have checked that [another pull request](https://github.com/pi-hole/pi-hole/pulls) for this purpose does not exist.
- [x ] I have considered, and confirmed that this submission will be valuable to others.
- [ x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [ x] I give this submission freely, and claim no ownership to its content.

- [ ] 1 (very unfamiliar)
- [ ] 2
- [ ] 3
- [ ] 4
- [x ] 5
- [ ] 6
- [ ] 7
- [ ] 8
- [ ] 9
- [ ] 10 (very familiar)

---
when you delete a whitelist we can asume you want it blocked as whitelisting a site allows it to work. so when whitelist delete is called go ahead and blacklist it

### DIFF
--- a/advanced/Scripts/whitelist.sh
+++ b/advanced/Scripts/whitelist.sh
@@ -33,7 +33,6 @@ fi
 #globals
 basename=pihole
 piholeDir=/etc/${basename}
-blacklistScript=/opt/pihole/blacklist.sh
 adList=${piholeDir}/gravity.list
 whitelist=${piholeDir}/whitelist.txt
 reload=true
@@ -100,8 +99,7 @@ RemoveDomain(){
   	fi
   else
     echo "$1" | sed 's/\./\\./g' | xargs -I {} perl -i -ne'print unless /'{}'(?!.)/;' ${whitelist}
-    #Blacklist unwhitelisted
-    ${blacklistScript} $1
+    pihole -g
   fi
 }
 
--- a/advanced/Scripts/whitelist.sh
+++ b/advanced/Scripts/whitelist.sh
@@ -33,6 +33,7 @@ fi
 #globals
 basename=pihole
 piholeDir=/etc/${basename}
+blacklistScript=/opt/pihole/blacklist.sh
 adList=${piholeDir}/gravity.list
 whitelist=${piholeDir}/whitelist.txt
 reload=true
--- a/advanced/Scripts/whitelist.sh
+++ b/advanced/Scripts/whitelist.sh
@@ -99,6 +99,7 @@ RemoveDomain(){
   	fi
   else
     echo "$1" | sed 's/\./\\./g' | xargs -I {} perl -i -ne'print unless /'{}'(?!.)/;' ${whitelist}
+    #reload gravity to pickup any entries this may have removed
     pihole -g
   fi
 }
--- a/advanced/Scripts/whitelist.sh
+++ b/advanced/Scripts/whitelist.sh
@@ -99,6 +99,8 @@ RemoveDomain(){
   	fi
   else
     echo "$1" | sed 's/\./\\./g' | xargs -I {} perl -i -ne'print unless /'{}'(?!.)/;' ${whitelist}
+    #Blacklist unwhitelisted
+    ${blacklistScript} $1
   fi
 }
 


dschaper:
Thanks, can you update the title of the PR so that it's a bit more descriptive of what it contains. And then detail what your changes are and what you want to accomplish with the changes in the template where is asks you `{replace this line with your pull request content}`

dschaper:
Thanks!

PromoFaux:
Trying to decide if this is necessary...

The way I see it...

Whitelisting a domain stops it from being included in `gravity.list` in the case that it was in one of the source lists, but was not necessarily in the user's custom blacklist to start with. Does un-whitelisting a domain automatically mean you want it blacklisted? 

The same could be said for blacklisting a domain, and then removing it from the blacklist. If you remove it from the blacklist, do you necessarily want to add it to your whitelist? If so, maybe we need another rethink of the white/blacklist scripts.

Perhaps we could take out the remove function of each script entirely, and just automatically remove anything added to the whitelist from the blacklist, and anything added to the whitelist from the blacklist.

Thoughts?

dschaper:
If it was on White, and was removed, then `pihole -g` would add it back if it existed on an adlist. No need to add it to black.

If it was black, and you remove it, there is some reason you removed it, and I would assume that would be because you want to see the traffic, so it should be whitelisted if not on an adlist. (There is the corner case of someone just cleaning an mistaken addition to the blacklist, but not sure how often that occurs.)

tuplink:
so maybe just write it back into list.preEventHorizon instead of oficialy blacklisting it

PromoFaux:
What if it was not in `list.preEventHorizon` to start with? :)

tuplink:
its an over analized scnario, i tend to do that.      then why did you whitelist it?

dschaper:
You don't have to close it out, we're thinking through the possible cases and options. I wasn't ready to pass over the PR yet.

PromoFaux:
Edge-case I guess, but some people might just add domains to whitelist as a force of habit, just in case...

e.g www.msftncsi.com doesn't actually appear to have been on any of the default lists for a while! But people still add it.

There's merit in this idea... I just think it needs a bit of a rethink! Like I mentioned above:

Remove delete mode from both scripts.

Whitelist add removes from blacklist.txt if it exists.
blacklist add removes from whitelist.txt if it exists.

tuplink:
so maybe instead if having it add it to the blacklist we just need to reparse the lists

tuplink:
if i only run gravity once a day and i unwhitelist a domain i would expect to see it work imedietly. if not im scratching my head trying to see why unwhitelisting the domain that i think is my ad source is still working

dschaper:
Removing an item from whitelist should immediately fire the pihole update. Same with any changes to blacklist, fire the script right away.

tuplink:
with this PR it modifies list.preEventHorizon instead of parsing the lists or downloading them for that matter  speeds things up      if im correct

dschaper:
On the `pihole -g` calls, does your PR allow for users to change the location of the scripts, or do they still install to the LFS location in `/opt`? It probably would be better to have the full path to `pihole` instead of just the bare call without the full path. Might not be a big difference, but for completion sake...

tuplink:
all of the rest of the calls are to pihole   that is in /usr/local/bin/ hopefully that is in PATH on all systems

tuplink:
so in what i want to do i can just edit pihole    change the pathnames and modifie the GLOBAL VAR PATH to include pihole i think

dschaper:
We'll need to update the rest of the calls. If we're going to allow users to place the files where they wish and then use variables to track them, then we need to make sure that we have full paths to what ever we can. It's just defensive programming. 

If you'd like to have the path to the scripts as part of a variable, that's okay too. We follow what's called the LFHS and it dictates what files are supposed to go in which directories. Some `/usr/local/bin|sbin` some `/var/opt`, etc... The section with all the `install` calls puts the files where the LFHS requests we install them.

PromoFaux:
Just a note, the last two commits are unnecessary! Reload is handled by the `Reload` function which fires every time unless `-nr` is passed to the whitelist/blacklist script.

dschaper:
Does whitelist.sh/blacklist.sh call the reload themselves? I'm wondering if for the sake of atomiticity (yeah, I know that's not a word but can't think of a better one,) we should just have white and black only modify the lists, and then have a controller call the reload. One function, one job kind of thing?

PromoFaux:
OK, in the mods in my pull, whitelist and blacklists call gravity (with the new -`sd`, or `--skip-downloads`) option to reload the lists and rebuild `gravity.list` from the cached `list.preEventHorizon`

PromoFaux:
Other than that they only directly edit the whitelist.txt and blacklist.txt files

dschaper:
Makes perfect sense, I just don't know exactly how far down the rabbit hole we should go with the refactor code. Basic_install is getting picked apart so that each function only does one thing, and takes/returns arguments. I'm bouncing back and forth... I'm just not a fan of a function calling a function calling a function... Anal retentive I am...

tuplink:
so on that note    pihole uninstalll dont work    /usr/local/bin/pihole: line 89: /opt/pihole/uninstall.sh: No such file or directory     looks like the install script misses it

tuplink:
but im also not on master 👍 atm

dschaper:
Sorry, for derailing that there.... Uninstall hasn't been worked on in a bit, and it doesn't know about all of the changes we've made in the last week or so. We need to work on cleaning up after ourselves next...

tuplink:
dont bother my    power off instance    link clone instance   power up ssh and BAM about a min fresh install

dschaper:
Lol, sounds like me, 5 SD cards lined up to have a fresh OS ready to go, and then all the VPS's we use for testing... Glad the new installer is working well, it's got a bit more work to do. I do like where you are going with your code commits though, should be easier to give users even more customization options.

tuplink:
are all of the webcalls to blacklist and whitelist directed through pihole?   if so would therre be any reason not to combine all 3 gravity scripts?

dschaper:
Factoring, in general you just want each script or each function within the script to do only one job. It's the Unix way "Do one thing and do it well." It makes it way easier to swap out parts when you want or need to make changes. Think of it like a car, easier to change the battery when it's not buried deep under and behind the engine.

PR #819, opened 2016-10-20
Remove problem list, sends in Unicode - Not adware

### LINKED ISSUE #816: binary files in gravity.list
some times gravity.list will end up being truncated. with a binary file matches error.    i belive this to be due to https://ransomwaretracker.abuse.ch/downloads/RW_DOMBL.txt.    taking the output file from that list  tells me its gzipped      i belive the solution to this is adding --compressed to curl   not sure if it hurts to do this on all of the lists but adding    
            "ransomwaretracker.abuse.ch")
                cmd_ext="--compressed"
            ;;
in gravity_spinup semes to have cured this but only time will tell as this is a random error

dschaper:

- [x] 10 (very familiar)

---
Remove problematic adlist from sources, does not block ads, is a Command and Control blocker, small number of domains and presents list in Unicode on occasion, breaking system. Cull.

Fixes #816

### DIFF
--- a/adlists.default
+++ b/adlists.default
@@ -42,9 +42,6 @@ https://raw.githubusercontent.com/quidsup/notrack/master/trackers.txt
 # Block the BBC News website Breaking News banner
 #https://raw.githubusercontent.com/BreakingTheNews/BreakingTheNews.github.io/master/hosts
 
-# List of known C&C malware servers (see https://github.com/pi-hole/pi-hole/issues/528)
-https://ransomwaretracker.abuse.ch/downloads/RW_DOMBL.txt
-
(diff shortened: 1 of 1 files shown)


Mcat12:
Approved

PromoFaux:
Approved

Note for future users wondering why their list has suddendly dropped by 2,258 (ish) domains. This is why.

PromoFaux:
Wow, that list has really grown, When it was first added it was only an additional 174 domains.

dschaper:
They're all random subdomains, not many domains...

dschaper:
:+1:

PR #828, opened 2016-10-21
Improved iptables error handling

brontide:
1. We test to make sure iptables -L INPUT does not fail, if it does
   we skip the process
2. We only add the iptables rules if they do not already exist which
   should prevent duplicate rules when update is run

For each firewall type I have existed the function.

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -775,14 +775,19 @@ configureFirewall() {
 	if [ -x "$(command -v firewall-cmd)" ]; then
 		firewall-cmd --state &> /dev/null && ( echo "::: Configuring firewalld for httpd and dnsmasq.." && firewall-cmd --permanent --add-port=80/tcp && firewall-cmd --permanent --add-port=53/tcp \
 		&& firewall-cmd --permanent --add-port=53/udp && firewall-cmd --reload) || echo "::: FirewallD not enabled"
-	elif [ -x "$(command -v iptables)" ]; then
+        return
+    fi
+    if [ -x "$(command -v iptables)" ]; then
+        iptables -L INPUT &> /dev/null && (
 		echo "::: Configuring iptables for httpd and dnsmasq.."
-		iptables -A INPUT -p tcp -m tcp --dport 80 -j ACCEPT
-		iptables -A INPUT -p tcp -m tcp --dport 53 -j ACCEPT
-		iptables -A INPUT -p udp -m udp --dport 53 -j ACCEPT
-	else
-		echo "::: No firewall detected.. skipping firewall configuration."
-	fi
+        # Insert these 3 rules at the front of the INPUT chain only if they do not already exist
+		iptables -C INPUT -p tcp -m tcp --dport 80 -j ACCEPT &> /dev/null || iptables -I INPUT 1 -p tcp -m tcp --dport 80 -j ACCEPT
+		iptables -C INPUT -p tcp -m tcp --dport 53 -j ACCEPT &> /dev/null || iptables -I INPUT 2 -p tcp -m tcp --dport 53 -j ACCEPT
+		iptables -C INPUT -p udp -m udp --dport 53 -j ACCEPT &> /dev/null || iptables -I INPUT 3 -p udp -m udp --dport 53 -j ACCEPT
+            ) || echo "::: iptables commands errored out, please confirm ports 53/udp, 53/tcp, and 80/tcp are open"
+        return
+    fi
+	echo "::: No firewall detected.. skipping firewall configuration."
 }
 
 finalExports() {


dschaper:
Like the OR checking, and I think if there's a failure in the subshell then `set -e` should bounce out of the shell and not out of the script. Thinking how to replicate this for testing....

brontide:
It's fully tested and it's mostly an edge case since it's pretty hard to not load those modules, you might be able to do it by blacklisting the ip_tables module.  There are still problems with the handling of iptables since it doesn't save across boots.  I've also got a handful of additional fixes at home for centos 6 handling since it's php is out of date.  The web ui requires 5.4 or higher.

dschaper:
May want to consider not touching the firewall for some/most installs and let the user handle the details of their own installs. PHP is another issue, we're working how to solve that one over on that repository.

bcambl:
Let me guess. hash_equals.  :P

brontide:
php is easy on rhel/centos 6

```
    # v6 variants php is too old, install repo for php7.1
    if grep -q -i "release 6." /etc/redhat-release
    then
      echo "::: running CentOS/RHEL 6.X, we need newer version of php"
      echo -n "::: installing additional repos for php 7.1... " 
      ( yum -y install http://rpms.famillecollet.com/enterprise/remi-release-6.rpm &> /dev/null  && yum -y install yum-utils &> /dev/null && yum-config-manager --enable remi-php71 &> /dev/null && echo "done!" ) ||
 exit -1
    fi

```

I put this in the rpm block and php7.1 installs during the package installation.

dschaper:
From a personal standpoint, and as a user, I would veto any blind addition of non-official repos, and I know our users would be very vocal about us allowing that to occur. The IPTables stuff is great and needs testing, but installing Remi for PHP is not something that I would approve of.

brontide:
Centos 6 is still a well supported platform.  Would you be more comfortable with a dialog for approval or just an error message with resolution?

    echo `yum list -q php | grep php.x86_64 | awk '{print $2}' | awk -F. '{print $1 "." $2 }'` ">=" 5.4 | bc

Will return 1 if the version is 5.4 or higher ( supported ) and 0 or error otherwise.

As for iptables I'm leaning towards dump it since the support is kinda rickety unless we add it to the startup.

dschaper:
I think for now we're looking at rewriting the function that causes the dependency on 5.4 in a way that is operational on earlier versions of PHP. There's a lot of internal discussion on how to handle EOL packages/platforms/OSes. We originally were just a Raspberry Pi project and now are facing some questions that will shape how we develop for the future on all the different platforms that our users have found capable.

bcambl:
If the team does decide to add better support for CentOS by enabling third party repositories, the EPEL repository should be moved into the same logic block.  

The release check should be for Fedora or CentOS:
Fedora = supported due to package availability via default repos && continue installation
CentOS = prompt user && enable EPEL/Remi (also change a few package names like `procps`)
Other = notify user of un-tested distribution detected and a prompt to continue

You would also need to have logic to collect the `$releasever` and `$basearch`. so you enable the correct repository versions. That said, the repositories do not have ARM packages built and would require building a few packages from source in the event where someone is using CentOS on ARM (see: centos userland images). Fedora has had ARM support on alternative platforms for a while now and recently announced Pi support with Fedora 25. All required packages are available via default Fedora repositories.

brontide:
@bcambl the installer already installs epel-release as a dependency.

dschaper:
It's a matter of source. Can you provide a location from Remi that would show us the actual source he used and the process he uses to create the packages? I can get that from EPEL, so I can follow the chain back to something I can base my trust on.

brontide:
http://rpms.famillecollet.com/SRPMS/  webtastic is another good repo, but it's not a drop in because the php packages are named differently.

But I think that's putting the cart before the horse, I think we need to be addressing the question about support of other platforms.  Personally I would love to do testing, but I can't get lxd to run a debian image without choking on systemd issues so I'm hesitant to make more drastic changes to the installer.

I've been doing some thingking and personally I think basic-installer.sh needs to go on a diet.  Much of what you do now could easily be moved to a post-git step where it's executed from the installed codebase.  We could then move more of these smarts into individual scripts that are cleaner and easier to maintain.

As for dependencies I would say concentrate on "berries" and then for non-berries do a sanity check for the actual binaries like ps, php, and lighttpd rather than trying to get it right for every platform.  If you do add platforms them specific rather than general so that you will know positively that they will work.

So you might as well drop this patch and pull out, as well as the rest of the yum support.

dschaper:
I tend to agree with most of what you have to say. So I'll go ahead and put in my vote now.

dschaper:
Borg

PR #831, opened 2016-10-22
Consistent Style in Bash

airencracken:
_This template was created based on the work of [`udemy-dl`](https://github.com/nishad/udemy-dl/blob/master/LICENSE)._

**By submitting this pull request, I confirm the following (please check boxes, eg [X]):**

- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [x] I have checked that [another pull request](https://github.com/pi-hole/pi-hole/pulls) for this purpose does not exist.
- [x] I have considered, and confirmed that this submission will be valuable to others.
- [x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [x] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:**

- [x] 1 (very unfamiliar)

---
So it seems the bash here had about four or five different styles co-existing. I've tried to clean that up and make it look similar everywhere. Even if someone doesn't like the particular style I've picked (I tried to go with whatever seemed most common in the files already) it's easier to change now that it's all relatively the same. I suspect this was the result of a lot of copy/paste. ~~Going to hopefully work on another commit that pulls out some of the more common functions (like the spinner function) and puts them into a library that the respective scripts can just source.~~

**Edit: I'll do that in a separate pull request.

PR #835, opened 2016-10-23
Debug script tweaks.

PromoFaux:

- [x] 10 (very familiar)

---

Fix shebang line broken in #831

Remove root check and references to ${SUDO}

Check for /etc/pihole/setupVars.conf and use piholeInterace in IPv6 checks as suggest by @jacobsalmela https://github.com/pi-hole/pi-hole/pull/831#issuecomment-255606854

### DIFF
--- a/advanced/Scripts/piholeDebug.sh
+++ b/advanced/Scripts/piholeDebug.sh
@@ -1,4 +1,4 @@
-usr/bin/env bash
+#!/usr/bin/env bash
 # Pi-hole: A black hole for Internet advertisements
 # (c) 2015, 2016 by Jacob Salmela
 # Network-wide ad blocking via your Raspberry Pi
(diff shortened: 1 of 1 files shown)


PromoFaux:
Engage

advanced/Scripts/piholeDebug.sh
```diff
@@ -33,32 +33,15 @@ cat << EOM
 ::: and optionally upload the generated log to a unique and random directory on
 ::: Termbin.com. NOTE: All log files auto-delete after 1 month and you are the only
 ::: person who is given the unique URL. Please consider where you post this link.
-::: 
+:::
 EOM
 
-
```
dschaper: Why did this leave? Because it's only fired from `pihole -d` and `pihole` is a sudo script?

dschaper:
Approved

advanced/Scripts/piholeDebug.sh
PromoFaux: To get in line with the rest of the scripts (See #814)m `pihole` (well, and `basic-install.sh`) will be the only script to have the root check, as everything else is called by it. No need to keep checking for elevation in each script called by `pihole`! 

It also means that script's **must** be called using `pihole`, although if you wanted you could run `sudo /opt/pihole/somescript.sh`

jacobsalmela:
Approved

PR #849, opened 2016-10-25
[BugFix] A man has no name resolution

PromoFaux:

- [x] 10 (very familiar)

---

(should) fix an issue brought up here: https://github.com/pi-hole/pi-hole/pull/814#issuecomment-256037665

### DIFF
--- a/advanced/01-pihole.conf
+++ b/advanced/01-pihole.conf
@@ -20,6 +20,9 @@
 #                        OR IN /etc/dnsmasq.conf                              #
 ###############################################################################
 
+address=/pi.hole/@IPv4@
+address=/pi.hole/@IPv6@
+
 addn-hosts=/etc/pihole/gravity.list
 
 domain-needed
--- a/advanced/01-pihole.conf
+++ b/advanced/01-pihole.conf
@@ -23,6 +23,9 @@
 address=/pi.hole/@IPv4@
 address=/pi.hole/@IPv6@
 
+address=/@HOSTNAME@/@IPv4@
+address=/@HOSTNAME@/@IPv6@
+
 addn-hosts=/etc/pihole/gravity.list
 
 domain-needed
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -536,6 +536,18 @@ version_check_dnsmasq() {
 	else
 		sed -i '/^server=@DNS2@/d' ${dnsmasq_pihole_01_location}
 	fi
+
+	if [[ -f /etc/hostname ]]; then
+		hostname=$(</etc/hostname)
+		sed -i "s/@HOSTNAME@/$hostname/" ${dnsmasq_pihole_01_location}
+	elif [ -x "$(command -v hostname)" ]; then
+		hostname=$(hostname -f)
+		sed -i "s/@HOSTNAME@/$hostname/" ${dnsmasq_pihole_01_location}
+	else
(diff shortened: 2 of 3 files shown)


PromoFaux:
Downside to this method is of course:
```
adam@adam-debian:/etc/.pihole$ ping pi.hole
PING pi.hole (192.168.216.236) 56(84) bytes of data.
64 bytes from 0000mps.webpreview.dsl.net (192.168.216.236): icmp_seq=1 ttl=64 time=0.016 ms
```

But I'm sure we can find another way around this.

PromoFaux:
Hmm, this isn't working on my Pi at home. Yet it worked earlier at work....

PromoFaux:
Last commit fixed the issue I was having above.

dschaper:
Approved

jcandresen:
is there a way to fix the pi showing up as 0000mps.webpreview.dsl.net ?

PromoFaux:
Unfortunately, a side effect of this change was that locally the pi will show as whatever the top result in `/etc/pihole/gravity.list` is. 

However it made sense at the time to move the `pi.hole` and `<whateveryourhostnameis>` from `gravity.list` as they're not really blocking domains.

If we come up with a solution, I'll try and remember to update this thread!

tl87:
I've noticed the problem as well.
pihole error
To make things clear, this means that everytime I see "0000mps.webpreview.dsl.net", it's just the Pi-Hole, that is using the first line in /etc/pihole/gravity.list as it's hostname?

DL6ER:
Indeed, I can confirm this also as follows:

<pre>
me@local ~ $ host pi.hole
pi.hole has address 1.2.3.4
pi.hole has IPv6 address ....
me@local ~ $ host 1.2.3.4
1.2.3.4.in-addr.arpa domain name pointer <b>0.r.msn.com</b>.
me@local ~ $ 
</pre>

@PromoFaux Could we just add `ip.of.pi.hole pi.hole` (and the same for IPv6) in the first line of `gravity.list`?

PromoFaux:
@DL6ER 
We used to have it at the top of `gravity.list`, but when in `pihole disable` mode, you would no longer be able to resolve `pi.hole`.

The work around to that was to remove it from `gravity.list` and add it into `01-pihole.conf`

@tl87 that's correct, it doesn't matter what the first line is, it will return that name.

@pi-hole/gravity, is it overkill to have two entries for `pi.hole` ? One at the top of `gravity.list` for reasons mentioned in the above two comments, and one in `01-pihole.conf` for the purpose of still being able to resolve it when `pihole disable` mode is on.

Mcat12:
As long as it works, that sounds like a good idea. Only thing might be that pinging the Pi-hole while it's disabled would have a different hostname than when it's enabled. So maybe a better idea would be to create a small list with only hostname/`pi.hole` information and load that list before gravity.list? So basically instead of setting the information in the dnsmasq config we link to a small list with the same information. In that list we could optionally also implement the custom local domain names, if we want to do that.

PromoFaux:
so `hostnames.list` and `gravity.list` and have two `addn-hosts` lines?

Mcat12:
Ya, then the gravity count wouldn't be inflated by `pi.hole` and other unblocked hosts.

PromoFaux:
OK, let's have a play and see what happens...

PromoFaux:
Yep, that works. I'll stick in a pull request.

PromoFaux:
@jcandresen @tl87 your concerns should be resolved in the next release

tl87:
@PromoFaux much much appreciated :-)

PR #858, opened 2016-10-26
[discussion] Alpine support

brontide:

- [ ] 1 (very unfamiliar)
- [ ] 2
- [ ] 3
- [ ] 4
- [ ] 5
- [ ] 6
- [X] 7
- [ ] 8
- [ ] 9
- [ ] 10 (very familiar)

---

This is Alpha level support for alpine as I'm still getting a feel for it.  I have not yet run this branch through the meat grinder on the other distros.  Many of the small things I found along the way are solved by improving the base to use `install` rather than `cp`.  Busybox usually has a single way to do some of the stuff like user and group add/mod.

What works ( basically everything! )
- installer
- pihole
- webui
- [debug log](http://termbin.com/qhgo)

Not working/skipped
- Package update listing/count

Not tested
- sudo
- cron ( I'm guessing this will need some work based on what I saw but not much )


brontide:
Oh, and you need to `apk add bash ncurses` to run the installer!

PromoFaux:
I was going to test this but I'm struggling with the setup of alpine!

Bugs out at the disk section (installing as vm on hyper-v)

brontide:
I'm neither a hyper-v nor Alpine on VM expert but the tips looked like a easy place to start ( kernel command like options ).

- [tips which might help](https://technet.microsoft.com/en-us/windows-server-docs/compute/hyper-v/best-practices-for-running-linux-on-hyper-v?f=255&MSPPError=-2147217396)
- [Alpine discussion](https://forum.alpinelinux.org/forum/installation/alpine-linux-ms-hyper-v)
- [RHEL 6 numa](https://access.redhat.com/solutions/436883)

I'll take some time today and run this branch against the other 9 hosts I have to see if there are any reversions.

brontide:
ncurses is no longer a pre-install dep.

automated install/basic-install.sh
```diff
@@ -32,9 +32,10 @@ useUpdateVars=false
 IPv4_address=""
 IPv6_address=""
 
-# Find the rows and columns
-rows=$(stty size | awk '{print $1}')
-columns=$(stty size | awk '{print $2}')
+# Find the rows and columns will default to 80x24 is it can not be detected
```
PromoFaux: Nice addition. What do the whiptail dialogs look like at 80x24?

brontide:
The update/reconfigure spills, but the rest work perfectly

automated install/basic-install.sh
```diff
@@ -40,6 +40,9 @@ columns=$(echo $screen_size | awk '{print $2}')
 # Divide by two so the dialogs take up half of the screen, which looks nice.
 r=$(( rows / 2 ))
 c=$(( columns / 2 ))
+# Unless the screen is tiny
+r=$(( r < 20 ? 20 : r ))
+c=$(( c < 70 ? 70 : c ))
```
PromoFaux: TIL you can do `statement` `?` `result if` `:` `result if not`  conditionals in bash.

PromoFaux: I like it! I have a soft spot for one-liners

diginc:
Is the goal supporting Alpine x86 installations , Raspberry Pi alpine, or both?  My docker container runs anywhere docker runs FYI :D

My x86 optimized docker is alpine based for reference though I don't to worry about making sure the install script works since the Dockerfile replaces install functionality https://github.com/diginc/docker-pi-hole/blob/master/alpine.docker

brontide:
@diginc FYI, pi-hole seems to request 53 tcp as well as udp from the firewall but your dockerfile only passes udp.

diginc:
No worries, Dockerfile ports aren't used since they only ever get auto mapped to random host ports.  All the README.md run commands say to pass both TCP  & UDP

advanced/lighttpd.conf.debian
```diff
@@ -52,10 +52,11 @@ $HTTP["url"] =~ "^/admin/" {
     )
 }
 
+# Rewite js requests, must be out of $HTTP block due to bug #2526
```
diginc: I did not have any problem with javascripts getting rewritten with my alpine docker.

are you sure this isn't fixed in the newest alpine lighttpd?

brontide: Got pulled in from my other work it's for debian wheezy and ubuntu 14.04 lts

diginc:
> Is the goal supporting Alpine x86 installations , Raspberry Pi alpine, or both?

still wondering

brontide:
I was curious how hard it would be.  Alpine was actually a lot easier than I expected.  The testing of minimal platforms also allows me to pinpoint other hidden dependencies which might be overlooked.

Personally i don't care much for docker as it's neither fish nor fowl and it doesn't sit right with me.  Call it a matter of taste or style I don't see myself running docker by choice.

I put the pull in to get some visibility and generate discussion, if it's not wanted then it will age and be deleted.

diginc:
Support for installing natively on top of OSs is fine and dandy and some people will always prefer it..  This one is actually pretty minimal and non intrusive to add and I like it.  It is just my opinion that it is a slippery slope to support too many OSs.

The reason I keep reiterating my question 'is it Raspberry Pi or x86 alpine' is it'll increase that compatibility matrix (test matrix) by 2 instead of just 1 if you answer both and I want the pi-hole official guys to keep that in mind since they're the ones who will need to test a whole list of features still work times however many OSs are officially supported.  These are not problems I personally have in docker land.

brontide:
As far as the documentation goes this is for raspbain, the support for other platforms is a plus.  Testing on a wide selection of hosts is not too difficult once automated.  I'm trying to add a arm based VM with raspbain which would bring my list up to 11 different distros.

dschaper:
I'm hoping to get a look at this in the near future, I've changed my dev box over to Fedora so I'm set up for libvirt and next thing to tackle is the VM/VE's.

dschaper:
Closing as outdated.

PR #870, opened 2016-10-27
Revert "Remove `listen-address` directive"

dschaper:
Reverts pi-hole/pi-hole#869


dschaper:
Don't believe we will be needing this option.

PR #891, opened 2016-11-02
Bugfix/chronometer domain count

### LINKED ISSUE #674: LCD Script (chronometer) - Displaying different data than OS or WEBIF
##### Expected Behaviour:

Blocking, Queries, Pi-holed values match between webif, fresh OS login, and chronometer script running on LCD screen.
##### Actual Behaviour:

"Queries" value resets every day and remains at a very low value such as 2 or 3 until a reboot. Even though WEBIF or fresh login to the OS reflects the true "Queries" value, in the thousands.

run chronometer script on LCD screen following the pi-hole guide, and wait approximately 24 hours or so.
https://pi-hole.net/faq/install-the-real-time-lcd-monitor-chronometer/
##### Forum Thread Reference:

https://pi-hole.net/forums/topic/lcd-script-chronometer-displaying-different-data-than-os-or-webif/
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/37776882-lcd-script-chronometer-displaying-different-data-than-os-or-webif?utm_campaign=plugin&utm_content=tracker%2F3011939&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F3011939&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jacobsalmela:

- [x] 10 (very familiar)

---
Fixes chronometer output so it doesn't display twice as many domains as are actually being blocked.

Fixes #674

### DIFF
--- a/advanced/Scripts/chronometer.sh
+++ b/advanced/Scripts/chronometer.sh
@@ -61,14 +61,6 @@ CalcPercentBlockedToday() {
 	fi
 }
 
-CheckIPv6() {
-	piholeIPv6file="/etc/pihole/.useIPv6"
-	if [[ -f ${piholeIPv6file} ]];then
-		# If the file exists, then the user previously chose to use IPv6 in the automated installer
-		piholeIPv6=$(ip -6 route get 2001:4860:4860::8888 | awk -F " " '{ for(i=1;i<=NF;i++) if ($i == "src") print $(i+1) }')
-	fi
-}
-
 outputJSON() {
 	CalcQueriesToday
 	CalcblockedToday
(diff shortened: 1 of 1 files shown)


dschaper:
Looks sound. Do we want to block it in and do `if IPv6 elif IPv4 else err`, or does that check already happen by the time we get to Chono?

PromoFaux:
Actually thinking about it (and this applies to Admin page aswell)

We actually need to expand this check:
```
if ipv4 only
    numdomains = numdomains
if ipv6 only
    numdomains = numdomains
if ipv4 && ipv6
   numdomains = numdomains /2
```

Although could that actually be:
```
if ipv4 && ipv6
    numdomains = numdomains /2
else
    numdomains = numdomains
fi
```

dschaper:
Good catch!

dschaper:
Approved

PR #899, opened 2016-11-03
Configure firewall unittest

diginc:
**By submitting this pull request, I confirm the following**

- [] 1 (very unfamiliar)
- [] 2
- [] 3
- [] 4
- [] 5
- [] 6
- [x] 7
- [] 8
- [] 9
- [] 10 (very familiar)

---

Expanding unit testing to configure firewall function.  I might expand more on this function's test still, I just wanted to open this to show off mock_command possibilities in bash testing, enabling whiptail / firewal changes that would otherwise not be possible easily.

Found the testinfra Docker.run() (Pihole in this code base) function defaults to using [dash](https://github.com/philpep/testinfra/blob/55f25e30da3f09b11adc4f596e21ed6823513b90/testinfra/backend/docker.py#L35) which is why I had some trouble with the earlier tests and had to write bash script files into the docker and them execute them.  I should be able to remove the need for `_write_test_script` function in an upcoming refactor...simply things a bit. Edit: DONE!


diginc:
I'd like to get this in before adding any more tests since this has changes beyond just the one firewall test (optimizing test framework overall).  Plus, I might not have time to add anymore in the next couple days so shipping as is seems fine.  If you have questions about anything let me know or hit me up in chat.

PromoFaux:
What sets the value `PHTEST`?

Also, as it's a boolean, could we match it to the other flag checks? e.g `if [[ "${PHTEST}" == false ]]` ?

diginc:
Before I source basic-install.sh I usually set PHTEST in the shell of the docker tests so I can get functions but not execute anything.  I may just move it to the Dockerfiles to prevent repeatedly setting it every test though.

I'll apply your other suggestions tonight.  I went string because I was sure it would work, whereas I remembered [bash booleans being a minefield of gotchas](http://stackoverflow.com/questions/2953646/how-to-declare-and-use-boolean-variables-in-shell-script/21210966#21210966) ... after reading that, fairly easy when following the [[ ]] convention I guess.

dschaper:
`ENV` a possibility?

diginc:
Yup, `ENV` is how you'd move it to the Dockerfile to make it permanent in all the test dockers' sessions.

diginc:
updated, couldn't have done PH_TEST == false since it's a flag you flip on in the tests - not a flag we always have turned off out in the real world (it would be null string/var in the wild)

dschaper:
Okay for code review?

dschaper:
Approved, will wait till okay for merge from @diginc

diginc:
I'm okay with merging.

dschaper:
Need one more developer approval and then a cla agreement and we can merge it.

Mcat12:
Approved

PromoFaux:
@diginc Please sign the CLA.

PR #907, opened 2016-11-06
replace hardcoded interface with var

jacobsalmela:
setupVars.conf is already sourced, so we just need to replace the
hardcoded value with the variable name.


PromoFaux:
Silly question, why not use IPV4_ADDRESS ?

dschaper:
Approved

PR #961, opened 2016-12-04
Custom Pi-Hole block page

DL6ER:

- [X] 10 (very familiar)

---

Add a custom Pi-Hole block page with more details for a user that does not know why the page he wants to visit won't load / is blocked.


DL6ER:
It now tells the user why the requested domain is blocked

screenshot at 2016-12-05 11-27-25

For this to work you have to have the changes of https://github.com/pi-hole/AdminLTE/pull/222 in the web UI scripts.

Note that loading of the list of results is asynchronous, so the page will immediately be displayed while the searching through the lists might still be running (especially important on my slow Pi B+).

automated install/basic-install.sh
```diff
@@ -955,6 +964,9 @@ accountForRefactor() {
 	sed -i 's/piholeDNS1/PIHOLE_DNS_1/g' ${setupVars}
 	sed -i 's/piholeDNS2/PIHOLE_DNS_2/g' ${setupVars}
 
+	# Account for change in lighttpd config file
+	sed -i 's:pihole/index.html:pihole/index.php:' ${lighttpdConfig}
```
Mcat12: Shouldn't the config get overwritten on line 715, making it already have the right path?

DL6ER: Good spot, totally overlooked that we overwrite it anyway...

DL6ER:
I should point out that we do no succeed with showing the blocking page in all scenarios.

Example: www.google-analytics.com enforces HTTP Strict Transport Security (HSTS) with long duration and therefore the browser will always enforce SSL for a connection. However, we cannot deliver a blocking page via SSL with the correct SSL subject and therefore we cannot show the blocking page in those cases.
screenshot at 2016-12-06 12-18-18

There is nothing we can really do about this.

Mcat12:
How about in page ads? Is the message ever shown?

DL6ER:
I haven't seen it in page for some days. However, this morning I've seen this ...

screenshot at 2016-12-06 13-48-29

These parts of the webpage are loaded as iframes:
```
<iframe src="http://rcm-de.amazon.de/e/cm?t=unterhaltun07-21&amp;o=3&amp;p=16&amp;l=st1&amp;mode=books-de&amp;search=weihnachten%20advent&amp;fc1=&amp;lt1=&amp;lc1=&amp;bg1=FFDF7F&amp;f=ifr" marginwidth="0" marginheight="0" width="468" height="336" border="0" frameborder="0" style="border:none;" scrolling="no"></iframe>
```
Not so nice.

DL6ER:
Found a solution to detect if the blocking page in display as in page ad.

Compare to screenshot shown above:
screenshot at 2016-12-06 14-07-09

Mcat12:
Something like [this](https://github.com/pi-hole/pi-hole/blob/master/advanced/lighttpd.conf.debian#L52) would also work to prevent iframes.

DL6ER:
@Mcat12 That works as well and is a much cleaner version. Thanks!

advanced/lighttpd.conf.debian
```diff
@@ -59,7 +59,10 @@ url.rewrite = ( "^(?!/admin/).*\.js$"  => "pihole/index.js"   )
 # If the URL does not start with /admin, then it is a query for an ad domain
 $HTTP["url"] =~ "^(?!/admin)/.*" {
     # Create a response header for debugging using curl -I
-    setenv.add-response-header = ( "X-Pi-hole" => "A black hole for Internet advertisements." )
+    setenv.add-response-header = (
+        "X-Pi-hole" => "A black hole for Internet advertisements.",
+        "X-Frame-Options" => "DENY"
```
Mcat12: Since all ads are going through here, you might want to just add a regex for `pihole/ ` separately.

DL6ER: Why is it bad to have this property for all ads?

DL6ER:
@Mcat12 As far as I can see it will just prevent opening anything that is offered by the Pi-Hole in a frame environment. That seems fine to me. However, it creates an error in the log.

screenshot at 2016-12-06 17-28-55

My JavaScript solution did not cause any errors.

advanced/lighttpd.conf.debian
Mcat12: Some people want to replace ads with an image or other content, so we can't set the frame options unless we take away that ability. However, since we're already replacing that option with this page, it might not matter anymore.

DL6ER:
Wit some help from others I was able to get a browser test for the new blocking page on various platforms. Everything seems to work quite nicely almost always, however the "suppress in page" feature does **not** work for IE 8 and below. Firefox and Chrome always work (even in old versions like Firefox 4.0).

PromoFaux:
>however the "suppress in page" feature does not work for IE 8 and below

Then those people should update to a modern browser. Sometimes it's more hassle than it is worth trying to achieve complete backward compatibility with older browsers.

Tested Working OK on all other browsers on my network 👍

Mcat12:
[This](http://stackoverflow.com/a/10979107) might be interesting to test on IE 8

C4Wiz:
this seems nice for webpages but what about for ads in a page like this?

this is mine:

screen shot 2016-12-10 at 9 42 49 am

this is my full page:

screen shot 2016-12-10 at 9 43 38 am

DL6ER:
@C4Wiz We want to preserve the current behavior, i.e. don't display anything in page by default.

DL6ER:
@Mcat12 Just tested it, doesn't work with IE 8 either.

C4Wiz:
is there a easy way to toggle this new behavior?

DL6ER:
@C4Wiz We are basically only doing two things:
1. We change the displayed page from index.**html** to index.**php**
2. Instead of an empty index page, there is now some content in this page.

So, you just have to replace the new index page with your custom page you already have.

I'm currently working on a detailed settings page. I could imagine to add an option where the user can chose between e.g.

1. Display the blocking page (new behavior)
2. Display nothing (empty page, current behavior)
3. Display in page content (like what you have shown [here](https://github.com/pi-hole/pi-hole/pull/961#issuecomment-266221907))

advanced/index.php
```diff
@@ -0,0 +1,103 @@
...
+	<meta name='robots' content='noindex,nofollow'/>
+</head>
+<body id="body">
+<header>
+	<h1><a href='/'>Website Blocked</a></h1>
+</header>
+<main>
+	<div>Access to the following site has been blocked:<br/>
+	<span class='pre msg'><?php echo $serverName.$uri; ?></span></div>
+	<div>If you have an ongoing use for this website, please ask the owner of the Pi-Hole in your network to have it whitelisted.</div>
```
Mcat12: Pi-Hole -> Pi-hole

jpgpi250:
I've tested the new index.php, blockingpage.css and modified lighttpd.conf. Since I've changed the pi-hole's admin page to be password protected (see Jacob salmela's [link](https://jacobsalmela.com/2014/05/25/password-protect-a-lighttpd-web-server-on-a-raspberry-pi-using-mod-auth/), I had to move (blockingpage.css) /copy (favicon.png, logo.svg, jquery.min.js and queryads.js) the files out of the admin directory (moved them to /var/www/html/pihole), otherwise I had to authenticate multiple times.
This requires editing both blockingpage.css and index.php

DL6ER:
@jpgpi250 Thanks for the test. We implement a password protection into the web UI with the upcoming release. This one is very convenient, easy to set up and secure (no plaintext password stored anywhere), hence recommended for all users. You will have to remove your existing authentication to prevent having to log in twice.

DL6ER:
Resolved merge conflicts. Ready for review and merge.

dschaper:
Approved

jcastro:
This is great! might be interesting also that I could blacklist it as well? I'm not seeing it being added to the blacklist! 

thanks a lot!

PR #1008, opened 2016-12-22
Feature/track pre installed packages

dschaper:
**By submitting this pull request, I confirm the following _Failure to fill the template will close your PR_:**

- [x] 10 (very familiar)

---
Track which packages we install, and keep the package managers from grabbing unneeded downloads.


dschaper:
Looks like this fixes the slow install on Fedora droplets.

PromoFaux:
>Track which packages we install

How about on Pi-hole updates?

dschaper:
Right now it just checks each package for installation status, and packages that are not installed get added to an array. Then the array is passed as a string to the package installer. So with things like Fedora dnf/yum that update the cache every time `dnf` is called, this is big savings in time. Plus some Debians will download every package passed to it, so again this will only download the packages not currently installed.

It's a future thing to make this sticky so that we can track installed packages (in setupVars? or another file?) so that they could be selectively removed on uninstall, if someone was to get the crazy idea of uninstalling Pi-hole...

PromoFaux:
OK, I get that.. I think. So if all Pi-hole dependencies are already installed by a previous installation (from before this PR), and the user then updates... Are the packages added to the list, even though they weren't installed by the update? *brain fart*

`setupVars.conf` is as good a place as any, but it's probably best we move that file to `/opt/pihole/pihole.cfg` or something

dschaper:
It it's a reinstall, then the array is of 0 length, so the dialog to install isn't even shown. It still runs down the list of dependencies and does a `dpkg` to check if it's already there (equivalent in Fedora). So on reinstall, there wouldn't be any packages that show as uninstalled, unless the user uninstalled or we added a dependency with a new release. 

And yeah, config files usually go in `/etc/default/<packagename>` or `/etc/<packagename>/*` if there are a few of them. Would be good to start to get aligned with the LFH at some point.

PromoFaux:
>So on reinstall, there wouldn't be any packages that show as uninstalled, unless the user uninstalled or we added a dependency with a new release.

Yeah, this is what I mean. I think. So if it's a reinstall, nothing will get populated in the config file, unless it's a new dependency we added in that update?

> `/etc/<packagename>/*` ah, like we already do then!

dschaper:
Exactly, if there's no new dependencies added, then the scan will show all packages installed, and no need to call the installer and nothing added to the array. (There's a check in the install section that bypasses if the array is length 0)

PR #1024, opened 2016-12-23
Make grep case insensitive

DL6ER:
Internal. Make grep case-insensitive.
screenshot at 2016-12-23 17-04-34


jacobsalmela:
Approved

PR #1009, opened 2016-12-22
replace 'git -C' with long version

bcambl:
- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [x] I have checked that [another pull request](https://github.com/pi-hole/pi-hole/pulls) for this purpose does not exist.
- [x] I have considered, and confirmed that this submission will be valuable to others.
- [x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [x] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:**
- [x] 8

---

The -C argument was introduced in git 1.8.4. CentOS 7.3 at the time of this commit provides v1.8.3.1.
see: https://git.kaarsemaker.net/git/commit/44e1e4d67d5148c245db362cc48c3cc6c2ec82ca/

Fixes #1004

PR #1033, opened 2016-12-24
update.sh code refactoring

DL6ER:

---

Code refactoring of the existing `update.sh` without changing its external behavior. Internally, the behavior changes (getting completely independent of GitHub tags).

The old logic has been:

1. Get tag from GitHub
2. Get local tag
3. Compare tags and update the corresponding parts, if necessary

Proposed new logic (we discussed about that):

1. Fetch (possible) changes from `origin`
2. Get local git status
3. Update if local copy is behind remote version

Note that the new behavior is independent of tags but relies on the git log, only. Hence, it is much more reliable and trustworthy.

### DIFF
--- a/advanced/Scripts/update.sh
+++ b/advanced/Scripts/update.sh
@@ -22,13 +22,15 @@ readonly PI_HOLE_FILES_DIR="/etc/.pihole"
 is_repo() {
   # Use git to check if directory is currently under VCS, return the value
   local directory="${1}"
+  local curdir
+  local rc
 
-  curdir=$PWD
-  cd $directory
+  curdir="${PWD}"
+  cd "${directory}" || { echo "Unable to change to ${directory}, exiting."; exit 1; }
   git status --short &> /dev/null
   rc=$?
-  cd $curdir
-  return $rc
+  cd "${curdir}" || { echo "Unable to change to ${curdir}, exiting."; exit 1; }
+  return "${rc}"
 }
 
 prep_repo() {
--- a/advanced/Scripts/update.sh
+++ b/advanced/Scripts/update.sh
@@ -40,22 +40,19 @@ make_repo() {
   local remoteRepo="${2}"
   local directory="${1}"
 
-  (prep_repo "${directory}" && git clone -q --depth 1 "${remoteRepo}" "${directory}" > /dev/null)
+  (prep_repo "${directory}" && git clone -q --depth 1 "${remoteRepo}" "${directory}")
      return
 }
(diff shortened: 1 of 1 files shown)


DL6ER:
Note that this will also make the updater work for other branches, like:
```
root@raspberrypi:/etc/.pihole# git pull
Already up-to-date.
root@raspberrypi:/etc/.pihole# git reset --hard HEAD^
HEAD is now at 4632b0f Updated updater logic
root@raspberrypi:/etc/.pihole# pihole -up
::: Checking for updates...
::: Pi-hole Core:   update available
::: Web Interface:  up to date
:::
::: Pi-hole core files out of date
:::
[... rest not shown ...]
```

advanced/Scripts/update.sh
```diff
@@ -40,22 +44,19 @@ make_repo() {
...
   # Stash all files not tracked for later retrieval
-  git -C "${directory}" stash --all --quiet &> /dev/null || ${retVal}=1
+  git -C "${directory}" stash --all --quiet
   # Force a clean working directory for cloning
-  git -C "${directory}" clean --force -d &> /dev/null || ${retVal}=1
+  git -C "${directory}" clean --force -d
   # Fetch latest changes and apply
-  git -C "${directory}" pull --quiet &> /dev/null || ${retVal}=1
-  return ${retVal}
+  git -C "${directory}" pull --quiet
```
Mcat12: We should use the long version of `-C` #1009

dschaper:
`pihole -v` would show the tag for the release, but not the hash (currently.) Ideally, the update script shouldn't output any kind of information, just a return value that the caller would identify and then let the caller print the information. `update.sh` shouldn't ever be run bare, it should be called as a utility script/library function. Will the user ever see any of the output from this script?

DL6ER:
It is run bare for updating.

https://github.com/pi-hole/pi-hole/blob/development/pihole#L51

dschaper:
Right, it's called via a script though, we don't advise users to go to a shell prompt and run `update.sh` by itself.

DL6ER:
Does that matter if is called via a script? Would you like to rewrite more fundamental?

dschaper:
I'm over on MM now, but I think these should be utilities or the beginnings of a library and we handle all the input/output from `pihole`. So we keep the interface with the user in one place, instead of spreading it out. Longer explanation in chat...

dschaper:
Approved

dschaper:
Approved

PR #1048, opened 2016-12-28
Major simplification of webpage.sh + extend DHCP to IPv6

### LINKED ISSUE #1034: Nest thermostat offline after installing Pi-Hole
**In raising this issue, I confirm the following (please check boxes, eg [X]) Failure to fill the template will close your issue:**

- [x] The issue I am reporting can be *replicated*
- [x] The issue I'm reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

_1_

---
**[FEATURE REQUEST | QUESTION | OTHER]:**

Please [submit your feature request here](https://discourse.pi-hole.net/c/feature-requests), so it is votable by the community.  It's also easier for us to track.

**[BUG | ISSUE] Expected Behaviour:**
Nest thermostat should be online and should be controlled by Nest mobile app/web UI.  I also have Nest cameras and they are working fine.

**[BUG | ISSUE] Actual Behaviour:**
Nest thermostat shows offline and not cannot be accessed from Mobile app/web UI.

**[BUG | ISSUE] Steps to reproduce:**
Install Nest thermostat.  Install Pi-Hole connected to the router.
-
-
-
-

**(Optional) Debug token generated by `pihole -d`:**
1rn87fe1wd

DL6ER:

---
Internal PR. Will make it much simpler to add additional settings in the future.

Will also add IPv6 DHCP functionality.

Since DNS settings will be reprocessed after an update and DHCP settings will be stored in a different file, this will fix #1034 

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -179,6 +179,8 @@ SetQueryLogOptions(){
 
 ProcessDHCPSettings() {
 
+	if [[ "${DHCP_ACTIVE}" == "true" ]]; then
+
 	source "${setupVars}"
 	interface=$(grep 'PIHOLE_INTERFACE=' /etc/pihole/setupVars.conf | sed "s/.*=//")
 
@@ -211,6 +213,9 @@ dhcp-range=::100,::1ff,constructor:${interface},ra-names,slaac,${leasetime}
 ra-param=*,0,0
 " > "${dhcpconfig}"
 
+	else
+		rm "${dhcpconfig}"
+	fi
 }
 
 EnableDHCP(){
@@ -239,7 +244,7 @@ DisableDHCP(){
 	delete_dnsmasq_setting "dhcp-"
 	delete_dnsmasq_setting "quiet-dhcp"
 
-	rm "${dhcpconfig}"
+	ProcessDHCPSettings
 
 	RestartDNS
 }
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -9,8 +9,6 @@
 # the Free Software Foundation, either version 2 of the License, or
 # (at your option) any later version.
 
-args=("$@")
-
 readonly setupVars="/etc/pihole/setupVars.conf"
 readonly dnsmasqconfig="/etc/dnsmasq.d/01-pihole.conf"
 readonly dhcpconfig="/etc/dnsmasq.d/02-pihole-dhcp.conf"
@@ -275,30 +273,35 @@ ResolutionSettings() {
 	fi
 }
 
-case "${args[1]}" in
-	"-p" | "password"   ) SetWebPassword;;
-	"-c" | "celsius"    ) unit="C"; SetTemperatureUnit;;
-	"-f" | "fahrenheit" ) unit="F"; SetTemperatureUnit;;
-	"-k" | "kelvin"     ) unit="K"; SetTemperatureUnit;;
-	"setdns"            ) SetDNSServers;;
-	"setexcludedomains" ) SetExcludeDomains;;
-	"setexcludeclients" ) SetExcludeClients;;
-	"reboot"            ) Reboot;;
-	"restartdns"        ) RestartDNS;;
-	"setquerylog"       ) SetQueryLogOptions;;
-	"enabledhcp"        ) EnableDHCP;;
-	"disabledhcp"       ) DisableDHCP;;
-	"layout"            ) SetWebUILayout;;
-	"-h" | "--help"     ) helpFunc;;
-	"domainname"        ) SetDHCPDomainName;;
-	"privacymode"       ) SetPrivacyMode;;
-	"resolve"           ) ResolutionSettings;;
-	*                   ) helpFunc;;
-esac
+main() {
 
-shift
+	args=("$@")
 
-if [[ $# = 0 ]]; then
-	helpFunc
-fi
+	case "${args[1]}" in
+		"-p" | "password"   ) SetWebPassword;;
+		"-c" | "celsius"    ) unit="C"; SetTemperatureUnit;;
+		"-f" | "fahrenheit" ) unit="F"; SetTemperatureUnit;;
+		"-k" | "kelvin"     ) unit="K"; SetTemperatureUnit;;
+		"setdns"            ) SetDNSServers;;
+		"setexcludedomains" ) SetExcludeDomains;;
+		"setexcludeclients" ) SetExcludeClients;;
+		"reboot"            ) Reboot;;
+		"restartdns"        ) RestartDNS;;
+		"setquerylog"       ) SetQueryLogOptions;;
+		"enabledhcp"        ) EnableDHCP;;
+		"disabledhcp"       ) DisableDHCP;;
+		"layout"            ) SetWebUILayout;;
+		"-h" | "--help"     ) helpFunc;;
+		"domainname"        ) SetDHCPDomainName;;
+		"privacymode"       ) SetPrivacyMode;;
+		"resolve"           ) ResolutionSettings;;
+		*                   ) helpFunc;;
+	esac
 
+	shift
+
+	if [[ $# = 0 ]]; then
+		helpFunc
+	fi
+
+}
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -189,10 +189,15 @@ ProcessDHCPSettings() {
 		interface="eth0"
 	fi
 
+	if [[ "${PIHOLE_DOMAIN}" == "0" ]]; then
+		PIHOLE_DOMAIN="local"
+		change_setting "PIHOLE_DOMAIN" "${PIHOLE_DOMAIN}"
+	fi
+
 	if [[ "${DHCP_LEASETIME}" == "0" ]]; then
 		leasetime="infinite"
 	elif [[ "${DHCP_LEASETIME}" == "" ]]; then
-		leasetime="24"
+		leasetime="24h"
 		change_setting "DHCP_LEASETIME" "${leasetime}"
 	else
 		leasetime="${DHCP_LEASETIME}h"
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -191,6 +191,9 @@ ProcessDHCPSettings() {
 
 	if [[ "${DHCP_LEASETIME}" == "0" ]]; then
 		leasetime="infinite"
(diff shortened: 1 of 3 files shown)


darkrain-nl:
Is it possible/a good idea to make the type of DHCPv6 configurable? Now it is always DHCPv6 while Android does not support that, for Android SLAAC is needed. So it would be nice if we could choose between SLAAC and DHCPv6...

My configuration for SLAAC looks like this b.t.w.:

`# Do stateless DHCP, SLAAC, and generate DNS names for SLAAC addresses
# from DHCPv4 leases.
dhcp-range=2001:here my prefix::, ra-stateless, ra-names
ra-param=*,0,0
dhcp-option=option6:dns-server,[::]`

This works on Android, Debian Jessie and Windows. One thing is not working without a 3rd party solution and that's RDNSS on Windows. But that is no big problem as ipv6 DNS request can go via ipv4 DNS servers...

DL6ER:
@darkrain-nl I'm completely open for suggestions concerning our implementation of IPv6.
@dschaper might also want to comment on this.

dschaper:
screenshot from 2016-12-28 10-03-03

dschaper:
Although generating a SLAAC from the IPv4 (privacy extentions) would be a better approach than MAC'ing it, but I think that's set at the client stack level and not at the DHCP level. LEDE(OpenWRT) allows for a hint in hexadecimal which is the last `n` bits of the host portion of the address. So I can add a hint of `0000 000A` and that's either used directly as the LSB of the host, or masked on to the LSBs of the host.

darkrain-nl:
@dschaper yes, you are correct about the privacy extensions. That is a client level configuration, Windows enables privacy extensions by default. 
I'm not sure if you can configure dnsmasq to be stateless + stateful? That would be great, because then you don't have to make a choice anymore...

dschaper:
I'll have to pull out the configs and see what comes up, I think odhcpd is used on the router solutions, but if there's a way to do it in `dnsmasq` we should be able to find it. (The documentation is quite extensive actually.)

darkrain-nl:
If you can find it, it just means I wasn't searching good engough ;-)
I did search for this some months ago...

dschaper:
Night time reading material http://www.thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html

darkrain-nl:
@dschaper thanks I was aware of the manual.

dschaper:
That was a post to remind me of where I needed to start looking, and I'll check that WP site as well. Thanks!

darkrain-nl:
Ok I did some testing and found out it is possible to do stateful DHCPv6 and SLAAC.

My new configuration:
`dhcp-range=::2,::500,constructor:eth0,ra-names,slaac

ra-param=*,0,0
dhcp-option=option6:dns-server,[::]`

This works on both Android and Windows. No need for any 3rd party applications anymore.

DL6ER:
If you find something that will make all of us happy and both of you agree that this is good, then I'll just put it in. The less settings we expose the our users the better their experience. And I have to do less programming on the Settings page ;-)

dschaper:
If it does both stateful and stateless, then bases are covered, and the clients can battle it out for addressing.

DL6ER:
Is there something where we can be sure that they catch the correct IPv6 DNS server? On my Ubuntu system at home, sometimes the automatic IPv6 config finds the router (although it shouldn't do anything concerning DHCP, etc.) and sometimes the Pi.

dschaper:
I think that's more in the RA portion. (Admittedly I haven't gone to deep into router advertisements and payloads...) But `option6` is the DNS server field for DHCP so even if they go SLAAC, they should be able to get that information.

darkrain-nl:
@DL6ER how is your router configured? In mine it is possible to just do SLAAC + stateless DHCPv6 which means it only advertises the default gateway for the network and it handles the SLAAC portion (Android devices in my case) It doesn't send the option6 field.
Also what you're describing is a problem only for Linux based systems, Windows doesn't have that problem cause Windows does nothing with DNS servers from Router Advertising.

DL6ER:
Well, my router is the cheap one I got for free. No settings at all concerning IPv6. The only thing I can change is the IPv4 DHCP server (which is disabled).

darkrain-nl:
That's a shame, cause what happens now is that both dnsmasq and your router do Router Advertisements. This is somewhat a good thing, because you need the Router Advertisement from the router for your default gateway. But on the other hand you don't want it to do something with the DNS server which should be coming from dnsmasq.
I'm not sure how to prevent this from happening other then buying a new router ;-)

DL6ER:
Are we good to go now with the current version of this PR?

dschaper:
We'll have to see if the others can help test this out. I have a 5 vLAN configuration that I'd rather not break down quite yet. I'll see if I can set up something smaller and sniff a few packets.

darkrain-nl:
I'm not sure if I look correct, but is the IPv4 DNS server address (the pi-hole) also distributed by DHCP?
Something like `dhcp-option=6,0.0.0.0` 

Another thing in the future could be the option to create static leases based on MAC address or hostname, but that can be something for later..

DL6ER:
> Another thing in the future could be the option to create static leases based on MAC address or hostname, but that can be something for later..

We almost have this. You can simply edit the `dhcp.leases` file.

> I'm not sure if I look correct, but is the IPv4 DNS server address (the pi-hole) also distributed by DHCP?
Something like dhcp-option=6,0.0.0.0

The DHCP server of `dnsmasq` defaults to itself as DNS server. Use the IPv4 part already with great success at home and I know of many others where it works just fine.

darkrain-nl:
> The DHCP server of dnsmasq defaults to itself as DNS server. Use the IPv4 part already with great success at home and I know of many others where it works just fine.

I did not know this.
I did know I can just edit the dhcp.leases file, or add some dhcp-host lines in the dhcp config file.

Looks good from my side.

darkrain-nl:
One more question, why use an infinite lease time?
I think it's better to use something like 12 or 24 hours, if a device leaves the network I think it's nice it cleans up after some time...

DL6ER:
I can only speak for myself but I like to know that the IPs in my network don't change. Imagine I'm away for a week and switch on my devices afterwards. Would like them to be reachable under the same address. I doubt that you would normally see enough devices in a net to outrun the /24 range

dschaper:
My vote is for user configurable. I normally run 7 day leases, but if I know there's going to be some changes in the infrastructure, being able to step down the lease time to a short half-life is nice, then I can change things out and set the lease back up to a high level. I also have guests on net that I don't need hanging around so I don't need them to be perma-leases. I'd rather have my servers on static/reserved lease and then the transient devices on a timed lease. I don't know if a client would attempt to rebind if given an infinite lease, there's no real TTL for it to calculate a time period to send out a rebind request?

darkrain-nl:
> I can only speak for myself but I like to know that the IPs in my network don't change. Imagine I'm away for a week and switch on my devices afterwards. Would like them to be reachable under the same address.
That's what DNS is for in my opinion.

> I don't know if a client would attempt to rebind if given an infinite lease, there's no real TTL for it to calculate a time period to send out a rebind request?
I know the DHCP server sends out infinite as lease time, but I don't know how a client behaves with this.

My vote is also for user configurable.

advanced/Scripts/webpage.sh
```diff
@@ -149,123 +173,136 @@ RestartDNS(){
...
+dhcp-authoritative
+dhcp-range=${DHCP_START},${DHCP_END},infinite
+dhcp-option=option:router,${DHCP_ROUTER}
+dhcp-leasefile=/etc/pihole/dhcp.leases
+domain=${PIHOLE_DOMAIN}
+quiet-dhcp
+quiet-dhcp6
+#enable-ra
+dhcp-option=option6:dns-server,[::]
+dhcp-range=::100,::1ff,constructor:${interface},ra-names,slaac
```
darkrain-nl: Can you please add the lease time for dhcpv6 also? It seems to default to 1 hour which is a bit fast in my opinion...
Maybe make it configurable for both dhcpv4 and dhcpv6.

DL6ER:
From the `dnsmasq` man page:

> When replying with information from /etc/hosts or configuration or the DHCP leases file dnsmasq by default sets the time-to-live field to zero, meaning that the requester should not itself cache the information. This is the correct thing to do in almost all situations.

advanced/Scripts/webpage.sh
DL6ER: From the `dnsmasq` man page:
> When replying with information from /etc/hosts or configuration or the DHCP leases file dnsmasq by default sets the time-to-live field to zero, meaning that the requester should not itself cache the information. This is the correct thing to do in almost all situations.

darkrain-nl:
OK, I see now that it doesn't make a difference in relation to how often a client requests a rebind.
So the default should be sufficient.

DL6ER:
Okay, I added the lease time to the settings page (see PR https://github.com/pi-hole/AdminLTE/pull/309)

I managed to apply it for IPv4 + IPv6:
```
Dec 29 14:24:51 dnsmasq-dhcp[32081]: DHCP, IP range 10.1.0.2 -- 10.1.9.2, lease time 1d
Dec 29 14:24:51 dnsmasq-dhcp[32081]: DHCPv6, IP range ::100 -- ::1ff, lease time 1d, template for eth0
Dec 29 14:24:51 dnsmasq-dhcp[32081]: DHCPv4-derived IPv6 names on eth0
Dec 29 14:24:51 dnsmasq-dhcp[32081]: router advertisement on eth0
```

DL6ER:
Enabled the DHCP logging by default. Not sure if we want to do this, but we can discuss. I don't want to add another setting for this on the Settings page...

darkrain-nl:
> Enabled the DHCP logging by default. Not sure if we want to do this, but we can discuss. I don't want to add another setting for this on the Settings page...

I like it as you know, so my vote goes to enable.

darkrain-nl:
I just found out I had no IPv6 DNS server address configured on my Windows 10 machine, at first I thought that there was a problem with dnsmasq, but it seems Microsoft broke DHCPv6 after their Anniversary Update... https://social.technet.microsoft.com/Forums/windows/en-US/b16e7d78-e390-4ada-a24b-3ccba60fa571/no-ipv6-dns-statelessdhcp-since-windows-10-anniversary-update?forum=win10itpronetworking&prof=required

I'm posting this just so you know.

PromoFaux:
Approved

PR #1074, opened 2017-01-01
[RELEASE] Pi-hole Core 2.11

PromoFaux:
Staging


jpgpi250:
@dschaper may i ask why you removed the QuidsUp tracker list?

DL6ER:
It blocks some domains which appear to be of value for some users (e.g. `support.microsoft.com`). For such domains it is not clear if they do any tracking or ad serving, but many users running Win or using their video game console seem to have several problems:
- They cannot install updates
- The OS tells them that internet is not working for them (although it does, but the domain MS uses for checking connectivity is blocked)
- They cannot so certain things on the video game consoles

Since we don't want to preset a whitelist for all users we decided to remote the QuidsUp list to improve the experience for our general user. In the next (not this) release we'll have added wildcard blacklisting support which can be used to block `microsoft.com` + all subdomains if you want to do this.

Also we will add a web interface for customizing the ad lists very conveniently.

dschaper:
Approved

jacobsalmela:
Approved

Mcat12:
Approved

PR #1174, opened 2017-01-27
Improve logrotate support

DL6ER:

---

When Pi-hole tries to rotate its logs on Ubuntu, the following message appears:

```
error: skipping "/var/log/pihole.log" because parent directory has insecure permissions
(It's world writable or writable by group which is not "root") Set "su" directive in
config file to tell logrotate which user/group should be used for rotation.
```
Different operating systems have different user / group settings for `logrotate` that makes it impossible to create a static `logrotate` file that will work with e.g. Raspbian and Ubuntu at the same time. Hence, we have to customize the `logrotate` script dynamically in order to reflect the local properties of the /var/log directory.

The reason why this only affects our `logrotate` config and not others is that we explicitly call our config file. In this case, `logrotate` skips reading the global config file and misses this critical information.

On Raspbian this adds
```
su root root
```
to our `logrotate` file (`pihole flush` is working fine)

On Ubuntu this adds
```
su root syslog
```
to our `logrotate` file (`pihole flush` is working fine)

See also [this](https://bugs.launchpad.net/ubuntu/+source/logrotate/+bug/1278193) bug report ob Launchpad.

### DIFF
--- a/advanced/Scripts/piholeLogFlush.sh
+++ b/advanced/Scripts/piholeLogFlush.sh
@@ -13,7 +13,7 @@
 echo -n "::: Flushing /var/log/pihole.log ..."
 # Test if logrotate is available on this system
 if command -v /usr/sbin/logrotate &> /dev/null; then
-  /usr/sbin/logrotate --force /etc/.pihole/advanced/logrotate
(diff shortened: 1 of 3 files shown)


dschaper:
Approved

PR #1177, opened 2017-01-27
Fix CLI domain validation

DL6ER:
Internal.

Here string redirection is not "binary safe" and Bash will perform expansion on the here string. That is not good.


Mcat12:
Approved

DL6ER:
@dschaper Is `grep -P` available on all platforms we support? Might be GNU `grep`. We could probably switch to `sed` if support is not guaranteed on all platforms, but that would need some more work then.

PR #1178, opened 2017-01-27
Use perl instead of grep for domain validation

DL6ER:
Internal


Mcat12:
Returns `example.` for `example.com`

DL6ER:
We don't use the returned value. Just look if anything is getting returned.

DL6ER:
@Mcat12 Please review again

dschaper:
Approved

PR #1216, opened 2017-02-04
elevate permissions and execute this local script 

stevenharradine:
instead of downloading from the internet

_1_

---
_sudo bash basic-install.sh "$@"_

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1095,7 +1095,11 @@ main() {
 
     if command -v sudo &> /dev/null; then
       echo "::: Utility sudo located."
-      sudo bash basic-install.sh "$@"
+      if [[ -a basic-install.sh ]]; then
+        sudo bash basic-install.sh "$@"
+      else
+        exec curl -sSL https://raw.githubusercontent.com/pi-hole/pi-hole/master/automated%20install/basic-install.sh | sudo bash "$@"
+      fi
       exit $?
     else
       echo "::: sudo is needed for the Web interface to run pihole commands.  Please run this script as root and it will be automatically installed."
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1095,7 +1095,7 @@ main() {
 
     if command -v sudo &> /dev/null; then
       echo "::: Utility sudo located."
-      exec curl -sSL https://raw.githubusercontent.com/pi-hole/pi-hole/master/automated%20install/basic-install.sh | sudo bash "$@"
+      sudo bash basic-install.sh "$@"
       exit $?
     else
       echo "::: sudo is needed for the Web interface to run pihole commands.  Please run this script as root and it will be automatically installed."


Mcat12:
`basic_install.sh` is only there if you download the script first.

dschaper:
Yeah, `curl |bash` doesn't leave anything to re-run. If you can find another solution, we'd be very glad to see it.

Declined

dschaper:
Denied

dschaper:
What's the PA NAck code??

dschaper:
Rejected   ?

stevenharradine:
I added a test if the file exists to run it locally.

it caught me off guard when it was redownloading the script from the internet after I had downloaded it and was editing it.  Just running with sudo is probably a good work around, but if someones intentionally downloaded it i don't think they expect the script to download itself again and run that with elevated permissions.

dschaper:
What do you think about `curl -sSL https://install.pi-hole.net  -o basic-install.sh; sudo bash basic-install.sh` as the one-line installer method? That would be very explicit in what it's going to do. We do put a very large emphasis on asking the users to read through the code before executing, but that advice is sometimes at a level higher than our users are comfortable with. 

I'm going through the code right now with the assisstance of the rest of the development team to try to reduce or elmiminate the need for elevate priv's and only call for superuser if required. (Daemon install, any function that needs ports in the <1024 protected range, etc.)

We've been looking for ways to reduce our security footprint, and I really do like the idea, I'm just not sure the best way to implement the whole process. (And bash -x would just leave most users running in abject terror :) )

stevenharradine:
just ran some tests it appears that will still overwrite the file and run it.  there does not appear to be flags to not overwrite with curl.  wget however has --no-clobber but im unsure of the reasoning for using curl here in the first place (but i see wget installed in the script elsewhere)

wget --no-clobber https://raw.githubusercontent.com/pi-hole/pi-hole/master/automated%20install/basic-install.sh; sudo bash basic-install.sh "$@"

dschaper:
The script is `curl`ed to bash to start the install. If `main()` is called as an unprivileged user, then we have to restart the script as `sudo` since we install packages and affect changes to the networking stack. Due to the fact that the script does not exist anywhere as it is piped directly to bash, the only way to call the script a second time with `sudo` is to curl the script a second time as the code does now.

Options would be download the script and not pipe anymore, but my concerns are already in this thread about the issues with that. 

I'm not sure what the difference in `curl` or `wget` would be, and the `raw.githubusercontent.com` address is `install.pi-hole.net`, I just changed the address in the script on the secondary curl to be more specific that we aren't pulling in backdoor code.

I also think we can get rid of the `$@` trailer, as the install is called with no arguments, it doesn't have the code necessary to handle any passed arguments.

dschaper:
Ah, I understand now the basis for the PR. As this would just be an edge case for Ansible I'll stay with my initial declined comment. This opens too many holes that we've spent time to make sure are closed.

dschaper:
After some internal communication, some ideas. What is it that is being changed in the `basic-install.sh` file that needs customization? If we offered the ability to do something along the lines of a preseed file where you could answer the installer questions and set things like file installation paths and the installer would use the preseed declarations, would that help? So you could `curl` the file but still have customization of the process?

dschaper:
Re-opening for further discussion on the issue, feel free to unfollow the issue if you'd like to not be notified of our conversations.

dschaper:
No discussion ensued. Closing.

PR #1223, opened 2017-02-06
FTL integration

DL6ER:

---

Clone, compile, and install `FTL`.

A log file is created:
```
pi@eden: ~# less /etc/pihole/FTL_build.log 
rm -f obj/*.o pihole-FTL
Making FTL version on branch master - 75a3085 (2017-02-06 12:36:02 +0100)
gcc -c -o obj/main.o main.c -I. -Wall -g
gcc -c -o obj/structs.o structs.c -I. -Wall -g
gcc -c -o obj/log.o log.c -I. -Wall -g
gcc -c -o obj/daemon.o daemon.c -I. -Wall -g
gcc -c -o obj/parser.o parser.c -I. -Wall -g
gcc -c -o obj/signals.o signals.c -I. -Wall -g
gcc -c -o obj/socket.o socket.c -I. -Wall -g
gcc -c -o obj/request.o request.c -I. -Wall -g
gcc -c -o obj/grep.o grep.c -I. -Wall -g
gcc -c -o obj/setupVars.o setupVars.c -I. -Wall -g
gcc -o pihole-FTL obj/main.o obj/structs.o obj/log.o obj/daemon.o obj/parser.o obj/signals.o obj/socket.o obj/request.o obj/grep.o obj/setupVars.o -I. -Wall -g 
install -m 0755 pihole-FTL /usr/local/bin
touch /var/log/pihole-FTL.log /var/run/pihole-FTL.pid /var/run/pihole-FTL.port
chmod 0666 /var/log/pihole-FTL.log /var/run/pihole-FTL.pid /var/run/pihole-FTL.port
```

### DIFF
--- a/advanced/Scripts/update.sh
+++ b/advanced/Scripts/update.sh
@@ -89,7 +89,6 @@ main() {
   fi
 
   if ! is_repo "${FTL_LOCAL_REPO}"; then
-    echo "::: FTL repo not present on system"
     getGitFiles ${FTL_LOCAL_REPO} ${FTL_GIT_URL} || \
     { echo "!!! Unable to clone ${FTL_GIT_URL} into ${FTL_LOCAL_REPO}, unable to continue."; \
         exit 1; \
@@ -116,10 +115,10 @@ main() {
 
   if GitCheckUpdateAvail "${FTL_LOCAL_REPO}" ; then
     web_update=true
-    echo "::: FTL:  update available"
+    echo "::: FTL:            update available"
   else
     web_update=false
-    echo "::: FTL:  up to date"
+    echo "::: FTL:            up to date"
   fi
 
   # Logic
--- a/advanced/Scripts/update.sh
+++ b/advanced/Scripts/update.sh
@@ -18,6 +18,8 @@ readonly ADMIN_INTERFACE_GIT_URL="https://github.com/pi-hole/AdminLTE.git"
 readonly ADMIN_INTERFACE_DIR="/var/www/html/admin"
 readonly PI_HOLE_GIT_URL="https://github.com/pi-hole/pi-hole.git"
 readonly PI_HOLE_FILES_DIR="/etc/.pihole"
+# FTL_GIT_URL sourced from basic-install.sh
+# FTL_LOCAL_REPO sourced from basic-install.sh
 
 PH_TEST=true
 source ${PI_HOLE_FILES_DIR}/automated\ install/basic-install.sh
@@ -60,7 +62,7 @@ GitCheckUpdateAvail() {
     git status
     exit
   fi
-  
+
   # Change back to original directory
   cd "${curdir}"
 
@@ -86,6 +88,14 @@ main() {
     exit 1;
   fi
 
+  if ! is_repo "${FTL_LOCAL_REPO}"; then
+    echo "::: FTL repo not present on system"
+    getGitFiles ${FTL_LOCAL_REPO} ${FTL_GIT_URL} || \
+    { echo "!!! Unable to clone ${FTL_GIT_URL} into ${FTL_LOCAL_REPO}, unable to continue."; \
+        exit 1; \
+    }
+  fi
+
   echo "::: Checking for updates..."
 
   if GitCheckUpdateAvail "${PI_HOLE_FILES_DIR}" ; then
@@ -102,6 +112,14 @@ main() {
   else
     web_update=false
     echo "::: Web Interface:  up to date"
+  fi
+
+  if GitCheckUpdateAvail "${FTL_LOCAL_REPO}" ; then
+    web_update=true
+    echo "::: FTL:  update available"
+  else
+    web_update=false
+    echo "::: FTL:  up to date"
   fi
 
   # Logic
--- a/advanced/Scripts/update.sh
+++ b/advanced/Scripts/update.sh
@@ -114,10 +114,10 @@ main() {
   fi
 
   if GitCheckUpdateAvail "${FTL_LOCAL_REPO}" ; then
-    web_update=true
+    FTL_update=true
     echo "::: FTL:            update available"
   else
-    web_update=false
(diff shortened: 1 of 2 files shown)


PromoFaux:
Do we need to add a new dependency?

DL6ER:
Excellent question. Everything should be there, but let me test that assumption on a droplet.

Including `build-essentials` cannot harm, but I don't know if a similar package exists for Fedora.

DL6ER:
@PromoFaux Yes!

---

Installer:
```
::: Compiling FTL... grep: /etc/pihole/setupVars.conf: No such file or directory
```

FTL buildlog
```
cat /etc/pihole/FTL_build.log
./automated install/basic-install.sh: line 973: make: command not found
```

---

```
sudo apt-get install build-essential
```

---

Installer
```
::: Compiling FTL... done
::: Installing FTL... done
```

FTL buildlog
```
rm -f obj/*.o pihole-FTL
Making FTL version on branch master - b01d6b6 (2017-02-06 13:19:17 +0100)
gcc -c -o obj/main.o main.c -I. -Wall -g
gcc -c -o obj/structs.o structs.c -I. -Wall -g
gcc -c -o obj/log.o log.c -I. -Wall -g
gcc -c -o obj/daemon.o daemon.c -I. -Wall -g
gcc -c -o obj/parser.o parser.c -I. -Wall -g
gcc -c -o obj/signals.o signals.c -I. -Wall -g
gcc -c -o obj/socket.o socket.c -I. -Wall -g
gcc -c -o obj/request.o request.c -I. -Wall -g
gcc -c -o obj/grep.o grep.c -I. -Wall -g
gcc -c -o obj/setupVars.o setupVars.c -I. -Wall -g
gcc -c -o obj/args.o args.c -I. -Wall -g
gcc -o pihole-FTL obj/main.o obj/structs.o obj/log.o obj/daemon.o obj/parser.o obj/signals.o obj/socket.o obj/request.o obj/grep.o obj/setupVars.o obj/args.o -I. -Wall -g 
install -m 0755 pihole-FTL /usr/local/bin
touch /var/log/pihole-FTL.log /var/run/pihole-FTL.pid /var/run/pihole-FTL.port
chmod 0666 /var/log/pihole-FTL.log /var/run/pihole-FTL.pid /var/run/pihole-FTL.port
```

diginc:
What are the technical advantages of building this client side vs continuous integration building and uploading it to github releases for download?

> build-essential

Ouch, right in my docker image size!  Here is what build-essentials or g++/make will do to my image size:

* diginc/pi-hole:alpine is 147MB uncompressed (after downloaded from hub)
  * installing g++ and make says it adds : `158 MiB in 27 packages`
  * 107% increase :anguished: 
* diginc/pi-hole:debian is 320MB
  * installing build essential says: `After this operation, 182 MB of additional disk space will be used.`
  * 56% increase :-1:
* average compression of my images by docker hub (soon to be moved to docker store) is around 40%, I don't want to upload there just to find the actual numbers of what the download will be.  The point is my download and uncompress times will increase by a not insignificant number.
* non docker raspberry pis will also have to download and install these requirements, increase the install time and OS footprint there too.

How big is the binary after compilation?  I'm guessing < 1MB.  please please please take the more efficient path here.  Build it once, distribute the built binaries instead.  

If build-essential takes 2 minutes to download + install for the average pi-hole client and a downloading the binary takes 2 seconds, think of that difference multiplied by the number of installs that'll happen...that's a lot of time.

DL6ER:
No, you entirely misunderstood my point here. The reason why we don't want to simply grab a binary is that it does not work - not my unwillingness to do this - I invested over one week investigating how to do this best, because not having to compile on some unknown machine appears best.
We tried the recommended way for building a kernel for Raspberry Pi on an arbitrary Linux machine and that worked for a small test program. I was able to compile the binary using Travis CI and successfully download and run it on the Raspberry Pi.

Then, I enabled the very same procedure for `FTL`. The binaries have been created and uploaded successfully. However, the binary did not work on a Raspberry Pi 1 and Pi Zero, while they worked fine for a Raspberry Pi 3 (note that the cross compiled simple test program ran on all three devices). It entirely failed for my OrangePi Zero and NanoPi NEO. I complied `FTL` locally and the file size of the generated binary differs considerably. Who knows how many different variants are there for which I'd have to generate a customized binary.

The problem extends (I already said that I invested over one week): On the systems where the binary fails, it simply get's killed by `SIGSERV`. `gdb` does not help, since the binary is compiled with a too new version of the crosscompilers. I have found no way to either install the corresponding older version of the crosscompiler on Travis CI, nor to install a much more recent `gdb` on my Raspbian devices. I always end up with dozens of unfulfillable package dependencies, which will likely break the whole system. Hence, I had no possibility to track down *where* the binary fails exactly:
```
(gdb) run
Starting program: /home/pi/pi-hole-daemon/pihole-FTL-arm-linux-gnueabihf 

Program received signal SIGSEGV, Segmentation fault.
0x00058026 in ?? ()
(gdb) where
#0  0x00058026 in ?? ()
#1  0x00008e72 in _start ()
```

Before you say `valgrind`, let me tell you that it is of no help, either. There seems to be a rather severe bug with ARM which is only solved in a very recent version, which I - again - cannot install without tearing the whole system apart (like also including bleeding-edge versions of `libc`). I wouldn't even know if my system would behave the same after doing so many and such drastic changes.

DL6ER:
Interestingly enough, none of the problems mentioned in my previous post, exist when I clone and compile the code directly on the target device. I agree that `build-essential` might not be the way to go, but I came from the point of presenting something that works. It is very likely that we don't need most of the stuff that `build-essential` installs, but I have the feeling that we cannot circumvent the need for locally compiling the source code - we will at least need `gcc`, `make` and a set of `...-dev` packages.

Further details would have to be worked out.

DL6ER:
This code has become entirely obsolete as are now able to provide binaries for all systems we can test.

PR #1265, opened 2017-02-21
Redirect pi.hole to pi.hole/admin/

ProtoFoo:

**How familiar are you with the codebase?:** 5

---
Navigating to "pi.hole" redirects to the admin interface. (See pull request #1241)

I currently use this configuration in my setup and tested it with Firefox 51.0.1 and Internet Explorer 11. Output of `curl -v pi.hole`:
```
* Rebuilt URL to: pi.hole/
[...]
* Connected to pi.hole (192.168.1.2) port 80 (#0)
> GET / HTTP/1.1
> User-Agent: curl/7.34.0
> Host: pi.hole
> Accept: */*
>
< HTTP/1.1 301 Moved Permanently
< Location: /admin/
< Content-Length: 0
< Date: Tue, 21 Feb 2017 19:49:29 GMT
* Server lighttpd/1.4.35 is not blacklisted
< Server: lighttpd/1.4.35
<
* Connection #0 to host pi.hole left intact
```

### DIFF
--- a/advanced/lighttpd.conf.debian
+++ b/advanced/lighttpd.conf.debian
@@ -73,7 +73,6 @@ $HTTP["url"] =~ "^(?!/admin)/.*" {
 # Entering just "pi.hole" into a browser redirects to "pi.hole/admin/"
 $HTTP["host"] == "pi.hole" {
     $HTTP["url"] == "/" {
-        setenv.add-response-header = ( "X-Pi-hole" => "Redirecting pi.hole to /admin/." )
         url.redirect = ( "" => "/admin/" )
     }
 }
--- a/advanced/lighttpd.conf.debian
+++ b/advanced/lighttpd.conf.debian
@@ -70,5 +70,13 @@ $HTTP["url"] =~ "^(?!/admin)/.*" {
     setenv.add-response-header = ( "X-Pi-hole" => "A black hole for Internet advertisements." )
 }
 
+# Entering just "pi.hole" into a browser redirects to "pi.hole/admin/"
+$HTTP["host"] == "pi.hole" {
+    $HTTP["url"] == "/" {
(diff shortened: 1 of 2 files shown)


PromoFaux:
We were trying to figure this out the other day! (You may have seen)

I take it this does not cause issues when navigating directly to `pi.hole/admin`? That's where we were falling over

jacobsalmela:
Nice!  I'll have to try this out.  I was banging my head over the regex, but this looks like it could work

ProtoFoo:
@PromoFaux Yes, I saw the changes and was happy about a good new feature until this was reverted. I testet the admin interface intensly and I cannot see anything being broken. The new check only triggers if the URL is exactly "/". (Which is the default if only the hostname was provided.)

@jacobsalmela No need for a RegEx in this one. Just two simple string comparisons. You also tried to rewrite the URL (instead of redirecting the browser). Rewriting the URL would mean to hide the "/admin" part from the browser, which in my opinion is not needed here.

Just now I saw that Lighttpd seems to ignore the additional response header when redirecting. So the line
`setenv.add-response-header = ( "X-Pi-hole" => "Redirecting pi.hole to /admin." )`
does nothing. Feel free to remove it. Or do you want me to add two more commits to this PR?

jacobsalmela:
I recall seeing a bug with `lighttpd` where some nested conditionals did not work properly...

Nonetheless, if they ever fix it, it should just start working.  

I also tested out your pull and it seems to work for me--it doesn't cause issues like my pull did.

PromoFaux:
Tested working OK.

Approved.

@ProtoFoo Thanks for the submission!

jacobsalmela:
Approved

PromoFaux:
Approved(again!)

jacobsalmela:
Approved

PR #1266, opened 2017-02-22
Add host-record option

DL6ER:

---

Implementation of a feature request on [Discourse](https://discourse.pi-hole.net/t/domain-redirect/1826/) .

Curent behavior: The Android OpenVPN client is not able to connect to the VPN server when the Android device is within the internal network.

```plain
pi@raspberrypi:~# pihole -a hostrecord home.mydomain.de 192.168.1.10
Setting host record for home.mydomain.de -> 192.168.1.10

pi@raspberrypi:~# grep 'home' /etc/dnsmasq.d/01-pihole.conf
host-record=home.mydomain.de,192.168.1.10
```
Afterwards: The Android OpenVPN client is able to connect to the VPN server and shows its internal IP (`192.168.1.10`).

---

This command is IPv6-aware. Use:

```plain
pi@raspberrypi:~# pihole -a hostrecord home.mydomain.de 192.168.1.10,2003:89:xxxx:xxxx::xxx
Setting host record for home.mydomain.de -> 192.168.1.10,2003:89:xxxx:xxxx::xxx
```

`dig aaaa home.mydomain.de` will return the IPv6 address on the corresponding device in the internal network. Note that should not be necessary when using the proper global scope IPv6 address.

---

It is also possible to remove the configuration by not specifying any host/IP:
```plain
pi@raspberrypi:~# pihole -a hostrecord
Removing host record

pi@raspberrypi:~# grep 'home' /etc/dnsmasq.d/01-pihole.conf
```

Result: The configuration has been removed. The OpenVPN server is, again, not able to connect.

---

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -127,6 +127,12 @@ trust-anchor=.,19036,8,2,49AAC11D7B6F6446702E54A1607371607A1A41855200FD2CE1CDDE3
 " >> "${dnsmasqconfig}"
 	fi
 
+	delete_dnsmasq_setting "host-record"
+
+	if [ -n "${#HOSTRECORD}" ]; then
+		add_dnsmasq_setting "host-record" "${HOSTRECORD}"
+	fi
+
 }
 
 SetDNSServers(){
@@ -313,6 +319,7 @@ ResolutionSettings() {
 	elif [[ "${typ}" == "clients" ]]; then
 		change_setting "API_GET_CLIENT_HOSTNAME" "${state}"
 	fi
+
 }
 
 AddDHCPStaticAddress() {
@@ -331,12 +338,30 @@ AddDHCPStaticAddress() {
 		# Full info given
 		echo "dhcp-host=${mac},${ip},${host}" >> "${dhcpstaticconfig}"
 	fi
+
 }
 
 RemoveDHCPStaticAddress() {
 
 	mac="${args[2]}"
 	sed -i "/dhcp-host=${mac}.*/d" "${dhcpstaticconfig}"
+
+}
+
+SetHostRecord(){
+
+	if [ -n "${args[3]}" ]; then
+		change_setting "HOSTRECORD" "${args[2]},${args[3]}"
+		echo "Setting host record for ${args[2]} -> ${args[3]}"
+	else
+		change_setting "HOSTRECORD" ""
+		echo "Removing host record"
+	fi
+
+	ProcessDNSSettings
+
+	# Restart dnsmasq to load new configuration
+	RestartDNS
 
 }
 
@@ -363,6 +388,7 @@ main() {
 		"resolve"           ) ResolutionSettings;;
 		"addstaticdhcp"     ) AddDHCPStaticAddress;;
 		"removestaticdhcp"  ) RemoveDHCPStaticAddress;;
+		"hostrecord"        ) SetHostRecord;;
 		*                   ) helpFunc;;
 	esac
 


jacobsalmela:
Approved

cusco:
Just a question: this allows only one custom dns entrie??

autumnwalker:
Wondering the same thing - can this be used to add multiple host records?

mikegleasonjr:
@cusco @autumnwalker only one, just tested it 😢

mikegleasonjr:
For those wondering how to add multiple entries, add your entries to `/etc/hosts` then `$ pihole restartdns`

cusco:
What I did, in /etc/dnsmasq.conf I added a line:

conf-file=/etc/dnsmasq.more.conf

and in this file I added lines such as:
host-record=hostname.com,192.168.1.1

cby016:
This feature is not very useful if it only works for one host entry. I went through and added about 10 entries and then tested and only the last one was working. When I checked the config files only the last one had been saved.

cusco:
Exactly!!

I found a workaround but this is for dnsmasq only. No pihole interaction.

Include a extra configuration file with several lines.

cby016:
Yeah I ended up doing this https://discourse.pi-hole.net/t/howto-using-pi-hole-as-lan-dns-server/533

joshbgosh10592:
I'd really, really like the ability to add multiple host records... One reason is that I frequently call my Windows server something else, depending on what I'm working on at that moment (Apollo/media/share) and often forget that it's name is actually Apollo. I used to have a host record on my router's DNS to make it all point to the same IP.

Second reason is that for some reason, Pi-hole doesn't always resolve my hostnames. My MacBook Pro, for example, shows as the IP address only, and no client on my network can resolve to it. I want to be able to manually assign it.

Mcat12:
Add extra records to `/etc/hosts`. Make sure you only have one DNS server sent to your clients. If you have other questions, make a new issue.

mikegleasonjr:
I would remove the feature then to avoid confusion.

DL6ER:
Well, @mikegleasonjr you can read in the original PR description for what users are using this feature. I know of at least five people using it exactly for this purpose (including me). Everything else should be done as @Mcat12 suggested: In early times, all name resolution was done by having up-to-date `hosts` files on the servers. As the Internet grew, the DNS service was introduced, eventually. Our DNS resolver (the current and the new one) both hand out all knowledge you specify in this one single place (the `/etc/hosts`) file and this is the place where any non-trivial configuration should be installed.

xenithorb:
This option makes zero sense if it's only for one record

gmask:
Yeah I don't understand, does anybody who needs a custom DNS entry for local IPs only need one?

pralor:

https://discourse.pi-hole.net/t/can-i-use-pi-hole-dns-as-a-online-hosts-manager/25504/1

pralor:

https://discourse.pi-hole.net/t/request-dns-rewrites-planned/27350/11

pralor:

https://discourse.pi-hole.net/t/request-dns-rewrites-planned/27350/14

DL6ER:
We're sorry guys, this feature got down-voted very often and we missed it. We'll remove it to avoid further confusion.
For users of this feature:
- We will not remove existing host records.
- The Custom DNS page is an adequate replacement. It can be accessed from the dashboard.
- If you cannot use the dashboard, please add your records to `/etc/pihole/custom.list` or `/etc/hosts` and run `pihole restartdns reload` afterwards.

pralor:

https://discourse.pi-hole.net/t/request-dns-rewrites-planned/27350/18

pralor:

https://discourse.pi-hole.net/t/request-dns-rewrites-planned/27350/23

cusco:
> * n be accessed from th

We understand removing a feature that is not serving its purpose. But how about adding a feature that does? Just a custom hosts entry like you mentioned, but manageable on the web gui (and pihole cli)

dschaper:
>     * The Custom DNS page is an adequate replacement. It can be accessed from the dashboard.

Screenshot_2020-05-02 Pi-hole Admin Console

DL6ER:
And to add what @dschaper quoted,

> If you cannot use the dashboard, please add your records to /etc/pihole/custom.list or /etc/hosts and run pihole restartdns reload afterwards.

This is likely even much easier than having to remember the Pi-hole syntax for the same task and will always give you an overview of what is already there.

You'll be asked for the `sudo` password, but you'll also be asked for it when using the `pihole` command (and don't have passwordless `sudo` such as the user `pi` on Raspbian).

Conclusion: Everything is still available. On the dashboard it is even much improved.

Anuskuss:
Late to the party but while I agree that Local DNS Records / `/etc/hosts` / `/etc/pihole/custom.list` are fine for job, it would be great to have the same flexability as white/blacklists but for redirects, mainly regex and groups. I personally have a niche but valid use case for this: I have two identical devices in my network and I'd like to prevent the first one from auto-updating itself. So I'll setup my Pi-hole to redirect to a custom "You're on the latest firmware" page while at the same time allowing the second device to be on the actual latest firmware. That is just not possible to achieve with only one Pi-hole right now.

PR #1282, opened 2017-02-28
Add local-service to dnsmasq's configuration template

DL6ER:

---

Add `local-service` to `dnsmasq`'s configuration template

From `dnsmasq`'s `man` page:

> Accept DNS queries only from hosts whose address is on a local subnet, i.e. a subnet for which an interface exists on the server. This option only has effect if there are no `interface` `except-interface`, `listen-address` or `auth-server` options.

> It is intended to be set as a default on installation, to allow unconfigured installations to be useful but also safe from being used for DNS amplification attacks.

Now that we listen on all devices for easing the setup in combination with local networking and distant networking (via OpenVPN) ( #1273 ), this option should prevent opening security holes on VPS (or similar) setups which lack a proper firewall setup.

### DIFF
--- a/advanced/01-pihole.conf
+++ b/advanced/01-pihole.conf
@@ -40,3 +40,5 @@ log-facility=/var/log/pihole.log
 local-ttl=300
 
 log-async
+
+local-service


DL6ER:
This is also used in OpenWRT routers for some years to not listen on the WAN interface and seems to work properly: https://dev.openwrt.org/changeset/43982

DL6ER:
Replaced by https://github.com/pi-hole/pi-hole/pull/1284 which implements this behavior as a standard

PR #1294, opened 2017-03-04
Tweak DNS debugging 

dschaper:

10

---

-Check localhost, IPv4 address and Google in DNS check. 
-Short check, quieter and we don't need the extra information.

### DIFF
--- a/advanced/Scripts/piholeDebug.sh
+++ b/advanced/Scripts/piholeDebug.sh
@@ -264,9 +264,11 @@ daemon_check() {
 }
 
 testResolver() {
-	header_write "Resolver Functions Check"
   local protocol="${1}"
+  header_write "Resolver Functions Check (IPv${protocol})"
   local IP="${2}"
+  local g_addr
+  local l_addr
   local url
   local testurl
   local localdig
@@ -275,8 +277,10 @@ testResolver() {
 
   if [[ ${protocol} == "6" ]]; then
     g_addr="2001:4860:4860::8888"
+    l_addr="::1"
   else
     g_addr="8.8.8.8"
+    l_addr="127.0.0.1"
   fi
 
 	# Find a blocked url that has not been whitelisted.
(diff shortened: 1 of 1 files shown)


advanced/Scripts/piholeDebug.sh
```diff
@@ -440,6 +398,50 @@ finalWork() {
...
+processor_check || echo "Processor Check soft fail"
+
+ip_check 6 ${IPV6_ADDRESS}
+ip_check 4 ${IPV4_ADDRESS}
+
+daemon_check lighttpd http
+daemon_check dnsmasq domain
+daemon_check pihole-FTL 4711
+checkProcesses
+testResolver 4 "${IPV4_ADDRESS%/*}"
```
DL6ER: No IPv6 test?

DL6ER:
Your debug token is : `syqlklfgs6`

DL6ER:
Approved

advanced/Scripts/piholeDebug.sh
```diff
@@ -264,52 +264,70 @@ daemon_check() {
...
-				done < "${WHITELISTMATCHES}"
-			fi
-		done < "${GRAVITYFILE}"
+	url=$(shuf -n 1 "${GRAVITYFILE}" | awk -F ' ' '{ print $2 }')
+
+	testurl="${url:-doubleclick.com}"
+
+
+	log_write "Resolution of ${testurl} from Pi-hole (${l_addr}):"
+	if localdig=$(dig -"${protocol}" "${testurl}" @${l_addr} +short "${r_type}"); then
```
DL6ER: Is there a specific reason why you test resolution of A only via IPv4 and AAA only via IPv6? They should always work both, regardless of the protocol. This might obscure a problem with resolving A vs. AAAA records. Maybe we should check for both on both protocols.

advanced/Scripts/piholeDebug.sh
```diff
@@ -264,52 +264,70 @@ daemon_check() {
...
-	log_write "Resolution of ${TESTURL} from 8.8.8.8:"
-	REMOTEDIG=$(dig "${TESTURL}" @8.8.8.8)
-	if [[ $? = 0 ]]; then
-		log_write "${REMOTEDIG}"
+	log_write "Resolution of ${testurl} from ${g_addr}:"
+	if remotedig=$(dig -"${protocol}" "${testurl}" @${g_addr} +short "${r_type}"); then
+		log_write "${remotedig:-NXDOMAIN}"
 	else
-		log_write "Failed to resolve ${TESTURL} on 8.8.8.8"
+		log_write "Failed to resolve ${testurl} on upstream server ${g_addr}"
```
DL6ER: Interestingly, I got in my test run:

```
Resolution of suzi-champ.us from 8.8.8.8:
NXDOMAIN
[...]
Resolution of ieissue20.xyz from 2001:4860:4860::8888:
NXDOMAIN
```

My debug token is `1uxvopz5j3`

dschaper: That's a positive result, it's in fact working. 
```
root@vultr:/tmp# dig ieissue20.xyz

; <<>> DiG 9.10.3-P4-Ubuntu <<>> ieissue20.xyz
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 34387
```

advanced/Scripts/piholeDebug.sh
dschaper: We could, but then we'd have to create logic to determine if the IPv4 AAAA failure is because of lack of record, or because of lack of Internet connectivity. Pi-hole blocked domains would have both A and AAAA records in `gravity.list` but only if IPv6 blocking is enabled. If disabled, then `gravity.list` would not have AAAA records. We'd be setting up known fails in those cases.

DL6ER:
Approved

dschaper:
And for the folks watching at home: DL and I talked about this offline, he's not just approving off of my replies here.

PR #1297, opened 2017-03-05
Redirect to admin panel when accessing 'http://pi.hole/'

raincoats:

2

---
If someone tries to access 'http://pi.hole/', it will take them to the "Website blocked" page. Very confusing if you don't know to go to 'http://pi.hole/admin/'. This just redirects them to the admin panel.

### DIFF
--- a/advanced/index.php
+++ b/advanced/index.php
@@ -8,7 +8,7 @@ $serverName = escapeshellcmd($_SERVER['SERVER_NAME']);
 // Let's be nice and redirect them.
 if ($serverName === 'pi.hole')
 {
-    header('HTTP/1.1 302 Found');
+    header('HTTP/1.1 301 Moved Permanently');
     header("Location: /admin/");
 }
 
--- a/advanced/index.php
+++ b/advanced/index.php
@@ -4,6 +4,14 @@
 $uri = escapeshellcmd($_SERVER['REQUEST_URI']);
 $serverName = escapeshellcmd($_SERVER['SERVER_NAME']);
 
+// If the server name is 'pi.hole', it's likely a user trying to get to the admin panel.
+// Let's be nice and redirect them.
+if ($serverName === 'pi.hole')
+{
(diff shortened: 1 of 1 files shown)


dschaper:
I believe that has already been addressed the latest release. Check here and see if this is the same resolution to the problem: https://github.com/pi-hole/pi-hole/blob/master/advanced/lighttpd.conf.debian#L73

raincoats:
Oh well there's my issue, I'm not using the lighttpd that gets installed, I'm using nginx.

DL6ER:
HTTP response code **302 Found** is usually used for pages that have only moved temporarily (like due to maintance). This is not the case as the dashboard will actually always only be available under `admin/`, so I'd suggest response code **301 Moved Permanently**

I remember having read that in Nginx, you should always only accomplish redirects with the built-in `rewrite` directive. A quick look into the `man` page suggests something like

```
server {
    [... something else ...]
    rewrite ^/$ http://pi.hole/admin permanent;
    [... something more ...]
}
```

I cannot test this, since I have no Nginx running, but it should be easy for you to test this yourself.

DL6ER:
This solution should work across a number a different web servers. Please change your code to respond with code 301 (`Moved permanently`) instead of 302 (`Found`), like this:
```
if ($serverName === 'pi.hole')
{
	header("Location: admin/", true, 301);                                     
}
```

raincoats:
You're right, to be honest I never gave the 301/302 codes much thought. I've gone ahead and made the change to my code (I think? Clumsy with git, worse with pull requests).

DL6ER:
Approved

dschaper:
Do we want to merge this now, or wait until after 3.0?

DL6ER:
I vote for merging it now.

dschaper:
Okay, sounds good, I just didn't know if this would conflict with anything that's currently staged.

PR #1299, opened 2017-03-06
Update `apt` lock check function.

dschaper:

10

---

Cloud deployments using unattended-upgrades needs a timer to check for `apt` locking. 

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -95,7 +95,7 @@ if command -v apt-get &> /dev/null; then
 
     for i in $(seq 1 $maxtries);
     do
-      apt-get -y update
+      "${UPDATE_PKG_CACHE}"
       retVal=$?
       # if we find lock condition
       if [[ $retVal -eq 100 ]]; then
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -769,7 +769,6 @@ update_package_cache() {
   if eval "${UPDATE_PKG_CACHE}" &> /dev/null; then
     echo " done!"
   else
-    echo -en "\n!!! ERROR - Unable to update package cache. Please try \"${UPDATE_PKG_CACHE}\""
     return 1
   fi
 }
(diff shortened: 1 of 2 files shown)


automated install/basic-install.sh
```diff
@@ -88,16 +88,27 @@ if command -v apt-get &> /dev/null; then
...
+    local maxtries
+    local retVal
+    maxtries=75
+
+    for i in $(seq 1 $maxtries);
+    do
+      "${UPDATE_PKG_CACHE}"
+      retVal=$?
+      # if we find lock condition
+      if [[ $retVal -eq 100 ]]; then
```
DL6ER: Why no `else` condition? I could image a number of errors that could happen (like one or more mirrors not available, etc.)

automated install/basic-install.sh
```diff
@@ -88,16 +88,27 @@ if command -v apt-get &> /dev/null; then
...
+        sleep 5
+      elif [[ $retVal -eq 0 ]]; then
+        return 0
+      fi
     done
-    # Always return success, since we only return if there is no
-    # lock (anymore)
-    return 0
+    # Counter timed out, and we didn't return already, error
+    return 1
```
DL6ER: I see we wait up to 6 minutes - is this sufficient in case that a huge bunch of security updates is installed on low-performance (possibly including more than one kernel and some `dkms`) are downloaded and compiled/installed? 

From the `man` page:

> apt-get returns zero on normal operation, decimal 100 on error.

Maybe we can still check for the lock and at least tell the user *why* he has to wait?

dschaper: We can push out the time, I've found on Vultr that it fires after one loop anyways, (apparently it only runs an update on boot, and waits till stable before actually doing any security upgrades. Otherwise boot times would be 5 minutes to spin up an instance.) I just don't want to get into a race where we're locking and then `unattended-updates|cloud-init` is updating and we're playing pingpong with the lockfile.

Thats also why I only call the lock test on the first `apt update` and dependency install, there shouldn't be a lock condition on the `PIHOLE_DEPS` installer.

So, sure, timout to 10 minutes, but just not infinite. It's a pain to SSH in to a second term and try to chase down the first shell to `kill -9` it.

dschaper: 10 minutes now set for the clock.

dschaper:
This should already be resolved, the issues with cloud-init spinups can be resolved through other means. Closing.

Issue #830, opened 2016-10-21
Migrate away from DHCPCD5 on non-Berry platforms.


dschaper:
_This template was created based on the work of [`udemy-dl`](https://github.com/nishad/udemy-dl/blob/master/LICENSE)._

**In raising this issue, I confirm the following (please check boxes, eg [X]):**
- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [x] The issue I am reporting can be _replicated_
- [x] The issue I'm reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the codebase?:**
- [x] 10 (very familiar)

---

**[FEATURE REQUEST]:**

Setting static IP's via DHCPCD5 on non-Rasbperry platforms is problematic. We should be stetting static addresses via `/etc/network/interfaces.d` or `/etc/network/interfaces`.

Also should take into consideration the situation where static reservations are set for Pi-hole installs and not touch the networking at all, let the users own DHCP server handle the static lease/reserved IP assignment.

dschaper:
Platform and default manual static configuration software
- Raspbian Jessie: DHCPCD configuration controlled static IP.
- Raspbian Wheezy: EOL - Suggested removal from supported platforms

Fedora-Core:
This works for me,
on a stock Raspian Jessie,
and on a stock Debian Jessie,
(no reservations in DHCP server required)
 =>

[on GitHub : /etc/network/interfaces](https://github.com/Fedora-Core/Pi-hole-adblock-multiple-IP-addresses)

Greetings, F.C.

dschaper:
Thanks F.C., trying to compile a list of OSes and what they use for stock addressing, either DHCP acquired or through static addressing.

dschaper:
High Priority. Affects just about any Debian install that isn't Raspbian, and kills other installs dead pretty quickly. Thanks to the RPF for this abomination of a system, it's appreciated.

brontide:
Why not make user responsible for static IP on non-Pi platforms?

dschaper:
[...]

PR #1306, opened 2017-03-06
[WIP] Detect if this is a Raspbian Jessie install

dschaper:

10

---

Only depend on `dhcpcd` if it's a real Raspbian Jessie installation.


automated install/basic-install.sh
```diff
@@ -98,6 +98,12 @@ if command -v apt-get &> /dev/null; then
     return 0
   }
 
+  source /etc/os-release || true
+    if [[ ${PRETTY_NAME} == "Raspbian GNU/Linux 8 (jessie)" ]]; then
```
DL6ER: Will this finally break support for `wheezy`?

And what about `stretch`? Will it still be using `dhcpcd5`?

dschaper: Wheezy uses `/etc/network/interfaces` and I'm betting that `stretch` will keep `dhcpcd`. The RPF (Raspbian) uses that as an assistant for their GUI configuration of Raspbian, and I don't see them changing Pixel now. 

The other fun fact is that they have released Raspbian Jessie for x86 (Live environment only I believe, I don't think you can use it as a full desktop replacement yet.) Haven't the motivation to boot that thing on any hardware I have here to check it out.

dschaper: I'm almost inclined to just leave that out of the dependencies now, anyone that is installing Raspbian Jessie would already have the package installed, we don't need to add it anymore. Then the `/etc/dhcpcd.conf` file detection would be enough to check if that file needed modification.

dschaper:
Fixes #830

dschaper:
This is turning into a total rewrite of the IPv4 configuration detection and correction of the problems with multiple entries in the `/etc/dhcpcd.conf` file that we've put in there. Might just be something for v3 and just overhaul the whole process.

jacobsalmela:
This should fix #830. Just posting this here for documentation.

unclehowell:
Guys, is there a solution for auto selecting the defaults with the whiptail menu? I want to just deploy this beauty, it works, I trust it, why make users go through manual install

PromoFaux:
@unclehowell our, [Discourse Forums](https://discourse.pi-hole.net), or [Subreddit](https://reddit.com/r/pihole) are a better place for discussion than unrelated issue threads

dschaper:
This should be covered under the new installer plans. I don't see a need to keep this open.

PR #1320, opened 2017-03-13
Use Firewalld --add-service instead of port

georou:

6

---
Hi, Firewalld offers the use of services instead of adding straight ports. It's much more cleaner and more verbose to see "http" and "dns" as allowed rather than port numbers. Port numbers, especially 53 require the user to know it's dns - which is ambiguous for novices.

In any case, if it's an aesthetic choice to use numbers, that's fine. At the very least, I can do a new PR using brace expansion for numbers, as I have here. Rather than duplicating add-port twice.

Cheers

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -965,7 +965,7 @@ configureFirewall() {
     whiptail --title "Firewall in use" --yesno "We have detected a running firewall\n\nPi-hole currently requires HTTP and DNS port access.\n\n\n\nInstall Pi-hole default firewall rules?" ${r} ${c} || \
     { echo -e ":::\n::: Not installing firewall rulesets."; return 0; }
     echo -e ":::\n:::\n Configuring FirewallD for httpd and dnsmasq."
-    firewall-cmd --permanent --add-port=80/tcp --add-port=53/tcp --add-port=53/udp
+    firewall-cmd --permanent --add-service={http,dns}
     firewall-cmd --reload
     return 0
   # Check for proper kernel modules to prevent failure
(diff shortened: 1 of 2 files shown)


dschaper:
First, I must complement you on the fine, fine choice of Avatar. And I'll take a look at the code :)

georou:
Haha, cheers. I'd be lying if I said I also have a replica portrait of Kramer from Seinfeld above my fireplace. Alas, that's a future dream. I do enjoy these portrait styles!

Something I overlooked, brace expansion isn't exactly supported in POSIX bash. To make it more portable, I changed it to just using add service twice. As ugly as it is. But at least the service names are still usable if desired.

Also, I noticed Travis failed and it looks like a line ending error maybe? I checked my autocrlf is on input and cleaned out my existing files, [according to Github.](https://help.github.com/articles/dealing-with-line-endings/#platform-linux) But it seemed nothing needed to change. Sorry, first time doing a PR, unless I've missed something.

dschaper:
I to prefer the brace expansion, but for clarity in the script, the long hand way probably is better.

The Travis failures are because of how the test is written, it's expecting that the call to `firewalld` is formatted with port numbers instead of the standardized names, it's just something that needs to be updated within the test itself, and is not a failure of your code to function. (We can only do so much to mock the commands within a container, and firewalls are not typically built inside containers, the test just checks to see what the string is that's passed to `firewalld` and does a straight string comparison.) 

If you'd like, you can update the file at https://github.com/pi-hole/pi-hole/blob/development/test/test_automated_install.py#L81 and add that to this PR. You would just need to change that line so that it reflects the new commands being sent in place of the port number calls.

georou:
Awesome, Dan. Thanks for pointing that out.

I've changed them as you've said. Hopefully that was correct. Not familiar with Travis CI. I'll need to read up on it further. 

Thanks again!

diginc:
Quick comment / question: is this adding a dependency on /etc/services to translate names to ports and is /etc/services universal across more minimal OS installations?

georou:
That's a good point, diginc. I can't off the top of my head, think of a situation where the services file wouldn't be present. Or the http and dns changed from the standard ports. 

As for dependent, firewalld ships with the http and dns xml files as standard, even on minimal(from what I've seen). If you check inside them, it's referencing port numbers, nothing special: firewall-files: /usr/lib/firewalld/services/http.xml OR firewall-cmd --info-services=http

Pretty much just a more verbose way of doing it. Still a good question, dignic. Cheers!

PR #1338, opened 2017-03-17
updated log rotate to be able to forward logs

jalogisch:

5
---
If you use rsyslog - at least on Jessie PI to forward the logfile to some central location rsyslog does not notice the updated logfile. 

After I had added the restart to the post rotate this is working with no issue. As I did not test if this is working on other type of installation that might only be save for a pi based installation.

### DIFF
--- a/advanced/logrotate
+++ b/advanced/logrotate
@@ -1,10 +1,40 @@
 /var/log/pihole.log {
-	# su #
-	daily
-	copytruncate
-	rotate 5
-	compress
-	delaycompress
-	notifempty
-	nomail
+    # Rotate the log files daily
+    daily
+
+    # Don't return an error if the log file is missing
+    missingok
+
+    # Do not rotate if the log file is empty
+    notifempty
+
+    # Rotate the log file twice before deleting it
+    rotate 2
+
+    # Compress the log file using gzip
+    compress
+
+    # Delays compression when some program cannot be told to immediately close it's logfile
+    delaycompress
+
+    # Include the date in the rotated log file's name
+    dateext
+
+    # The postrotate script will only be run once (after the old
+    #   logs  have  been  compressed), not once for each log which is rotated.
+    sharedscripts
+
+    # The  lines  between 'postrotate' and 'endscript' are executed after the log is rotated
+    postrotate
+        # If dnsmasq is running, kill it using the user defined signal
+        #[ ! -f /var/run/dnsmasq/dnsmasq.pid ] || kill -USR2 cat /var/run/dnsmasq/dnsmasq.pid
+    	# use pihole restart function
+	/usr/local/bin/pihole restartdns
+    endscript
+
+    # Rotate the original file and create the new file with specified permission, user, and group
+    create 0644 dnsmasq root
+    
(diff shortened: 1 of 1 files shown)


advanced/logrotate
```diff
@@ -1,10 +1,40 @@
...
+    # The postrotate script will only be run once (after the old
+    #   logs  have  been  compressed), not once for each log which is rotated.
+    sharedscripts
+
+    # The  lines  between 'postrotate' and 'endscript' are executed after the log is rotated
+    postrotate
+        # If dnsmasq is running, kill it using the user defined signal
+        #[ ! -f /var/run/dnsmasq/dnsmasq.pid ] || kill -USR2 cat /var/run/dnsmasq/dnsmasq.pid
+    	# use pihole restart function
+	/usr/local/bin/pihole restartdns
```
DL6ER: Restarting `dnsmasq` should not be done. The current method works w/o problems for basically all users.

DL6ER:
You could also look at this post here which is weakly related:
https://discourse.pi-hole.net/t/moving-the-pi-hole-log-to-another-location-device/2041

advanced/logrotate
jalogisch: Maybe I'm just holding it wrong:

Read the log file with rsyslog:

```
module(load="imfile" PollingInterval="10")

input(type="imfile"
      File="/var/log/pihole.log"
      StateFile="/var/run/pihole.log.state"
      Tag="pihole"
      Severity="info"
      Facility="local7")
```

And then notice that this file is not picked up after the `echo "" > `.

Maybe it is just my environments where I'm responsible for, but the restart is the only way to have syslog recognize the new added entries.

jalogisch:
thank you @DL6ER for pointing to that post. 

But I just want to forward that log to my central log environment and not move it to another disk/place on the device.

For your [references (lang=de)](https://jalogisch.de/2017/der-eigene-dns-resolver-zuhause/) why I want this.

DL6ER:
> And then notice that this file is not picked up after the echo "" >.
> Maybe it is just my environments where I'm responsible for, but the restart is the only way to have syslog recognize the new added entries.

So the problem is that `rsyslog` is not behaving nice when the log file is flushed. How about restarting this service?

> For your references (lang=de) why I want this.

German is not an issue in our team as we have a native speaker (which is me :smiley: ) on board. We thought about implementing similar statistics like you show with graylog, but we found severe licensing barriers which the others project simply seems to ignore. However, we want to stay out of curt and hence refrained from implementing the nice and fancy IP maps, etc.

jalogisch:
> So the problem is that rsyslog is not behaving nice when the log file is flushed. How about restarting this service?

valid point - will test some modified script before I modify the PR.

DL6ER:
Any progress on this?

jalogisch:
currently I was not able to find a working solution other than restart the DNS. 

I'm in doubt that his can be fixed with the given version on the default raspdian syslog.

dschaper:
Thank you for the submission, I'm closing this for now, but if you find a solution please reopen the PR.

PR #1351, opened 2017-03-23
Firewall config changes

flo80:

2

---
I added firewall rules to reject https (port 443) traffic - this avoids the timeout issues when https pages try to load ads. I also added IPv6 firewall rules - these might not be required (except for https to avoid timeout again) since most people will run dual stack but can't harm.

### DIFF
--- a/advanced/firewall.sh
+++ b/advanced/firewall.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+# Pi-hole: A black hole for Internet advertisements
+# (c) 2017 Pi-hole, LLC (https://pi-hole.net)
+# Network-wide ad blocking via your own hardware.
+#
+# Sets the firewall rules after Pi-hole reboot
+#
+# This file is copyright under the latest version of the EUPL.
+# Please see LICENSE file for your rights under this license.
+
+set -e
+setupVars=/etc/pihole/setupVars.conf
+
+source ${setupVars}
+
+# Allow HTTP and DNS traffic
+if firewall-cmd --state &> /dev/null; then
+    firewall-cmd --permanent --add-service=http --add-service=dns
+    firewall-cmd --reload
+    return 0
+    # Check for proper kernel modules to prevent failure
+elif modinfo ip_tables &> /dev/null && command -v iptables &> /dev/null; then
+    # If chain Policy is not ACCEPT or last Rule is not ACCEPT
+    # then check and insert our Rules above the DROP/REJECT Rule.
+    if iptables -S INPUT | head -n1 | grep -qv '^-P.*ACCEPT$' || iptables -S INPUT | tail -n1 | grep -qv '^-\(A\|P\).*ACCEPT$'; then
+	# Check chain first, otherwise a new rule will duplicate old ones
+	iptables -C INPUT -p tcp -m tcp --dport 80 -j ACCEPT &> /dev/null || iptables -I INPUT 1 -p tcp -m tcp --dport 80 -j ACCEPT
+	iptables -C INPUT -p tcp -m tcp --dport 53 -j ACCEPT &> /dev/null || iptables -I INPUT 1 -p tcp -m tcp --dport 53 -j ACCEPT
+	iptables -C INPUT -p udp -m udp --dport 53 -j ACCEPT &> /dev/null || iptables -I INPUT 1 -p udp -m udp --dport 53 -j ACCEPT
+	# Reject https to avoid timeout issues for blocked https adds
+	iptables -C INPUT -p tcp -m tcp --dport 443 -j REJECT &> /dev/null || iptables -I INPUT 1 -p tcp -m tcp --dport 443 -j REJECT
+    fi
+    
+    if [[ ! -z "${IPV6_ADDRESS}" ]]; then
+	# Configure IPv6 firewall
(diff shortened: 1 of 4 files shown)


flo80:
I now also added a script to re-enable the firewall rules after a reboot (as most systems do not automatically save and restore them). It is only installed during the setup process and called after a reboot by cron.

dschaper:
Sorry, I haven't abandoned this, I've just been hit with one thing after another in real life that's pulled me off of doing a lot of Pi-hole things.

dschaper:
Thanks for including the updated tests for the pull. We'll have more time to investigate this now.

dschaper:
Looks good to me, but I'd like to get @DL6ER to take a look as well.

DL6ER:
I'm very sorry - I haven't seen this PR at all... In the meantime, the firewall rules have been updated (added port rules for `FTL`). Please adjust your code and I'll review it ASAP.

flo80:
FTL firewall changes merged

Mcat12:
Approved. @DL6ER

DL6ER:
See [line 965](https://github.com/pi-hole/pi-hole/pull/1351/files#diff-689793e68cde6270d074695b71c969afR965). Shall we notify the user that these rules will be enabled on every startup and where your script is located so that s/he can change it?. E.g. if you install a VPN on a droplet, you *don't want* port 80, 53, etc. to be open at all after a reboot.

PromoFaux:
Pinging @DL6ER for update

DL6ER:
My concerns are still there. Pinging @dschaper for his comment on this. It would actually make it harder to secure any Internet facing device, as after each reboot, everything will be opened to the public if the users doesn't change the rules.

Maybe it would be better to work with `iptables-restore` and friends?

dschaper:
Persistence should be handled by the user, but looking back at the whole concept, I don't like punching holes in peoples firewalls anymore. Raspbian comes without a firewall enabled, but other OSes may include a firewall and/or SELinux. I really think the entire function should be pulled out and replaced with a dialog that notes which ports need to be opened.

flo80:
@dschaper I think, persistence should follow the same logic like setting firewall rules. If the initial install changes firewall rules, these should also be made persistent. Otherwise, this will be highly confusing for users (i.e., suddenly pi-hole stops working for them if they had a drop rule in place). 

@DL6ER ```iptables-save``` / ```iptables-restore```: in my experience they don't work too well, as the rules are only stored the first time when the package gets installed. Any changes afterwards require manual efforts again (either editing the rules files or overwriting them). Automating this is also dangerous as the user might have changed current rules but does not want to make them persistent. (E.g., in my setup outside traffic except port 53 is not allowed in the default rules though when I update my Raspi I temporarily allow port 80 traffic as well, though don't want this rule to be persistent).   


Generally, I agree, only telling the user which ports should be opened is a good idea. I'm just not sure if the 'regular' pi-hole user does have the Linux knowledge (though as you say, Raspbian has no default rules anyways). I think this is a question of who is the target audience for pi-hole and how much technical / Linux background do they have. 

---
In any case, I though would recommend to keep the 'https reject' rule in all cases (initial install as well as reboot) as it really makes sense (otherwise pi-hole really does not work well with https sites) and should not disrupt anything else.  

Please feel free to push changes or advise how I should modify the PR.

DL6ER:
> the rules are only stored the first time when the package gets installed

Well, I run `sudo ip(6)tables-save > /home/myuser/iptables.save` whenever I think that I need to backup. I have a `@reboot` `cron` job that restored from this file a few seconds after bootup.

> Generally, I agree, only telling the user which ports should be opened is a good idea.

They will not need any firewall at all when they work on a well-shielded device in their home environment. When they install on a public Internet facing server (like one of those cheap VPS services), they this rules won't save them at all (they would make it even worse if the server was securely firewalled before). Users will have to learn that they have use one or the other VPN solution as everything else is either insecure or even much more complicated (talking about more sophisticated e.g. origin-based firewalling).

> In any case, I though would recommend to keep the 'https reject' rule in all cases (initial install as well as reboot) as it really makes sense (otherwise pi-hole really does not work well with https sites) and should not disrupt anything else.

Well, except when users already host a `https` service on this server! We could detect if there is something listening on port 443 and then suggest something like:

> Hey, we detected you have no service that is listening on port 443. Due to reasons A and B, it is advised that you install a rule that blocks this port. Say Yes if you would like to have this done in an automated fashion for you.

Also, if the firewall in entirely open (`APPECT` anything), there is no need to explicitly reject access to port 443, as the kernel will do that by itself:
```
$ curl https://pi.hole
curl: (7) Failed to connect to pi.hole port 443: Connection refused
```

This rule is only necessary of the default policy is set to `DROP` if I'm not mistaken. But we have encountered quite a number of cases where explicitly rejecting port 443 helped the users and I don't recall of the the details for all of the occurrences of this issue.

dschaper:
I do not like scripting any modifications to users firewalls. I further think we need to remove the code that already exists from the install script and not get involved in security of users systems. Having installed on multiple OSes with multiple firewall methodologies I am in the firm belief that we either cover all possible cases, or none at all. 

Borg

dschaper:
Rejected

PromoFaux:
Agreed.

PR #988, opened 2016-12-15
[WIP] Allow enabling/disabling of adlists via webUI

DL6ER:
**By submitting this pull request, I confirm the following (please check boxes, eg [X])Failure to fill the template will close your PR:**

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] I have checked that [another pull request](https://github.com/pi-hole/pi-hole/pulls) for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:**

- [X] 10 (very familiar)

---
Backend for https://github.com/pi-hole/AdminLTE/pull/252

_This template was created based on the work of [`udemy-dl`](https://github.com/nishad/udemy-dl/blob/master/LICENSE)._

PR #1359, opened 2017-03-26
Clarify adlists default

PromoFaux:
[Apparently it is not clear enough](https://discourse.pi-hole.net/t/how-do-i-add-additional-block-lists-to-pi-hole/259/17?u=promofaux)


DL6ER:
It is unclear to me to what extend this improves the situation. If you don't know that you first have to `ssh` to your Pi-hole, you are most likely not even able to copy/edit some files even with proper description what to do.

flokX:
The best solution would be #988 ...

DL6ER:
Obsolete due to #1365

Issue #1368, opened 2017-04-02
Poor exception handling for missing packages


adamcamp:
**In raising this issue, I confirm the following (please check boxes, eg [X]) Failure to fill the template will close your issue:**

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*
- [X] The issue I'm reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the codebase?:**

3

---

**[BUG | ISSUE] Expected Behaviour:**

- Install script fails in a deterministic manner when packages are missing or yum installation fails
- Install script outputs useful debugging information when failures occur

**[BUG | ISSUE] Actual Behaviour:**
- Install script only fails when the missing package is the only one installed
- Install script just terminates with no output to identify the error.

**[BUG | ISSUE] Steps to reproduce:**
Prereq:
1. Ensure that one of the following dependencies is missing from yum, or add a fake dependency to the end of the list (for me procps-ng was missing):

`(git iproute net-tools newt procps-ng)`

2. Manually uninstall git if it is already installed.
`yum remove git`

Part A (successful install):
1. Run through install and it should complete successfully.
`bash basic-install.sh`

**Note that the installation succeeded despite this missing package** 

Part B (failed update):
1. Run pihole -r to reconfigure pihole
`pihole -r`
2. Install script will fail with the following output:
```::: Reconfigure option selected
::: Verifying free disk space...
:::
::: Updating local cache of available packages... done!
:::
::: Checking yum for upgraded packages.... done!
:::
::: There are 2 updates available for your system!
::: We recommend you update your OS after installing Pi-Hole!
:::
:::    Checking for git... installed!
:::    Checking for iproute... installed!
:::    Checking for net-tools... installed!
:::    Checking for newt... installed!
:::    Checking for procps-ng... added to install list!
:::    Checking for epel-release... installed!
```

**Note that there is no information about why it failed or how to resolve the issue.** 

**Analysis**
Basically what is happening is that the successful installation of git is overriding the return code of the failed installation of procps-ng, and so the initial install is successful.

[...]

PR #1369, opened 2017-04-02
Fix - Poor exception handling for missing packages #1368

adamcamp:
https://github.com/pi-hole/pi-hole/issues/1368

3

---
See https://github.com/pi-hole/pi-hole/issues/1368

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -838,17 +838,13 @@ install_dependent_packages() {
   for i in "${argArray1[@]}"; do
     echo -n ":::    Checking for $i..."
     if ${PKG_MANAGER} -q list installed "${i}" &> /dev/null; then
-      echo " installed!"
+      echo " already installed!"
     else
-      echo " added to install list!"
-      installArray+=("${i}")
+      echo " installing!"
+      "${PKG_INSTALL[@]}" "${i}"
     fi
   done
-    if [[ ${#installArray[@]} -gt 0 ]]; then
-      "${PKG_INSTALL[@]}" "${installArray[@]}" &> /dev/null
-      return
-    fi
-    return 0
+  return 0
 }
 
 CreateLogFile() {


automated install/basic-install.sh
```diff
@@ -838,17 +838,13 @@ install_dependent_packages() {
   for i in "${argArray1[@]}"; do
     echo -n ":::    Checking for $i..."
     if ${PKG_MANAGER} -q list installed "${i}" &> /dev/null; then
-      echo " installed!"
+      echo " already installed!"
     else
-      echo " added to install list!"
-      installArray+=("${i}")
+      echo " installing!"
+      "${PKG_INSTALL[@]}" "${i}"
```
DL6ER: There is a lot of overhead if you run the package manager for every package on its own.

DL6ER:
Please explain why the title of this PR is

> Fix - Poor exception handling for missing packages

It is not clear to me from looking at the changes you suggest

adamcamp:
@DL6ER ,

Please see the associated bug report for a detailed analysis and to understand the purpose of the changes: 
https://github.com/pi-hole/pi-hole/issues/1368

The problem with running the package manager for all the packages at once is that it doesn't give the proper return code. If you install git and procps-ng at the same time then even if procps-ng fails, it will say the entire install was successful and continue as if nothing went wrong.

I see your point that it could potentially slow down the install, but to me the increased reliability of the install outweighs the performance impact.

Perhaps someone can suggest a flag that can be passed to yum to ensure that it aborts the install after any failures, but I'm not aware of any such flag.

dschaper:
So this is `yum` failing by trying to install a package that doesn't exist? I'm trying to see where the successful install of one package masks out the failure of another, but I think the better approach would be to determine if a package exists prior to trying to install it. Is this with CentOS 6 or 7 that you are seeing the failures?

adamcamp:
See the below example for clarification. Note that the first return code is successful, but the second is failed. The test was on CentOS 7, but it could be the same result on any host with yum and custom repos.

Checking if the package exists beforehand would cover the failure scenario when an RPM doesn't exist, but I'm not sure what would happen if the RPM did exist but it failed to install.

```
[adam@localhost ~]$ sudo yum install git fake_package; echo $?
Loaded plugins: fastestmirror, langpacks
Loading mirror speeds from cached hostfile
 * base: mirror.5ninesolutions.com
 * extras: mirror.5ninesolutions.com
 * updates: mirror.5ninesolutions.com
Package git-1.8.3.1-6.el7_2.1.x86_64 already installed and latest version
No package fake_package available.
Nothing to do
0
[adam@localhost ~]$ sudo yum install fake_package; echo $?
Loaded plugins: fastestmirror, langpacks
Loading mirror speeds from cached hostfile
 * base: mirror.5ninesolutions.com
 * extras: mirror.5ninesolutions.com
 * updates: mirror.5ninesolutions.com
No package fake_package available.
Error: Nothing to do
1

```

DL6ER:
The current yum behavior is quite unfriendly for script usage... See [here](https://bugzilla.redhat.com/show_bug.cgi?id=736694) for discussion I found on this.

In the case of `yum` we might loop over all packages with `yum info <package> >> /dev/null 2>&1 || exit` beforehand? I don't think we have to change the behavior for `apt-get` (but I haven't tested that right now).

dschaper:
`apt` uses the same approach that the Ubuntu/Debian installer uses, `debconf-apt-progress` as a parent to the underlying process. So this would just be for the `yum` (not sure if `dnf` has the same behaviour). But I do like the changes to the installed/installing verbiage, but that would need to be used for both blocks so that users for any of the supported OSes get the same display.

Mcat12:
What's the status on this? It looks like some changes are requested?

dschaper:
Yes, changes were requested, but in the comments, not in a commit review.

DL6ER:
@adamcamp Do you have any update for us?

PromoFaux:
Closing to tidy up, as this appears to have gone stale. Can be reopened if @adamcamp wishes to continue work on it!

PR #1378, opened 2017-04-11
Add web frontend shutdown command

gedetil:

_1_

---

Implement "Halt system" button, next to "Restart system" button, on
admin/settings page.  Useful for doing clean shutdown before powering off.

This affects 4 files, 3 for the web content (elsewhere), 1 for backend script (here).

Gilbert Detillieux <gedetil> 2017-04-11

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -211,6 +211,10 @@ SetExcludeClients(){
 
 }
 
+Halt(){
+        nohup bash -c "sleep 5; halt" &> /dev/null </dev/null &
+}
+
 Reboot(){
 
 	nohup bash -c "sleep 5; reboot" &> /dev/null </dev/null &
@@ -456,6 +460,7 @@ main() {
 		"setdns"            ) SetDNSServers;;
 		"setexcludedomains" ) SetExcludeDomains;;
 		"setexcludeclients" ) SetExcludeClients;;
+		"halt"              ) Halt;;
 		"reboot"            ) Reboot;;
 		"restartdns"        ) RestartDNS;;
 		"setquerylog"       ) SetQueryLogOptions;;


advanced/Scripts/webpage.sh
```diff
@@ -211,6 +211,10 @@ SetExcludeClients(){
 
 }
 
+Halt(){
+        nohup bash -c "sleep 5; halt" &> /dev/null </dev/null &
```
DL6ER: Now I have found the time to think about this: In general, it seems to be best practice to use the `shutdown` command instead of `halt` on Linux and UNIX as it e.g. sends around messages to all users that are logged in to notify them of a shutdown or reboot so they can properly finish what they are working on or save their work.

You have to take into account that this will be implemented into Pi-hole and not all users run it on a Pi-like device. I'll have to change the `reboot` command as well, now that I think about it.

Also, you won't need `sleep 5` anymore, because it has an internal timer and will also prevent users from loging in any more (since the system will shut down soon, anyway).

advanced/Scripts/webpage.sh
```diff
@@ -456,6 +460,7 @@ main() {
 		"setdns"            ) SetDNSServers;;
 		"setexcludedomains" ) SetExcludeDomains;;
 		"setexcludeclients" ) SetExcludeClients;;
+		"halt"              ) Halt;;
```
DL6ER: Thinking of what I wrote above you might want to change the name to `shutdown` (not a hard requirement, though).

DL6ER:
Edited the title of this PR. Connected web PR is:

github.com/pi-hole/AdminLTE/pull/470

DL6ER:
In general, you could find behavior like:

- `halt` = Ungraceful shutdown. Performs an unconditional shutdown for system processes. It does not shut down any services because it does not run through the rc scripts, it only syncs the hard drives and halts the processor! There is a good chance of lost data and corrupt files.

- `shutdown` = Shuts down the system in a clean orderly fashion.

On the Raspberry Pi, `shutdown` just calls `halt` - but that does not have to be the case anywhere else.

advanced/Scripts/webpage.sh
gedetil: To me, "Halt" and "Reboot" are less ambiguous, whereas "shutdown" can imply either action.

DL6ER: Maybe `poweroff`, since that is definitely distinct from rebooting the system.

advanced/Scripts/webpage.sh
gedetil: Hadn't thought of that, and was trying to implement my change in a way that was as consistent as possible with what was there already.  But I agree that it might be better practice to use shutdown command (in both cases) with an appropriate message, in case this is running on a more general-purpose system and not a dedicated box.

Did you want to implement the change to both the Halt and Reboot functions?  Or should I do that?  (Keep in mind that I'm new to GitHub, and you could probably do this way faster than me!)

advanced/Scripts/webpage.sh
gedetil: True, but I avoided "poweroff" since it's kind of misleading on the Pi, since it doesn't actually power the device off, but just complete the halt action.  (Of course, this would likely behave differently on different hardware.)

Your call...  I'm fine with whatever you decide.

advanced/Scripts/webpage.sh
DL6ER: Yes! Go ahead and try it. Learning by doing is a good thing :-)

gedetil: Hmm... Don't think I can specify a delay in seconds to "shutdown"...  "+m" specifies it in minutes. Do we really want to add a one-minute delay? Using "+0" or "now" would be immediate, so the message is perhaps of limited value then.

DL6ER:
Coming back to this, I still don't think that `poweroff` is that misleading. Admittedly, that is not strictly what is going to happen on the Raspberry, but in the end you can safely unplug the Pi afterwards. Might even be more intuitive than `halt` for users that are inexperienced with those technical terms.
Furthermore, on some "real" hardware, you most likely want this option to shut down the device physically so you don't have to go there and press the button to have the system finally switch off.

dschaper:
@DL6ER Have the issues on this been addressed?

gedetil:
Sorry, I've been busy with other stuff and haven't had a chance to get 
back to this.  It may not happen until the last week of this month...

On 11/05/2017 2:20 AM, Dan Schaper wrote:
> @DL6ER <https://github.com/DL6ER> Have the issues on this been addressed?
-- 
Gilbert E. Detillieux		E-mail:	<gedetil@cs.umanitoba.ca>
Dept. of Computer Science	Web:	http://www.cs.umanitoba.ca/~gedetil/
University of Manitoba		Phone:	(204)474-8161
Winnipeg MB CANADA  R3T 2N2	Fax:	(204)474-7609

DL6ER:
@gedetil Any chance we still find a solution? ;-)

gedetil:
On 29/06/2017 1:32 PM, DL6ER wrote:
> @gedetil <https://github.com/gedetil> Any chance we still find a 
> solution? ;-)

Sorry, swamped at work and this is not a priority for me.  At this 
point, I should probably just drop the fork since the codebase will have 
changed a lot since I started this.

-- 
Gilbert E. Detillieux		E-mail:	<gedetil@cs.umanitoba.ca>
Dept. of Computer Science	Web:	http://www.cs.umanitoba.ca/~gedetil/
University of Manitoba		Phone:	(204)474-8161
Winnipeg MB CANADA  R3T 2N2	Fax:	(204)474-7609

DL6ER:
There hasn't been much of a change, but we can look into implementing this separately from you, if you like.

gedetil:
On 29/06/2017 1:41 PM, DL6ER wrote:
> There hasn't been much of a change, but we can look into implementing 
> this separately from you, if you like.

That would be great.  Thanks.

-- 
Gilbert E. Detillieux		E-mail:	<gedetil@cs.umanitoba.ca>
Dept. of Computer Science	Web:	http://www.cs.umanitoba.ca/~gedetil/
University of Manitoba		Phone:	(204)474-8161
Winnipeg MB CANADA  R3T 2N2	Fax:	(204)474-7609

PromoFaux:
See #1572

PR #1388, opened 2017-04-16
Fixed #1386: 'Block page' broken when custom hostname is used

### LINKED ISSUE #1386: "Block page" broken when custom hostname is used

- [X] The issue I am reporting can be *replicated*
- [X] The issue I'm reporting isn't a duplicate

2

**[BUG] Expected Behaviour:**

I changed my Raspberry to respond at "server" hostname (not default pi.hole hostname).
I expected to go to http://server/ and see the "block page" with all css/js loaded.

**[BUG] Actual Behaviour:**

If i go to http://server/ I see a broken page (no css/js loaded because loaded from pi.hole domain that doesn't exist)

**[BUG] Steps to reproduce:**

- Change the hostname (ex: server)
- Go to http://server/
- See the broken page (no css/js loaded)

SnaKKo:

2

### DIFF
--- a/advanced/blockingpage.css
+++ b/advanced/blockingpage.css
@@ -57,7 +57,7 @@ header h1 {
 	width: 100%;
 	padding: 8px 0;
 	text-indent: 32px;
-	background: url("http://pi.hole/admin/img/logo.svg") left no-repeat;
+	background: url("/admin/img/logo.svg") left no-repeat;
 	background-size: 30px 22px;
 }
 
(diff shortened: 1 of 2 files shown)


SnaKKo:
Issue #1386 seems to be a my LAN fault (DNS config, etc...) but I leave here the pull request because IMHO it's better to use relative path than absolute anyway.

DL6ER:
I'm not entirely sure that this will work also when users have set up Pi-hole behind some other web server (speaking of apache, nginx, etc.)

DL6ER:
I'm closing this PR for now since it is unclear if it has unforeseen side effects. It can be re-opened whenever we want this.

PR #1407, opened 2017-04-30
Update 01-pihole.conf

deHakkelaar:
Avoid dnsmasq being populated with loopback IP addresses from the "/etc/hosts" file that are of no use for resolving host to host:
"no-hosts"

Discource:
https://discourse.pi-hole.net/t/pi-hole-server-lose-awareness-of-it-self/2715/26

_{2}_

---
_{Avoid dnsmasq being populated with loopback IP addresses in the "/etc/hosts" file that are of no use for resolving host to host}_

### DIFF
--- a/advanced/01-pihole.conf
+++ b/advanced/01-pihole.conf
@@ -27,6 +27,7 @@ domain-needed
 
 bogus-priv
 
+no-hosts
 no-resolv
 
 server=@DNS1@


dschaper:
I think the `no-hosts` directive would stop `dnsmasq` from using any hosts files, including `gravity.list`? Will have to check.

deHakkelaar:
@dschaper , I have it configured this way for a while now and just now run "pihole -g" without errors.
Anything else I need to pay attention to ?

dschaper:
Thanks, I'll have to pull some research, I believe we tried that flag in a previous version and it lead to some errors. Or I could just be remembering another configuration flag.

deHakkelaar:
That could be the other ignore:
"no-resolv"

dschaper:
Thanks for your first PR though, we'll take a look and see!

DL6ER:
We often recommended users to put host names they want to get resolved in `/etc/hosts` as this is the place where they belong. If you think back to the beginning of the Internet and before the advent of DNS you will see that the `hosts` file was *the* place for this kind of config.

I'll brainstorm for alternative ideas.

deHakkelaar:
I dont believe thats valid.
Like explained on discourse, the "/etc/hosts" file is intended for resolution from the localhost point of view.
Not to be used by other hosts on the network for resolution.
I've setup few clusters (10 nodes max) where I had to give them all unique "/etc/hosts" files.
I know it is the easy way of fooling the clients but its not the proper way.
If you want DNS records, one should use the addn-hosts directive to create a list ... is my opinion ?

DL6ER:
> Not to be used by other hosts on the network for resolution.

Yeah, I came to the same conclusion after thinking about it a second time.

> I've setup few clusters (10 nodes max) where I had to give them all unique "/etc/hosts" files.

I did the same (albeit with a much higher number of nodes) but didn't even bothered to setup the `hosts` files, since name resolution is done by the local DHCP/DNS server. This is no bottleneck, since I used dedicated MPI hardware in all nodes so Ethernet was barely used except for some few job management instructions. `ssh`ing into the machines was not possible, of course.

> If you want DNS records, one should use the addn-hosts directive to create a list ... is my opinion ?

I tend to agree, but my main point is here that I don't want to see a (potentially large) number of perfectly working fine setup loosing the ability to resolve the specified hosts after an update. In other words, I think simply adding `no-hosts` isn't a sufficient solution and one would have to think more about code that

1. Create a new file for `addn-hosts`
2. Add this to `dnsmasq`'s config
3. Looks at the current `/etc/hosts` file and copy all entries that are not `127.0.0.0/8` over to the other file

Does that sound like a good plan to you?
Also asking @dschaper about his opinion.

deHakkelaar:
Some poor sod pulled 3 or maybe even 4 wires (cant remember) from my cluster.
I was glad I put in the extra effort so they kept on resolving and no down time was experienced.

For me sounds simpler to touch file "/etc/pihole/list.99.custom" and point users to use that one ?
If they make a mistake in that file, how much harm can it do ?

dschaper:
Standard is to use `/etc/hosts` files. Changing that to now point to another file in a non-standard directory is not something I'm kin to be in favor of. (I don't want yet another `/etc/networks/interfaces` vs `/etc/dhcpcd` situation that I'm battling now. 

There would have to be good reason, and a majority of the standard installations to warrant a change from accepted practices.

deHakkelaar:
/etc/hosts.custom ?

DL6ER:
> Some poor sod pulled 3 or maybe even 4 wires (cant remember) from my cluster.

Nobody is able to enter the room if they are not within the team. Solves such problems ;-)

> For me sounds simpler to touch file "/etc/pihole/list.99.custom" and point users to use that one ?
If they make a mistake in that file, how much harm can it do ?

If you browser carefully through, lets say only the most recent six months of connected posts on Discourse, you will find an enormous number of enthusiatsic Pi-hole users that are not familiar with using the Linux terminal at all! We usually had long(er) discussions with them and have been able to restore access to e.g. their router by giving them an one-liner that adds the required line to theirs `hosts` file.
If your change gets into action w/o a proper fix so that functionality is still guaranteed afterwards (e.g. like I suggested with my 1.-2.-3. above), then those users will loose connectivity again and this will cause *a lot* of unnecessary confusion and work on all possible ends.

dschaper:
Still a non-standard file that most of the users would not look to place hostnames in. Standard `/etc/resolv.conf` points to `127.0.0.1` which is `dnsmasq` for our installs. 

What exactly would be gained by our users with this deviation from standards?

dschaper:
Not saying no at this point, but I need to know that our home users that only know how to run the basic-install script do not flood us with additional support requests for things that are Pi-hole specific. And do not create issues with other packages that may be installed with Pi-hole concurrently.

DL6ER:
> What exactly would be gained by our users with this deviation from standards?

I think @deHakkelaar is talking about this:
```
me@desktop ~ $ dig raspberrypi

; <<>> DiG 9.10.3-P4-Ubuntu <<>> raspberrypi
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 63661
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;raspberrypi.			IN	A

;; ANSWER SECTION:
raspberrypi.		300	IN	A	192.168.2.10
raspberrypi.		300	IN	A	127.0.1.1

;; Query time: 0 msec
;; SERVER: 192.168.2.100#53(192.168.2.10)
;; WHEN: Sun Apr 30 12:54:57 CEST 2017
;; MSG SIZE  rcvd: 72
```

Where you clearly don't want the `127.0.1.1.` answer, which comes from the local entry in `/etc/hosts` (standard entry at least in Raspbian)

deHakkelaar:
Yeah  :)

But yes I can see the problems this might cause.
Its the same thing as the FQDN issue, if you dont set it up correct right from the start, it will bite you at one stage and will be harder to fix.

dschaper:
But `127.0.1.1` is reserved for the FQDN of the host machine. You shouldn't have an entry of `raspberry.` as the FQDN. That is kind of a mis-configuration?

dschaper:
I can see moving non `127/8` addresses, but then we'd need to do that on schedule to make sure that people following the advice of Ubuntu and putting FQDN's with IP's in that file get the information moved to the new location.

DL6ER:
Brainstorming and looking up everything important in `dnsmsaq`'s source code finished.

I now agree with @deHakkelaar that we have to do something here. The question is only where, what and how.

Problem:
- `/etc/hosts` *will* contain something like `127.0.1.1 raspberrypi` on Raspberry and possibly any Debian derivative. Note that this does not affect `pi.hole`, but only the actual host name of the device
- `dnsmasq` will read this file and happily hand out everything that is in there as there is nothing in the code that stops it from handing out `127.0.0.0/8` and/or `::1` addresses

Solutions:
1. Fix `/etc/hosts` by modifying it: replace `127.*.*.*` with the current IP
 I don't like that at all since it can break things in many places and any software update or other software might easily revert our "fix"
2. Use a secondary file (like `/etc/hosts.dnsmasq`) and copy all entries from `/etc/hosts` (stripping `127.0.0.0/8` addresses) on a run of gravity over

I'm open for further discussions.

deHakkelaar:
$ hostname -f
noads.dehakkelaar.nl

$ cat /etc/hosts
127.0.0.1       localhost
::1             localhost ip6-localhost ip6-loopback
ff02::1         ip6-allnodes
ff02::2         ip6-allrouters
127.0.1.1       noads.dehakkelaar.nl

C:\>nslookup noads.dehakkelaar.nl
Server:  noads.dehakkelaar.nl
Address:  10.0.0.8

Name:    noads.dehakkelaar.nl
Addresses:  10.0.0.8
          127.0.1.1

deHakkelaar:
I am not in favor of option 1 as you lose the advantages that come with using loopback.

DL6ER:
We are discussing a different method in the background and will open a separate PR for discussing it, shortly.

DL6ER:
@deHakkelaar Please have a look at #1408

deHakkelaar:
Closing the PR as bellow is a better solution:

https://github.com/pi-hole/pi-hole/pull/1408

PR #1415, opened 2017-05-02
Block Page 2.0

### LINKED ISSUE #1330: Potential optimizations for queryFunc()
**In raising this issue, I confirm the following (please check boxes, eg [X]) Failure to fill the template will close your issue:**

- [x] The issue I am reporting can be *replicated*
- [x] The issue I'm reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

5

---
**[ISSUE] Expected Behaviour:**
When using `pihole -q`, returning results for manually blacklisted (or wildcard listed) should be fast. 

**[ISSUE] Actual Behaviour:**
Returning results for manually blacklisted sites are slow, as [`queryFunc()`](https://github.com/pi-hole/pi-hole/blob/de6aaf18abf414f297abd09f30e767a37341bc37/pihole#L109) scans `/etc/pihole/list.*` before everything else.

**[ISSUE] Steps to reproduce:**
Manually blacklist a site (either exact or wildcard), and run `pihole -q`

**Extra thoughts:**
While developing my block page, I've been considering *first* checking `/etc/pihole/blacklist.txt` and `/etc/dnsmasq.d/03-pihole-wildcard.conf` via PHP to increase the speed in which the block page loads, but I think it'd be better to provide a fix rather than a workaround.

This appears to achieve the result I'm looking for:

```
queryFunc() {
  domain="${2}"
  method="${3}"
  
  # Scan for possible wildcard matches
  if [ -e "${wildcardlist}" ]; then
    local wildcards=($(processWildcards "${domain}"))
    for wdomain in ${wildcards[@]}; do
      result=$(scanList "\/${domain}\/" ${wildcardlist})
      # Remove empty lines before couting number of results
      count=$(sed '/^\s*$/d' <<< "$result" | wc -l)
      if [[ ${count} > 0 ]]; then
        echo "::: Wildcard blocking ${domain} (${count} results)"
        echo "${result}"
        echo ""
        exit 0
      fi
    done
  fi
  
  lists=(/etc/pihole/blacklist.txt /etc/pihole/list.*)
  for list in ${lists[@]}; do
    if [ -e "${list}" ]; then
      result=$(scanList ${domain} ${list} ${method})
      # Remove empty lines before couting number of results
      count=$(sed '/^\s*$/d' <<< "$result" | wc -l)
      echo "::: ${list} (${count} results)"
      if [[ ${count} > 0 ]]; then
        echo "${result}"
        [ ${list} = "/etc/pihole/blacklist.txt" ] && exit 0
      fi
      echo ""
    else
      echo "::: ${list} does not exist"
      echo ""
    fi
  done

  exit 0
}
```

The idea is to search wildcards and `blacklist.txt` first (running `exit 0` after a match so that `/admin/scripts/pi-hole/php/queryads.php` can return the final result faster), before scanning `list.*` files.

As I don't have a full understanding of the Pi-hole codebase (and therefore any potential ramifications this may incur), I feel it'd be better to raise this as an issue to be discussed, rather than a pull request.

**Differences in loading time:**
[Current `queryFunc()`](https://github.com/pi-hole/pi-hole/blob/de6aaf18abf414f297abd09f30e767a37341bc37/pihole#L109):
* test.com: 1.98s/1.96s/1.98s
* lb.usemaxserver.de (Manually blacklisted):  1.67s/1.66s/1.64s
* gas.com (Wildcard blacklisted): 1.95s/1.94s/1.96s

Above `queryFunc()`:
* test.com: 1.97s/1.97s/1.98s
* lb.usemaxserver.de (Manually blacklisted): 0.22s/0.22s/0.22s
* gas.com (Wildcard blacklisted): 0.15s/0.15s/0.15s

Above `queryFunc()`, changing `/etc/pihole/list.*` to read `/etc/pihole/list.*.domains` (Bypassing preEventHorizon):
* test.com: 1.75s/1.73s/1.75s
* lb.usemaxserver.de (Manually blacklisted): Unchanged
* gas.com (Wildcard blacklisted): Unchanged

WaLLy3K:

6

---
Modifications for `/usr/local/bin/pihole`:
* Allow scanList() to search files using a wildcard by removing quotes wrapped around `${list}`
* scanList() will not provide a domain ouput on each string if exact is specified (`grep -l`)
* Remove processWildcards() function
* Return a message if no domain is specified
* IDN domains are converted to punycode when running a `pihole -q` search if the `python` package is available, otherwise will revert to current behaviour
* Scan Blacklist & Wildcards first, exiting from search if a match is found (Fixes #1330)
* Use one `grep` subshell to search for all "*.domains" lists at once (opposed to looping to get every matching file name, and then spawning a `grep` instance for every matching file)
* queryFunc() will not return "(0 results)" output from files where no match is found
* Sort results based off list number
* Return a message if no results are found

Modifications for `/var/www/html/pihole/index.php` (Block Page):
* Set pcre.recursion_limit, 3x the limit necessary to process a valid 253 character domain name
* An "About Pi-hole" link on the block page provides an ELI5 explanation to those not familiar with Pi-hole
* An email contact link on the block page provides users of your Pi-hole with a means to easily get in touch with you
* Browsing to your Pi-hole's address will show a simple "landing page", which can be replaced by adding "landing.php" within "/var/www/html"
* Users manually browsing to file/image based content (i.e: non HTML based content) on blocked sites will be greeted with a small "Blocked by Pi-hole" image
* Sites that are manually blacklisted will display a notice of this on the block page
* Sites that aren't directly blocked, but have a CNAME record, will show a notification on the block page (e.g: If raw.githubusercontent.com is not blocked, but github.map.fastly.net is)
* On the block page, "Back to Safety" now directs the user to "about:home" if Javascript is disabled
* Whitelisting is disabled for installs without a password, or if a client does not have Javascript

* Known issues:
  * Admin Console needs a text field under "Web User Interface" where the admin can enter a preferred contact email when a site needs to be whitelisted, to be saved to setupVars.conf with the key "ADMIN_EMAIL"
  * Admin Console needs a text field under "Networking" where the admin can enter their Pi-hole's externally contactable FQDN, allowing access to their landing page when browsing to mypi.duckdns.org, to be saved to setupVars.conf with the key "FQDN"
  * I am not aware of expected output of `$_SERVER["VIRTUAL_HOST"]`, so I have assumed it should be filtered as if it's a domain

Modifications for `/var/www/html/pihole/blockingpage.css`:
* Block page UI overhaul to replicate the style of the Admin Console
* Block page UI is now mobile friendly
* Users can safely customise text in order to make the block page more friendly for their household

Modifications for `/etc/lighttpd/lighttpd.conf`:
* Disable `include-conf-enabled.pl`, as blindly enabling HTTPS (as Let's Encrypt does by having a file in that folder) makes loading web pages slower (and in the case of macOS/iOS, provides certificate error popups)
* Make Block page handle JS request rewrite, allowing users to better utilise their `lighttpd` service
* Make Block page handle debugging Pi-hole header
* Make Block page redirect users from `pi.hole` to `http://pi.hole/admin`

Supporting changes needed to allow users to enable HTTPS without affecting page load speeds:
```
# Only enable SSL for primary FQDN
$HTTP["host"] == "rpi.ddns.com" {
  include_shell "cat conf-enabled/letsencrypt.conf 2>/dev/null"
}
```

* Using the key `FQDN` (as found within `setupVars.conf`) to replace "rpi.ddns.com", will enable Let's Encrypt configuration if the file exists
* Expected behaviour: Querying a blocked page via HTTPS returns an error similar to: `curl: (35) error:14077458:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 unrecognized name`, but the request will still be dropped immediately

### DIFF
--- a/advanced/blockingpage.css
+++ b/advanced/blockingpage.css
@@ -1,136 +1,383 @@
-/* CSS Reset */
-html, body, div, span, applet, object, iframe, h1, h2, h3, h4, h5, h6, p, blockquote, pre, a, abbr, acronym, address, big, cite, code, del, dfn, em, img, ins, kbd, q, s, samp, small, strike, strong, sub, sup, tt, var, b, u, i, center, dl, dt, dd, ol, ul, li, fieldset, form, label, legend, table, caption, tbody, tfoot, thead, tr, th, td, article, aside, canvas, details, embed, figure, figcaption, footer, header, hgroup, menu, nav, output, ruby, section, summary, time, mark, audio, video { margin: 0; padding: 0; border: 0; font-size: 100%; font: inherit; vertical-align: baseline; }
-article, aside, details, figcaption, figure, footer, header, hgroup, menu, nav, section { display: block; }
-body { line-height: 1; }
-ol, ul { list-style: none; }
-blockquote, q { quotes: none; }
-blockquote:before, blockquote:after, q:before, q:after { content: ''; content: none; }
-table { border-collapse: collapse; border-spacing: 0; }
-html { height: 100%; overflow-x: hidden; }
+/* Pi-hole: A black hole for Internet advertisements
+*  (c) 2017 Pi-hole, LLC (https://pi-hole.net)
+*  Network-wide ad blocking via your own hardware.
+*
+*  This file is copyright under the latest version of the EUPL.
+*  Please see LICENSE file for your rights under this license. */
 
-/* General Style */
-a	{ color: rgba(0,60,120,0.95); text-decoration: none; } /* 1E3C5A */
-a:hover { color: rgba(210,120,0,0.95); transition-duration: .2s; } /* 255, 128, 0 */
-divs a		{ border-bottom: 1px dashed rgba(30,60,90,0.3); }
-b	{ font-weight: bold; }
-i	{ font-style: italic; }
+/* Text Customisation Options ======> */
+.title:before { content: "Website Blocked"; }
+.altBtn:before { content: "Why am I here?"; }
+.linkPH:before { content: "About Pi-hole"; }
+.linkEmail:before { content: "Contact Admin"; }
 
-footer, pre, td { font-family: monospace; padding-left: 15px; }
-/*body, header { background: #E1E1E1; }*/
+#bpOutput:before { content: "Info"; }
+.error:before, .exception:before { content: "Error" !important; }
+.success:before { content: "Success" !important; }
+
+.add:after { content: "The domain is being whitelisted..."; }
+.exception:after { content: "An unhandled exception occured. This may happen when your browser is unable to load jQuery, or when the webserver is denying access to the Pi-hole API."; }
+.success:after { content: "Website has been whitelisted! You may need to flush your DNS cache"; }
+
+.recentwl:after { content: "This site appears to have been recently whitelisted. Please flush your DNS cache and/or restart your browser."; }
+.cname:before { content: "This site is an alias for "; } /* <a href="http://cname.com">cname.com</a> */
+.cname:after { content: ", which may be blocked by Pi-hole."; }
+
+.blacklist:before { content: "Manually Blacklisted"; }
+.wildcard:before { content: "Manually Blacklisted by Wildcard"; }
+.noblock:before { content: "Not found on any Blacklist"; }
+
+#bpBlock:before { content: "Access to the following website has been denied:"; }
+#bpFlag:before { content: "This is primarily due to being flagged as:"; }
+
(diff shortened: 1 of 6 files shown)


dschaper:
Can we break this out to a couple of PR's? I/We will have to change some of the testing code to account for the new changes, and doing that on a big pull like this is going to be a bit of a beast.

WaLLy3K:
I was considering doing this *after* the PR, so yeah, can do!

Issue #1414, opened 2017-05-01
Installation of v3 on a freshly installed Debian Jessie x64 server - no password set


Ardakilic:
**In raising this issue, I confirm the following (please check boxes, eg [X]) Failure to fill the template will close your issue:**

- [x] The issue I am reporting can be *replicated*
- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [x] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the codebase?:**

_4_ I know PHP, but I'm sure my issue isn't related the the PHP related sides of the project.

---

After v3 fresh installation on Debian remote server, I could access to admin interface without password. I've been using Pi-hole for months with my local setup, so I'm not a newbie here about usage of the application.

I'm using Debian Jessie as my remote VPS, only [SoftEtherVPN](https://github.com/SoftEtherVPN/SoftEtherVPN) was already installed, but it shouldn't affect this behaviour.

Initial installation failed because my VPS didn't have `dialog` app installed. so I installed it.

After that, I re-run the installation one-liner, then I got “RTNETLINK answers: File exists” issue during setup, but this was an issue I also got on my Orange Pi Zero (which also runs ARMBian Debian distro), so I've assumed It's a bug in Debian, and re-run the installation, this time it worked. It's possible it's not remotely related, but I'm mentioning anyways.

All steps of the setup went smoothly after this, the dialog screen provided me with the auto generated password.

**[BUG | ISSUE] Expected Behaviour:**
http://remoteip/admin should ask password for critical sections after fresh installation

**[BUG | ISSUE] Actual Behaviour:**
http://remoteip/admin is running without admin password even though dialog provided me with one, administration interface was open to public. I had to `pihole -a -p` to set a new password.

I have another pi-hole running on my local network on Orange Pi Zero, if that would matter (maybe cookies somehow messed up?).

**[BUG | ISSUE] Steps to reproduce:**

* Install pi-hole v3 to a freshly VPS (I'm using on HostUS if that would matter) Debian Jessie after it .
* Administration panel password is not set.

[...]

PR #1423, opened 2017-05-02
Double hash the password directly in the install script

PromoFaux:
Alternative fix for https://github.com/pi-hole/pi-hole/issues/1414

Hash the password in the install script, rather than adding an undocumented method to change the password into the password script

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -67,6 +67,13 @@ SetTemperatureUnit(){
 
 }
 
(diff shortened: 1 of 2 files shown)


PromoFaux:
Changed to functioni(s/z)e the hashing part of `webpage.sh`, and source it in the install script. No duplication of code.

pralor:

https://discourse.pi-hole.net/t/pi-hole-3-0-1-fixes-and-tweaks/2921/1

PR #1437, opened 2017-05-05
Ensure any changes to blocking page are updated.

PromoFaux:

- [x] 10 (very familiar)

---

Looks for a value of `CUSTOMBLOCKPAGE` (boolean) in `setupVars.conf`, if it is true then it does not replace the block page (this allows for users to have their own custom block page. If it is false or does not exist in `setupVars.conf` (set to false at the top of the script), then the block pages will be overwritten with updated files.


dschaper:
Who wants to update the test rig for the changes ;)

PR #1440, opened 2017-05-06
Revert "Ensure any changes to blocking page are updated."

PromoFaux:
Reverts pi-hole/pi-hole#1437


dschaper:
Handling in house, no need for reversion.

PromoFaux:
@pi-hole/gravity Strongly suggest we revert this pull request and think about it properly, rather than trying to code around it. We (okay, I) merged it by mistake without updating the tests, but it has since become clear that we need to further discuss how we implement it.

DL6ER:
Approved

PR #1453, opened 2017-05-10
Improvements to custom block page updates

PromoFaux:

- [x] 10 (very familiar)

---

Users may miss announcement of new flag for preserving their custom block pages. This makes sure they do not. Wording on the whiptail may need some work.

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -35,7 +35,7 @@ IPV4_ADDRESS=""
 IPV6_ADDRESS=""
 QUERY_LOGGING=true
 INSTALL_WEB=true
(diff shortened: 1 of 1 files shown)


DL6ER:
What happens for the users that run update automated (e.g. via `cron`)?

PromoFaux:
I mean, really they should not be, for the very same reason we don't include an auto-update setting. 

But, we have to respect that some people _may_ be doing this (chap on discourse run's it every 1/2 hour!) So this is far from perfect. I suggest we roll back #1437 in the meantime.

PromoFaux:
closed by #1440

PR #1460, opened 2017-05-12
Solve piholeLogFlush.sh having to be issued 2 x to clear logs

cybrnook:
Simplified the command -v syntax, and added a sleep 3 timer to the first execution of the log rotation. The second execution was being issued while the first was still running, thus it would fail and you would have to issue the "Flush Logs" command a second time.

6

### DIFF
--- a/advanced/Scripts/piholeLogFlush.sh
+++ b/advanced/Scripts/piholeLogFlush.sh
@@ -10,9 +10,9 @@
 
 echo -n "::: Flushing /var/log/pihole.log ..."
 # Test if logrotate is available on this system
-if command -v /usr/sbin/logrotate &> /dev/null; then
+if command -v /usr/sbin/logrotate >/dev/null; then
   # Flush twice to move all data out of sight of FTL
-  /usr/sbin/logrotate --force /etc/pihole/logrotate
+  /usr/sbin/logrotate --force /etc/pihole/logrotate; sleep 3
   /usr/sbin/logrotate --force /etc/pihole/logrotate
 else
   # Flush both pihole.log and pihole.log.1 (if existing)


dschaper:
The `&>` notation was to redirect both `stderr` and `stdout` to `/dev/null` so that all responses from `command -v` were silenced. Does the simplification on that line still do the same behavior?

cybrnook:
It does, the response is silent.

This was my test case:

```
pi@pihole:/opt/pihole $ cat test.sh 
if command -v /usr/sbin/logrotate &> /dev/null; then
echo "Yep"
	else
echo "Nope"
fi
pi@pihole:/opt/pihole $ sh test.sh 
Yep

```

dschaper:
Confirmed with two tests:
dschaper@nanopineo:~$ cat a.sh
```
if command -v /bin/bash &> /dev/null; then
	echo "Bash"
else
	echo "NoBash"
fi

if command -v /bin/nonexists > /dev/null; then
	echo "Nonexists"
else
	echo "No nonexists"
fi
```

```
dschaper@nanopineo:~$ bash a.sh
Bash
No nonexists
```

cybrnook:
Please note I have no space between ">/dev/null"

dschaper:
Noted, updated test and behavior is consistent.

DL6ER:
We have to re-approve?

Approved

DL6ER:
This pull request may have to be reverted or at least changed, since it seems that the daily flushing at midnight is now done *twice* as well!

from `/etc/cron.d/pihole`:
```
# Pi-hole: Flush the log daily at 00:00 so it doesn't get out of control
#          Stats will be viewable in the Web interface thanks to the cron job above
#          The flush script will use logrotate if available
00 00   * * *   root    PATH="$PATH:/usr/local/bin/" pihole flush
```

cybrnook:
I'm not following.... If we refer this PR, then we go back to a non functioning syntax? So the real issue would be back all the way in the original change you made.

If you refer this, then we are back to having to issue flush logs rice from the GUI because the timing is wrong. Then I think your desired effect of only gained because of a bug in your first, original syntax.

DL6ER:
I'll open a new PR that fixes the behavior in a few moments. You are invited to review it!

DL6ER:
@cybrnook see #1475 and subsequently #1476

cybrnook:
Will do my friend, give me a little time. Heading out to work right now.

cybrnook:
I think this looks better. I like that you have removed the original first occurrence of:

"if command -v /usr/sbin/logrotate >/dev/null; then"

I was thinking on my way to work, that I don't see how this condition would ever not be met. So I don't think adding an "else" to anything after that is meaningful, since it would never make it that far.

I like your correction of adding the pre "if" statement:

"if [[ "$@" == *"once"* ]]; then"

I will test your three PR's this evening if you wish. Visually I think it looks okay.

PR #1465, opened 2017-05-13
Update list.sh for busybox compatibility

diginc:
I believe this has feature parity with `sed /foo/ Id` but also supports busybox, and my alpine docker ;)

_8_

---
uses grep to find the line numbers matching the domain (case insensitive) and feeds them into sed for deletion.

### DIFF
--- a/advanced/Scripts/list.sh
+++ b/advanced/Scripts/list.sh
@@ -157,8 +157,8 @@ RemoveDomain() {
         if [[ "${bool}" == true ]]; then
             # Remove it from the other one
             echo "::: Removing $1 from $list..."
-            # /I flag: search case-insensitive
-            sed -i "/${domain}/Id" "${list}"
+            # Busybox sed compatible case-insensitive domain removal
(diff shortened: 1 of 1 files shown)


advanced/Scripts/list.sh
```diff
@@ -157,8 +157,8 @@ RemoveDomain() {
         if [[ "${bool}" == true ]]; then
             # Remove it from the other one
             echo "::: Removing $1 from $list..."
-            # /I flag: search case-insensitive
-            sed -i "/${domain}/Id" "${list}"
+            # Busybox sed compatible case-insensitive domain removal
+            sed -i "$(grep -in "^${domain}$" ${list} | awk -F':' '{print $1}' | tr '\n' ',' | sed 's/,$/\n/')d" ${list}
```
dschaper: Does `^${domain}$` work even though the lists start with more information before the domains are listed?

diginc: I thought the gravity.list had the IPs on the front, but whitelist/blacklists just had the domain them selves, one per line.

dschaper: It does, I was testing against the wrong lists, I was using the source lists, and got switched up.

dschaper: Edited comment to strikeout and clarify.

PR #1469, opened 2017-05-14
Tricorder: Insecure Opt-out

WaLLy3K:

9

---

* Check to see if Tricorder is being called directly
* Provide opt-out for insecure transmission of debug log
* Remove mention of internal function from help menu

### DIFF
--- a/pihole
+++ b/pihole
@@ -301,7 +301,7 @@ Branches:
 }
 
 tricorderFunc() {
-  if [ ! -p "/dev/stdin" ]; then
+  if [[ ! -p "/dev/stdin" ]]; then
     echo "Please do not call Tricorder directly."
     exit 1
   fi
@@ -319,7 +319,7 @@ tricorderFunc() {
     echo "There is a possibility that this could be intercepted by a third party"
     echo "If you wish to cancel, press Ctrl-C to exit within 10 seconds"
     secs="10"
-    while [ "$secs" -gt 0 ]; do
+    while [[ "$secs" -gt "0" ]]; do
        echo -ne "."
        sleep 1
        : $((secs--))
--- a/pihole
+++ b/pihole
@@ -305,9 +305,15 @@ tricorderFunc() {
     echo "Please do not call Tricorder directly."
     exit 1
   fi
+  
+  if ! timeout 2 nc -z tricorder.pi-hole.net 9998 &> /dev/null; then
+    echo "Unable to connect to Pi-hole's Tricorder server."
+    exit 1
+  fi
 
   if command -v openssl &> /dev/null; then
     openssl s_client -quiet -connect tricorder.pi-hole.net:9998 2> /dev/null < /dev/stdin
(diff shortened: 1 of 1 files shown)


pihole
```diff
@@ -269,10 +269,24 @@ piholeCheckoutFunc() {
...
+  if [ ! -p "/dev/stdin" ]; then
+    echo "Please do not call Tricorder directly."
+    exit 1
+  fi
+
   if command -v openssl &> /dev/null; then
     openssl s_client -quiet -connect tricorder.pi-hole.net:9998 2> /dev/null < /dev/stdin
   else
-    nc tricorder.pi-hole.net 9999 < /dev/stdin
+    echo "The debug log will be transmitted insecurely via plain-text"
```
DL6ER: Is "insecure" really the right wording here? I guess you are talking about "insecure" in the way that others could listen to what you send. I know that with a rogue access point this can be achieved fairly easily.
I guess we should be clear to the user and say something along the lines of "your data will be transported unencrypted and hence there is a possibility that it could be intercepted by a third party".

pihole
```diff
@@ -269,10 +269,24 @@ piholeCheckoutFunc() {
...
+    echo "If you wish to cancel, press Ctrl-C to exit within 10 seconds"
+    secs="10"
+    while [ "$secs" -gt 0 ]; do
+       echo -ne "."
+       sleep 1
+       : $((secs--))
+    done
+    echo " "
+    nc tricorder.pi-hole.net 9999 < /dev/stdin < /dev/stdin
```
DL6ER: Why `< /dev/stdin < /dev/stdin` ?

WaLLy3K: Good find, I have no idea how that slipped in there :smile:

pihole
WaLLy3K: You're absolutely right! Check the most recent version, and let me know what you think of the suggested wording.

pihole
```diff
@@ -300,6 +300,29 @@ Branches:
   checkout "$@"
 }
 
+tricorderFunc() {
+  if [ ! -p "/dev/stdin" ]; then
+    echo "Please do not call Tricorder directly."
+    exit 1
+  fi
+
+  if command -v openssl &> /dev/null; then
```
dschaper: Should we add in a check like `nc -v -z -w5 tricorder.pihole.net 9998` at the start to check and see if the port is open and accesible? Will retcode 0|1 for status...

WaLLy3K: Oh, in the event that a local firewall is blocking the port? That's a good thought.

dschaper: Yeah, or connectivity is down. The debug used to have that check in it, I can't find it now, but may have been in a version that never merged. And it left a message that the server could not be connected to. (In which case, it just left the log on the local machine...) But that second part could be handled in debug.sh

dschaper: So I guess another request would be to add retcodes to the functions so they can be evaluated in other scripts.

PR #1481, opened 2017-05-18
Retrieve local repos on repair

WaLLy3K:

4

---

**[BUG] Expected Behaviour:**

When running `pihole -r` for a repair, `/etc/.pihole` and `/var/www/html/admin` should be updated.

**[BUG] Actual Behaviour:**
```
::: You are root.
::: Repair option selected.
::: Verifying free disk space...
[]
:::    Checking for whiptail... installed!
::: --reconfigure passed to install script. Not downloading/updating local repos
:::    Checking for bc... installed!
```

**[BUG] Steps to reproduce:**

`pihole -r`, and select repair.

**PR fixes:**

* Change conditional to check for repair
* Change wording of Update/Reconfigure message
* Fixed indenting

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1179,8 +1179,9 @@ update_dialogs() {
 }
 
 clone_or_update_repos() {
-  if [[ "${UpdateCmd}" != "Repair" ]]; then
-    echo "::: Skipping update of local repositories"
+  if [[ "${reconfigure}" == true ]]; then
+    echo "::: --reconfigure passed to install script. Resetting changes to local repos"
+    git reset --hard
   else
     # Get Git files for Core and Admin
     getGitFiles ${PI_HOLE_LOCAL_REPO} ${piholeGitUrl} || \
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1179,22 +1179,22 @@ update_dialogs() {
 }
 
 clone_or_update_repos() {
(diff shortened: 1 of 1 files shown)


Mcat12:
Wouldn't this turn repair into running an update? Not updating the repos is useful for when you switch branches.

PromoFaux:
maybe a git reset --hard?

WaLLy3K:
Thinking about the code that was already there, I'd say you're right @Mcat12. As @PromoFaux, `git reset --hard` would be necessary to revert any changes, and then maybe of these two solutions:

* Repair becomes an "update and repair"
* Repair attempts to clone the current locally operating version

PromoFaux:
When I first implemented the repair option, the thinking was that nobody should really be editing files in the `.pihole` directory, but they may have messed things up elsewhere. Really it's only purpose was to redistribute the correct files to where they should be.

I use it when I'm doing experimental development, edit into the repo directory, and then install. Though, that said, I'm now using Dom's symlink script, so I will use the repair option less frequently.

Is it used by the `pihole checkout` function? I'd look but I'm on mobile and lazy!

WaLLy3K:
It looks like `git reset --hard` is the right tool for the job here!

technicalpyro:
not sure if this is a result of this commit or something else going on but this is what is returned when trying to update 
```
::: --reconfigure passed to install script. Resetting changes to local repos
fatal: Not a git repository (or any of the parent directories): .git
Unable to complete update, contact Pi-hole
```

PR #1565, opened 2017-06-27
Revamp of debug script with logic and colours

jacobsalmela:

5
---
This is an attempt to make the debug log easier to read and to disseminate the information it contains--with the end goal being that users will be able to help themselves based on the output of the script.  As much as we try to help every user who has a question, we're all volunteering our time, so if the user can run the debug script and self-diagnose the issue, not only will they learn something, but can hopefully fix any issue they may run into by following the URLs the script provides.

The script will:
- display the same information to the user that can be found in the log file
- display common information the developers need when assisting a user
- display important information first
- display information in a logical order
- display any caution or warnings in yellow
- display any errors in red
- link to an FAQ if the script detects any warnings or errors
- make the debug token highly-visible so users can clearly see they need to let us know what it

In addition to the revamp, detailed comments are included on nearly every line.  Pi-hole is a great beginners project (especially in the RPi-realm) and we want Pi-hole to be a project people feel they can contribute to.  To that end, each portion of code is explained so someone trying to learn about shell scripts can use our project as a springboard.

Intensive testing should be done on this before merging as the variety of systems and issues people may run into are difficult for one person to recreate.

And it's been a while since I've coded anything so forgive any rustiness you may find... and feel free to add any changes you see fit.

TODO
- [x] Strip off the color codes before uploading to `tricorder` since the colors show up like `\e[0;36m some more information here`, which makes it difficult to read once on the server
- [ ] Add the disclaimer at the beginning of the debugger
- [ ] Search the forums for more common issues to test for and create FAQs if necessary

### DIFF
--- a/advanced/Scripts/piholeDebug.sh
+++ b/advanced/Scripts/piholeDebug.sh
@@ -50,21 +50,21 @@ echo_succes_or_fail() {
   # Set the first argument passed to tihs function as a named variable for better readability
   local message="${1}"
   # If the command was successful (a zero),
-  if [ $? -eq 0 ]; then
+  if [[ $? -eq 0 ]]; then
     # show success
-    echo -e "    ${TICK} ${message}"
+    echo -e "    ${TICK} ${message}" 2>&1 | tee -a "${DEBUG_LOG}"
   else
     # Otherwise, show a error
-    echo -e "    ${CROSS} ${message}"
+    echo -e "    ${CROSS} ${message}" 2>&1 | tee -a "${DEBUG_LOG}"
   fi
 }
 
 initiate_debug() {
   # Clear the screen so the debug log is readable
   clear
-  echo -e "${COL_LIGHT_PURPLE}*** [ INITIALIZING ]${COL_NC}"
+  echo -e "${COL_LIGHT_PURPLE}*** [ INITIALIZING ]${COL_NC}" 2>&1 | tee "${DEBUG_LOG}"
   # Timestamp the start of the log
-  echo -e "    ${INFO} $(date "+%Y-%m-%d:%H:%M:%S") debug log has been initiated."
+  echo -e "    ${INFO} $(date "+%Y-%m-%d:%H:%M:%S") debug log has been initiated." 2>&1 | tee -a "${DEBUG_LOG}"
 }
 
 # This is a function for visually displaying the curent test that is being run.
@@ -72,7 +72,7 @@ initiate_debug() {
 # Colors do not show in the dasboard, but the icons do: [i], [✓], and [✗]
 echo_current_diagnostic() {
   # Colors are used for visually distinguishing each test in the output
-  echo -e "\n${COL_LIGHT_PURPLE}*** [ DIAGNOSING ]:${COL_NC} ${1}"
+  echo -e "\n${COL_LIGHT_PURPLE}*** [ DIAGNOSING ]:${COL_NC} ${1}" 2>&1 | tee -a "${DEBUG_LOG}"
 }
 
 file_exists() {
@@ -106,16 +106,16 @@ check_core_version() {
   local error_msg="git status failed"
   if_directory_exists "${PIHOLEGITDIR}" && \
     cd "${PIHOLEGITDIR}" || \
-    echo "pihole repo does not exist"
+    echo -e "pihole repo does not exist" 2>&1 | tee -a "${DEBUG_LOG}"
     if git status &> /dev/null; then
       PI_HOLE_VERSION=$(git describe --tags --abbrev=0);
       PI_HOLE_BRANCH=$(git rev-parse --abbrev-ref HEAD);
       PI_HOLE_COMMIT=$(git describe --long --dirty --tags --always)
       echo -e "    ${INFO} Core: ${PI_HOLE_VERSION}
         ${INFO} Branch: ${PI_HOLE_BRANCH}
-        ${INFO} Commit: ${PI_HOLE_COMMIT}"
+        ${INFO} Commit: ${PI_HOLE_COMMIT}" 2>&1 | tee -a "${DEBUG_LOG}"
     else
-      echo "${error_msg}"
+      echo -e "${error_msg}" 2>&1 | tee -a "${DEBUG_LOG}"
       return 1
     fi
 }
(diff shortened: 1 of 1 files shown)


advanced/Scripts/piholeDebug.sh
```diff
@@ -9,534 +9,1071 @@
...
-log_write() {
-    echo "${@}" >&3
-}
+# These variables would normally be next to the other files
+# but we need them to be first in order to get the colors needed for the script output
+PIHOLE_SCRIPTS_DIRECTORY="/opt/pihole"
+PIHOLE_COLTABLE_FILE="${PIHOLE_SCRIPTS_DIRECTORY}/COL_TABLE"
+
+# These provide the colors we need for making the log more readable
+if [[ -f ${PIHOLE_COLTABLE_FILE} ]]; then
```
PromoFaux: If you merge in development, then you should be able to get rid of this if/else check, as `COL_TABLE` will have been installed by the installer at this point.

WaLLy3K: The idea is that the debug script be as "independent" from the main install as possible, even if that means duplicating functions that can be sourced from other scripts.

advanced/Scripts/piholeDebug.sh
```diff
@@ -9,534 +9,1071 @@
...
-    while read -r line; do
-		  if [ ! -z "${line}" ]; then
-			  [[ "${line}" =~ ^#.*$  || ! "${line}" || "${line}" == "WEBPASSWORD="* ]] && continue
-				log_write "${line}"
-			fi
-		done < "${1}"
-		log_write ""
+log_write() {
+  # echo arguments to both the log and the console
+  echo -e "${@}" | tee -a /proc/$$/fd/3
```
WaLLy3K: * The indenting is not consistent with the rest of `pihole` (There should be two spaces before `[i]`)
* The debug log gets spammed with escape characters. Perhaps you could try [this `sed` code](https://github.com/pi-hole/pi-hole/blob/95150a134ed363afe360cd8bca1bb29e8fc11a7e/advanced/Scripts/chronometer.sh#L50) to strip colour codes before `tee`ing it?

advanced/Scripts/piholeDebug.sh
```diff
@@ -9,534 +9,1071 @@
...
 
-	if [[ ${IPV6_READY} ]]; then
-	  port_check 6 "${2}" "${1}"
+  # Same logic applies to the dashbord as above, if the X-Header matches what a working system shoud have,
+  if [[ $dashboard == "$dashboard_working" ]]; then
+    # then we can show a success
+    log_write "$TICK ${COL_LIGHT_GREEN}${dashboard}${COL_NC}"
+  else
+    # Othewise, it's a failure since the X-Headers either don't exist or have been modified in some way
+    log_write "$CROSS ${COL_LIGHT_RED}X-Header does not match or could not be retrieved.${COL_NC}"
```
WaLLy3K: I had this error occur on my install. It's not very descriptive - is it an issue with my Admin Console not showing the right header?

advanced/Scripts/piholeDebug.sh
```diff
@@ -9,534 +9,1071 @@
...
+
+  # If the gateway variable has a value (meaning a gateway was found),
+  if [[ -n "${gateway}" ]]; then
+    log_write "${INFO} Default IPv${protocol} gateway: ${gateway}"
+    # Let the user know we will ping the gateway for a response
+    log_write "   * Pinging ${gateway}..."
+    # Try to quietly ping the gateway 3 times, with a timeout of 3 seconds, using numeric output only,
+    # on the pihole interface, and tail the last three lines of the output
+    # If pinging the gateway is not successful,
+    if ! ${cmd} -c 3 -W 2 -n ${gateway} -I ${PIHOLE_INTERFACE} >/dev/null; then
```
WaLLy3K: I'm having issues with this error on my install:

```
[i] Default IPv4 gateway: 10.0.0.1
10.0.0.1
   * Pinging 10.0.0.1
10.0.0.1...
[✗] Gateway did not respond. (https://discourse.pi-hole.net/t/why-is-a-default-gateway-important-for-pi-hole/3546)
```

```
ping -c 3 -W 2 -n 10.0.0.1 -I eth0
PING 10.0.0.1 (10.0.0.1) from 10.0.0.2 eth0: 56(84) bytes of data.
64 bytes from 10.0.0.1: icmp_seq=1 ttl=64 time=0.478 ms
64 bytes from 10.0.0.1: icmp_seq=2 ttl=64 time=0.443 ms
64 bytes from 10.0.0.1: icmp_seq=3 ttl=64 time=0.419 ms

--- 10.0.0.1 ping statistics ---
3 packets transmitted, 3 received, 0% packet loss, time 2117ms
rtt min/avg/max/mdev = 0.419/0.446/0.478/0.034 ms
```

advanced/Scripts/piholeDebug.sh
Mcat12: There's some X headers set in the lighttpd config, and this checks if those headers are being added correctly. One is for the web interface, another is for anything else (most likely blocked ad)

jacobsalmela: Sorry about that.  The X-Headers have long been a backend test for checking if things are working properly on the Web interface side.  I'll have to add a comment in there for anyone else who does not know about them.

advanced/Scripts/piholeDebug.sh
jacobsalmela: - Good point; I'll fix that.
- I recall seeing your `sed` code, but couldn't find it so thanks for the link

advanced/Scripts/piholeDebug.sh
jacobsalmela: @WaLLy3K makes a good point since the debugger needs to be more independent to check for problems with an install.  But I can see both sides as repeating code is not usually very useful...

dschaper:
When did Codacy start enforcing ShellCheck? I though we just had Travis handling SC linting?

jacobsalmela:
> From a usability standpoint, would it be possible to duplicate all the messages in order at the end of the script? That would prevent "TL;DR" when a user inevitably finds they don't want to have to sift through 200+ lines to find individual errors that whizzed passed their screen when the debug was active.

I'll play around with it--maybe just "TL;DR"ing only errors...because the user should be able to clearly see where there is green across the board.

jacobsalmela:
>When did Codacy start enforcing ShellCheck? I though we just had Travis handling SC linting?

Yeah, I'm not sure why...

advanced/Scripts/piholeDebug.sh
WaLLy3K: I think that for maintenance and future bug-fixing purposes, we may want to consider something like the following (if this makes sense):

* In the debug script, put a comment above a duplicated function that references the original function name (if different) and the file it is located in
* In the original file, put a comment above the original function that references that it is also duplicated within the debug script

jacobsalmela: Or maybe attempt to `source` the function if possible, but if it fails, the duplicated code can still run--even on a "broken" install.

dschaper:
Just checked Codacy and there is partial ShellCheck enforcement there now. We can disable it and just go with Travis, which would turn off linting since Travis does not have `debug` script testing enabled currently.

dschaper:
Oh, and you may want to merge in development when you get a chance, some of the Travis ShellCheck tests have been disabled and that should clear out the Blocker errors that you are seeing on this branch with the Travis PUSH failures.

jacobsalmela:
I suppose we could if we start running `shellcheck` locally and making it has happy as possible before submitting a PR, which should probably just be doing anyway.

dschaper:
It makes things a lot easier to run locally and then commit once you have some things changed up. But we can collaborate on things and as long as we're talking with each other we can work the different failures as they come along. And then we can turn up the testing for the scripts just as a final check once we know things are getting dialed in. I just don't want to have 15 notices from Codacy and then have Travis red test things at the same time :)

advanced/Scripts/piholeDebug.sh
jacobsalmela: Stripping the character codes off in the `log_write` function will stop the colours from being visible, so I'll implement the `sed` command somewhere else.  This should also help not having to make a new subshell everytime the `log_write` is called.  Should be easier to just strip the characters off once before upload.

advanced/Scripts/piholeDebug.sh
jacobsalmela: Will you send me the `bash -x` output from that section of the log after running the script?  Feel free to send it to `tricorder` so we don't clutter up this comment thread.

advanced/Scripts/piholeDebug.sh
```diff
@@ -9,534 +9,1082 @@
...
+# Files required by Pi-hole
+# https://discourse.pi-hole.net/t/what-files-does-pi-hole-use/1684
+PIHOLE_CRON_FILE="${CRON_D_DIRECTORY}/pihole"
+
+PIHOLE_DNS_CONFIG_FILE="${DNSMASQ_D_DIRECTORY}/01-pihole.conf"
+PIHOLE_DHCP_CONFIG_FILE="${DNSMASQ_D_DIRECTORY}/02-pihole-dhcp.conf"
+PIHOLE_WILDCARD_CONFIG_FILE="${DNSMASQ_D_DIRECTORY}/03-wildcard.conf"
+
+WEB_SERVER_CONFIG_FILE="${WEB_SERVER_CONFIG_DIRECTORY}/lighttpd.conf"
+WEB_SERVER_CUSTOM_CONFIG_FILE="${WEB_SERVER_CONFIG_DIRECTORY}/external.conf"
```
DL6ER: This variable doesn't seem to be used at all.

advanced/Scripts/piholeDebug.sh
```diff
@@ -9,534 +9,1082 @@
...
+PIHOLE_DEBUG_LOG="${LOG_DIRECTORY}/pihole_debug.log"
+PIHOLE_DEBUG_LOG_SANITIZED="${LOG_DIRECTORY}/pihole_debug-sanitized.log"
+PIHOLE_FTL_LOG="${LOG_DIRECTORY}/pihole-FTL.log"
+
+PIHOLE_WEB_SERVER_ACCESS_LOG_FILE="${WEB_SERVER_LOG_DIRECTORY}/access.log"
+PIHOLE_WEB_SERVER_ERROR_LOG_FILE="${WEB_SERVER_LOG_DIRECTORY}/error.log"
+
+# An array of operating system "pretty names" that we officialy support
+# We can loop through the array at any time to see if it matches a value
+SUPPORTED_OS=("Raspbian" "Ubuntu" "Fedora" "Debian" "CentOS")
```
DL6ER: This variable doesn't seem to be used at all.

advanced/Scripts/piholeDebug.sh
```diff
@@ -9,534 +9,1082 @@
...
+${RUN_DIRECTORY}
+${LOG_DIRECTORY}
+${WEB_SERVER_LOG_DIRECTORY}
+${WEB_SERVER_CONFIG_DIRECTORY}
+${HTML_DIRECTORY}
+${WEB_GIT_DIRECTORY}
+${BLOCK_PAGE_DIRECTORY})
+
+# Store the required directories in an array so it can be parsed through
+mapfile -t array <<< "$var"
```
DL6ER: I don't see what this line is supposed to do here.

advanced/Scripts/piholeDebug.sh
```diff
@@ -9,534 +9,1082 @@
...
+  get_program_version "php"
+}
 
-	return "${error_found}"
+is_os_supported() {
+  local os_to_check="${1}"
+  # Strip just the base name of the system using sed
+  the_os=$(echo ${os_to_check} | sed 's/ .*//')
+  # If the variable is one of our supported OSes,
+  case "${the_os}" in
```
DL6ER: Ah, you probably want to use `$SUPPORTED_OS` here

advanced/Scripts/piholeDebug.sh
jacobsalmela: Yeah, some are unused right now, but we may use them in the future.

advanced/Scripts/piholeDebug.sh
jacobsalmela: Yes, that is correct.  Sorry for the confusion.  It' a beastly script akin to the installer....

pralor:

https://discourse.pi-hole.net/t/pi-hole-v3-2-introduces-long-term-statistics-an-audit-log-colours-and-more/5772/1

PR #1567, opened 2017-06-27
Add pihole checkout ftl

DL6ER:

## 10

---

Add `ftl` choice for `pihole checkout`. A lot of code had to be added as `ftl` is a different thing than `core` and `web` as it is not a local repository.

### DIFF
--- a/advanced/Scripts/piholeCheckout.sh
+++ b/advanced/Scripts/piholeCheckout.sh
@@ -303,12 +303,12 @@ checkout() {
     fi
     webbranches=($(get_available_branches "${webInterfaceDir}"))
 
-    if [[ "${corebranches[@]}" == *"master"* ]]; then
+    if [[ "${webbranches[@]}" == *"master"* ]]; then
       echo -e "${OVER}  ${TICK} $str
   ${INFO} ${#webbranches[@]} branches available for Web Admin"
     else
       # Print STDERR output from get_available_branches
-      echo -e "${OVER}  ${CROSS} $str\n\n${corebranches[*]}"
+      echo -e "${OVER}  ${CROSS} $str\n\n${webbranches[*]}"
       exit 1
     fi
 
@@ -329,11 +329,10 @@ checkout() {
 
     if check_download_exists "$path"; then
         echo "  ${TICK} Branch ${2} exists"
+        FTLinstall "${binary}" "${path}"
     else
         echo "  ${CROSS} Requested branch \"${2}\" is not available"
     fi
-
-    FTLinstall "${binary}" "${path}"
 
   else
     echo -e "  ${INFO} Requested option \"${1}\" is not available"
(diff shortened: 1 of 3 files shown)


advanced/Scripts/piholeCheckout.sh
```diff
@@ -24,6 +24,93 @@ update="false"
...
+
+  machine=$(uname -m)
+
+  local str="Detecting architecture"
+  echo -ne "  ${INFO} ${str}..."
+  if [[ ${machine} == arm* || ${machine} == *aarch* ]]; then
+    # ARM
+    local rev=$(uname -m | sed "s/[^0-9]//g;")
+    local lib=$(ldd /bin/ls | grep -E '^\s*/lib' | awk '{ print $1 }')
+    if [[ "$lib" == "/lib/ld-linux-aarch64.so.1" ]]; then
```
WaLLy3K: Question: should we not be quoting and bracing strings to be consistent with the rest of the codebase? (e.g: `[[ "${machine}" == "arm"* ||` (There's a couple of instances of differently formatted conditionals, but I won't flag each line for comment in case there's a reason for this!)

dschaper: Has this been run through a manual ShellCheck.net scan?

DL6ER: @WaLLy3K I use exactly the same code as is used in the installer script (where we know that it is working fine for all users).

@dschaper Yes, it reported two minor things (1. Add a check for failing `cd`, 2. Don't declare variable as `local` and assign value in the same line)

DL6ER:
@WaLLy3K You requested changes. How should they look like? Checking out `FTL` is basically another thing, as it is not a repository.

WaLLy3K:
Instead of saying "Branch [blank space] doesn't exist", could it match existing behaviour by saying "Requested Branch "" is not available" and then list the available branches? There's no reason why the output should be any different to Core or Web.

Optionally, I'd love to see the conditional formatting made consistent (quoting and bracing), but that's just me being extremely nit-picky, and will happily approve this even without this minor change! :)

WaLLy3K:
Just removing my review blocker, rather than explicitly approving it (but there's no difference according to GitHub). It'd still be good if the text of `pihole checkout ftl` was to look the same as `pihole checkout core`, but I understand that there's some limitations that prevent the actual branch readout.

PromoFaux:
Confirmed working, with the exception of switching back to `master` branch. But I don't think that is an issue with the code, rather an issue with the `.sha1` on `ftl.pi-hole.net` for the `master` branch files:

```
fap@fappotron:~$ pihole checkout ftl master
  Please note that changing branches severely alters your Pi-hole subsystems

  [✓] Detected x86_64 architecture
  [✓] Branch master exists
  [i] Installing FTL...sha1sum: pihole-FTL-linux-x86_64.sha1: No such file or directory
  [✗] Installing FTL
  Error: Download of binary from ftl.pi-hole.net failed
```

PR #1581, opened 2017-07-05
Add "php5-sqlite" to dependencies

WaLLy3K:

9

---

When loading Long Term Data > Query Log, this prevents the following message from popping up:

> An unknown error occured while loading the data.


WaLLy3K:
Given the various names for the package we need (and the potential uncertainty of the package not existing), we might want to consider doing a search for it beforehand with: `apt-cache search --names-only "^php" sqlite | cut -d " " -f1`

What do you think, @DL6ER?

DL6ER:
@WaLLy3K Sounds like a good idea.

dschaper:
Running on Armbian with `php 7` I get the following packages listed:

```
root@nanopineo:/var/log/lighttpd# apt-cache search --names-only "^php" sqlite | cut -d " " -f1
php-sqlite3
php7.0-sqlite3
```

The first is php5, the second is php7.

WaLLy3K:
Ah, so we're going to just have to amend the `--names-only` search to limit it to the users currently installed PHP version. Good catch, @dschaper!

dschaper:
https://github.com/pi-hole/pi-hole/blob/master/automated%20install/basic-install.sh#L102 This is the check we do during install to see what PHP version is available, would this be of any use for the issue in this thread?

WaLLy3K:
It could work, but I don't know how that'll affect anyone that's wanting to run PHP7 instead of 5.

WaLLy3K:
Currently, it looks like `apt-cache search --names-only "^${phpVer}" sqlite | cut -d " " -f1` will be the easiest way to *ensure* we're getting the right package, since the package name isn't exactly reliable. However, this won't work on the Fedora family. Could anyone confirm what the `apt-cache` equivalent is?

PromoFaux:
Is there a `yum` equivalent? @bcambl may know...

bcambl:
php-pdo is what you are looking for.

automated install/basic-install.sh
```diff
@@ -204,7 +209,7 @@ elif command -v rpm &> /dev/null; then
   PKG_COUNT="${PKG_MANAGER} check-update | egrep '(.i686|.x86|.noarch|.arm|.src)' | wc -l"
   INSTALLER_DEPS=(dialog git iproute net-tools newt procps-ng)
   PIHOLE_DEPS=(bc bind-utils cronie curl dnsmasq findutils nmap-ncat sudo unzip wget)
-  PIHOLE_WEB_DEPS=(lighttpd lighttpd-fastcgi php php-common php-cli)
+  PIHOLE_WEB_DEPS=(lighttpd lighttpd-fastcgi php php-common php-cli php-pdo)
```
Mcat12: Is PDO necessary?

PromoFaux:
Please note the three additional commits are from the `development` branch. From a PR that probably should have gone into 3.2 in the first place (cosmetic bugs)

automated install/basic-install.sh
PromoFaux: Based off of @bcambl 's comment here:
https://github.com/pi-hole/pi-hole/pull/1581#issuecomment-318190684

automated install/uninstall.sh
```diff
@@ -45,7 +45,7 @@ if [ -x "$(command -v rpm)" ]; then
 		PKG_MANAGER="yum"
 	fi
 	PKG_REMOVE="${PKG_MANAGER} remove -y"
-	PIHOLE_DEPS=( bind-utils bc dnsmasq lighttpd lighttpd-fastcgi php-common git curl unzip wget findutils )
+	PIHOLE_DEPS=( bind-utils bc dnsmasq lighttpd lighttpd-fastcgi php-common php-sqlite git curl unzip wget findutils )
```
Mcat12: I think this should be `php-pdo`?

PromoFaux: Well caught. Uninstall script neglected as-per! One sec...

automated install/basic-install.sh
Mcat12: Verified that (at least on Fedora) it provides the same SQLite library.

PR #1583, opened 2017-07-05
temp files to a proper homedir not /etc

graysky2:

_6_

---
The purpose of this PR is to relocate the individual block lists `gravity.sh` pulls down to a proper home directory for the `pihole` user rather than keeping them in under `/etc/pihole/` which is technically a place for config files, not temp files.

I am not very familiar with the script-based installer so I do not know how to handle an update where the target system has the `pihole` user already created.  What would need to happen is for a call to `usermod` to create the homedir I propose.  It might be useful to also add a line somewhere to remove the temp lists from `/etc/pihole/` should they already exist.


DL6ER:
We store the lists in `/etc/pihole` to be able to re-use them when downloading the current version of the lists fails. This proved to be particularly important for the default list downloaded from `hosts-file.net` as this server is experiencing a downtime quite frequently.

If we would not keep the lists around, we would "loose" their content as the next run of `gravity.sh` could not consider the contents of these lists.

jacobsalmela:
In addition, storing the lists locally and not downloading them every time `gravity` runs respects the bandwidth of those hosting the lists.

graysky2:
@DL6ER @jacobsalmela  - Agreed.  This PR just relocates  them from `/etc/pihole/` (config dir) to `/var/lib/pihole/` (pihole user's homedir).  It does not delete them between runs.  My comment about that was just in the case that you agree and merge, they would not need to be in both places.

automated install/basic-install.sh
```diff
@@ -1054,7 +1054,7 @@ create_pihole_user() {
     echo -ne "${OVER}  ${CROSS} ${str}"
     local str="Creating user 'pihole'"
     echo -ne "  ${INFO} ${str}..."
-    useradd -r -s /usr/sbin/nologin pihole
+		useradd -c 'pihole user' -d /var/lib/pihole -r -m -s /usr/sbin/nologin pihole
```
PromoFaux: In the case of an existing installation, what will this line do?

graysky2:
@PromoFaux  - I believe the modified call to `useradd` will be ignored in the `create_pihole_user` function because the call to `id ...` will end in a 0 code.  We do need to modify it further as I said to allow existing installs to both:
* Make the new homedir
* Optionally set `/etc/passwd` to point to it

Something like:
```
mkdir /var/lib/pihole/
chown pihole:pihole /var/lib/pihole/
usermod -d /var/lib/pihole/ pihole
```
The only issue I see what those 3 lines is that the last will error out if user pihole has any active processes (which it will on a active system... I believe it runs iphole-FTL) and thus not modify `/etc/passwd` ... in this case I don't think it matters since you defined the shell for pihole as `/usr/sbin/nologin` in the current version of the script.

For example on my ubuntu lxc, the current stable version script sets:
```
# grep pihole /etc/passwd
pihole:x:999:999::/home/pihole:/usr/sbin/nologin
```
We could consider a little sed magic on that line in `/etc/passwd` but I don't think it's worth it unless you can guarantee all distros have exactly the same line in that file per your install script.  Alternatively, we could check and set it elsewhere in the code for example when services are stopped. I am not familiar enough with the code to know where that point is though.

These steps are exactly what I will do to the Arch Linux package via a `post_install` call by the way.

graysky2:
All - I think it's more straightforward: the current script from the master branch does not create `/home/pihole`  since you did so with the `-r` option so I will adjust my PR accordingly.  This means no modification should be needed to me.  Let me test the code a few times (fresh install from stable then update with development as 1 case and fresh install from development as case 2).  Stay tuned.

graysky2:
OK.  Tested and it works on an update.  My modifications work on a clean install (creating `pihole` user and setting-up the homedir).  It should be noted that at least on my system, running the current development branch of https://github.com/pi-hole/pi-hole/blob/development/automated%20install/basic-install.sh fails on my system (see below).

The only thing I am unclear how to do is to have the redundant `/etc/pihole/list*.domains` deleted.  The script running in update mode by passes the check for `/var/lib/lxc` in this PR.  Perhaps someone with more knowledge about the code can suggest a place for it.

Note that starting on line 1031 (the runGravity function) looks as if it should accomplish this for us.

```
sudo ./basic-install.sh 
[sudo] password for ubuntu: 

        .;;,.
        .ccccc:,.
         :cccclll:.      ..,,
          :ccccclll.   ;ooodc
           'ccll:;ll .oooodc
             .;cll.;;looo:.
                 .. ','.
                .',,,,,,'.
              .',,,,,,,,,,.
            .',,,,,,,,,,,,....
          ....''',,,,,,,'.......
        .........  ....  .........
        ..........      ..........
        ..........      ..........
        .........  ....  .........
          ........,,,,,,,'......
            ....',,,,,,,,,,,,.
               .',,,,,,,,,'.
                .',,,,,,'.
                  ..'''.


  [✓] Root user check
  [✓] Disk space check

  [✓] Update local cache of available packages

  [✗] Checking apt-get for upgraded packages
       Kernel update detected. If the install fails, please reboot and try again

  [i] Installer Dependency checks...
  [✓] Checking for apt-utils
  [✗] Checking for dialog (will be installed)
  [✓] Checking for debconf
  [✗] Checking for dhcpcd5 (will be installed)
  [✗] Checking for git (will be installed)
  [✓] Checking for iproute2
  [✓] Checking for whiptail

  [✓] Stopping dnsmasq service...

  [✓] Stopping lighttpd service...
  [i] Using Google DNS servers
  [✓] Set IP address to 10.9.8.108
       You may need to restart after the install is complete
::: Found neither IPv6 ULA nor GUA address, blocking IPv6 ads will not be enabled
  [i] IPv4 address: 10.9.8.108/24
  [i] IPv6 address: 
  [i] Web Interface On.
  [i] Logging On.
  [✗] Check for existing repository in /etc/.pihole
  [✓] Clone https://github.com/pi-hole/pi-hole.git into /etc/.pihole

  [✗] Check for existing repository in /var/www/html/admin
  [✓] Clone https://github.com/pi-hole/AdminLTE.git into /var/www/html/admin

  [i] Main Dependency checks...
  [✗] Checking for bc (will be installed)
  [✓] Checking for cron
  [✓] Checking for curl
  [✗] Checking for dnsmasq (will be installed)
  [✗] Checking for dnsutils (will be installed)
  [✓] Checking for iputils-ping
  [✗] Checking for lsof (will be installed)
  [✗] Checking for netcat (will be installed)
  [✓] Checking for sudo
  [✗] Checking for unzip (will be installed)
  [✓] Checking for wget
  [✗] Checking for lighttpd (will be installed)
  [✗] Checking for php-common (will be installed)
  [✗] Checking for php-cgi (will be installed)
  [✓] Creating user 'pihole'  [i] Installing scripts from /etc/.pihole...install: cannot stat './advanced/Scripts/COL_TABLE': No such file or directory
grep: /etc/pihole/setupVars.conf: No such file or directory
  [i] Restarting services...

  [✓] Starting dnsmasq service

  [✓] Enabling dnsmasq service to start on reboot

  [✓] Starting lighttpd service

  [✓] Enabling lighttpd service to start on reboot

  [i] Preparing to run gravity.sh to refresh hosts...
  [i] Running gravity.sh
:::
::: WARNING: /etc/pihole/setupVars.conf missing. Possible installation failure.
:::          Please run 'pihole -r', and choose the 'reconfigure' option to reconfigure.
```

graysky2:
EDIT: I just commited and rebased solving this.  I believe the PR is good to merge as-is.

graysky2:
I am happy to correct the merge conflict if you guys find value in the PR.

PromoFaux:
Could you hold off until we've released 3.1.1 /3.2, got a lot of merge conflicts we need to sort out in the base :)

graysky2:
@PromoFaux - No problem.  My first time dealing with merge conflicts actually.  I think I just brought the PR back into sync.

dschaper:
This is being addressed in #1773

dschaper:
Sorry, I thought this was addressing the location of log files, not the lists themselves. Reopening.

PromoFaux:
Sorry @graysky2, would you mind going over the conflicts again :)

Mcat12:
Closing for now, unless this PR becomes active again.

Issue #1591, opened 2017-07-09
tmpfs FTL start order


runningman84:
My pihole rpi (first gen using rasbian lite image) uses a tmpfs for /var/log in order to limit writes to sd card and increase data privacy...:
```
# rc.local
mkdir /var/log/apt
mkdir /var/log/fscks
mkdir /var/log/ntpstats && chown ntp.ntp /var/log/ntpstats
mkdir /var/log/lighttpd && chown www-data.www-data /var/log/lighttpd

touch /var/log/pihole.log && chmod 664 /var/log/pihole.log && chown dnsmasq.root /var/log/pihole.log
```
It looks like the start order of FTL is wrong:
```
root@pihole:/var/log# grep dns /var/log/daemon.log 
Jul  9 14:05:15 pihole systemd[1]: Starting dnsmasq - A lightweight DHCP and caching DNS server...
Jul  9 14:05:16 pihole dnsmasq[553]: dnsmasq: syntax check OK.
Jul  9 14:05:19 pihole systemd[1]: Started dnsmasq - A lightweight DHCP and caching DNS server.
root@pihole:/var/log# grep FTL /var/log/daemon.log 
Jul  9 14:05:12 pihole systemd[1]: Starting LSB: pihole-FTL daemon...
Jul  9 14:05:13 pihole pihole-FTL[368]: Not running
Jul  9 14:05:14 pihole systemd[1]: Started LSB: pihole-FTL daemon.
```
This causes an error in FTL
```
root@pihole:/var/log# tail pihole-FTL.log 
[2017-07-09 14:05:13.961] FATAL: Opening of /var/log/pihole.log failed!
[2017-07-09 14:05:13.961]        Make sure it exists and is readable by user pihole
```

I do not really understand the issue because my rc.local script creates an empty log which should be readable. The pihole installer should ensure that FTL is started after dnsmasq.

Futhermore it looks like FTL constantly writes its database file (pihole-FTL.db) in /etc/pihole. I have now linked it to a file in /tmp.

dschaper:
If `pihole-FTL.db` is linked to a temporary file in your /tmp directory that is deleted upon restart/reboot you will lose all historical data and the Pi-hole will not function correctly.

runningman84:
yes that's perfect from a privacy perspective...

dschaper:
We have a feature that disables all logging and log records if the goal is to not have any logging recorded.

runningman84:
[...]

PR #1593, opened 2017-07-09
Update pihole-FTL.service

runningman84:
fixes https://github.com/pi-hole/pi-hole/issues/1591

2

---
pihole-FTL fails to start if the dnsmasq logfile is not in place:
```
[2017-07-09 14:05:13.961] FATAL: Opening of /var/log/pihole.log failed!
[2017-07-09 14:05:13.961]        Make sure it exists and is readable by user pihole
```

### DIFF
--- a/advanced/pihole-FTL.service
+++ b/advanced/pihole-FTL.service
@@ -1,8 +1,8 @@
 #!/bin/bash
 ### BEGIN INIT INFO
 # Provides:          pihole-FTL
-# Required-Start:    $remote_fs $syslog
-# Required-Stop:     $remote_fs $syslog
+# Required-Start:    $remote_fs $syslog dnsmasq
+# Required-Stop:     $remote_fs $syslog dnsmasq
 # Default-Start:     2 3 4 5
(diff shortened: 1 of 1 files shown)


dschaper:
2nd look at this, I don't think we should link FTL with `dnsmasq` as they are functionally separate and each one works okay on its own. The existence of `/var/log/pihole.log` is a sanity check, and that file should exist. The only time would be on a bare `/var/log` directory, which would mean other problems.

DL6ER:
The current version of `FTL` (or any version since end of June) checks for `/var/log/pihole.log` up to two times (it waits for 15 seconds in between): https://github.com/pi-hole/FTL/pull/106

Also, as Dan already said, we don't need a running `dnsmasq` necessarily (it may not be able to start if you have a bug in its configuration and that should really not stop `FTL` from starting).

Hence, I close your PR now. Please don't see this as a sign of disregard, but I just think that we have found a more universal way of doing what you want that has less secondary effects. Feel free to contribute anything you want anytime in the future again!

PR #1617, opened 2017-07-17
Hotfix/cidr (Pi-hole 3.1.1)

PromoFaux:
Let's try this again. 

_Please ensure this is tested thoroughly before approving._

### DIFF
--- a/advanced/Scripts/update.sh
+++ b/advanced/Scripts/update.sh
@@ -19,8 +19,14 @@ readonly ADMIN_INTERFACE_DIR="/var/www/html/admin"
 readonly PI_HOLE_GIT_URL="https://github.com/pi-hole/pi-hole.git"
 readonly PI_HOLE_FILES_DIR="/etc/.pihole"
 
+# shellcheck disable=SC2034
 PH_TEST=true
-source ${PI_HOLE_FILES_DIR}/automated\ install/basic-install.sh
+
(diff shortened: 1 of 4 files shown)


PromoFaux:
Test results:
https://pastebin.com/raw/rX1uuVKX

advanced/Scripts/update.sh
```diff
@@ -62,7 +66,7 @@ GitCheckUpdateAvail() {
   fi
 
   # Change back to original directory
-  cd "${curdir}"
+  cd "${curdir}" || exit
```
WaLLy3K: Unless I'm mistaken in its purpose, if `cd` fails, we'll want to use `exit 1` to signify that it quit with an error.

dschaper: That whole function could use a rewrite, there's no need to store the PWD as the function happens in a subshell so we don't need to move back to the PWD, it's not needed. But that can come in another PR.

dschaper:
`pihole -up` looks okay and `gravity.list` doesn't have any CIDR in it. So I think that covers things?

advanced/Scripts/update.sh
dschaper: I'd fix that now, but that would reset all the current approvals, and it's really not breaking anything it's just extra steps that don't need to be taken.

dschaper:
Approved (since PA doesn't accept `shipit` anymore :( )

PR #1647, opened 2017-07-28
[DevPatch] Stop an extra `h` being added to leasetime of DHCP server. 

PromoFaux:
Same as  #1646 , but into the development branch rather than `release/3.2`


Mcat12:
Bugfixes go into release, and release can be merged with dev to get bug fixes there.

PR #1631, opened 2017-07-24
Clean up and optimise Gravity

WaLLy3K:
**By submitting this pull request, I confirm the following:**

- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [x] I have checked that [another pull request](https://github.com/pi-hole/pi-hole/pulls) for this purpose does not exist.
- [x] I have considered, and confirmed that this submission will be valuable to others.
- [x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [x] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:**

9

---
Some pictures of Gravity running can be [found here](https://imgur.com/a/kf8iT).

Changes for `gravity.sh`
* Shellcheck validation
* Made variable names, function names, comments and output clearer to understand
* Quoted and braced variables and conditionals
* Fix adlists.list handling logic, and remove any CR line endings
* Make CP/MV/RM provide user-friendly output upon failure
* Change adlists.list retrieval logic
* Moved and fixed adlists.list domain parsing logic
* Create gravity_ParseFileAsDomains() function to handle parsing of source files
* If no changes to a list is detected, print no output
* Ensure each source blocklist has a final newline
* Format number output as currency
* Make array of adlists domain sources unique to prevent redundant whitelisting
* Merged bash IPv4/IPv6 hosts formatting IF statement into an awk one-liner
* Trap Ctrl-C cancellations and run gravity_Cleanup()
* Use new gravity_Cleanup() function on errors and script completion
* Ensure that dnsmasq uses force-reload when gravity is invoked
* Add --wildcard option to ensure dnsmasq is restarted upon b/wlisting of a wildcard

Changes for `list.sh`
* Add "--blacklist-only" to only run essential gravity functions
* Pass "--wildcard" option to `gravity.sh` to ensure dnsmasq is restarted

Changes for `pihole`
* Made use of $PI_HOLE_SCRIPT_DIR
* Fix #1610 & #1624
* Add "$2" to restartDNS options

Changes for `webpage.sh`
* Make RestartDNS() use `pihole restartdns`

PR #1656, opened 2017-08-11
added sleep before gravity for slower hardware 

technicalpyro:

_{5}_

---
_{attempt to sleep for 15s to allow dnsmasq to fully start before calling gravity 
use case is pi2B or lower with anything more than 900K domains}_

users with faster hardware will still be slept as im not sure how to code this on a per use case 

ALSO no clue why the hotfix is showing in this PR but if someone with the ability could delete that from it i would appreciate it 

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -983,7 +983,8 @@ runGravity() {
     cp ${PI_HOLE_LOCAL_REPO}/adlists.default /etc/pihole/adlists.default
   fi
   echo "::: Running gravity.sh"
-  { /opt/pihole/gravity.sh; }
+  sleep 15s
+ { /opt/pihole/gravity.sh; }
 }
 
 create_pihole_user() {


dschaper:
We're actually debating this right now on how to handle the problem with `dnsmasq` and having this timing verified in a real use case scenario is very helpful. WaLLy was working on something similar so I'll ping him to check this out. @WaLLy3K 

Thanks! (And it's showing a merge conflict, looks like it's a difference in the echo style, we've moved from the older `:::` format to a new checkbox format. I'll tweak the code in the pull to resolve the conflict, but you'll need to pull that change in to your branch.)

technicalpyro:
yeah i made some mistakes when i forked this morning to make the pull request in my experience if we wait 10-15s before calling gravity it allows dnsmasq to finish its startup unless something elaborate that checks the dnsmasq status before running gravity is coded i think this really is the most simple solution

WaLLy3K:
While the issue that this attempts to resolve is a very valid problem, your proposed solution is a bit ham-fisted as it only affects a specific minority of users while creating a huge delay for others running default blocklists.

As Dan said, this is something I'm already attempting to resolve - via [this section of code](https://github.com/pi-hole/pi-hole/blob/tweak/gravity/gravity.sh#L71).

DL6ER:
Will be resolved by #1631 in a more elegant way (users on fast hardware or with small blocking lists won't have to wait an extra 15 seconds)

PR #1676, opened 2017-08-31
Replace superseded netstat command

WaLLy3K:

9

---

* Make colfile readonly, and use path of PI_HOLE_SCRIPT_DIR
* Rename piholeStatus function to statusFunc for function name consistency
* Replace superseded `netstat` command with `nc`
* Perform addn-hosts check using a single `grep` subshell

### DIFF
--- a/pihole
+++ b/pihole
@@ -1,4 +1,5 @@
 #!/bin/bash
+
 # Pi-hole: A black hole for Internet advertisements
 # (c) 2017 Pi-hole, LLC (https://pi-hole.net)
 # Network-wide ad blocking via your own hardware.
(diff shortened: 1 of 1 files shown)


pihole
```diff
@@ -481,41 +482,41 @@ Options:
...
-piholeStatus() {
-  if [[ "$(netstat -plnt | grep -c ':53 ')" -gt "0" ]]; then
-    if [[ "${1}" != "web" ]]; then
-      echo -e "  ${TICK} DNS service is running"
-    fi
+statusFunc() {
+  local addnConfigs
+
+  # Determine if service is running on port 53
+  if nc -z 127.0.0.1 53; then
```
dschaper: `if [[ nc -z 127.0.0.1 53 ]]; then`

But I'm not sure if this is just goal post stuff, `nc` isn't stock on a lot of systems. Some have BSD `netcat` and some don't have anything installed at all. Will have to check a few systems to make sure this is available.

pihole
```diff
@@ -481,41 +482,41 @@ Options:
...
+    [[ "${1}" != "web" ]] && echo -e "  ${TICK} DNS service is running"
```
dschaper: Just for style this would be better in an `if [[ ]]; then; else` in either a multi line statement or if you want to keep it oneliner then in that format.

pihole
WaLLy3K: I'll be happy to sort the style issue out once there's more word on this. It'll also mean [this code](https://github.com/pi-hole/pi-hole/blob/feb412b02af9b0b0465e096f4141ac0d23e7349d/pihole#L564) will need to be changed.

dschaper:
Very nice, looks good.

Approved

pralor:

https://discourse.pi-hole.net/t/pi-hole-v3-2-introduces-long-term-statistics-an-audit-log-colours-and-more/5772/1

PR #1420, opened 2017-05-02
Disable HTTPS & shift functions to block page

WaLLy3K:
**By submitting this pull request, I confirm the following:**

- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [x] I have checked that [another pull request](https://github.com/pi-hole/pi-hole/pulls) for this purpose does not exist.
- [x] I have considered, and confirmed that this submission will be valuable to others.
- [x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [x] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:**

5

---

The changes here primarily shift the work of header functions, JS redirects, etc, to the block page as per https://github.com/pi-hole/pi-hole/pull/1416.

Also, by commenting out `include-conf-enabled.pl` and replacing it with `include_shell "find /etc/lighttpd/conf-enabled -name '*.conf' -a ! -name 'letsencrypt.conf' -printf 'include \"%p\"\n' 2>/dev/null"`, the following can be added to `external.conf` to enable HTTPS, without affecting web browsing load speeds:
```
# Only enable SSL for primary FQDN
$HTTP["host"] == "rpi.ddns.com" {
  include_shell "cat conf-enabled/letsencrypt.conf 2>/dev/null"
}
```

A blocked HTTPS query will be instantly dropped, but may provide an error message similar to: `curl: (35) error:14077458:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 unrecognized name`. Autofilling `$HTTP["host"] == ""` within `external.conf` with the proposed `FQDN` key from `setupVars.conf` is not a part of this PR and can be developed separately.

PR #1692, opened 2017-09-16
Fix for bug on block page caused by CORS mixed content when behind reverse proxy using SSL

andofrjando:

_5_

---
**This fixes the following bug:**
If Pi-Hole is behind a reverse proxy that uses SSL, then the block page will not load resources such as `blockingpage.css` and `jquery.min.js` as the insecure `http://` is hard coded. Browsers will block attempts to load insecure resources if the page is loaded over SSL.
The fix is achieved by checking `$_SERVER['HTTPS']` and setting the variable `$proto` to either `http` or `https` accordingly. The hardcoded `http` is replaced by the contents of this variable.

___
_Bug:_
Bug
___
_Fixed_
Fixed
___


WaLLy3K:
Hey @andofrjando, thanks for bringing this to our attention!

The block page you've made the fix for is the old version available in the `master` branch, which has recently been updated in our `development` branch. If you can confirm the issue persists after running `pihole checkout core development` (which I believe it will), then we'd appreciate the fix pushed against the latest version. :smile:

andofrjando:
I'm sorry - I took the screenshots from the `master` branch, but made the code changes in `development`. 
When I was retaking the screen shots there was an error in my original pull request, but the has been fixed with a 2nd commit.
I can confirm the bug exists in the `development` branch as well. See screen shots below:
___
_Before Fix_
Bug
___
_With fix_
Fixed

WaLLy3K:
It looks like Codacy isn't in love with how the code's written, but that's okay because you should be able to appease it by replacing instances of `<?php echo $proto; ?>` with `<?=$proto ?>`.

I suspect it's expecting quotes around `"$proto"`, and the latter is just in line with how short variables are already printed anyway.

andofrjando:
Thanks for that. Looks like Codacy is happier now with the commit incorporating you suggestions.

WaLLy3K:
Thanks for that! :smile:

Based off eyeballing the code, I have a *very strong* feeling that non reverse proxy installations with SSL enabled (e.g: Let's Encrypt & Self Signed) are going to complain that there's no valid certificate for `pi.hole`, so I'll have to do some testing when I've got some spare time and post my findings back here.

andofrjando:
Yes, the browser is always going to complain about the certificate if `https` is used. And not just about loading `*.js` and `*.css` resources from `https://pi.hole`, but also the block page itself, since lighttp (or Nginx as the actual web server or as reverse proxy) is not going to have certificates for the blocked domain from a trusted certificate authority. 
Personally, I use a valid domain name (e.g. `server_name pihole.mydomain.com;`) to serve the admin interface, with a Letsencrypt certificate for that domain. Then use a `default_server` (`server_name _;`) section of the Nginx config using a self signed certificate with `*` as the common name and add it to the trust store of my browser. The `default_server` section serves up both the block page for any blocked domains, and also resources from `https://pi.hole`. Not very elegant, but a self signed certificate is the only way to serve a DNS blocked domain over `https`, and it works fine for a personal setup. 
This pull request won't break anything for people who just continue to use the default setup with `http`, but at least it will fix a small problem for anyone who wants to use `https` while providing their own solution to the mismatched domain names.

WaLLy3K:
> but also the block page itself, since lighttp (or Nginx as the actual web server or as reverse proxy) is not going to have certificates for the blocked domain from a trusted certificate authority.

I'm fully aware of this, which is why [this `lighttpd` config](https://github.com/pi-hole/pi-hole/blob/a0bd517380d6e52205387c670ae14e2db54e41f1/advanced/lighttpd.conf.debian#L54) excludes `letsencrypt.conf`, so that one can [wrap it in a conditional for `pihole.mydomain.com`](https://github.com/pi-hole/pi-hole/pull/1420) ensuring that blocked sites are not served over HTTPS, but their valid Pi-hole domain name is.

This is the primary HTTPS-enabled use-case I'm considering (primarily so VPS users can have secure remote administration); though it's really interesting to read how you've cleverly gone about it!

WaLLy3K:
Apologies for the delay, it's taken me a bit to get around to installing Let's Encrypt.

Everything's working as I'd expect it to, so your fix is good! :+1:

PR #1697, opened 2017-09-18
Use full path for git on install

totu:

1

---
Some versions of git do not like running as root (as actual root or with
sudo) when invoked from cli with just `$ git`. However using git from
default path (i.e. `$ /usr/bin/git`) works fine with root privileges.

### DIFF
--- a/advanced/Scripts/piholeCheckout.sh
+++ b/advanced/Scripts/piholeCheckout.sh
@@ -122,8 +122,8 @@ fully_fetch_repo() {
 
   cd "${directory}" || return 1
   if is_repo "${directory}"; then
-    git remote set-branches origin '*' || return 1
-    git fetch --quiet || return 1
+    /usr/bin/git remote set-branches origin '*' || return 1
+    /usr/bin/git fetch --quiet || return 1
   else
     return 1
   fi
@@ -138,7 +138,7 @@ get_available_branches() {
 
   cd "${directory}" || return 1
   # Get reachable remote branches, but store STDERR as STDOUT variable
-  output=$( { git remote show origin | grep 'tracked' | sed 's/tracked//;s/ //g'; } 2>&1 )
+  output=$( { /usr/bin/git remote show origin | grep 'tracked' | sed 's/tracked//;s/ //g'; } 2>&1 )
   echo "$output"
   return
 }
@@ -152,10 +152,10 @@ fetch_checkout_pull_branch() {
 
   # Set the reference for the requested branch, fetch, check it put and pull it
   cd "${directory}" || return 1
-  git remote set-branches origin "${branch}" || return 1
-  git stash --all --quiet &> /dev/null || true
-  git clean --quiet --force -d || true
-  git fetch --quiet || return 1
+  /usr/bin/git remote set-branches origin "${branch}" || return 1
+  /usr/bin/git stash --all --quiet &> /dev/null || true
+  /usr/bin/git clean --quiet --force -d || true
+  /usr/bin/git fetch --quiet || return 1
   checkout_pull_branch "${directory}" "${branch}" || return 1
 }
 
@@ -169,19 +169,19 @@ checkout_pull_branch() {
 
   cd "${directory}" || return 1
 
-  oldbranch="$(git symbolic-ref HEAD)"
+  oldbranch="$(/usr/bin/git symbolic-ref HEAD)"
 
   str="Switching to branch: '${branch}' from '${oldbranch}'"
   echo -ne "  ${INFO} $str"
-  git checkout "${branch}" --quiet || return 1
+  /usr/bin/git checkout "${branch}" --quiet || return 1
   echo -e "${OVER}  ${TICK} $str"
 
 
-  if [[ "$(git diff "${oldbranch}" | grep -c "^")" -gt "0" ]]; then
+  if [[ "$(/usr/bin/git diff "${oldbranch}" | grep -c "^")" -gt "0" ]]; then
     update="true"
   fi
 
-  git_pull=$(git pull || return 1)
+  git_pull=$(/usr/bin/git pull || return 1)
 
   if [[ "$git_pull" == *"up-to-date"* ]]; then
     echo -e "  ${INFO} ${git_pull}"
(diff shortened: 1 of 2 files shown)


b264:
What benefit does this change provide?

WaLLy3K:
We also use elevated `git` within [checkout.sh](https://github.com/pi-hole/pi-hole/blob/development/advanced/Scripts/piholeCheckout.sh), so it'd be best to apply your changes here too.

For the sake of testing and reproducing, which versions of `git` have this issue?

dschaper:
This was an issue on some OSes (OctoPi/OctoPrint) where the path has been changed per the OS and a modified version of `git` is preferred and will not run root.

totu:
@dschaper This is still issue in OctoPi, which is how I noticed it, but this could mean that this is not an issue in other distributions/images. 

@WaLLy3K The git version is `2.1.4`, but I suspect it might be issue with just some images (i.e. OctoPi). I've also amended the commit to include `piholeCheckout.sh`

WaLLy3K:
That's great information, thanks for letting us know!

Code wise, I'm just not sure whether I'm sold on the idea of appending the full `git` path to each use of the command, and I know that @DL6ER feels something along these lines too.

There's another solution I'd like to look into first: perhaps we can detect if the non-sudo-functioning-`git` is found on the system and if so, we change PATH to prefer the version found in `/usr/bin`?

What's the output of `which git` and `echo $PATH` on the OctoPi?

totu:
`git` just points to `/usr/bin/git`, so the magical behaviour happens somewhere else

```
pi@octopi:~ $ which git
/usr/bin/git
```

`$PATH` looks pretty normal

```
pi@octopi:~ $ echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/games:/usr/games
```

I do agree that appending the path kind a smells, but I don't have a better solution for the problem.

WaLLy3K:
Based off [this commit](https://github.com/guysoft/OctoPi/commit/2cd53c44bf6242f9d68ddac0b2b9d471b2e69866), it looks like there's two options that I can see:

* Deleting `/root/bin:` from `secure_path` inside of `/etc/sudoers` (Dangerous, and could lead to an inexperienced user breaking OctoPrint's permissions)
* Adding detection for this type of setup, and using variables to define either `git` or `/usr/bin/git`

```
# Determine if OctoPi git wrapper is in effect
if [[ ! -e "/root/bin/git" ]]; then
  git="git"
else
  git="/usr/bin/git"
fi
```

Thoughts or other options, @pi-hole/approvers?

Mcat12:
I like the latter approach (conditional detection)

totu:
Do you want me to amend the commit, according WaLLy3k’s suggestion, or just close this PR?

WaLLy3K:
We're having some discussion internally, so we'll get back to you on this :smile:

dschaper:
Do we know why OctoPi changes the `git` binary to one that has nosudo restrictions on it?

dschaper:
https://github.com/guysoft/OctoPi/issues/373

dschaper:
Looking through the issues, one is to use `sudo /usr/bin/git` which means we would have to both hardcode the path and append a `sudo` call for an already root install. The other option is to `sudo rm /root/bin/git` which is not an optimal choice. I would say at this point, my personal view is that we can not support OctoPi as a platform as the only solutions are workarounds to something artificially imposed by a non-standard method by the OctoPrint maintainers.

Mcat12:
It seems that they officially support removing the wrapper as a method to get around the restriction, if users really want to install Pi-hole on OctoPi:
https://github.com/guysoft/OctoPi/pull/374

WaLLy3K:
Sounds like we'd need to present this to the user (informing them as to why it's there, possibly upon installation) and give them the option to remove it.

circa10a:
Its very possible (especially RHEL users, using redhat git packages) are dependent  on git installations that don't reside in /usr/bin, but still update their PATH. Maybe check the PATH instead of hard coding by using 

`$(which git)`

Instead of 

`/usr/bin/git`

WaLLy3K:
> Its very possible (especially RHEL users, using redhat git packages) are dependent on git installations that don't reside in /usr/bin, but still update their PATH.

Things like this is exactly why any solution we end up going with would need to specifically target OctoPi, rather than go full ham-fist. :smile:

WaLLy3K:
Given the very informative discussion we've had here, I feel it'd be best we go for the supported option of removing the "sanity check" from OctoPi (`sudo rm /root/bin/git`), but *make sure we provide an adequate notice to warn people of what we've done*, as anything else would be reckless IMHO.

Due to this, I'm going to have to close this PR and request that someone submit a PR with the preferred change.

Thanks! :smile:

dschaper:
Can we get a `ls -lach /root/bin/git` and a `ls -lach /usr/bin/git` to see if there are any chattr's being done between the two different `git` binaries?

dschaper:
Other option. `git clone` to `/tmp` and then `sudo mv` the cloned repository to `/etc/.pihole`, but that leads to a question of if we have to make any other calls later on once the repo has been moved to `/etc/.pihole` and if those calls would need to be made as `sudo git <action>`.

dschaper:
On one of the systems that is showing this behaviour, can I get the output from:

    type -a git

Please?

totu:
```
pi@octopi:~ $ ls -lach /root/bin/git 
-rwxr-xr-x 1 root root 155 Jun 25 00:47 /root/bin/git
```
```
pi@octopi:~ $ ls -lach /usr/bin/git
-rwxr-xr-x 1 root root 1.4M Jun 25 00:17 /usr/bin/git
```
```
pi@octopi:~ $ type -a git
git is /usr/bin/git
```
Hope this helps.

As for cloning to `/tmp`, I think it would just create more problems with future updates (i.e. `git pull`s). At least I'm guessing you can't just do pull once the repo is moved over.

dschaper:
Revisiting this, I actually think that if `which git` does return the proper binary, then we could go with `$(which git)` would be an acceptable solution.

dschaper:
@totu Do you want to update this to conform to the comment suggestions? Mostly storing the full path in a variable that is the path from `which git` and then using that variable for the `git` calls?

Mcat12:
@totu Have you seen the suggestions? This PR will probably be closed unless there's further activity.

dschaper:
This PR is now stale. We have not received any issue reports for this situation in a long time. Closing.

PR #1701, opened 2017-09-21
Decrease `local-ttl` to 2

WaLLy3K:

9

---

* Decreasing value should benefit clients when whitelisting blocked sites, [as per this discussion](https://github.com/pi-hole/pi-hole/pull/1698#issuecomment-331134576).

### DIFF
--- a/advanced/01-pihole.conf
+++ b/advanced/01-pihole.conf
@@ -42,6 +42,6 @@ cache-size=10000
 log-queries
 log-facility=/var/log/pihole.log
 
-local-ttl=300
+local-ttl=2
 
 log-async


WaLLy3K:
I've noticed an issue where on slower hardware with large blocklists (e.g: Pi Zero (class 6 Micro SD card) with 1.2M blocked domains), `dnsmasq` can take up to two minutes to be resolvable as `/etc/pihole/gravity.list` is still being read - this can result in an end user having their browser show that the DNS is unavailable, instead of the Block Page (and the notice which says that they should flush their DNS cache).

Altering `local-ttl` only exposes an issue *that was already there*, but more importantly, creates the benefit of a more seamless whitelisting experience for those with smaller lists and/or more powerful hardware.

With this in mind, I'm still merging this PR.

PR #1730, opened 2017-10-07
Disable password field on block page if no password is set

Mcat12:

10

---

For when you want to whitelist a domain and no password is set.
Requires pi-hole/AdminLTE#597

screenshot from 2017-10-07 16-50-02

### DIFF
--- a/advanced/index.php
+++ b/advanced/index.php
@@ -218,6 +218,7 @@ if (explode("-", $phVersion)[1] != "0")
         // Enable password input if necessary
         if (!empty($svPasswd)) {
           echo '$("#bpWLPassword").attr("placeholder", "Password");';
+          echo '$("#bpWLPassword").css("display", "inline-block");';
           echo '$("#bpWLPassword").prop("disabled", false);';
         }
       }
@@ -273,7 +274,7 @@ if (explode("-", $phVersion)[1] != "0")
     
     <form id="bpWLButtons" class="buttons">
       <input id="bpWLDomain" type="text" value="<?=$serverName ?>" disabled/>
-      <input id="bpWLPassword" type="password" placeholder="<?=$wlPlaceHolder ?>" disabled/><button id="bpWhitelist" type="button" disabled></button>
+      <input id="bpWLPassword" type="password" placeholder="<?=$wlPlaceHolder ?>" disabled style="display: none"/><button id="bpWhitelist" type="button" disabled></button>
     </form>
   </div>
 </main>
--- a/advanced/index.php
+++ b/advanced/index.php
@@ -98,9 +98,6 @@ if ($serverName === "pi.hole") {
 
 /* Start processing Block Page from here */
 
-// Determine placeholder text based off $svPasswd presence
-$wlPlaceHolder = empty($svPasswd) ? "" : "Javascript disabled";
-
 // Define admin email address text based off $svEmail presence
 $bpAskAdmin = !empty($svEmail) ? '<a href="mailto:'.$svEmail.'?subject=Site Blocked: '.$serverName.'"></a>' : "<span/>";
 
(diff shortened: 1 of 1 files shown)


advanced/index.php
```diff
@@ -209,11 +209,17 @@ function queryAds($serverName) {
     window.onload = function () {
       <?php
       // Remove href fallback from "Back to safety" button
-      if ($featuredTotal > 0) echo '$("#bpBack").removeAttr("href");';
-      // Enable whitelisting if $svPasswd is present & JS is available
-      if (!empty($svPasswd) && $featuredTotal > 0) {
-          echo '$("#bpWLPassword, #bpWhitelist").prop("disabled", false);';
+      if ($featuredTotal > 0) {
```
WaLLy3K: This was a single line statement, why does this now cover the other statement?

WaLLy3K:
From a minimalist standpoint, having the password field stay there when there's no password seems tacky and unnecessary. From a "I don't understand why users on my network are able to whitelist stuff" use-case, it kinda makes sense, but *probably* should be a notification for elsewhere.

I'm of two minds on this matter - what are your thoughts?

advanced/index.php
Mcat12: Now the whitelist button is always enabled (as long as `$featuredTotal > 0`), but the password input is only enabled if there's a password.

Mcat12:
Do you mean you would want to remove the password input when there's no password, and put some notice on the web interface that there is no security over the whitelist? Keeping the password input reminds them that you can limit access to the whitelist.

WaLLy3K:
> Do you mean you would want to remove the password input when there's no password, and put some notice on the web interface that there is no security over the whitelist?

Yeah, I think that's the better option as long as the notice can be dismissed - some people would be happy to accept the "risks", especially if they're the sole user of Pi-hole.

> Keeping the password input reminds them that you can limit access to the whitelist.

It reminds the *admin*, sure, but it's irrelevant (and due to the wording, potentially confusing) to a user who is accessing the block page.

technicalpyro:
you gentlemen are the experts but would it be possible to have a whitelist button that then either causes a  dialog for password entry of if no password is set just whitelists ?

Mcat12:
Do you want that just as a UI change? What benefits would that bring?

technicalpyro:
> Do you want that just as a UI change? What benefits would that bring?

in my mind it would solve some of the present problem either way you have a button to whitelist if the admin password is set it then asks for it if not it just does the whitelist procedure.  unless i was misunderstanding the discussion above and way off in left field

Mcat12:
It might make sure the user doesn't get confused, sure.

WaLLy3K:
Currently getting the following error: 

> No password set - whitelisting without password not supported

I believe this message is pulled from `/admin/scripts/pi-hole/php/add.php`, but I thought we already merged an update that fixed this?

Mcat12:
Are you on the correct web branch? This requires devel.

Mcat12:
@WaLLy3K What do you think of this PR now?

WaLLy3K:
Functionality wise, it looks good. My only issue is that removing the password field hides the "[Javascript disabled](https://github.com/pi-hole/pi-hole/blob/009fa2f1b7f44ce983eece60a19ec1c109dcdcfe/advanced/index.php#L89)" text, which is used to tell a user that they need to enable JS before whitelisting. As it stands, the "Whitelist" button is just grey with no prompt as to why - this could create user confusion.

Mcat12:
Do you want me to add the password box back so it's always there, but disabled unless there's a password?

WaLLy3K:
That might be the easiest solution for now! 😄

WaLLy3K:
With JS disabled, I'm only getting the "No Admin Password Set" text, and not the Enable Javascript text, which is a bit odd. Has there been any changes to this functionality that could account for why this isn't working correctly? Also, if it's possible to remove the password field when JS *is* enabled, I think that'd be great. :smile: 

untitled

Mcat12:
The reason it's showing no password set is because of that line you referenced which decides which message to show by default based on the presence of a password.

PR #1751, opened 2017-10-25
Send SIGHUP to pihole-FTL when restarting/reloading dnsmasq

DL6ER:

## 10

---

See title. Due to a breaking change in commit https://github.com/pi-hole/pi-hole/commit/a2825be81961cee4ddbb37f9e984fe6a27faa99a, we have to inform `FTL` of the fact that we restarted `dnsmasq` resp. had it reload its files as `FTL` cannot detect this. The actual bug fix is in https://github.com/pi-hole/FTL/pull/151

### DIFF
--- a/pihole
+++ b/pihole
@@ -346,9 +346,6 @@ restartDNS() {
     svc="service dnsmasq ${svcOption}"
   fi
 
-  # Send signal to FTL to have it re-parse the gravity files
-  killall -s SIGHUP pihole-FTL
-
   # Print output to Terminal, but not to Web Admin
   str="${svcOption^}ing DNS service"
   [[ -t 1 ]] && echo -ne "  ${INFO} ${str}..."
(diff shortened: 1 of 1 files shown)


DL6ER:
Can we assume `killall` to be always around on all supported platforms?

Mcat12:
I think we've checked in the past, and `killall` is on all the platforms we support. Since it's already used elsewhere in the code, I assume it is still a valid command to use.

pihole
```diff
@@ -346,6 +346,9 @@ restartDNS() {
     svc="service dnsmasq ${svcOption}"
   fi
 
+  # Send signal to FTL to have it re-parse the gravity files
+  killall -s SIGHUP pihole-FTL
```
Mcat12: This should go after we restart dnsmasq (see where `svc` is executed).

PR #1752, opened 2017-10-28
Add bash / cron based update checker for Pi-hole

DL6ER:

## 10

---

Checks for remote (GitHub) versions as well as local versions and branches and stores them in files where `PHP` can later access them for comparison:

```
$ cat /etc/pihole/GitHubVersions
v3.1.4 v3.1 v2.11.1

$ cat /etc/pihole/localbranches
development devel tweak/GIT_BRANCH

$ cat /etc/pihole/localversions
v3.1.4-423-g2c33c64-dirty v3.1-263-g62e3e22 v2.9.1-78-g72b4844
```

TODO:

- [x] Run `pihole updatechecker` once on update so that the files are initialized before the user opens the dashboard (which needs these files)
- [x] Ensure that there are no errors on system without a web interface (i.e. `/var/www/html/admin` is missing)

### DIFF
--- a/advanced/Scripts/updatecheck.sh
+++ b/advanced/Scripts/updatecheck.sh
@@ -46,7 +46,7 @@ function get_local_branch() {
 
 CORE_BRANCH="$(get_local_branch /etc/.pihole)"
 WEB_BRANCH="$(get_local_branch /var/www/html/admin)"
-FTL_BRANCH="$(pihole-FTL tag)"
+FTL_BRANCH="$(pihole-FTL branch)"
 
 echo "${CORE_BRANCH} ${WEB_BRANCH} ${FTL_BRANCH}" > "/etc/pihole/localbranches"
 
--- a/advanced/Scripts/updatecheck.sh
+++ b/advanced/Scripts/updatecheck.sh
@@ -46,7 +46,10 @@ function get_local_branch() {
 
 CORE_BRANCH="$(get_local_branch /etc/.pihole)"
 WEB_BRANCH="$(get_local_branch /var/www/html/admin)"
-FTL_BRANCH="$(pihole-FTL branch)"
+#FTL_BRANCH="$(pihole-FTL branch)"
+# Don't store FTL branch until the next release of FTL which
+# supports returning the branch in an easy way
+FTL_BRANCH="XXX"
 
 echo "${CORE_BRANCH} ${WEB_BRANCH} ${FTL_BRANCH}" > "/etc/pihole/localbranches"
 
--- a/advanced/Scripts/updatecheck.sh
+++ b/advanced/Scripts/updatecheck.sh
@@ -1,0 +1,70 @@
+#!/usr/bin/env bash
+# Pi-hole: A black hole for Internet advertisements
+# (c) 2017 Pi-hole, LLC (https://pi-hole.net)
+# Network-wide ad blocking via your own hardware.
+#
+# Checks for updates via GitHub
+#
+# This file is copyright under the latest version of the EUPL.
+# Please see LICENSE file for your rights under this license.
+
+# Credit: https://stackoverflow.com/a/46324904
+function json_extract() {
+  local key=$1
+  local json=$2
+
+  local string_regex='"([^"\]|\\.)*"'
+  local number_regex='-?(0|[1-9][0-9]*)(\.[0-9]+)?([eE][+-]?[0-9]+)?'
+  local value_regex="${string_regex}|${number_regex}|true|false|null"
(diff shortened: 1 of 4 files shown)


PromoFaux:
LGTM

advanced/Scripts/updatecheck.sh
```diff
@@ -0,0 +1,73 @@
...
+
+echo "${GITHUB_CORE_VERSION} ${GITHUB_WEB_VERSION} ${GITHUB_FTL_VERSION}" > "/etc/pihole/GitHubVersions"
+
+function get_local_branch() {
+  # Return active branch
+  local directory
+  directory="${1}"
+  local output
+
+  cd "${directory}" 2> /dev/null || return 1
```
dschaper: Could either explicitly redirect STDERR to `/dev/null` or if you wanted to use shorthand, could redirect to the closed filehandle.

```
cd "${directory}" 2>&- || return 1
```

dschaper: Also would prefer if all variable are `UPPERCASED` just per convention instead of having some lowercased and some uppercased.

advanced/Scripts/updatecheck.sh
```diff
@@ -0,0 +1,73 @@
...
+function get_local_version() {
+  # Return active branch
+  local directory
+  directory="${1}"
+  local output
+
+  cd "${directory}" 2> /dev/null || return 1
+  # Store STDERR as STDOUT variable
+  output=$( { git describe --long --dirty --tags; } 2>&1 )
+  echo "$output"
```
dschaper: Did we set this up for a reason? (Save the output from the command to a variable only to echo out the variable and then not use the variable again?) Can we just echo out the command directly instead of storing it in a variable and remove the need for the local `output` variable completely?

advanced/Scripts/updatecheck.sh
dschaper: After review of http://wiki.bash-hackers.org/scripting/obsolete I'm going to say just leave it as is, and be more readable and understandable to users trying to understand the script.

PromoFaux:
From a functionality testing standpoint, I've not been able to find any oddities, so am happy to put my approval stamp on it, but just want to confirm @dschaper is happy with the code beforehand

advanced/Scripts/updatecheck.sh
Mcat12: It probably doesn't need to go in a variable first.

advanced/pihole.cron
```diff
@@ -28,3 +28,6 @@
 00 00   * * *   root    PATH="$PATH:/usr/local/bin/" pihole flush once quiet
 
 @reboot root /usr/sbin/logrotate /etc/pihole/logrotate
+
+# Pi-hole: Grab remote version and branch every 10 minutes
+*/10 *  * * *   root    PATH="$PATH:/usr/local/bin/" pihole updatechecker
```
Mcat12: This might be better run every 30 minutes, or even once a day. Though, I can see this being subjective and up to the user.

advanced/Scripts/updatecheck.sh
```diff
@@ -0,0 +1,73 @@
...
+  # Return active branch
+  local directory
+  directory="${1}"
+  local output
+
+  cd "${directory}" 2> /dev/null || return 1
+  # Store STDERR as STDOUT variable
+  output=$( { git rev-parse --abbrev-ref HEAD; } 2>&1 )
+  echo "$output"
+  return
```
Mcat12: Why do we need an empty return at the end of the function?

advanced/Scripts/updatecheck.sh
```diff
@@ -0,0 +1,73 @@
...
+}
+
+CORE_BRANCH="$(get_local_branch /etc/.pihole)"
+WEB_BRANCH="$(get_local_branch /var/www/html/admin)"
+#FTL_BRANCH="$(pihole-FTL branch)"
+# Don't store FTL branch until the next release of FTL which
+# supports returning the branch in an easy way
+FTL_BRANCH="XXX"
```
Mcat12: Do you expect the new version of FTL to not be released when 3.2 comes out?

advanced/pihole.cron
DL6ER: Six times an hour seems to be fair. The reason for this frequency is not to check for updates very often, but to detect the branch the user is currently on (the determination of the current hash is also outsourced into `cron` as well)

advanced/Scripts/updatecheck.sh
DL6ER: Unfortunately, there is no nice way for a transition as `FTL` has to be released before we active this. I recommend to do this later via a separate pull request.

PR #1774, opened 2017-11-17
Add support for Quad9 secure dns server

mterron:
See www.quad9.net

_1_

---
_Add support for Quad9 DNS provider_

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -767,7 +767,7 @@ setDNS() {
       Quad9 ""
       Custom "")
   # In a whiptail dialog, show the options
-  DNSchoices=$(whiptail --separate-output --menu "Select Upstream DNS Provider. To use your own, select Custom." ${r} ${c} 6 \
+  DNSchoices=$(whiptail --separate-output --menu "Select Upstream DNS Provider. To use your own, select Custom." ${r} ${c} 7 \
     "${DNSChooseOptions[@]}" 2>&1 >/dev/tty) || \
     # exit if Cancel is selected
     { echo -e "  ${COL_LIGHT_RED}Cancel was selected, exiting installer${COL_NC}"; exit 1; }
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -764,6 +764,7 @@ setDNS() {
       Norton ""
       Comodo ""
       DNSWatch ""
+      Quad9 ""
       Custom "")
   # In a whiptail dialog, show the options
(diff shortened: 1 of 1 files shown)


automated install/basic-install.sh
```diff
@@ -764,6 +764,7 @@ setDNS() {
       Norton ""
       Comodo ""
       DNSWatch ""
+      Quad9 ""
       Custom "")
   # In a whiptail dialog, show the options
   DNSchoices=$(whiptail --separate-output --menu "Select Upstream DNS Provider. To use your own, select Custom." ${r} ${c} 6 \
```
Mcat12: You should increment the `6` value so all options are made sure to be shown at once.

mcattle:
I'd suggest also adding Quad9's [secure IPv6 address](https://www.quad9.net/#/faq#is-there-ipv6-support-for-quad9) **2620:fe::fe**

It may also be worth adding a "Quad9 Unsecure" option, with the IPv4 and IPv6 addresses of 9.9.9.10 and 2620:fe::10 respectively, but that could be a new pull request.

mterron:
@mcattle , I can't find a place to add IPv6 DNS servers anywhere. It seems to me that the place where I added Quad9 only supports IPv4 addresses as per the rest of the options and there is no other screen to set up IPv6 DNS servers.

mcattle:
@mterron I'd have to dig around the code, but the web UI for pi-hole includes checkboxes for IPv6 options for Google, OpenDNS, and DNS.WATCH, so I'm guessing they're in there somewhere.  It could just be a web UI thing, though.

mcattle:
@mcattle Upon further reflection, it's probably not worth adding Quad9's unsecure addresses as options, as it goes against the value proposition of the Quad9 service.  Those addresses mainly exist to help test false positives from their secure resolvers.  Anyone who wants to use them could always add them manually.

I have also not found IPv6 defaults in Pi-hole itself, but have submitted a PR for Pi-hole's AdminLTE web UI to add Quad9 as an option.

Mcat12:
You would add the IPv6 version in the same area that the IPv4 upstreams are set. However, we have not added IPv6 upstreams in the installer yet because it does not differentiate between IPv4 and IPv6 installs in that function at the moment.

Mausy5043:
No option to select this from the WebUI?

mcattle:
@Mausy5043 The change to the WebUI is here: https://github.com/pi-hole/AdminLTE/pull/619

PR #1776, opened 2017-11-19
Pi-hole core v3.2

DL6ER:
Release

PR #1792, opened 2017-12-06
Add in an extra check to enable lighttpd.

PromoFaux:
Internal.

Should fix, [this](https://github.com/pi-hole/pi-hole/pull/1776#discussion_r155155350)


PromoFaux:
Confirmed does not break debian

PromoFaux:
This now addresses all of @bcambl's concerns brought up on #1776

PR #1796, opened 2017-12-06
added the tested logic for the password validation for the issue # 1781…

sri786:

8

Following is the link for detailed documentation of the issue and the changes with the test results:
[pihole-issue-1781-documentation.docx](https://github.com/pi-hole/pi-hole/files/1536793/pihole-issue-1781-documentation.docx)

Following is the pull request content (line 100 to 121):
image

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -89,25 +89,38 @@ SetWebPassword() {
 		exit 1
 	fi
 
-  if (( ${#args[2]} > 0 )) ; then
-    readonly PASSWORD="${args[2]}"
-    readonly CONFIRM="${PASSWORD}"
-  else
-    # Prevents a bug if the user presses Ctrl+C and it continues to hide the text typed.
-    # So we reset the terminal via stty if the user does press Ctrl+C
-    trap '{ echo -e "\nNo password will be set" ; stty sane ; exit 1; }' INT
-    read -s -p "Enter New Password (Blank for no password): " PASSWORD
-    echo ""
-
-    if [ "${PASSWORD}" == "" ]; then
-      change_setting "WEBPASSWORD" ""
-      echo -e "  ${TICK} Password Removed"
-      exit 0
-    fi
-
-    read -s -p "Confirm Password: " CONFIRM
-    echo ""
-  fi
+	if (( ${#args[2]} > 0 )) ; then
+		readonly PASSWORD="${args[2]}"
+		readonly CONFIRM="${PASSWORD}"
+	else
+	# Prevents a bug if the user presses Ctrl+C and it continues to hide the text typed.
+	# So we reset the terminal via stty if the user does press Ctrl+C
+	trap '{ echo -e "\nNo password will be set" ; stty sane ; exit 1; }' INT
+		
+	# Added the logic for password validation using standard password requirements to strengthen the password - issue#1781
+	# Password should be minimum 8 characters long and contain atleast 1 uppercase alphabet and 1 digit 
+	# And, i have added -r when password or confirm password is read, because read without -r will mangle backslashes
+	echo "Note: Password should be minimum 8 characters long and contain atleast 1 uppercase alphabet and 1 digit."
+	read -sr -p "Enter New Password (Blank for no password): " PASSWORD
+	echo ""
+		if [ "${PASSWORD}" == "" ]; then
+			echo "Password Removed"
+			exit 0
+		fi
+		if [[ ${#PASSWORD} -le 7 ]]; then
+			echo "Password should be atleast 8 characters long ! "
+			exit 1;
+		elif [[ "${PASSWORD}" != *[[:upper:]]* ]]; then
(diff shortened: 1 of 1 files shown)


diginc:
I had left a comment about maybe using `cracklib-check` previously.  I really like it because then we don't have to write an opinionated re-implementation of how to make a strong password in our code base.  I was imagining a flow like:

> Do you want to set a web password? [ Y / N ]

Y

> User enters password once 

any cracklib warnings show up at this stage, similar to when changing a linux password

> User entered password again to confirm

> The password you entered is weak because of \<cracklib-check output minus the password\>
> Tips for a strong password:  ...
> 
> Press Y to confirm you want to use a weak password <if Y proceed>
> Press N to attempt a different, stronger, password <if N loop back to beginning>

Some example cracklib output:

```
$ echo "1a2z" | cracklib-check
1a2z: it is too short
$ echo "pass123" | cracklib-check
pass123: it is based on a dictionary word
$ echo "pass1234" | cracklib-check
pass1234: OK
$ echo "JDSjfdjjeJJEJE3332#" | cracklib-check
JDSjfdjjeJJEJE3332#: OK
```

sri786:
@dschaper I have updated the terms and agreements (template), and signed the CLA agreement. sorry to miss that earlier.

@diginc since it is a basic validation, i did not go for crack-lib check. And, user might need to have knowledge about cracklib also.

dschaper:
Thanks for the submission and the template fill out. Here are my thoughts:

This needs to be optional in my view. I like the idea of strong passwords, but I use pass phrases that are way more secure than just requiring something along the lines of PAM enforcement. (The time to crack `p4SSw0rd!` is way less than `This is my passphrase` and the phrase is easier to remember and type. But that is a discussion on use cases, and the discussion of strong passwords vs pass phrases has been in the news lately as the author of the original work that suggested obfuscated passwords has said it's just a bad idea.) 

 I like diginc's flow, let the user enter a password and then alert them that it may be insecure and if they would still like to use it, then at least they know it is insecure. For someone that is using it for just themselves or for trusted users that have the option to access the control panel, they don't need strong password requirements and forcing them to use strong passwords would end up with them just disabling passwords completely. 

As for user knowledge of `crack-lib`, they wouldn't see the back end of how the entropy was validated so they wouldn't need to know anything about it. 

So, I'm not a fan of hard coded requirements for passwords as it's not something that I see as being helpful for a majority of our users. If the person running the Pi-hole wants a strong password, they will just set it up that way. But I do like giving the users the option to have a strong password and I do like notifying users that they may be using a weak password, and still allow them to use weak if they should decide that is how they want to run their Pi-hole. We do not endorse or suggest using the Pi-hole in an open resolver or on a publicly accessible interface, and we go out of our way to let people know that we do not suggest using the Pi-hole in an insecure way.

Can you give us some of your thoughts on why we should enforce and require strong passwords without the option of bypassing that choice?

robinagandhi:
> @dschaper: Can you give us some of your thoughts on why we should enforce and require strong passwords without the option of bypassing that choice?

Here are a few reasons to consider:
* Enforcement of strong passwords may deter brute-force attempts by worms or other malware seeking to pivot in a home network. 
* There is something to be said about best-practices for cybersecurity that should be followed regardless of context. Just like good physical hygiene. Options to set a weak password may reinforce bad cybersecurity practices on part of users. In this case, a large password length does not seem to be a burden for end user as they will have a full keyboard at their disposal.
* Advanced users who really care about cybersecurity may be turned off by the lack of any enforcement of password complexity. May suggest similar lax behaviors throughout the product.

I work with @sri786 on this project.

dschaper:
Thank you for the comments, we are taking them under advisement and will discuss as a team. At this point I am not in agreement with the current process. My view is as stated. Complex passwords are nice and we can alert users to the use of less than secure passwords, but I'm not going to force them on users. Alerting users to the state of their passwords shows that we are indeed concerned about the usage of insecure passwords and shows that we are security minded and that is enough for me. I will not approve the PR as it stands currently, but I can be overridden by the other developers.

robinagandhi:
Agreed. Notifying users about the state of their password will be a good compromise.

dschaper:
That I would have no problem with, as with Diginc's approach. The reason I like `crack-lib` instead of a custom coded check function is that `crack-lib` is a tried and true tested utility that provides us with a level of assurance that there isn't something that may have been inadvertently overlooked. It also provides us with some feedback on why the password is not a good choice without having to add more bulk to the bash code to cover all of the conditions under which the password would fail a check.

sri786:
@dschaper @diginc  Thanks for the detailed explanation and the reasons stated regarding the crack-lib over the use the usage of custom coded function. Agreed, use of crack-lib instead of custom coded validation can be a better choice. It also help users to learn what kind of password is most secure against threats.

Mcat12:
@sri786 Since the discussion seems to conclude with a consensus on using cracklib, do you plan on updating this PR, or should it be closed?

dschaper:
Stale. Closing.

PR #1811, opened 2017-12-08
check NetworkManager status prior to using the cli (#1653)

### LINKED ISSUE #1653: CentOS 7 installer crashes out if nmcli is not available
The CentOS/Fedora portion of the installer should do a basic sanity check to verify the NetworkManager service is running, if not it should start the service for the user via:` systemctl start NetworkManager`

Currently if the service is installed but not running the installer will just crash back to bash with a `Error: NetworkManager is not running.` Error message which leaves the user to search the web for an answer.

bcambl:

---

Install script checks for `nmcli` prior to reloading sysconfig interface files but fails when NetworkManager is not running.

Closes #1653

Instead of changing the state of the `NetworkManager` service, we will check the status via `nmcli`.
The reasoning for this approach:
 -  protect users who do not want to use `NetworkManager`.
 - `NetworkManager` is not a requirement of the project and does not need to be enabled
 - We only want to reload the network information from the sysconfig files if `NetworkManager` is enabled.
 - We already checked for the `nmcli` command, so check the status using `nmcli`. (avoid checking `systemctl` vs `service`)

Line comment has been updated.

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -709,8 +709,8 @@ setStaticIPv4() {
       }> "${IFCFG_FILE}"
       # Use ip to immediately set the new address
       ip addr replace dev "${PIHOLE_INTERFACE}" "${IPV4_ADDRESS}"
-      # If NetworkMangler command line interface exists,
-      if command -v nmcli &> /dev/null;then
+      # If NetworkMangler command line interface exists and ready to mangle,
+      if command -v nmcli &> /dev/null && nmcli general status &> /dev/null; then
         # Tell NetworkManagler to read our new sysconfig file
         nmcli con load "${IFCFG_FILE}" > /dev/null
       fi


automated install/basic-install.sh
```diff
@@ -709,8 +709,8 @@ setStaticIPv4() {
       }> "${IFCFG_FILE}"
       # Use ip to immediately set the new address
       ip addr replace dev "${PIHOLE_INTERFACE}" "${IPV4_ADDRESS}"
-      # If NetworkMangler command line interface exists,
-      if command -v nmcli &> /dev/null;then
+      # If NetworkMangler command line interface exists and ready to mangle,
```
jacobsalmela: Glad to see you used the correct name here.

pralor:

https://discourse.pi-hole.net/t/pi-hole-v3-2-1-released-with-lots-of-fixes/6096/1

PR #1851, opened 2017-12-16
[WIP] Pi-hole core v3.2.1

dschaper:
Release v3.2.1


DL6ER:
Pi-hole core v3.2.1 should **not** be released without having modified the debugger to contain **both** a `head` and a `tail` of `dnsmasq`'s and `FTL`'s log files

dschaper:
Closing due to change in policy.

PR #1856, opened 2017-12-21
Pi-hole core v3.2.1

dschaper:
Release for version 3.2.1


.stickler.yml
```diff
@@ -0,0 +1,3 @@
+linters:
```
PromoFaux: Commenting to approve

Issue #1909, opened 2018-01-13
Error at bottom of gravity.list: Binary file /etc/pihole/list.preEventHorizon matches


StarPicard:
**In raising this issue, I confirm the following:** `{please fill the checkboxes, e.g: [X]}`

- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the the source code relevant to this issue?:**

`4`

---
**Expected behaviour:**

The file `gravity.list` should be a normal host-file.

**Actual behaviour:**

The file `gravity.list` is ended by a  `Binary file /etc/pihole/list.preEventHorizon matches` errror.

**Steps to reproduce:**

Currently I have the following host-files added:
```
https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts
http://localhost/adblock.hosts
http://mirror1.malwaredomains.com/files/justdomains
http://sysctl.org/cameleon/hosts
https://zeustracker.abuse.ch/blocklist.php?download=domainblocklist
https://s3.amazonaws.com/lists.disconnect.me/simple_tracking.txt
https://s3.amazonaws.com/lists.disconnect.me/simple_ad.txt
https://hosts-file.net/ad_servers.txt
https://raw.githubusercontent.com/crazy-max/WindowsSpyBlocker/master/data/hosts/win10/spy.txt
http://securemecca.com/Downloads/hosts.txt
https://raw.githubusercontent.com/quidsup/notrack/master/trackers.txt
https://raw.githubusercontent.com/Dawsey21/Lists/master/main-blacklist.txt
https://raw.github.com/notracking/hosts-blocklists/master/hostnames.txt
https://media.kuketz.de/lists/mobile_list.txt
https://www.netguard.me/hosts
https://hosts-file.net/download/hosts.txt
```
... and some special own lists on localhost. 
When I run pihole -g the error occurs.

**Troubleshooting undertaken, and/or other relevant information:**

Adding` -a` to grep at line 452 in gravity.sh helps me:
```
[...]

PR #1910, opened 2018-01-13
Potential fix for #1909

PromoFaux:

---

When a domain contains accented letters/unicode (e.g `www.hualañe.cl`), the current method of using `grep` prints `Binary file /etc/pihole/list.preEventHorizon matches` into `gravity.list`, whereas this prints the actual domain

using `comm` instead of `grep`. However, I'm unsure if solution passes muster, as it is essentially the same as using `grep` with the `-a` switch. I think. I'm no expert, I'll leave it down to the scrutiny of those with more knowledge than me!

n/a

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -449,7 +449,7 @@ gravity_Whitelist() {
   echo -ne "  ${INFO} ${str}..."
 
   # Print everything from preEventHorizon into whitelistMatter EXCEPT domains in $whitelistFile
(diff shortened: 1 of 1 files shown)


StarPicard:
Have checked it manually on my pi and it seems to solve the problem of [#1909](#1909 )
I would recommend replacing [line 333](https://github.com/pi-hole/pi-hole/blob/876170767c72d3db9b319214593d0a8921773a78/gravity.sh#L333) as well.

dschaper:
Fixes #1909

PR #1954, opened 2018-02-07
Prevent redundant entries from being added to Adlists.list

ryanknapper:
**By submitting this pull request, I confirm the following:** `{please fill any appropriate checkboxes, e.g: [X]}`

`{Please ensure that your pull request is for the 'development' branch!}`

- [] I have commented my proposed changes within the code.

---

Presently a user can add redundant entries to adlists.list, each of which Gravity will check for updates.

A simple grep for the proposed change. If true, do nothing.

None needed.

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -182,6 +182,10 @@ trust-anchor=.,20326,8,2,E06D44B80B8F1D39A95C0B0D7C65D08458E880409BBC68345710423
 
 		add_dnsmasq_setting "interface" "${PIHOLE_INTERFACE}"
 	fi
+	if [[ "${CONDITIONAL_FORWARDING}" == true ]]; then
+		add_dnsmasq_setting "server=/${CONDITIONAL_FORWARDING_DOMAIN}/${CONDITIONAL_FORWARDING_IP}"
+		add_dnsmasq_setting "server=/${CONDITIONAL_FORWARDING_REVERSE}/${CONDITIONAL_FORWARDING_IP}"
+	fi
 
 }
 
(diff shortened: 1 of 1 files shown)


DL6ER:
Does this look for exact matches? I don't think so. The likeliness of the following might be small, but assume I have a list

http://blocking-list-provider.com/tracker

Assume now, they also offer a list like

http://blocking-list-provider.com/tracker-german

I think with your proposed solution, the latter one cannot be added if the former is already present. I might be mistaken here (I'm not in the position to test this right now), but it is something I would like to be through about.

jacobsalmela:
I didn't think of this, but it still appears to work with in the situation you described.
screenshot 2018-02-18 16 08 17

advanced/Scripts/webpage.sh
```diff
@@ -346,7 +346,9 @@ CustomizeAdLists() {
 	elif [[ "${args[2]}" == "disable" ]]; then
 		sed -i "\\@${args[3]}@s/^http/#http/g" "${list}"
 	elif [[ "${args[2]}" == "add" ]]; then
-		echo "${args[3]}" >> ${list}
+		if [[ $(grep -c "${args[3]}" "${list}") -eq 0 ]] ; then
```
DL6ER: You should probably be doing an *exact* comparison here to avoid the situation I described above from happening. Just use the proper anchors here:
```bash
grep -c "^${args[3]}$" "${list}"
```

ryanknapper:
Suggested corrections made.

Mcat12:
Do we ant to finish this PR? If so, @DL6ER, please re-review.

PR #2025, opened 2018-03-06
Disable dnsmasq FTLDNS

PromoFaux:
Internal

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1112,8 +1112,7 @@ installConfigs() {
 stop_service() {
   # Stop service passed in as argument.
   # Can softfail, as process may not be installed when this is called
-  local str="Stopping ${1} service"
-  echo ""
+  local str="Stopping ${1} service"  
   echo -ne "  ${INFO} ${str}..."
   if command -v systemctl &> /dev/null; then
     systemctl stop "${1}" &> /dev/null || true
@@ -1126,8 +1125,7 @@ stop_service() {
 # Start/Restart service passed in as argument
 start_service() {
   # Local, named variables
-  local str="Starting ${1} service"
-  echo ""
+  local str="Starting ${1} service"  
   echo -ne "  ${INFO} ${str}..."
   # If systemctl exists,
   if command -v systemctl &> /dev/null; then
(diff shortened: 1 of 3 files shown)


automated install/basic-install.sh
```diff
@@ -202,21 +200,20 @@ elif command -v rpm &> /dev/null; then
...
   fi
 
-  # Fedora and family update cache on every PKG_INSTALL call, no need for a separate update.
+  # Fedora and family update cache on every PKG_INSTALL call, no need for a separate update. 
   UPDATE_PKG_CACHE=":"
   PKG_INSTALL=(${PKG_MANAGER} install -y)
   PKG_COUNT="${PKG_MANAGER} check-update | egrep '(.i686|.x86|.noarch|.arm|.src)' | wc -l"
   INSTALLER_DEPS=(dialog git iproute net-tools newt procps-ng)
-  PIHOLE_DEPS=(bc bind-utils cronie curl findutils nmap-ncat sudo unzip wget libidn2 psmisc)
+  PIHOLE_DEPS=(bc bind-utils cronie curl findutils nmap-ncat sudo unzip wget libidn2 psmisc libnettle.so.4)
```
DL6ER: I don't think we need `libnettle` now any longer as we ship with it

PromoFaux: Ah wait, yeah OK I'll get rid of that

automated install/basic-install.sh
```diff
@@ -1754,44 +1770,78 @@ FTLinstall() {
...
+    
+    # Make the tempory binary executable so that we can test the --resolver flag
+    chmod +x "${binary}"
+
+    # If the --resolver flag returns True (exit code 0), then we can safely stop & disable dnsmasq
+    if ./"${binary}" --resolver > /dev/null; then
+      if [[ $(which dnsmasq 2>/dev/null) ]]; then
+        stop_service dnsmasq
+        disable_service dnsmasq                
+        #ensure /etc/dnsmasq.conf contains `conf-dir=/etc/dnsmasq.d`
```
DL6ER: Are we sure that this line is now always there? Assume there was never a `dnsmasq` install, then I think this would be skipped, no?

PromoFaux: The whole if block will be skipped if `dnsmasq` was not installed. However, `dnsmasq` _may` be pre-installed, and therefore we need to make sure that line exists

PR #2108, opened 2018-04-13
Fix sed: can't read /etc/pihole/pihole-FTL.conf: No such file or directory

LudovicRousseau:

---
Each night since the upgrade to version v3.3.1 the script ``piholeLogFlush.sh`` generates the following error sent by cron in an email:
```
Subject: Cron <root@PiHole>    PATH="$PATH:/usr/local/bin/" pihole flush once quiet

sed: can't read /etc/pihole/pihole-FTL.conf: No such file or directory
```

The script now checks that the file ``/etc/pihole/pihole-FTL.conf`` exists before trying to read it.

I also changed the regular expression used to parse the file ``/etc/pihole/pihole-FTL.conf``. It was not working (for me).

 No documentation change is needed.

### DIFF
--- a/advanced/Scripts/piholeLogFlush.sh
+++ b/advanced/Scripts/piholeLogFlush.sh
@@ -17,7 +17,9 @@ source ${colfile}
 # a) the setting is not present in the config file, or
 # b) the setting is commented out (e.g. "#DBFILE=...")
 FTLconf="/etc/pihole/pihole-FTL.conf"
-[ -e "$FTLconf" ] && DBFILE="$(sed -n -e 's/^\s*DBFILE\s*=\s*//p' ${FTLconf})"
+if [ -e "$FTLconf" ]; then
+  DBFILE="$(sed -n -e 's/^\s*DBFILE\s*=\s*//p' ${FTLconf})"
+fi
 # Test for empty string. Use standard path in this case.
 if [ -z "$DBFILE" ]; then
   DBFILE="/etc/pihole/pihole-FTL.db"
--- a/advanced/Scripts/piholeLogFlush.sh
+++ b/advanced/Scripts/piholeLogFlush.sh
@@ -17,7 +17,7 @@ source ${colfile}
 # a) the setting is not present in the config file, or
 # b) the setting is commented out (e.g. "#DBFILE=...")
 FTLconf="/etc/pihole/pihole-FTL.conf"
-[ -e "$FTLconf" ] && DBFILE="$(sed -n -e 's/^\s*DBFILE\s*=\s*//p' $FTLconf)"
+[ -e "$FTLconf" ] && DBFILE="$(sed -n -e 's/^\s*DBFILE\s*=\s*//p' ${FTLconf})"
 # Test for empty string. Use standard path in this case.
 if [ -z "$DBFILE" ]; then
   DBFILE="/etc/pihole/pihole-FTL.db"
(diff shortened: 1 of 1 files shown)


advanced/Scripts/piholeLogFlush.sh
```diff
@@ -16,7 +16,8 @@ source ${colfile}
 # Constructed to return nothing when
 # a) the setting is not present in the config file, or
 # b) the setting is commented out (e.g. "#DBFILE=...")
-DBFILE="$(sed -n -e 's/^\s^.DBFILE\s*=\s*//p' /etc/pihole/pihole-FTL.conf)"
+FTLconf="/etc/pihole/pihole-FTL.conf"
+[ -e "$FTLconf" ] && DBFILE="$(sed -n -e 's/^\s*DBFILE\s*=\s*//p' $FTLconf)"
```
dschaper: Just a stylistic issue, not a functional change, but can you set the variable expansion to `${FTLconf}`?

LudovicRousseau: Done in a new commit.

PromoFaux:
Hi there,

This has actually already been fixed in the `FTLDNS` branch (which is due to be merged into `development` soon™ [here](https://github.com/pi-hole/pi-hole/commit/82d5afe9961a7964bc22e70f44ec8fdd504fa855), although I think I prefer this version maybe. As you are testing for the file before setting `DBFILE` rather than just silencing the error...

Any additional thoughts @dschaper ?

dschaper:
I like this one as it's in the same style as the other `if` clauses in the file. I'd like to keep them the same instead of having a couple of different styles. Functionally I think they all work, it's really just a matter of style and readability.

There's only one other change I'd make, and that is to break that `[ ] && <code>` into an:

```
if [ ]; then
  code
fi
```

dschaper:
Need to test it later on today when I can but it looks good.

jacobsalmela:
We'll have to take this into account because we were thinking of using this bit of sed code in the debugger.

advanced/Scripts/piholeLogFlush.sh
```diff
@@ -16,7 +16,10 @@ source ${colfile}
 # Constructed to return nothing when
 # a) the setting is not present in the config file, or
 # b) the setting is commented out (e.g. "#DBFILE=...")
-DBFILE="$(sed -n -e 's/^\s^.DBFILE\s*=\s*//p' /etc/pihole/pihole-FTL.conf)"
```
dschaper: I think this has an errant `^`, so the new replacement regex works better.

dschaper:
For the debugger config grabber I'd use the regex suggested in this PR. Should handle the cases we need it to. 

Quoting a commit message that was collapsed, just for record:

```
The regex ^\s^.DBFILE\s*=\s* does not work to match "DBFILE="
I don't know what the second '^' is used for.

With the change I now have the results:

DBFILE=/foo/bar
-> /foo/bar

  DBFILE=/foo/bar
-> /foo/bar

 # DBFILE=/foo/bar
-> /etc/pihole/pihole-FTL.db

xDBFILE=/foo/bar
-> /etc/pihole/pihole-FTL.db

Signed-off-by: Ludovic Rousseau <ludovic.rousseau@free.fr>
```

pralor:

https://discourse.pi-hole.net/t/pihole-ftl-conf-no-such-file-or-directory-its-back/11188/5

PR #2112, opened 2018-04-14
Install either init.d script or systemd unit depending on used init

DL6ER:

---
Replace the ancient `init.d` script for starting and controlling the `pihole-FTL` process by a modern `systemd` unit. This should resolve some problems as discussed [here](https://discourse.pi-hole.net/t/service-pihole-ftl-active-exited/8335) or [here](https://discourse.pi-hole.net/t/pihole-ftl-is-working-but-webmin-says-the-service-is-not-running/8368).

This now also makes the `status` command behave correctly, as `systemd` can use the PID file of FTL to find out if it is actually running or not:
screenshot at 2018-04-14 11-24-00

Also, since *FTL*DNS is a very important part of your system, the unit ensures that it will be automatically restarted whenever something unexpected happens. As an example, I sent `SIGKILL` to FTL and as you can see, `systemd` took care about immediately restarting the service:
screenshot at 2018-04-14 11-27-08
Note that we do not restart `pihole-FTL` automatically after having received `SIGTERM` as this is the correct way of terminating the resolver (e.g. through `service pihole-FTL stop`).

I wrote a systemd unit for `pihole-FTL` and modified the installer to replace the old init.d script.

All commands should work as before.


pralor:

https://discourse.pi-hole.net/t/service-pihole-ftl-active-exited/8335/8

pralor:

https://discourse.pi-hole.net/t/pihole-ftl-is-working-but-webmin-says-the-service-is-not-running/8368/18

advanced/pihole-FTL.service
```diff
@@ -1,89 +1,38 @@
...
+
+[Service]
+Restart=on-abnormal
+User=root
+Group=root
+
+Type=forking
+PIDFile=/run/pihole-FTL.pid
+
+ExecStartPre=/bin/bash /etc/.pihole/advanced/pihole-FTL-prestart.sh
```
Mcat12: Should this be installed somewhere instead of referencing the cloned repository?

advanced/pihole-FTL-prestart.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+/bin/touch /var/log/pihole-FTL.log /run/pihole-FTL.pid /run/pihole-FTL.port /var/log/pihole.log
+/bin/mkdir -p /var/run/pihole /var/log/pihole
+/bin/chown pihole:pihole /var/run/pihole /var/log/pihole
+if [ -e "/var/run/pihole/FTL.sock" ]; then
+  /bin/rm /var/run/pihole/FTL.sock
+fi
+/bin/chown pihole:pihole /var/log/pihole-FTL.log /run/pihole-FTL.pid /run/pihole-FTL.port /etc/pihole /etc/pihole/dhcp.leases /var/log/pihole.log
+/bin/chmod 0644 /var/log/pihole-FTL.log /run/pihole-FTL.pid /run/pihole-FTL.port /var/log/pihole.log
+#/sbin/setcap "CAP_NET_BIND_SERVICE,CAP_NET_RAW,CAP_NET_ADMIN+eip" "/usr/bin/pihole-FTL"
```
Mcat12: If this is only run during install, then this can be removed.

test/test_automated_install.py
```diff
@@ -310,15 +310,16 @@ def test_FTL_download_unknown_fails_no_errors(Pihole):
...
-def test_FTL_binary_installed_and_responsive_no_errors(Pihole):
-    ''' confirms FTL binary is copied and functional in installed location '''
-    installed_binary = Pihole.run('''
-    source /opt/pihole/basic-install.sh
-    FTLdetect
-    pihole-FTL version
-    ''')
-    expected_stdout = 'v'
-    assert expected_stdout in installed_binary.stdout
+# Temporarily disabled as we cannot use setcap on Travis CI
```
Mcat12: Will this be re-enabled when we use CircleCI then?

dschaper:
This may be a breaking change, replacing init scripts with systemd services would cause issues with some widely used, but not currently supported distributions. There is an open PR for Gentoo support and that is not a systemd distribution by default. Also some of the chroot jail installs rely on init scripts. Will need to do some investigation and get some feedback from the Synology users to see if we need to do both, or just stay with init for the time being.

test/test_automated_install.py
DL6ER: Yes

advanced/pihole-FTL-prestart.sh
DL6ER: I will remove it

advanced/pihole-FTL.service
DL6ER: Should it? It is something a user shouldn't change, I guess

Mcat12: It should be easy to pop into `/opt/pihole`

test/test_automated_install.py
Mcat12: If it's running in the docker container, woudn't it be fine to run on Travis?

DL6ER:
Marking this WIP for the moment as we have to clarify when to use `systemd` and when `init.d`. My suggestion so far is:
```
$ file $(which init) | grep "systemd" &> /dev/null
$ echo $?
0
```

test/test_automated_install.py
DL6ER: Travis doesn't like us playing with `setcap`

DL6ER:
Thank you very much for your comments. I did try `AmbientCapabilities` but had problems with it not being recognized on Raspbian (Jessie).

Thank you very much for suggesting `PermissionsStartOnly`. I wasn't aware of this instruction. It is exactly what I was looking for and will help getting rid of the `su` call (although I haven't had any problem with it).

> Just my 2¢ on the changes, although I get that drive-by comments on pull requests often aren't that helpful given the need to understand the surrounding context.

This PR is sort of stand-alone and not really embedded into any greater picture. I'd happily read any other comments from you and will work on your suggestions. I'd be really happy if you could review my upcoming changes (expect them in a few minutes).

DL6ER:
@hvindin I wasn't successful with

```
ExecStartPre=/sbin/setcap cap_net_bind_service,cap_net_raw,cap_net_admin+eip /usr/bin/pihole-FTL
```
(I commented out the exact line I used if you want to have a look)

After `enable` + `restart`, the job failed with the following message:
```
Apr 21 16:42:02 raspberrypi setcap[30353]: Failed to set capabilities on file `/usr/bin/pihole-FTL' (Invalid argument)
Apr 21 16:42:02 raspberrypi setcap[30353]: The value of the capability argument is not permitted for a file. Or the file is not a regular (non-symlink) file
```
---
It appears that previously I didn't use `AmbientCapabilities` correctly (I assumed a comma-, not space-separated list). This actually seems to work. I'm currently testing more and will push a commit if no problems show up. Do you know which would be the minimum version of `systemd` needed?
---

One other thing: Do you know what it means by the first line?
```
Apr 21 16:43:18 raspberrypi systemd[1]: pihole-FTL.service: Supervising process 30694 which is not our child. We'll most likely not notice when it exits.
Apr 21 16:43:18 raspberrypi systemd[1]: Started Pi-hole FTLDNS.
Apr 21 16:43:18 raspberrypi pihole-FTL[30691]: FTL started!
```

hvindin:
Ok, well I think there may be one issue there, I'm running this on a pi with:
```
PRETTY_NAME="Raspbian GNU/Linux 9 (stretch)"
NAME="Raspbian GNU/Linux"
VERSION_ID="9"
VERSION="9 (stretch)"
ID=raspbian
ID_LIKE=debian
HOME_URL="http://www.raspbian.org/"
SUPPORT_URL="http://www.raspbian.org/RaspbianForums"
BUG_REPORT_URL="http://www.raspbian.org/RaspbianBugs"
```
and doing a bit of testing for an outdated systemd install on my laptop running:
```
NAME="Red Hat Enterprise Linux Workstation"
VERSION="7.5 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VARIANT="Workstation"
VARIANT_ID="workstation"
VERSION_ID="7.5"
PRETTY_NAME="Red Hat Enterprise Linux"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.5:GA:workstation"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.5
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.5"
```

But it looks like that's not quite synonymous enough.

Can you have a look at this gist: [Alternatively-pihole-FTL.service](https://gist.github.com/hvindin/969ac05f1e3181d3caf0f49e4403bb6a)

I've added the systemd service I used on my pi running stretch (ie. including the AmbientCapabilities line) as well as the screen dump of me starting from a clean slate, wiping my raspberry pi and reinstalling pihole, checking out this branch and just removing the init services and systemd stuff and adding my own thing to /lib/systemd/system/pihole-FTL.service

The service which I added to my stretch installation is [This One](https://gist.github.com/hvindin/969ac05f1e3181d3caf0f49e4403bb6a#file-pihole-ftl-service)

and the one which seems to work on my laptop running rhel (ie with ancient systemd) is [This One](https://gist.github.com/hvindin/969ac05f1e3181d3caf0f49e4403bb6a#file-alternatively-pihole-ftl-service)  

Unrelated to that specific problem, I also got rid of that bash script that was being executed as part of the startup, IMO it makes the systemd service opaque, if there are a bunch of startup tasks you need to run PreExec then that's what the PreExec stuff is for, so why not just use it?  

Anyway, have a look at those files as they might help.  

The only major difference I can think of is that you have:
```
ExecStartPre=/sbin/setcap cap_net_bind_service,cap_net_raw,cap_net_admin+eip /usr/bin/pihole-FTL
```

Whereas I have:
```
ExecStartPre=/sbin/setcap cap_net_bind_service,cap_net_raw,cap_net_admin=+eip /usr/bin/pihole-FTL
```
(note: the = sign)

Both of those configurations seem to work ok for me.

hvindin:
The other thing it's probably worth doing if we're using systemd is adding
```
Wants=lighttpd.service
```

To make sure that's started when pihole-FTL starts up, but I'm not that familiar with how everything is hooked together so that might be a redundant step.

mbiebl:
While a native .service file is obviously the best solution, there is a trivial way to make the SysV init script work better by using the ```# pidfile:``` pseudo header.
```
index 959b779..5762e61 100755
--- a/init.d/pihole-FTL
+++ b/init.d/pihole-FTL
@@ -8,6 +8,7 @@
 # Short-Description: pihole-FTL daemon
 # Description:       Enable service provided by pihole-FTL daemon
 ### END INIT INFO
+# pidfile: /var/run/pihole-FTL.pid
 
 FTLUSER=pihole
 PIDFILE=/var/run/pihole-FTL.pid
```
This results then in
```
# systemctl status pihole-FTL.service 
● pihole-FTL.service - LSB: pihole-FTL daemon
   Loaded: loaded (/etc/init.d/pihole-FTL; generated; vendor preset: enabled)
   Active: active (running) since Sat 2018-05-12 04:24:26 CEST; 4s ago
     Docs: man:systemd-sysv-generator(8)
  Process: 5934 ExecStop=/etc/init.d/pihole-FTL stop (code=exited, status=0/SUCCESS)
  Process: 5954 ExecStart=/etc/init.d/pihole-FTL start (code=exited, status=0/SUCCESS)
 Main PID: 5977 (pihole-FTL)
    Tasks: 0 (limit: 19660)
   CGroup: /system.slice/pihole-FTL.service
           ‣ 5977 /usr/bin/pihole-FTL
```

DL6ER:
@mbiebl Thanks for your comments! It seems that inside our team, the idea of going away from the init.d script (even if implemented in the way suggested here to choose either the old or the new script).

As I don't see this merged any time soon, we decided to change the behavior to not fork at all, but have the init.d script keep control over the process to be able to report if it is running or not: https://github.com/pi-hole/FTL/pull/274

mbiebl:
If you use non-forking mode in the SysV init script, this will have undesired behaviour when running init script under SysV, as the script will never return. That's not how SysV init scripts usually work.

That said, please reconsider adding a native systemd .service file.

DL6ER:
Okay, so let's try the following: I'll reopen and try to motivate this PR more in our internal team.

Could you check if you find this systemd unit suitable?
There is still the uncertainty about the `setcap` but how I did it (using a script in contrast to `AMBIENT_CAPABILITIES`, etc.) works on all platforms I have tested.

Having said that, I will also add the `# pidfile: /var/run/pihole-FTL.pid` line you suggested.

DL6ER:
> the lockdown

What do you mean? We need to ensure a few things regarding ownership and permissions I would like to not have an intermediate version that runs `pihole-FTL` as `root` because we don't have the proper permission control in place.

mbiebl:
Some high level comments: 
- I'd probably use a tmpfile instead of a hand crafted script to setup the runtime folders
- I'd let systemd do the user switching via ```User=``` instead of using a combination of su+bash. This is frowned upon.

If time permits, I'll do a more thorough review.

automated install/basic-install.sh
```diff
@@ -1754,6 +1754,11 @@ FTLinstall() {
   # Move into the temp ftl directory
   pushd "$(mktemp -d)" > /dev/null || { echo "Unable to make temporary directory for FTL binary download"; return 1; }
 
+  # Remove old init.d script if present as it cannot coexist with the systemd unit we are installing here
```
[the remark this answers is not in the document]
DL6ER: Okay, still, I think it is cleaner to have either or. I will rephrase this comment.

mbiebl:
> What do you mean? We need to ensure a few things regarding ownership and permissions I would like to not have an intermediate version that runs pihole-FTL as root because we don't have the proper permission control in place.

You can let the main process (The one started via ExecStart=) run as unprivileged user and do the setup  either via a tmpfile or a separate ExecStartPre= which, can be run with root privileges.

automated install/basic-install.sh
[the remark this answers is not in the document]
mbiebl: Make sure to clean up the symlinks in /etc/rc?.d/ as well when you do the clean up

DL6ER:
> I'd let systemd do the user switching via User= instead of using a combination of su+bash. This is frowned upon.

This we already do (we use `User` and `PermissionsStartOnly=true`). Only in the old init.d script we use `su`.

Any review is highly appreciated at any time!

mbiebl:
What's the reason for ```PermissionsStartOnly=true```? That's rather uncommon and i would drop that.

DL6ER:
As I understand it this gives us `root` permissions on the `ExecStartPre=` line so we can arrange all permissions when needed (and we needed to ensure this almost always), whereas the `ExecStart=/usr/bin/pihole-FTL` is run under the `User=` defined user.

mbiebl:
Ah right, you need root privs for ```ExecStartPre=/bin/bash /opt/pihole/pihole-FTL-prestart.sh```
I don't think you strictly need it for ```ExecReload=``` though.
An alternative to ```PermissionsStartOnly=true``` would be, to only mark the Exec lines you need to run with elevated privileges using the '!' executable prefix. 
See https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStart=
I personally like that syntax a bit better, as it makes it more obvious, which parts of the .service file need elevated privs.

I guess the mkdir/chown bits from pihole-FTL-prestart.sh could be implemeted using a tmpfile. That would only leave ```/bin/echo "nameserver 127.0.0.1" | /sbin/resolvconf -a lo.piholeFTL``` and the which needs to run with root privs. The sock file removal doesn't strictly need root privs.

mbiebl:
As for the pid file:
It would be nicer to create a ```/run/pidhole/``` directory owned by pihole:pihole and then the unprivileged daemon can write into this directory

advanced/pihole-FTL.systemd
```diff
@@ -0,0 +1,40 @@
+[Unit]
+Description=Pi-hole FTLDNS
+; This unit is supposed to indicate when network functionality is available, but it is only
+; very weakly defined what that is supposed to mean, with one exception: at shutdown, a unit
+; that is ordered after network.target will be stopped before the network
+After=network.target
+Requires=network.target
```
mbiebl: The Requires line should be dropped.
network.target is a so-called passive target (see man systemd.special) and should not be pulled in actively by units.
It is pulled in by the service implementing network.target. Other units should simply order itself against network.target.

advanced/pihole-FTL.systemd
```diff
@@ -0,0 +1,40 @@
...
+Type=forking
+PIDFile=/run/pihole-FTL.pid
+
+ExecStartPre=/bin/bash /opt/pihole/pihole-FTL-prestart.sh
+;ExecStartPre=/sbin/setcap cap_net_bind_service,cap_net_raw,cap_net_admin+eip /usr/bin/pihole-FTL
+ExecStart=/usr/bin/pihole-FTL
+ExecReload=/bin/kill -HUP $MAINPID
+
+; Use graceful shutdown with a reasonable timeout
+KillMode=mixed
```
mbiebl: We only spawn a single pihole-FTL process, right? Any reason we need to override the default ```KillMode=``` and use ``mixed```?

automated install/basic-install.sh
```diff
@@ -1754,8 +1754,25 @@ FTLinstall() {
   # Move into the temp ftl directory
   pushd "$(mktemp -d)" > /dev/null || { echo "Unable to make temporary directory for FTL binary download"; return 1; }
 
-  # Always replace pihole-FTL.service
-  install -T -m 0755 "${PI_HOLE_LOCAL_REPO}/advanced/pihole-FTL.service" "/etc/init.d/pihole-FTL"
+  # Determine if systemd is used on this system
+  if file "$(which init)" | grep "systemd" &> /dev/null; then
```
mbiebl: The canonical check whether systemd is the active init system is
``` test -e /run/systemd/system```
See https://www.freedesktop.org/software/systemd/man/sd_booted.html

The only case I can imagine where this would fail is, if you install pihole in a chroot.

File: automated install/basic-install.sh [hunk 8]
```diff
@@ -1754,8 +1754,25 @@ FTLinstall() {
...
+    # Use systemd unit
+    # Always replace pihole-FTL.service (systemd unit)
+    install -T -m 0644 "${PI_HOLE_LOCAL_REPO}/advanced/pihole-FTL.systemd" "/etc/systemd/system/pihole-FTL.service"
+    install -T -m 0755 "${PI_HOLE_LOCAL_REPO}/advanced/pihole-FTL-prestart.sh" "/opt/pihole/pihole-FTL-prestart.sh"
+
+    # Remove old init.d script if present as it cannot coexist with the systemd unit we are installing here
+    if [ -e "/etc/init.d/pihole-FTL" ]; then
+      rm "/etc/init.d/pihole-FTL"
```
mbiebl: This will leave dangling /etc/rc?.d/pihole-FTL symlinks.
Those should be cleaned up using ```update-rc.d pihole-FTL remove```

mbiebl:
On a related note: I notice that the SysV init script uses ```su``` to start the pihole-FTL process. It would be better if it uses runuser for that. See https://danwalsh.livejournal.com/55588.html

DL6ER:
I tried prefacing the `ExecStartPre` lines like
```
ExecStartPre=!/bin/touch /var/log/pihole-FTL.log /run/pihole-FTL.pid /run/pihole-FTL.port /var/log/pihole.log
```
however, it doesn't like the `!` at the beginning:
```
[/etc/systemd/system/pihole-FTL.service:21] Executable path is not absolute, ignoring: /bin/touch /var/log/pihole-FTL.log /run/pihole-FTL.pid /run/pihole-FTL.port /var/log/pihole.log
... the same for the other lines ...
```

without the `!` and with `PermissionsStartOnly=true` everything works fine

advanced/pihole-FTL.systemd
```diff
@@ -20,13 +19,19 @@ PermissionsStartOnly=true
...
 PIDFile=/run/pihole-FTL.pid
 
-ExecStartPre=/bin/bash /opt/pihole/pihole-FTL-prestart.sh
-;ExecStartPre=/sbin/setcap cap_net_bind_service,cap_net_raw,cap_net_admin+eip /usr/bin/pihole-FTL
+ExecStartPre=/bin/touch /var/log/pihole-FTL.log /run/pihole-FTL.pid /run/pihole-FTL.port /var/log/pihole.log
+ExecStartPre=/bin/mkdir -p /var/run/pihole /var/log/pihole
+ExecStartPre=/bin/chown pihole:pihole /var/run/pihole /var/log/pihole
+ExecStartPre=-/bin/rm /var/run/pihole/FTL.sock
+ExecStartPre=/bin/chown pihole:pihole /var/log/pihole-FTL.log /run/pihole-FTL.pid /run/pihole-FTL.port /etc/pihole /etc/pihole/dhcp.leases /var/log/pihole.log
+ExecStartPre=/bin/chmod 0644 /var/log/pihole-FTL.log /run/pihole-FTL.pid /run/pihole-FTL.port /var/log/pihole.log
```
mbiebl: Wouldn't it be much simpler, if there was a single /var/log/pihole/ directory. This directory would be created with the proper permissions, so the unprivileged pihole-FTL process can write into it. It should then be sufficient to create this directory once during installation.
As for the files in /run, having a /run/pihole directory with the proper permissions where all runtime files are stored would be much simpler as well. This directory could be managed by systemd directly via
```RuntimeDirectory=```. systemd would take care of creating that directory, chowning it, and cleaning it up when the service is stopped.

Comment (mbiebl) [reply advanced/pihole-FTL.systemd]: https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory=

**Comment by DL6ER** [comment]
> Wouldn't it be much simpler, if there was a single /var/log/pihole/ directory. This directory would be created with the proper permissions, so the unprivileged pihole-FTL process can write into it.

Yes, I'm very much in favor of this. However, we will always have to apply such changes to all repositories and should do it in separate pull requests.

File: automated install/basic-install.sh (hunk 7 above)
Comment (hvindin) [reply automated install/basic-install.sh]: I have actually tested running pihole on a synology NAS, I ended up just buying another pi and dedicating it to the purpose of being my DNS device.

If one was to assume that supporting pihole on synology NAS's (or really any proprietary platform) should be a desirable outcome, then there are only really two options:

1. Put a bunch of effort into make pihole run in a container (most things nowadays run docker in some form or another so that gets it into a known sandbox where it can do its thing)

2. Actually publish native packages for those proprietary environments, ie. synology NAS's basically use weirdly formed init services that are written as json things (I haven't looked particularly closes so could very well be very very wrong on this) and they are all handled by synoservicectl just to make things more difficult. So actually publishing a synology package, signing it and putting it into a synology package repo would be the most native way to do this.

File: advanced/pihole-FTL.systemd (hunk 9 above)
Comment (hvindin) [reply advanced/pihole-FTL.systemd]: I'd actually made the assumption that all the setup stuff that get's done every time pihole starts is to make up for the fact that it's not actually "installed" per se, it's just a script that needs to do it's own set up every time.

So it can never assume that the relevant directories exist nor that the relevant caps are set on binaries or that anything has been set up at all really, hence the need to do so much setup as root then try and drop privs down to a non-root user. I mean, might as well make this thing compilable in some language to get a binary and throw CAP_SYS_ADMIN onto it... just seems like a bunch of hacks which seem to work for now because conventional approaches to doing things seem to have all gone out the window 

If it where possible ideally there would be .deb .rpm .whatever packages that could handle the creation of /var/log/pihole /var/run/pihole /etc/pihole /whateverelsewe need as well as handing any permission munging or, alternatively, have the process run as a proper executable binary that can depermission itself once it's done its thing (similar to dnsmasq, runs as root but the process very clearly drops most of the privileges it has pretty quickly if you strace a startup).

It feels like making pihole work with systemd means making pihole work as an actual linux service that follows all associated conventions, which is a hell of an overhaul.

File: automated install/basic-install.sh (hunk 7 above)
Comment (hvindin) [reply automated install/basic-install.sh]: Also, probably worth noting, my synology ds716+II doesn't have `file` installed on it by default, so you can't just use file to check what the init system is on a synology nas because it just won't execute it.

You could something like:

```
if stat $(which init) | grep -q systemd; then
...
```
with the if file thing I get
```
hvindin@diskstation:~$ if file $(which init) | grep -q systemd; then echo "systemd"; else echo "not systemd"; fi
bash: file: command not found
not systemd
hvindin@diskstation:~$ if stat $(which init) | grep -q systemd; then echo "systemd"; else echo "not systemd"; fi
not systemd
```

But then I'm not sure if stat is more or less common than file... It's on all by debian, centos, rhel, synology etc. boxes but there are a shit load of distros and file might just be the more common thing to have installed by default.

But in particular on purpose built hardware being repurposed I feel like it's less likely that common nixy utilities are going to be present.

**Comment by pralor** [comment]

https://discourse.pi-hole.net/t/minor-issue-with-ftldns/10092/2

**Comment by dschaper** [comment]
Just to put this one to bed, let's rename the `.systemd` file to `.service` and just include the file as an example of what works. I've review a couple of users implementing the service file in systemd and have had no issues with it.

**Comment by DL6ER** [comment]
@dschaper Is this still the preferred way to go?

I'll fix the merge conflicts later today.

**Comment by dschaper** [comment]
There are a lot of unresolved discussions going on with this PR. It no longer does what the PR statement says it does, and I'm not convinced that we should even drop sysvinit support and force systemd, (if that still what this PR hopes to accomplish, I'm not exactly sure what it's doing anymore?)

**Comment by DL6ER** [comment]
The code this PR wants to add detects if the currently used init system is `sytemd`. If so, we replace the old script by the new one. If not, we copy the init.d file as before.

The main motivation for this PR was to fix several issue reports, commenting that `service pihole-FTL status` failed on more recent versions of `systemd`.

We should decide if we want to continue this (if this asks for reopening this PR for a new/updated description that's okay with me). If not, we should close this PR. Keeping it open until we're forced to switch at some point (maybe because some future version of `systemd` does not accept init.d scripts any longer).

**Comment by dschaper** [comment]
My immediate thought is to close this and start a new one that is clear in intent and direction. But I don't want to lose the insight from @mbiebl and his very constructive comments. 

I don't see the need to either/or this, `systemd` "is compatible with SysV and LSB init scripts." (from Debian Wiki) and changes to that would break a ton of systems. 
>From systemd.service(5):

```
If a service is requested under a certain name but no unit configuration
file is found, systemd looks for a SysV init script by the same name
(with the .service suffix removed) and dynamically creates a service
unit from that script. This is useful for compatibility with SysV. Note
that this compatibility is quite comprehensive but not 100%. For details
about the incompatibilities, see the Incompatibilities with SysV[1]
document.
```
(https://www.turnkeylinux.org/blog/debugging-systemd-sysv-init-compat)

DL6ER:
This PR may be reopened in the future, please do not delete the branch `FTLDNS-systemd` for future reference.

mbiebl:
That systemd supports legacy init scripts does not mean one should not provide native .service files.
While support for legacy init scripts is probably not going away anytime soon in systemd, you get a ton of benefits if you provide a .service file which is tailored to your specific service.

dschaper:
I completely agree, my point is that there's no need to have either .service or init.d, it's okay to have both.

mbiebl:
You mean installed on the target system or shipped by pi-hole?
This PR does keep both sysvinit and systemd .service file, it simply installs only one of them on the target system. Do you want to always have both installed?

dschaper:
If someone want's to go through all the open conversations and reviews and resolve them that would clear things up. I'm just going by my gut feeling and the age of this PR and saying it will never be approved in it's current state. Too many unanswered questions and changes in approach or goals. 

Yes, both should be installed unless you can, with certainty, know an installs init system. `chroot` systems will have misleading PIDs for init and cause confusion. There is no danger in having an init.d and a .service file as systemd's sysv-init compatibility handles that.

mbiebl:
Gotcha. It's indeed possible to install a /etc/init.d/foo and /lib/systemd/system/foo.service. systemd will always pick the native service, if the names match.
The only issue is, that when users are executing ```/etc/init.d//etc/init.d/pihole-FTL start``` directly, it would start a (second) instance which is not under the supervision of systemd. And you really want to avoid that situation.

Most distros I know intercept such direct executions of init scripts and will translate a ```/etc/init.d/foo start``` into a ```systemctl start foo.service```.
E.g. in Debian it's done by sourcing ```. /lib/lsb/init-functions``` which in turn calls /lib/lsb/init-functions.d/40-systemd which does the trickery.
Unfortunately, there is no distro agnostic way to write such a sysv init script.
I guess the best you could do is to have a ```[ -e /run/systemd/system ] && exit 0``` at the start of the init script, so it will become a nop if directly executed when systemd is the active PID 1.

mbiebl:
> The LSB shim is pretty universal from what I understand. Debian has it's package for it, there's an Arch package and I think it's standard on any platform that deploys systemd.

Unfortunately, this is not the case :-/
E.g. under Debian/Ubuntu the file that needs to be sourced is named ```/lib/lsb/init-functions```, under Fedora it is named ```/etc/init.d/functions``` and so on.

dschaper:
Great... In that case it may be beneficial to do something like `if exists then source`. But there may be a more elegant solution out there, we can't be the first to run in to this.

DL6ER:
I recall that this was an issue. Some distros even need to source something to provide the `status` command. Others (at least the ones we officially support) do not. This was one of the reasons for starting this PR in the first place.

mbiebl:
> Great... In that case it may be beneficial to do something like `if exists then source`. But there may be a more elegant solution out there, we can't be the first to run in to this.

Written portable init scripts which also properly integrate into distros is hard if not impossible, at least that is my experience.

mbiebl:
What pi-hole currently does is providing a portable init script, which is not very well integrated.

beren12:
Please don't only support systemd. Some of us like a choice and don't need the bloat/overhead/bugs that systemd brings, esp. in a chroot or lightweight vm. Lots of projects have an init script that works pretty well across distros. Lots of times a distro adds a native package or volunteers provide a tailored script for different system bases, like debian/rh/etc.

DL6ER:
> Please don't only support systemd.

I wonder where such rumors come from. This was never the plan. My first proposed solution used a rather sophisticated testing if the system does not only *have* systemd but is also actively *using* it as init system and then install *either* the systemd unit *or* the previously used init.d script. My newer proposal is to just install both and hand over the responsibility to the system to pick the one it likes most.

beren12:
> 
> On Nov 30, 2018, at 12:39 AM, DL6ER <notifications@github.com> wrote:
> 
> Please don't only support systemd.
> 
> I wonder where such rumors come from. This was never the plan. My first proposed solution used a rather sophisticated testing if the system does not only have systemd but is also actively using it as init system and then install either the systemd unit or the previously used init.d script. My newer proposal is to just install both and hand over the responsibility to the system to pick the one it likes most.
> 
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub, or mute the thread.



Sorry i read the last dozen comments and that seemed where it was going. I’ve used too many projects that go systemd and dump everything else. I realize it’s a closed bug I just wanted to put my vote in :-)

I just googled to see if systemd was required and your bug report came up. I’m planning on making a small pihole vm on my bsd firewall and I want it as lean as possible

hvindin:
If one wanted to allow the use of systemd to manage pihole, but the installation got to complicated given the range of distros and vendor custom stuff (looking at you synology) that pihole is commonly run on, and also to quash any thoughts of "only systemd" being a thing, perhaps it wouldn't be unreasonable to throw the systemd service file into the repo somewhere, or even just in some docs with a 'how to replace the init.d script with a systemd service'.

Obviously its a little more involved than the idea of automatically working out what needs to go where and which init system is in use, but personally I quite like being able to define my services as systemd units, although given how many people I personally know and how many blogs are around the internet by people who seem to absolutely hate systemd, its presumably a bit to polarising to make a default anywhere...

advanced/pihole-FTL.systemd
dschaper: That's a fair statement, and if you wanted to open a new Feature Request on https://discourse.pi-hole.net asking for the inclusion of a systemd service file in the pi-hole/pi-hole repo, I think that we most likely would be able to satisfy it.

dschaper:
That's a fair statement, and if you wanted to open a new Feature Request on https://discourse.pi-hole.net asking for the inclusion of a systemd service file in the pi-hole/pi-hole repo, I think that we most likely would be able to satisfy it.

PR #2119, opened 2018-04-15
add some more linters to stickler

PromoFaux:
Internal


PromoFaux:
Pretty sure this is as simple as it needs to be to start off with. There are some more advanced options, but I guess we just see what we need as the needs change..

Issue #1955, opened 2018-02-07
Block page or http://pi.hole returns HTTP Error 500


petererdos:
**In raising this issue, I confirm the following:** `{please fill the checkboxes, e.g: [X]}`

- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [x] The issue I am reporting can be *replicated*.
- [x] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the the source code relevant to this issue?:**
1

---
**Expected behaviour:**

Block page should load when an offending domain is hit; http//pi.hole should load the welcome screen

Instead of the block page, i get http error 500 instead of the block page, or if i try to access http://pi.hole:
This page isn’t working
pi.hole is currently unable to handle this request.
HTTP ERROR 500

![screenshot_6](https://user-images.githubusercontent.com/14872721/35945283-54c8dc8e-0c68-11e8-9556-27cdc3ec0a50.png)

Errors in the log, which appear when loading http://pi.hole or block page:
2018-02-06 20:09:20: (mod_fastcgi.c.434) FastCGI-stderr: PHP Fatal error: Can't use function return value in write context in /var/www/html/pihole/index.php on line 108

**Steps to reproduce:**

Install Pi-hole on CentOS 7.4 (fully updated):
centos-release-7-4.1708.el7.centos.x86_64
Kernel: Linux 3.10.0-693.17.1.el7.x86_64
SELinux disabled

Try to open: http://pi.hole or access a blocked domain with (tried) Edge and latest Google Chrome (from Win10 & Server 2016)

Issue: HTTP Error 500 page

**Debug token provided by [uploading `pihole -d` log](https://discourse.pi-hole.net/t/the-pihole-command-with-examples/738#debug):**

0c0r6bk4a2

**Troubleshooting undertaken, and/or other relevant information:**

pihole -up
pihole -r
.. although without effect.

Then I edited this in /var/www/html/pihole/index.php:
`if (empty(glob("/etc/pihole/list.0.*.domains")))`
Into this:
`if (!glob("/etc/pihole/list.0.*.domains"))`

[...]

PR #2176, opened 2018-05-11
Fix block page crash on PHP 5.4

rrobgill:

---
Prevent the pi-hole block page from returning 500 / server error on systems (such as Centos 7) which ship with old versions of PHP (5.4)

#1955 and [AdminLTE#746](https://github.com/pi-hole/AdminLTE/issues/746)

Avoid empty(glob()) which does not work under PHP < 5.5.

Avoiding calling empty() on a function allows this to work under PHP5. 
Making the check for blocklist generation in this way instead is compatible with both PHP5 and PHP7.

None. 

Signed-off-by: Rob Gill <rrobgill@protonmail.com>

### DIFF
--- a/advanced/index.php
+++ b/advanced/index.php
@@ -103,7 +103,7 @@ $bpAskAdmin = !empty($svEmail) ? '<a href="mailto:'.$svEmail.'?subject=Site Bloc
 
 // Determine if at least one block list has been generated
 $blocklistglob = glob("/etc/pihole/list.0.*.domains");
-if($blocklistglob === array()) {
+if ($blocklistglob === array()) {
     die("[ERROR] There are no domain lists generated lists within <code>/etc/pihole/</code>! Please update gravity by running <code>pihole -g</code>, or repair Pi-hole using <code>pihole -r</code>.");
 }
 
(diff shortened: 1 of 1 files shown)


advanced/index.php
```diff
@@ -102,8 +102,10 @@ function setHeader($type = "x") {
 $bpAskAdmin = !empty($svEmail) ? '<a href="mailto:'.$svEmail.'?subject=Site Blocked: '.$serverName.'"></a>' : "<span/>";
 
 // Determine if at least one block list has been generated
-if (empty(glob("/etc/pihole/list.0.*.domains")))
+$blocklistglob = glob("/etc/pihole/list.0.*.domains");
+if ($blocklistglob = "") {
```
DL6ER: `$blocklistglob = ""` will set `$blocklistglob` to `""`. This will always succeed hence the condition is always fulfilled. I guess you want `==` or (even better) `===`. Note that with `===` you will have to test for the exact match, so e.g. `if($blocklistglob === array())` to test against an empty array.

pralor:

https://discourse.pi-hole.net/t/new-install-block-page-not-displaying/10169/5

Issue #2179, opened 2018-05-14
DNS resolution fails on Ubuntu 18.04


RlndVt:
**In raising this issue, I confirm the following:** `{please fill the checkboxes, e.g: [X]}`

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the the source code relevant to this issue?:**

1

---
**Expected behaviour:**

Working pihole

**Actual behaviour:**

Install on (l)ubuntu 18.04 fails to activate DNS resolution. Installation fails on:

`[✗] DNS resolution is currently unavailable`

Need to manually change the nameserver in `/etc/resolv.conf` to 127.0.0.53 to regain DNS resolution. 

This allows for normal internet on the machine, but the pihole DNS resolution still doesn't work and cannot be started.

**Steps to reproduce:**

Standard/default installation options. 

**Debug token provided by [uploading `pihole -d` log](https://discourse.pi-hole.net/t/the-pihole-command-with-examples/738#debug):**

`62rwn2e69i`

**Troubleshooting undertaken, and/or other relevant information:**

rrobgill:
A more complete solution is to stop resolved from binding to port 53 (dnsmasq/ftl and resolved can't share the same port).  #2180 

This allows pihole dns resolution to work correctly (both for host and also other machines on the network) and allows normal internet connectivity on the host machine.

RlndVt:
Fwiw, with systemd-resolved stopped and/or with "DNSStubListener=no", dnsmasqd will not enable.

```
~ sudo systemctl stop systemd-resolved 
~ ping google.com
ping: google.com: Temporary failure in name resolution
~ sudo systemctl start dnsmasq.service   
Job for dnsmasq.service failed because the control process exited with error code.
See "systemctl status dnsmasq.service" and "journalctl -xe" for details.
```

With the same error, being unable to bind to port 53.

DL6ER:
[...]

PR #2180, opened 2018-05-14
Disable SystemD's DNS Stub Resolver On Install

rrobgill:
This deals with #2179

Allow pi-hole to work under Ubuntu Bionic 18.04

Stops systemd/resolved from binding to port 53.

None

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1266,21 +1266,28 @@ check_service_active() {
 }
 
 # Systemd-resolved's DNSStubListener and dnsmasq can't share port 53. 
-# Resolved needs to remain in place for installer to download needed files
+# Resolved needs to remain in place for installer to download needed files,
 # so this change needs to be made after installation is complete, but before resarting dnsmasq/ftl
 disable_resolved_stublistener() {
+  echo -en "  ${INFO} Testing if systemd-resolved is enabled"
   # Check if Systemd-resolved's DNSStubListener is enabled and active on port 53
   if check_service_active "systemd-resolved"; then
     # Check if DNSStubListener is enabled
+    echo -en "  ${OVER}  ${INFO} Testing if systemd-resolved DNSStub-Listener is active"
     if ( grep -E '#?DNSStubListener=yes' /etc/systemd/resolved.conf &> /dev/null ); then
       # Disable the DNSStubListener to unbind it from port 53
       # Note that this breaks dns functionality on host until dnsmasq/ftl are up and running
-      echo -e "Disabling systemd-resolved DNSStubListener"
+      echo -en "${OVER}  ${TICK} Disabling systemd-resolved DNSStubListener"
       # Make a backup of the original /etc/systemd/resolved.conf
       # (This will need to be restored on uninstallation)
       sed -r -i.orig 's/#?DNSStubListener=yes/DNSStubListener=no/g' /etc/systemd/resolved.conf
+      echo -e " and restarting systemd-resolved"
       systemctl reload-or-restart systemd-resolved
+    else
+      echo -e "${OVER}  ${INFO} Systemd-resolved does not need to be restarted"
     fi
+  else
+    echo -e "${OVER}  ${INFO} Systemd-resolved is not enabled"
   fi
 }
 
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1279,7 +1279,7 @@ disable_resolved_stublistener() {
       # Make a backup of the original /etc/systemd/resolved.conf
       # (This will need to be restored on uninstallation)
       sed -r -i.orig 's/#?DNSStubListener=yes/DNSStubListener=no/g' /etc/systemd/resolved.conf
-      systemctl restart systemd-resolved
+      systemctl reload-or-restart systemd-resolved
     fi
   fi
 }
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -2315,7 +2315,7 @@ main() {
     if (systemctl is-enabled systemd-resolved | grep -c 'enabled' || true); then
       # if resolveconf is running unbind it from port 53
       # Note that this breaks dns functionality on host until dnsmasq/ftl are up and running
-      echo -e "Unbinding resolved from port 53"
+      echo -e "Disabling systemd-resolved DNSStubListener"
       # Make a backup of the original /etc/systemd/resolveconf.d
       # (This will need to be restored on uninstallation)
       sed -i.orig 's/#DNSStubListener=yes/DNSStubListener=no/g' /etc/systemd/resolved.conf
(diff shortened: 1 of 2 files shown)


automated install/basic-install.sh
```diff
@@ -2304,6 +2304,25 @@ main() {
     fi
   fi
 
+  # resolved and dnsmasq can't share port 53. 
+  # resolved needs to remain in place for installer to download needed files
+  # so this change needs to be made after installation is complete, but before resarting dnsmasq/ftl
+
+  # Check if running ubuntu 18.04 bionic beaver, which ships with resolved active on port 53
+  # (This check may need to be broadened for other systems running resolved?)
+  if ( lsb_release -a | grep 'Ubuntu 18.04' &> /dev/null ); then
```
DL6ER: Why is this check needed at all (it feels very hacky)? Couldn't we just check if `systemd` is used as init and then if `systemd-resolved` is enabled?

automated install/basic-install.sh
```diff
@@ -2304,6 +2304,25 @@ main() {
...
+    # Running ubuntu 18.04, so check if resolved is running,
+    if (systemctl is-enabled systemd-resolved | grep -c 'enabled' || true); then
```
DL6ER: Is `systemctl` guaranteed to reply in English? Mind that there are tons of devices with German, Spanish, Dutch, ... locales and most Linux commands are actually capable of responding in the set locale. We should rely on return codes instead of return strings

automated install/basic-install.sh
```diff
@@ -2304,6 +2304,25 @@ main() {
...
+      # if resolveconf is running unbind it from port 53
+      # Note that this breaks dns functionality on host until dnsmasq/ftl are up and running
+      echo -e "Unbinding resolved from port 53"
```
DL6ER: Maybe the message should rather be something like `Disabling systemd-resolved DNSStubListener`? (this is just an idea, not a necessary request for change)

automated install/basic-install.sh
```diff
@@ -2304,6 +2304,25 @@ main() {
...
+      # Make a backup of the original /etc/systemd/resolveconf.d
+      # (This will need to be restored on uninstallation)
+      sed -i.orig 's/#DNSStubListener=yes/DNSStubListener=no/g' /etc/systemd/resolved.conf
+      systemctl restart systemd-resolved
```
DL6ER: Why do we not simply disable the service? What does it offer after we set `DNSStubListener=no`?

automated install/basic-install.sh
```diff
@@ -2304,6 +2304,25 @@ main() {
     fi
   fi
 
+  # resolved and dnsmasq can't share port 53. 
+  # resolved needs to remain in place for installer to download needed files
+  # so this change needs to be made after installation is complete, but before resarting dnsmasq/ftl
```
DL6ER: How about moving this into its own function and calling this inside `start_service pihole-FTL`?

automated install/basic-install.sh
rrobgill: That's intended as a stopgap "please don't break any other distro" measure. 

As systemd continues it's march onwards, this will certainly be an issue in other distros. If they keep /etc/systemd/resolved.conf where it belongs then there shouldn't be a problem with skipping the ubuntu check, but I'd err on the cautious side for the moment.

automated install/basic-install.sh
rrobgill: I was wondering about that. Could it be splitting it in two? One called early on to check for the presence of enabled resolved and set a flag there. Then just check that flag in the new start_servicepihole-FTL. Rather than having it tested and applied in the same section? It would allow more granularity in the test if it is needed without cluttering up the body of the script, and might make it cleaner to adjust things should any distros require any special tweaks.

automated install/basic-install.sh
rrobgill: done

automated install/basic-install.sh
```diff
@@ -1265,6 +1265,25 @@ check_service_active() {
...
 }
 
+# Systemd-resolved's DNSStubListener and dnsmasq can't share port 53. 
+# Resolved needs to remain in place for installer to download needed files
+# so this change needs to be made after installation is complete, but before resarting dnsmasq/ftl
+disable_resolved_stublistener() {
+  # Check if Systemd-resolved's DNSSTub listener is enabled and active on port 53
+  if check_service_active "systemd-resolved"; then
+    # Check if DNSStubListener is enabled
+    if ( grep '#DNSStubListener=yes' /etc/systemd/resolved.conf &> /dev/null ); then
```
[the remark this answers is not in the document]
rrobgill: Fixed to match either way.

automated install/uninstall.sh
```diff
@@ -165,6 +165,13 @@ removeNoPurge() {
 	${SUDO} rm -f /etc/sudoers.d/pihole &> /dev/null
   echo -e "  ${TICK} Removed config files"
 
+  # Restore resolved 
+  if [[ -e /etc/systemd/resolved.conf.orig ]]; then
+    systemctl disable systemd-resolved
+    ${SUDO} cp /etc/systemd/resolved.conf.orig /etc/systemd/resolved.conf
+    systemctl enable systemd-resolved
```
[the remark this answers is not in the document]
rrobgill: My understanding is that stop/start doesn't alwats enforce reloading of the conf file. But changing to reload-or-restart is probably neater than enable/disable.

automated install/basic-install.sh
```diff
@@ -1265,6 +1265,25 @@ check_service_active() {
...
+# Resolved needs to remain in place for installer to download needed files
+# so this change needs to be made after installation is complete, but before resarting dnsmasq/ftl
+disable_resolved_stublistener() {
+  # Check if Systemd-resolved's DNSStubListener is enabled and active on port 53
+  if check_service_active "systemd-resolved"; then
+    # Check if DNSStubListener is enabled
+    if ( grep -E '#?DNSStubListener=yes' /etc/systemd/resolved.conf &> /dev/null ); then
+      # Disable the DNSStubListener to unbind it from port 53
+      # Note that this breaks dns functionality on host until dnsmasq/ftl are up and running
+      echo -e "Disabling systemd-resolved DNSStubListener"
```
DL6ER: @PromoFaux @WaLLy3K Is this in agreement with your `${TICK}` magic in the installer?

PromoFaux: It'll stick out like a sore thumb if the tick/info/cross stuff isn't included, that's for sure!

rrobgill:
After discussion in #2179, in addition to the previous testing against Ubuntu 18.04, these changes have now also been tested against Lubuntu 18.04 (clean install), and also against an ubuntu 14.04 install updated to 16.04 and then 18.04. 

Master branch with these changes installed successfully on all without additional intervention. After running and testing master, they were changed over to the beta FTLDNS branch (via checkout as per discourse 8224) and continued working. (example debug tokens from the 14.04>>18.04 system - master : lbfv5f3jql - ftldns : voio2v46qu)

pralor:

https://discourse.pi-hole.net/t/dnsmasq-or-starting-ubuntu/9486/7

Mcat12:
@PromoFaux @DL6ER please re-review.

Issue #1805, opened 2017-12-07
QueryFunc is leaking errors from grep.


dschaper:
**In raising this issue, I confirm the following:** `{please fill the checkboxes, e.g: [X]}`

- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [x] The issue I am reporting can be *replicated*.
- [x] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the the source code relevant to this issue?:**

`8`

---
**Expected behaviour:**

`pihole -q` should just return usable information

**Actual behaviour:**

```
root@ubuntu-512mb-nyc3-01:/var/log# pihole -q -all google.com
grep: blacklist.txt: No such file or directory
grep: *.domains: No such file or directory
  [i] No results found for google.com found within block lists
```

**Steps to reproduce:**

`pihole -q`

Mcat12:
Introduced by https://github.com/pi-hole/pi-hole/commit/c2d3e99ddd7a3fbe65d2cfc2ebb8e5ad917b78de

DL6ER:
While we're at it, the whole `query` functionality should be moved into a separate file as we did also with nearly all other functions that are bulkier than a few lines.

dschaper:
Fixed in #2234

jacobsalmela:
This is fixed in v4.0

PR #2234, opened 2018-06-13
Fix grep query leak & relocate query

rrobgill:
- [] I have commented my proposed changes within the code.

---

Fixes the bug from #1805 
Remove the list query function (eg pihole -q) from the main pihole script

Leak fixed by redirecting stderr to null when calling grep.

Main pihole script now calls query.sh, which contains the code from queryFunc() and scanlist(), formerly in the main script.

None.

---


advanced/Scripts/query.sh
```diff
@@ -0,0 +1,222 @@
...
+Options:
+  -adlist             Print the name of the block list URL
+  -exact              Search the block lists for exact domain matches
+  -all                Return all query matches within a block list
+  -h, --help          Show this help dialog"
+  exit 0
+fi
+
+if [[ ! -e "$adListsList" ]]; then
+  echo -e "${COL_LIGHT_RED}The file '/etc/pihole/adlists.list' was not found${COL_NC}"
```
Mcat12: Should use `$adListsList`

PR #2236, opened 2018-06-15
Use regex lists instead of wildcards for blocking

DL6ER:

---

Change the CLI to support wildcard blocking through *FTL*DNS's regex filters. This is much more flexible than what we used before (defining zones in a `dnsmasq` config file) and replaces this method.

Use regex filters instead of building a `dnsmasq` config file.

Help and man pages.

---

### DIFF
--- a/advanced/Scripts/list.sh
+++ b/advanced/Scripts/list.sh
@@ -138,7 +138,7 @@ AddDomain() {
     bool=true
     domain="${1}"
 
-    [[ "${wildcard}" == true ]] && domain="((^)|(\\.))$(echo domain | sed "s/\./\\\./g")$"
+    [[ "${wildcard}" == true ]] && domain="((^)|(\\.))$(echo "${domain}" | sed "s/\./\\\./g")$"
 
     # Is the domain in the list?
     # Search only for exactly matching lines
@@ -186,7 +186,7 @@ RemoveDomain() {
     [[ -z "${type}" ]] && type="--wildcard-only"
     domain="${1}"
 
-    [[ "${wildcard}" == true ]] && domain="((^)|(\\.))$(echo domain | sed "s/\./\\\./g")$"
+    [[ "${wildcard}" == true ]] && domain="((^)|(\\.))$(echo "${domain}" | sed "s/\./\\\./g")$"
 
     bool=true
     # Is it in the list?
--- a/advanced/Scripts/list.sh
+++ b/advanced/Scripts/list.sh
@@ -222,7 +222,7 @@ Displaylist() {
     verbose=false
     echo -e "Displaying $string:\n"
     count=1
-    while IFS= read -r RD; do
+    while IFS= read -r RD || [ -n "${RD}" ]; do
       echo "  ${count}: ${RD}"
       count=$((count+1))
     done < "${listMain}"
--- a/advanced/Scripts/list.sh
+++ b/advanced/Scripts/list.sh
@@ -138,7 +138,7 @@ AddDomain() {
     bool=true
     domain="${1}"
 
-    [[ "${wildcard}" == true ]] && domain="((^)|(\\.))${domain}$"
+    [[ "${wildcard}" == true ]] && domain="((^)|(\\.))$(echo domain | sed "s/\./\\\./g")$"
 
     # Is the domain in the list?
     # Search only for exactly matching lines
@@ -186,7 +186,7 @@ RemoveDomain() {
     [[ -z "${type}" ]] && type="--wildcard-only"
     domain="${1}"
 
-    [[ "${wildcard}" == true ]] && domain="((^)|(\\.))${domain}$"
+    [[ "${wildcard}" == true ]] && domain="((^)|(\\.))$(echo domain | sed "s/\./\\\./g")$"
 
     bool=true
     # Is it in the list?
--- a/advanced/Scripts/list.sh
+++ b/advanced/Scripts/list.sh
@@ -17,6 +17,7 @@ readonly regexlist="/etc/pihole/regex.list"
 reload=false
 addmode=true
 verbose=true
+wildcard=false
 
 domList=()
 
@@ -68,7 +69,7 @@ HandleOther() {
 
   # Check validity of domain (don't check for regex entries)
   if [[ "${#domain}" -le 253 ]]; then
-    if [[ "${listMain}" == "${regexlist}" ]]; then
+    if [[ "${listMain}" == "${regexlist}" && "${wildcard}" == false ]]; then
       validDomain="${domain}"
     else
       validDomain=$(grep -P "^((-|_)*[a-z\\d]((-|_)*[a-z\\d])*(-|_)*)(\\.(-|_)*([a-z\\d]((-|_)*[a-z\\d])*))*$" <<< "${domain}") # Valid chars check
@@ -135,6 +136,9 @@ AddDomain() {
   elif [[ "${list}" == "${regexlist}" ]]; then
     [[ -z "${type}" ]] && type="--wildcard-only"
     bool=true
+
+    [[ "${wildcard}" == true ]] && domain="((^)|(\\.))${domain}$"
+
     # Is the domain in the list?
     # Search only for exactly matching lines
     grep -Fx "${domain}" "${regexlist}" > /dev/null 2>&1 || bool=false
@@ -179,6 +183,8 @@ RemoveDomain() {
     fi
   elif [[ "${list}" == "${regexlist}" ]]; then
     [[ -z "${type}" ]] && type="--wildcard-only"
+    [[ "${wildcard}" == true ]] && domain="((^)|(\\.))${domain}$"
+
     bool=true
     # Is it in the list?
     grep -Fx "${domain}" "${regexlist}" > /dev/null 2>&1 || bool=false
@@ -236,7 +242,8 @@ for var in "$@"; do
   case "${var}" in
     "-w" | "whitelist"   ) listMain="${whitelist}"; listAlt="${blacklist}";;
     "-b" | "blacklist"   ) listMain="${blacklist}"; listAlt="${whitelist}";;
-    "-wild" | "wildcard" ) listMain="${regexlist}";;
+    "-wild" | "wildcard" ) listMain="${regexlist}"; wildcard=true;;
+    "-regex" | "regex"   ) listMain="${regexlist}";;
     "-nr"| "--noreload"  ) reload=false;;
     "-d" | "--delmode"   ) addmode=false;;
     "-q" | "--quiet"     ) verbose=false;;
--- a/advanced/Scripts/list.sh
+++ b/advanced/Scripts/list.sh
@@ -33,7 +33,7 @@ helpFunc() {
     type="white"
   elif [[ "${listMain}" == "${regexlist}" ]]; then
     param="wild"
-    type="wildcard black"
+    type="regex black"
   else
     param="b"
     type="black"
--- a/advanced/Scripts/list.sh
+++ b/advanced/Scripts/list.sh
@@ -136,6 +136,7 @@ AddDomain() {
   elif [[ "${list}" == "${regexlist}" ]]; then
     [[ -z "${type}" ]] && type="--wildcard-only"
     bool=true
+    domain="${1}"
 
     [[ "${wildcard}" == true ]] && domain="((^)|(\\.))${domain}$"
 
@@ -145,13 +146,13 @@ AddDomain() {
 
     if [[ "${bool}" == false ]]; then
       if [[ "${verbose}" == true ]]; then
-      echo -e "  ${INFO} Adding ${1} to regex list..."
+      echo -e "  ${INFO} Adding ${domain} to regex list..."
       fi
       reload="restart"
-      echo "$1" >> "${regexlist}"
+      echo "$domain" >> "${regexlist}"
     else
       if [[ "${verbose}" == true ]]; then
-        echo -e "  ${INFO} ${1} already exists in regex list, no need to add!"
+        echo -e "  ${INFO} ${domain} already exists in regex list, no need to add!"
       fi
     fi
   fi
@@ -183,6 +184,8 @@ RemoveDomain() {
     fi
   elif [[ "${list}" == "${regexlist}" ]]; then
     [[ -z "${type}" ]] && type="--wildcard-only"
+    domain="${1}"
+
     [[ "${wildcard}" == true ]] && domain="((^)|(\\.))${domain}$"
 
     bool=true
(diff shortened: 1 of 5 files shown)


hoshsadiq:
Wildcards is often enough and usually much faster than regex. IMO this shouldn't replace wildcards but be an addition. Also, need to be careful as this would break people's black/white lists.

DL6ER:
> usually much faster than regex

How did you measure this? In our experiments, using about hundred regex filters didn't took longer than 1msec on a RPi 3. In contrast, `dnsmasq`'s internal lookups for the wildcards as we used them before (configured zones) were much slower They were on the order of 10 - 40msec *per* request whereas regex filtering is only done once per domain in *FTL*DNS, i.e. after a possible "slow" evaluation on first occurrence, it is available as blocked domain in the cache and will be answered blazingly fast. I see only speed benefits however I look at it (unless you may want to use thousands of regex filters, this is something we haven't tried ourselves so far).

> Also, need to be careful as this would break people's black/white lists.

That's the primary reason for this being marked as `[WIP]`. We will add an automated conversion from the existing format to the new one, but it needs time and more testing before it gets committed.

advanced/Scripts/list.sh
```diff
@@ -31,7 +31,7 @@ helpFunc() {
   if [[ "${listMain}" == "${whitelist}" ]]; then
     param="w"
     type="white"
-  elif [[ "${listMain}" == "${wildcardlist}" ]]; then
+  elif [[ "${listMain}" == "${regexlist}" ]]; then
     param="wild"
     type="wildcard black"
```
Mcat12: Do we want to change any of this from "wild"/"wildcard" to something else?

advanced/Scripts/list.sh
```diff
@@ -131,28 +128,22 @@ AddDomain() {
...
-    # Remove the /* from the end of the IP addresses
-    IPV4_ADDRESS=${IPV4_ADDRESS%/*}
-    IPV6_ADDRESS=${IPV6_ADDRESS%/*}
+  elif [[ "${list}" == "${regexlist}" ]]; then
     [[ -z "${type}" ]] && type="--wildcard-only"
     bool=true
     # Is the domain in the list?
-    grep -e "address=\/${domain}\/" "${wildcardlist}" > /dev/null 2>&1 || bool=false
+    # Search only for exactly matching lines
+    grep -E "^${domain}$" "${regexlist}" > /dev/null 2>&1 || bool=false
```
Mcat12: Will having a regex in regex mess things up? You could use `grep -F` to not interpret the regex input as regex.

advanced/Scripts/list.sh
```diff
@@ -183,20 +173,20 @@ RemoveDomain() {
         echo -e "  ${INFO} ${1} does not exist in ${listname}, no need to remove!"
       fi
     fi
-  elif [[ "${list}" == "${wildcardlist}" ]]; then
+  elif [[ "${list}" == "${regexlist}" ]]; then
     [[ -z "${type}" ]] && type="--wildcard-only"
     bool=true
     # Is it in the list?
-    grep -e "address=\/${domain}\/" "${wildcardlist}" > /dev/null 2>&1 || bool=false
+    grep -E "^${domain}$" "${regexlist}" > /dev/null 2>&1 || bool=false
```
Mcat12: See above

advanced/Scripts/list.sh
```diff
@@ -241,7 +231,7 @@ for var in "$@"; do
   case "${var}" in
     "-w" | "whitelist"   ) listMain="${whitelist}"; listAlt="${blacklist}";;
     "-b" | "blacklist"   ) listMain="${blacklist}"; listAlt="${whitelist}";;
-    "-wild" | "wildcard" ) listMain="${wildcardlist}";;
+    "-wild" | "wildcard" ) listMain="${regexlist}";;
```
Mcat12: See above

gravity.sh
```diff
@@ -462,13 +464,9 @@ gravity_ShowBlockCount() {
...
 
-  if [[ -f "${wildcardFile}" ]]; then
-    num=$(grep -c "^" "${wildcardFile}")
-    # If IPv4 and IPv6 is used, divide total wildcard count by 2
-    if [[ -n "${IPV4_ADDRESS}" ]] && [[ -n "${IPV6_ADDRESS}" ]];then
-      num=$(( num/2 ))
-    fi
-    echo -e "  ${INFO} Number of wildcard blocked domains: ${num}"
+  if [[ -f "${regexFile}" ]]; then
+    num=$(grep -c "^" "${regexFile}")
```
Mcat12: Could also use `wc -l` to count lines.

advanced/Scripts/wildcard_regex_converter.sh
```diff
@@ -0,0 +1,28 @@
...
+  fi
+  local addrlines domains uniquedomains
+  # Obtain wildcard domains from old file
+  addrlines="$(grep -oE "/.*/" ${wildcardFile})"
+  # Strip "/" from domain names and convert "." to regex-compatible "\."
+  domains="$(sed 's/\///g;s/\./\\./g' <<< "${addrlines}")"
+  # Remove repeated domains (may have been inserted two times due to A and AAAA blocking)
+  uniquedomains="$(uniq <<< "${domains}")"
+  # Automatically generate regex filters and remove old wildcards file
+  awk '{print "((^)|(\\.))"$0"$"}' <<< "${uniquedomains}" >> "${regexFile:?}" && rm "${wildcardFile}"
```
Mcat12: Might want to check the regex file for duplicates? Shouldn't be very likely.

advanced/Scripts/list.sh
DL6ER:
```
grep -F "${domain}" "${regexlist}"
```
wouldn't this also match if `$domain` is `abc` and `regex.list` contains `abcdefg`?

DL6ER: Problem solved, I should read the `man` page...
```
-x
Select only those matches that exactly match the whole line.
```

gravity.sh
DL6ER: This should cause no harm because `^` is the anchor for line begin, literal `^` in the file would have to be matched with `\^`.

advanced/Scripts/wildcard_regex_converter.sh
DL6ER: Even if they are there they wouldn't harm. *FTL*DNS stops the regex evaluation after the first match.

anudeepND:
@DL6ER some quick questions...
How does it affect whitelist? And what's the performance of regex blocking with lots of wildcard entries..like 1000+?

Mcat12:
@anudeepND Unlike with wildcards, you can whitelist specific domains even if they fall under a regex.

For performance, reference @DL6ER's reply to @hoshsadiq

WaLLy3K:
Conversion from the old format seems to work correctly (e.g: `((^)|(\.))moatads\.com$`), however if you add a new wildcard via the Web Interface, it doesn't appear to be parsed properly: 

```
...
((^)|(\.))moatads\.com$
foo.bar.com
```

DL6ER:
@WaLLy3K Users can add regex filters there, i.e. they'll have to enter `((^)|(\.))foo.bar.com$` instead of `foo.bar.com` if this is what they want.

anudeepND:
@DL6ER just to confirm....both wildcard and regex rules can be entered in the same place?

DL6ER:
@anudeepND Yes and No. Wildcards are removed for the sake of the much more powerful regex filters.

DL6ER:
@WaLLy3K @anudeepND see https://github.com/pi-hole/AdminLTE/pull/772#issuecomment-401537868

anudeepND:
@DL6ER sorry about the confusion....Do I need to add domains in regex format or normally enter `example.net` then system will automatically convert into regex format?

DL6ER:
@anudeepND it depends *where* you want to do it. assume you want to block `example.com` as a wildcard.

If you use

- the CLI (i.e., the `pihole` command) you will have to enter the full regex
- the dash board (i.e., the web interface) there are two buttons. One (Add wildcard) will generate the required regex for you and you can just enter your domain. The other (Add regex) will add exactly what you typed there as a regex filter and you'd have to enter the full regex here.

WaLLy3K:
> @WaLLy3K Users can add regex filters there, i.e. they'll have to enter ((^)|(\.))foo.bar.com$ instead of foo.bar.com if this is what they want.

To avoid user confusion, I would *very strongly* urge an additional option that allows users to continue adding wildcards in the same manner (where adding `foo.bar.com` will parse it as `((^)|(\.))foo\.bar\.com$`)

On top of potential user confusion, most users aren't going to realise that `.` means "any single character" instead of `\.` which is a literal dot - and Regex parsing `.` as any character is going to be far less efficient than `\.`.

I'm thinking a CLI flag & web checkbox that's automatically not selected that says "User specified Regex"?

DL6ER:
@WaLLy3K not sure about your last comment. Does this render **a CLI based method** to add the wildcards in the old style obsolete?

WaLLy3K:
> Does this render a CLI based method to add the wildcards in the old style obsolete?

I'm not sure I'm following your train of thought? So I guess I'll clarify.

I feel the best thing going forward would be if `-wild` via the CLI would use the "old style" wildcard method by wrapping a domain in `((^)|(\.))foo\.bar\.com$`, while a second option (e.g: `-regex`) would allow one to add in a domain via pure regex. This ensures that people's existing workflows aren't interrupted come `master` release, because all the existing behaviour continues to function as they expect.

The Web Interface can then have an option (e.g: a checkbox) that allows one to use the pure regex method.

Mcat12:
That's a good idea, and would avoid the confusion of using `-wild` to add regex filters.

The web interface does include a button for adding "wildcards". It will add a regex filter that emulates the wildcard.

manpages/pihole.8
```diff
@@ -273,11 +278,15 @@ Some usage examples
     Whitelist/blacklist manipulation
 .br
 
-    \fBpihole -w iloveads.example.com\fR  Add "iloveads.example.com" to whitelist
+    \fBpihole -w iloveads.example.com\fR       Add "iloveads.example.com" to whitelist
+.br
+    \fBpihole -b -d noads.example.com\fR       Remove "noads.example.com" from blacklist
 .br
-    \fBpihole -b -d noads.example.com\fR  Remove "noads.example.com" from blacklist 
+    \fBpihole -wild "^example.*$"\fR           Add "^example.*$" as a regex pattern - would
```
DL6ER: @Mcat12 is this description correct? I'd assume that `-wild` needs to be called with just a domain and should not take regex as input

dschaper:
Slating for 4.0

advanced/Scripts/list.sh
```diff
@@ -241,7 +245,8 @@ for var in "$@"; do
   case "${var}" in
     "-w" | "whitelist"   ) listMain="${whitelist}"; listAlt="${blacklist}";;
     "-b" | "blacklist"   ) listMain="${blacklist}"; listAlt="${whitelist}";;
-    "-wild" | "wildcard" ) listMain="${wildcardlist}";;
+    "-wild" | "wildcard" ) listMain="${regexlist}"; wildcard=true;;
```
dschaper: I'd prefer if these were `--wild` and `--regex`. This is a breaking change, but since this is going for 4.0 I think we can break on it.

manpages/pihole.8
```diff
@@ -5,7 +5,7 @@ Pi-hole : A black-hole for internet advertisements
 .br
 .SH "SYNOPSIS"
 
-\fBpihole\fR (\fB-w\fR|\fB-b\fR|\fB-wild\fR) [options] domain(s)
+\fBpihole\fR (\fB-w\fR|\fB-b\fR|\fB-wild\fR|\fB-regex\fR) [options] domain(s)
```
dschaper: `--wild` `--regex`

manpages/pihole.8
```diff
@@ -68,7 +68,12 @@ Available commands and options:
 
 \fB-wild, wildcard\fR [options] [<domain1> <domain2 ...>]
```
dschaper: `--wild`

manpages/pihole.8
```diff
@@ -68,7 +68,12 @@ Available commands and options:
...
 .br
-    Add or removes specified domain, and all subdomains to the blacklist
+    Add or removes specified domain to the wildcard blacklist
+.br
+
+\fB-regex, regex\fR [options] [<regex1> <regex2 ...>]
```
dschaper: `--regex`

manpages/pihole.8
```diff
@@ -273,11 +278,15 @@ Some usage examples
     Whitelist/blacklist manipulation
 .br
 
-    \fBpihole -w iloveads.example.com\fR  Add "iloveads.example.com" to whitelist
+    \fBpihole -w iloveads.example.com\fR       Add "iloveads.example.com" to whitelist
+.br
+    \fBpihole -b -d noads.example.com\fR       Remove "noads.example.com" from blacklist
 .br
-    \fBpihole -b -d noads.example.com\fR  Remove "noads.example.com" from blacklist 
+    \fBpihole -wild example.com\fR             Add example.com as a wildcard - would
```
dschaper: `--wild`

manpages/pihole.8
```diff
@@ -273,11 +278,15 @@ Some usage examples
...
+    block all subdomains of example.com, including example.com itself.
 .br
-    \fBpihole -wild example.com\fR        Add "example.com" as wildcard - would block ads.example.com, www.example.com etc.
+    \fBpihole -regex "ad.*\.example\.com$"\fR  Add "ad.*\.example\.com$" to the regex
```
dschaper: `--regex`

pihole
```diff
@@ -403,7 +393,8 @@ Add '-h' after specific commands for more information on usage
 Whitelist/Blacklist Options:
   -w, whitelist       Whitelist domain(s)
   -b, blacklist       Blacklist domain(s)
-  -wild, wildcard     Blacklist domain(s), and all its subdomains
+  -wild, wildcard     Wildcard blacklist domain(s)
```
dschaper: `--wild`

pihole
```diff
@@ -403,7 +393,8 @@ Add '-h' after specific commands for more information on usage
...
+  -regex, regex       Regex blacklist domains(s)
```
dschaper: `--regex`

pihole
```diff
@@ -445,9 +436,10 @@ fi
 
 # Handle redirecting to specific functions based on arguments
 case "${1}" in
-  "-w" | "whitelist"            ) whitelistFunc "$@";;
-  "-b" | "blacklist"            ) blacklistFunc "$@";;
-  "-wild" | "wildcard"          ) wildcardFunc "$@";;
+  "-w" | "whitelist"            ) listFunc "$@";;
+  "-b" | "blacklist"            ) listFunc "$@";;
+  "-wild" | "wildcard"          ) listFunc "$@";;
```
dschaper: `--wild`

pihole
```diff
@@ -445,9 +436,10 @@ fi
...
+  "-regex" | "regex"            ) listFunc "$@";;
```
dschaper: `--regex`

dschaper:
@jacobsalmela This will need to be documented on the release notes, if people previously were using `-wild` it is now `--wild`.

jacobsalmela:
Thanks for the ping.  I have added it to my draft release notes.

dschaper:
Let's go ahead and move this to development to see if there are any side effects that also need to be adjusted.

WaLLy3K:
We'll need to consider the help menu presented when typing `pihole --wild`, as [the example only shows one dash](https://github.com/pi-hole/pi-hole/blob/037989b8e42434e1ef8ff2489a48d8342a4f27a1/advanced/Scripts/list.sh#L44). This also makes me wonder: do we also want to change `-w` and `-b` to use `--`? If not, fixing the example should be as straight forward as adding another dash when `${param}` equals `wild` or `regex`.

dschaper:
Unix/Linux convention is to use one dash for single char arguments and two dashes for arguments that are more than one char. So `-w` --wild` `-b` `--regex` etc...

PR #2069, opened 2018-03-31
Fix error: /opt/pihole/gravity.sh: 385: Warning: command substitution: ignored null byte in input

jacobsalmela:
**By submitting this pull request, I confirm the following:** 
*please fill any appropriate checkboxes, e.g: [X]*

- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md), as well as this entire template.
- [x] I have made only one major change in my proposed changes.
- [x] I have commented my proposed changes within the code.
- [x] I have tested my proposed changes, and have included unit tests where possible.
- [x] I am willing to help maintain this change if there are issues with it later.
- [x] I give this submission freely and claim no ownership.
- [x] It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
- [x] I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

Please make sure you [Sign Off](https://github.com/pi-hole/pi-hole/wiki/How-to-signoff-your-commits.) all commits. Pi-hole enforces the [DCO](https://github.com/pi-hole/pi-hole/wiki/Contributing-to-the-project).

---
**What does this PR aim to accomplish?:**
Remove this error some users see:

```
/opt/pihole/gravity.sh: 385: Warning: command substitution: ignored null byte in input
```

**How does this PR accomplish the above?:**
Removed null bytes with `tr -d` as mentioned [here](https://stackoverflow.com/questions/46163678/get-rid-of-warning-command-substitution-ignored-null-byte-in-input?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa).

**What documentation changes (if any) are needed to support this PR?:**
None

---
* You must follow the template instructions. Failure to do so will result in your pull request being closed.
* Please respect that Pi-hole is developed by volunteers, who can only reply in their spare time.

PR #2249, opened 2018-06-24
Revert "Fix error: /opt/pihole/gravity.sh: 385: Warning: command substitution: ignored null byte in input"

dschaper:
Reverts pi-hole/pi-hole#2069


PromoFaux:
Nice and simple to review, so Approved (I'm still alive, by the way!)

Issue #2002, opened 2018-02-23
Query log is empty


heroes1412:
everything is up to date. i restarted pi-orange, but get same result

**In raising this issue, I confirm the following:

- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [x] The issue I am reporting can be *replicated*.
- [] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the the source code relevant to this issue?:**
2

---
**Expected behaviour:**

See full query log

**Actual behaviour:**

Query log is empty

**Steps to reproduce:**

Disable and Enable Log Query

**Debug token provided by [uploading `pihole -d` log](https://discourse.pi-hole.net/t/the-pihole-command-with-examples/738#debug):**

jspgofdboh

**Troubleshooting undertaken, and/or other relevant information:**

Disable + Enable Log Query
Restart Pi-Orange

dschaper:
Because of the number of support requests and the very limited resources we have as a free open-source project run by volunteers, we will be closing all Issues that do not fill out the Issue Template and do not provide debug tokens when possible. Thank you for your understanding.

Please search [open](https://github.com/pi-hole/pi-hole/issues) and [closed](https://github.com/pi-hole/pi-hole/issues?q=is:issue+is:closed) issues for others that may have your same question. Also, visit our [Discourse](https://discourse.pi-hole.net) site for additional assistance.

heroes1412:
edited

tehniemer:
same issue here, with the addition that graphs are not working either. I have pihole installed in a container running on a QNAP NAS. Debug log attached. 


[pihole_debug-sanitized.log.log](https://github.com/pi-hole/pi-hole/files/1756106/pihole_debug-sanitized.log.log)

Mcat12:
@heroes1412 Sorry, make a new debug token because the other one has expired.

[...]

PR #2277, opened 2018-07-15
Installer fixes for Fedora & CentOS

bcambl:
- [ ] I have squashed any insignificant commits.
---
- Update PHP via Remi after enabling the repository otherwise PHP will not get updated via Remi until next system update.
- Add `php-json` as a web dependency on Fedora. This package is pulled as a dependency with CentOS + Remi. Note that `php-json` is not available on CentOS without Remi (default PHP) so the dependency should not be added to `PIHOLE_WEB_DEPS` directly. (also discovered here: https://github.com/pi-hole/pi-hole/issues/2002#issuecomment-388687187)

- install/update PHP after Remi repository is enabled
- add `php-json` to web dependencies when installed on Fedora 

n/a

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -291,6 +291,13 @@ elif command -v rpm &> /dev/null; then
                 "${PKG_INSTALL[@]}" "yum-utils" &> /dev/null
                 yum-config-manager --enable ${REMI_REPO} &> /dev/null
                 echo -e "  ${TICK} Remi's RPM repository has been enabled for PHP7"
+                # trigger an install/update of PHP to ensure previous version of PHP is updated from REMI
+                if "${PKG_INSTALL[@]}" "php-cli" &> /dev/null; then
+                    echo -e "  ${TICK} PHP7 installed/updated via Remi's RPM repository"
+                else
+                    echo -e "  ${CROSS} There was a problem updating to PHP7 via Remi's RPM repository"
+                    exit 1
+                fi
             fi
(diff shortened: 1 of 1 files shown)


automated install/basic-install.sh
```diff
@@ -2330,6 +2338,9 @@ main() {
         fi
     fi
 
+    # Check if SELinux is Enforcing
+    checkSelinux
```
Mcat12: Should this be after the distro check?

bcambl: Reasoning?

Mcat12: If someone is on an unsupported distro with SELinux enabled, they will get the SELinux warning before knowing that their distro is not supported.

bcambl: Fair, I was trying to avoid any repos being enabled, packages installed, or other dialogs. I am fine with moving it behind distro check. The distro _check_ has slowly became more of a distro _config_.

dschaper: That will be rectified soon. We'll have a static list of distributions that we support and the installer will check against that before any actions are taken. Then the script can be a configuration checker.

bcambl: Sounds good, I will move the selinux check behind distro check for now.

bcambl: I just cherry-picked my other two commits and force pushed.  
The selinux check depends on `whiptail` so it must exist after installer deps have been installed.

PR #2315, opened 2018-07-29
Allow PHP to edit the regex file if web is installed

Mcat12:
---
Fix permissions error on the web interface when modifying the regex:
```
Failure! Something went wrong.
Unable to add regex "^example\.com$" to /etc/pihole/regex.list
Error message: file_put_contents(/etc/pihole/regex.list): failed to open stream: Permission denied
```

Use the correct lighttpd user group when setting regex file permissions. Not all distros use `www-data`.

None

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1239,7 +1239,7 @@ installConfigs() {
     # Install an empty regex file
     if [[ ! -f "${regexFile}" ]]; then
         # Let PHP edit the regex file, if installed
-        install -o pihole -g "${LIGHTTPD_GROUP:-pihole}" -m 644 /dev/null "${regexFile}"
+        install -o pihole -g "${LIGHTTPD_GROUP:-pihole}" -m 664 /dev/null "${regexFile}"
     fi
 
     # If the user chose to install the dashboard,
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1237,7 +1237,9 @@ installConfigs() {
     version_check_dnsmasq
 
     # Install an empty regex file
-    touch "${regexFile}"
+    if [[ ! -f "${regexFile}" ]]; then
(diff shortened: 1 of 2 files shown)


automated install/basic-install.sh
```diff
@@ -1235,6 +1236,11 @@ installConfigs() {
     # Make sure Pi-hole's config files are in place
     version_check_dnsmasq
 
+    # Install an empty regex file
+    touch "${regexFile}"
```
dschaper: Might want to wrap this in a guard just so that it's idempotent for future installs.

```
if [[ ! -f "${regexFile}" ]]; then
  do stuff
fi
```

automated install/basic-install.sh
```diff
@@ -1261,6 +1267,9 @@ installConfigs() {
         chown ${LIGHTTPD_USER}:${LIGHTTPD_GROUP} /var/cache/lighttpd/compress
         mkdir -p /var/cache/lighttpd/uploads
         chown ${LIGHTTPD_USER}:${LIGHTTPD_GROUP} /var/cache/lighttpd/uploads
+
+        # Let PHP edit the regex file
+        chown pihole:${LIGHTTPD_GROUP} "${regexFile}"
```
dschaper: And another guard on this call, just in case the file is missing, then we won't get any errors leaking. Or how ever you'd like to swallow errors.

Mcat12: The file should exist at this point, if we're creating it above. If this would be not run when the file already exists, it could lead to it not being editable from the web interface.

dschaper: I agree that it should exist at that point, but I'd like to have more error handling in the code as we go forward. We can add that now, or I can look at it when I go over the installer overhaul and add more error handling for the whole script. Just trying to get ahead on some things, the rest of the function will have protections added in later.

Issue #2330, opened 2018-08-07
Display fault in piHole -t function (colors)


eejeel:
**In raising this issue, I confirm the following:** `{please fill the checkboxes, e.g: [X]}`

- [] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [] The issue I am reporting can be *replicated*.
- [] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the the source code relevant to this issue?: 3

`{Replace this with a number from 1 to 10. 1 being not familiar, and 10 being very familiar}`

---
**Expected behaviour:**

In the pihole -t function the blocked URL must be displayed in red and the queries in white

**Actual behaviour:**

I miss the red and white lines in Pihole -t function
All loggings are displayed in grey lines in the SSH screen. See picture on https://pasteboard.co/Hy4rhTd.jpg with the old and the new version of PiHole

Cause:
In the tail function the retrieval of the IPv4/6 addresses from /etc/pihole/setupVars.conf is defined.
But this is not used for blocking anymore and so not logged in pihole.log. Now it is 0.0.0.0

This line in pihole is the cause:
    -e "s,(.*(gravity.list|black.list| config ).* is (${IPV4_ADDRESS%/*}|${IPV6_ADDRESS:-NULL}).*), ${COL_RED}&${COL_NC}," \

For my own I have changed it to
    -e "s,(.*(gravity.list|black.list| config ).* is (0.0.0.0).*),${COL_RED}&${COL_NC}," \

I don't have IPV6 so I think it must be (0.0.0.0|::). For me (0.0.0.0) works

But you have to make it universal depending on the blocking mode definition.

**Steps to reproduce:**

pihole -t

**Debug token provided by [uploading `pihole -d` log](https://discourse.pi-hole.net/t/the-pihole-command-with-examples/738#debug):**

`{Alphanumeric token}`

**Troubleshooting undertaken, and/or other relevant information:**

`{Steps of what you have done to fix this}`

> * `{Please delete this quoted section when opening your issue}`
> * You must follow the template instructions. Failure to do so will result in your issue being closed.
[...]

PR #2345, opened 2018-08-11
Fix colour tail for use with new blocking methods

WaLLy3K:

Resolves issue #2330 by including `NXDOMAIN`, `0.0.0.0` and `::`.

### DIFF
--- a/pihole
+++ b/pihole
@@ -303,7 +303,7 @@ tailFunc() {
   # Colour everything else as gray
   tail -f /var/log/pihole.log | sed -E \
     -e "s,($(date +'%b %d ')| dnsmasq[.*[0-9]]),,g" \
-    -e "s,(.*(gravity.list|black.list| config ).* is (0.0.0.0|::|NXDOMAIN|${IPV4_ADDRESS%/*}|${IPV6_ADDRESS:-NULL}).*),${COL_RED}&${COL_NC}," \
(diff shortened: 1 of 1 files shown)


WaLLy3K:
Instead of `<unknown>`, I feel you should be seeing `/etc/pihole/black.list` as per the following:

untitled

WaLLy3K:
Regex blacklisting probably shouldn't be replying as `<unknown>`, but rather `/etc/pihole/regex.list`. It's easy enough to account for either of these, but it does seem like a bug.

Mcat12:
@DL6ER Is it possible to use the regex file path instead of `<unknown>` in the log?

DL6ER:
I'm fully aware of this issues. It is caused because we don't mess around with `dnsmasq`'s own logging. Let me give some further details here:
The issue is here that we inject domains that are queried and do match a filter into `dnsmasq`'s cache as manually configured. However, as it is not tied to any `addn-hosts=<file>` option of `dnsmasq`, there is no name available for it. One solution is to leave it like it is now. The other solution would be to import an (empty) file just for establishing a filename we can use. I think the latter is overkill and that's why I haven't done it.

I will, nevertheless, look at this again in the next few days and maybe I can spark a new idea how to get this done without modifying `dnsmasq`'s logging routine.

DL6ER:
@Mcat12 @WaLLy3K https://github.com/pi-hole/FTL/pull/350

PR #2346, opened 2018-08-11
Fix regex wildcard blocklist queries

WaLLy3K:

---
A search using `pihole -q domain.com` should search `/etc/pihole/regex.list` wildcards

Updated code to search `regex.list` and print matching output

N/A


advanced/Scripts/query.sh
```diff
@@ -133,22 +116,17 @@ fi
...
-            fi
-            case "${blockpage}" in
-                true ) echo "π ${wildcardlist##*/}"; exit 0;;
-                *    ) echo "   *.${match}";;
-            esac
+    mapfile -t results <<< "$(scanList "${domainQuery}" "${wildcardlist}" "wc")"
+    if [[ -n "${results[*]}" ]]; then
+        if [[ -z "${wcMatch:-}" ]] && [[ -z "${blockpage}" ]]; then
+            wcMatch=true
+            echo " ${matchType^} found in ${COL_BOLD}Wildcards${COL_NC}:"
```
Mcat12: This should say "Regex" instead of "Wildcards" since using "wildcards" does not fit with the rest of the software.

advanced/Scripts/query.sh
```diff
@@ -131,24 +114,26 @@ if [[ -n "${results[*]}" ]]; then
...
+
+        # Remove first full stop if it exists (e.g: ".foo.bar" > "foo.bar")
+        [[ "${results[0]::1}" == "." ]] && results[0]="${results[0]:1}"
+
+        # Return matching regex.list line (Full stops escaped to check against wildcards)
+        result=( "$(grep -i "${results[0]//./\\\\.}" "${regexlist}")" "$(grep -i "${results[0]}" "${regexlist}")" )
+
+        case "${blockpage}" in
+            true ) echo "π ${regexlist##*/}"; exit 0;;
+            *    ) echo "   ${result[*]}";;
```
Mcat12: Only the first entry is indented:
```
$ pihole -q domain.co
 Match found in Regex list:
   .*domain\.com$
.*domain\.co
```

WaLLy3K: Thanks! I tried reproducing this prior to the PR, but only one line would print.

DL6ER:
@WaLLy3K This does not work
```
pihole -q a.bad
 Match found in Regex list:
   (^|\.)domain\.com$
   ((^)|(\.))domain\.com$
   ((^)|(\.))facebook\.
   ((^)|(\.))twitter\.
   ...
   [ I removed all further lines that are in my regex.list file ]
```

WaLLy3K:
Known edge case: Searching for a domain cannot return the correct regex if the line found in `regex.list` is not a whole word.

E.g: Search for `a.bad` against these lines in `regex.list` will mistakenly output `d\.bad$`

```
d\.bad$
\.b.*
b.d$
```
I'm currently not able to figure out a fix for this.

WaLLy3K:
[As per this line](https://github.com/pi-hole/pi-hole/blob/ab790864319ccabe75c94479b6a8d96d3b0667d2/advanced/Scripts/query.sh#L122), I'm looking for a way that I can return the matching regex pattern, instead of only printing the matching query output.

A solution to this would greatly simply this code, and ensure that users can always find the reason that a query was blocked by `regex.list`.

DL6ER:
Milestone cleared as still marked as WIP

dschaper:
Any new activity or is this just stalled out and needs to be closed?

mmotti:
@WaLLy3k could you possibly use awk? I can't think of any other appropriate (or faster) way to do this than iterating through each entry in regex.list, check if it matches the query string and if it does, print the current regex iteration and continue looping through in case it matches multiple?

mmotti:
Just a little demonstration:

```
1 result(s) matched ad.bbc.co.uk
^(.+[-_.])??m?ad[sxv]?[0-9]*[-_.]
```
```
2 result(s) matched tracking.ad.test.com
^track(ers?|ing)?[0-9]*[-.]
^(.+[-_.])??m?ad[sxv]?[0-9]*[-_.]
```
```
1 result(s) matched analytics.test.com
^analytics?[-.]
```
Code:

```
#!/usr/bin/env bash

# Download mmotti regex.list
file_regex=$(wget -qO - https://raw.githubusercontent.com/mmotti/pihole-regex/master/regex.list)

# Conditional exit if unable to obtain
[ -z "$file_regex" ] && echo 'Failed to download regexps' && exit

# Ask user for domain to match
read -r -p 'Enter domain: ' strDomain; clear

# Add each regexp to awk array
# For each target domain (1 at this time): iterate through each regexp, check if domain matches regexp pattern
# If match - output the current regexp iteration
matches=$(awk 'NR==FNR{regexps[$0]}{for (r in regexps)if($0 ~ r)print r}' <(echo "$file_regex") <(echo "$strDomain"))

# 0 match condition
[ -z "$matches" ] && echo 'No regex matches found' && exit

# Stats
echo "$(echo "$matches" | wc -l) result(s) matched $strDomain"

echo "$matches"
```

Mcat12:
Closed in favor of #2780

pralor:

https://discourse.pi-hole.net/t/faulty-regex-consumes-all-memory-and-causes-issues-with-pihole-q/28710/75

PR #2354, opened 2018-08-12
Create adlist file while running unattended setup

realtebo:
Signed-off-by: Realtebo <mirkotebaldi@yahoo.it>

---
The goal of this PR is to autocreate the standard file with adlists when using `--unattended` flag. 
Actually, If you do an unattended setup, there is no options to create this file.
The result is that at first run there is a php error in the settings page because the admin gui cannot open the adlist's file.

I created a new function in the setup code that is executed only when doing an unattended setup. Also, to avoid duplication of code, I created a simple helpers function accepting the 'name' of the list as first argument and adding the relative url the list address to the list files; do not fails if a unknow list name is passed.
I also avoided to override the lists file if it is already present.

Update unattended setup doc, if present, to explain that, if not exists, a default list file will be automatically created

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1087,17 +1087,40 @@ chooseBlocklists() {
     # For each choice available,
     for choice in ${choices}
     do
-        # Set the values to true
-        case ${choice} in
-            StevenBlack  )  echo "https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts" >> "${adlistFile}";;
-            MalwareDom   )  echo "https://mirror1.malwaredomains.com/files/justdomains" >> "${adlistFile}";;
-            Cameleon     )  echo "http://sysctl.org/cameleon/hosts" >> "${adlistFile}";;
-            ZeusTracker  )  echo "https://zeustracker.abuse.ch/blocklist.php?download=domainblocklist" >> "${adlistFile}";;
-            DisconTrack  )  echo "https://s3.amazonaws.com/lists.disconnect.me/simple_tracking.txt" >> "${adlistFile}";;
-            DisconAd     )  echo "https://s3.amazonaws.com/lists.disconnect.me/simple_ad.txt" >> "${adlistFile}";;
-            HostsFile    )  echo "https://hosts-file.net/ad_servers.txt" >> "${adlistFile}";;
-        esac
+        appendToListsFile choice
     done
+}
+
+# Accept a string parameter, it must be one of the default lists
+# This function allow to not duplicate code in chooseBlocklists and 
+# in installDefaultBlocklists
+appendToListsFile() {
+    case $1 in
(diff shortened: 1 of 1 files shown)


automated install/basic-install.sh
```diff
@@ -1087,19 +1087,42 @@ chooseBlocklists() {
...
+        DisconTrack  )  echo "https://s3.amazonaws.com/lists.disconnect.me/simple_tracking.txt" >> "${adlistFile}";;
+        DisconAd     )  echo "https://s3.amazonaws.com/lists.disconnect.me/simple_ad.txt" >> "${adlistFile}";;
+        HostsFile    )  echo "https://hosts-file.net/ad_servers.txt" >> "${adlistFile}";;
+    esac
+}
+
+# Used only in unattended setup
+# If there is already the adListFile, we keep it, else we create it using all default lists
+installDefaultBlocklists() {
+    # In unattended setup, could be usefull to use userdefined blocklist.
```
Mcat12: "usefull" only has one "l"

automated install/basic-install.sh
```diff
@@ -1087,19 +1087,42 @@ chooseBlocklists() {
...
+    # If this file exists, we avoid to override it.
```
Mcat12: > we avoid to override it.

Change to "we avoid overriding it."

realtebo: Updated as requested

automated install/basic-install.sh
realtebo: Updated as requested

automated install/basic-install.sh
```diff
@@ -1087,19 +1087,42 @@ chooseBlocklists() {
...
-        case ${choice} in
-            StevenBlack  )  echo "https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts" >> "${adlistFile}";;
-            MalwareDom   )  echo "https://mirror1.malwaredomains.com/files/justdomains" >> "${adlistFile}";;
-            Cameleon     )  echo "http://sysctl.org/cameleon/hosts" >> "${adlistFile}";;
-            ZeusTracker  )  echo "https://zeustracker.abuse.ch/blocklist.php?download=domainblocklist" >> "${adlistFile}";;
-            DisconTrack  )  echo "https://s3.amazonaws.com/lists.disconnect.me/simple_tracking.txt" >> "${adlistFile}";;
-            DisconAd     )  echo "https://s3.amazonaws.com/lists.disconnect.me/simple_ad.txt" >> "${adlistFile}";;
-            HostsFile    )  echo "https://hosts-file.net/ad_servers.txt" >> "${adlistFile}";;
-        esac
+        appendToListsFile choice
```
DL6ER: This PR introduces a bug. Instead of the array `${choice}`, the string `"choice"` is being passed to `appendToListsFile()` resulting in no file to be generated at all.

PR #2356, opened 2018-08-13
Change the enable/disable method

DL6ER:

---
**What does this PR aim to accomplish and how does it accomplish it?:**

Change the method we use to disable/enable blocking. So far, we commented the config lines
```
addn-hosts=/etc/pihole/gravity.list
addn-hosts=/etc/pihole/black.list
```
and then restarting `pihole-FTL`. This is very inefficient and slow, as, on shutdown of `pihole-FTL`, we update the database and then on restart, import all queries again from the database.

This PR changes this. Instead of changing the configuration to not load the files, we replace them by empty files which are then reread on `reload`ing `pihole-FTL`. The old files are kept as backup and restored on `pihole enable`. This is a much cleaner variant of handling the blocking status and also *much* faster than always restarting the daemon. It solves a number of delay issues recently reported with the Enable/Disable functionality of Pi-hole.

Pi-hole `gravity` is modified to update the backup files when blocking is disabled.

The current blocking status is tracked by setting the new variable `BLOCKED_ENABLED` in `setupVars.conf`.

### DIFF
--- a/pihole
+++ b/pihole
@@ -10,7 +10,8 @@
 # Please see LICENSE file for your rights under this license.
 
 readonly PI_HOLE_SCRIPT_DIR="/opt/pihole"
-readonly wildcardlist="/etc/dnsmasq.d/03-pihole-wildcard.conf"
+readonly gravitylist="/etc/pihole/gravity.list"
+readonly blacklist="/etc/pihole/black.list"
 readonly colfile="${PI_HOLE_SCRIPT_DIR}/COL_TABLE"
 source "${colfile}"
 
@@ -146,10 +147,13 @@ Time:
 
   elif [[ "${1}" == "0" ]]; then
     # Disable Pi-hole
-    sed -i 's/^addn-hosts=\/etc\/pihole\/gravity.list/#addn-hosts=\/etc\/pihole\/gravity.list/' /etc/dnsmasq.d/01-pihole.conf
-    sed -i 's/^addn-hosts=\/etc\/pihole\/black.list/#addn-hosts=\/etc\/pihole\/black.list/' /etc/dnsmasq.d/01-pihole.conf
-    if [[ -e "$wildcardlist" ]]; then
-      mv "$wildcardlist" "/etc/pihole/wildcard.list"
+    if [[ -e "${gravitylist}" ]]; then
+      mv "${gravitylist}" "${gravitylist}.bck"
+      echo "" > "${gravitylist}"
+    fi
+    if [[ -e "${blacklist}" ]]; then
+      mv "${blacklist}" "${blacklist}.bck"
+      echo "" > "${blacklist}"
     fi
     if [[ $# > 1 ]]; then
       local error=false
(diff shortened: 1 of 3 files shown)


pralor:

https://discourse.pi-hole.net/t/disable-enable-feature-in-pihole-admin-web-console-hanging/11784/18

diginc:
would `.disabled` be a more meaningful file extension than `.bck`

Does this impact the web interfaces updating of blacklist when the blacklist is in disabled mode?

diginc:
> I don't see how symlinks can be more efficient here.

The web interface impact got me thinking about keeping the file location consistent.  I'm not sure how many locations need updated to reference both the enabled/disabled versions and lookup the setupVar, but I was trying to think of ways to abstract that away by keeping the file location static.  

> Thanks for pointing this out. I addressed this in f794018

If this is the only other spot needing updated to reflect the new disabled file location, then my thought maybe overkill.  

What I had in my head was very similar to what nginx/apache do for webserver config folders being disabled and enabled: 

```
/etc/pihole/enabled-blacklists/gravity.list (symlink) -> /etc/pihole/available-blacklists/gravity.list (always present)
/etc/pihole/enabled-blacklists/black.list (symlink) -> /etc/pihole/available-blacklists/black.list (always present)
```

Disabling would just run an `unlink /etc/pihole/enabled-blacklists/gravity.list /etc/pihole/enabled-blacklists/black.list`.  Enabling would re create the symlink.  Not necessarily better, just follows a familiar pattern to keep consistent file locations.

DL6ER:
The web interface does it via a `pihole -b ...` call in the background so it is covered by my most recent change.

screenshot at 2018-08-13 18-48-43

PromoFaux:
Before:
lag

After:
nolag

pihole
```diff
@@ -187,19 +192,25 @@ Time:
...
     echo -e "  ${INFO} Enabling blocking"
     local str="Pi-hole Enabled"
 
-    sed -i 's/^#addn-hosts/addn-hosts/' /etc/dnsmasq.d/01-pihole.conf
-    if [[ -e "/etc/pihole/wildcard.list" ]]; then
-      mv "/etc/pihole/wildcard.list" "$wildcardlist"
+    if [[ -e "${gravitylist}.bck" ]]; then
+      mv "${gravitylist}.bck" "${gravitylist}"
+    fi
+    if [[ -e "${blacklist}" ]]; then
```
Mcat12: Shouldn't this check for `"${blacklist}.bck"`?

pihole
```diff
@@ -257,16 +268,17 @@ statusFunc() {
     return 0
   fi
 
-  # Determine if Pi-hole's addn-hosts configs are commented out
-  addnConfigs=$(grep -i "addn-hosts=/" /etc/dnsmasq.d/01-pihole.conf)
+  # Determine if Pi-hole's blocking is enabled
+
+  addnConfigs=$?
```
Mcat12: This line is no longer needed

pralor:

https://discourse.pi-hole.net/t/backup-dns-while-doing-tasks/14127/10

pralor:

https://discourse.pi-hole.net/t/pi-hole-v4-1-is-now-available/14990/1

Issue #2396, opened 2018-08-29
Scientific Linux support 


StackKorora:
**In raising this issue, I confirm the following:** `{please fill the checkboxes, e.g: [X]}`

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the the source code relevant to this issue?:**
8

---
Greetings,
This is a bit of a mix between a "Ack, what once worked is now broke" and a "feature request because I know I'm outside of the 'supported' realm".

In short, I have valid reasons for why I need to run Scientific Linux and am unable to run CentOS (another story for another day ;). For a long time, I've been running PiHole on Scientific Linux 7 without issue (although I apparently did edit my /etc/redhat-release to trick something at one point into thinking my server was CentOS...). Scientific is close enough to the supported CentOS that everything has just worked. Until I did an update tonight. After a lot of digging around I found the issue.  
Specifically this commit: 8ad37af70ee884c3d863755b32f87dea35681ae5  
Which impacts this file: /etc/.pihole/automated\ install/basic-install.sh  
Specifically the line (currently 257) which says `CURRENT_CENTOS_VERSION=$(rpm -q --queryformat '%{VERSION}' centos-release)`

Well, I may have faked /etc/redhat-release but I can't easily fake that (well, not without spinning my own rpm). So I just hard set that variable to 7 in the script and everything, of course, worked just fine.

My request, would it be possible to support Scientific Linux please? I'm pretty sure it wouldn't add much to do so (even if it is just a "you are running Scientific Linux - we will try but if this fails, well, its on you"). My limited testing shows that this should work:

Line 252: `elif grep -qiE 'centos|scientific' /etc/redhat-release; then`

And since you have the release information in that file anyway you could just remove the rpm check.

Line 257: `CURRENT_CENTOS_VERSION=$(grep -q 'release 7.' /etc/redhat-release && echo 7)`

Thoughts?
Thank you for taking the time to read my request.

PS: just for reference my release trickery
```
[...]

PR #2398, opened 2018-08-30
WIP: Update RPM distro checks

bcambl:

---
_Soften_ the RPM distro checking to allow installing on various Fedora/CentOS based distros.  
User issues:
#2396  
[FedBerry Issue](https://discourse.pi-hole.net/t/issues-installing-pi-hole-fedora27-spin-fedberry/6362)

- Add FedBerry to Fedora check
- Add Scientific to CentOS check
- Add prompt to allow user to continue installing on unsupported RPM based distro

TDB

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -307,13 +307,14 @@ elif command -v rpm &> /dev/null; then
     else
         # Warn user of unsupported version of Fedora or CentOS
         if ! whiptail --defaultno --title "Unsupported RPM based distribution" --yesno "Would you like to continue installation on an unsupported RPM based distribution?\\n\\nPlease ensure the following packages have been installed manually:\\n\\n- lighttpd\\n- lighttpd-fastcgi\\n- PHP version 7+" ${r} ${c}; then
+            echo -e "  ${CROSS} Aborting installation due to unsupported RPM based distribution"
             exit # exit the installer
         else
-            : # continue with unsupported RPM distribution
+            echo -e "  ${INFO} Continuing installation with unsupported RPM based distribution"
         fi
     fi
 
-# If neither apt-get or yum/dnf are found
+# If neither apt-get or yum/dnf package managers were found
 else
     # it's not an OS we can support,
     echo -e "  ${CROSS} OS distribution not supported"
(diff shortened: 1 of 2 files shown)


bcambl:
unsupported_rpm_dist_whiptail

automated install/basic-install.sh
```diff
@@ -305,13 +305,15 @@ elif command -v rpm &> /dev/null; then
...
-        # If not a supported version of Fedora or CentOS,
-        echo -e "  ${CROSS} Unsupported RPM based distribution"
-        # exit the installer
-        exit
+        # Warn user of unsupported version of Fedora or CentOS
+        if ! whiptail --defaultno --title "Unsupported RPM based distribution" --yesno "Would you like to continue installation on an unsupported RPM based distribution?\\n\\nPlease ensure the following packages have been installed manually:\\n\\n- lighttpd\\n- lighttpd-fastcgi\\n- PHP version 7+" ${r} ${c}; then
+            exit # exit the installer
+        else
+            : # continue with unsupported RPM distribution
+        fi
```
Mcat12: Is an empty else block necessary?

automated install/basic-install.sh
```diff
@@ -305,13 +305,15 @@ elif command -v rpm &> /dev/null; then
...
     fi
 
-# If neither apt-get or rmp/dnf are found
+# If neither apt-get or yum/dnf are found
```
Mcat12: The conditionals check for `rpm` and `dnf`, but don't explicitly check for `yum`. Does this comment still make sense then?

automated install/basic-install.sh
bcambl: True, its not necessary.

automated install/basic-install.sh
bcambl: I would say it makes more sense as it is now comparing _apple to apples_.
`dpkg` is to Debian as `rpm` is to Fedora.  
`apt-get` is to Debian as `yum` is to CentOS and `dnf` is to Fedora.

I just fixed a typo, we could re-word the whole thing imo.

Mcat12: It looks good, no change needed.

bcambl:
@Mcat12: I have addressed your review concerns.

Issue #2435, opened 2018-09-30
This breaks resetting a password on Ubuntu 16.04


cybrnook:
pihole -a -p no longer works:

pihole@piholeQ9-sec:~$ pihole -a -p
/opt/pihole/webpage.sh: line 13: setupVars: readonly variable
/opt/pihole/webpage.sh: line 113: read: `Enter New Password (Blank for no password): ': not a valid identifier

_Originally posted by @cybrnook in https://github.com/pi-hole/pi-hole/commit/8684356b4486f77bc7c49dd5e08620d2a20aa975#commitcomment-30709174_

Mcat12:
Thank you for quickly testing this.

cybrnook:
I happened to be in the middle of an install and switch to dev branch. Stumbled into it, so figured I would shoot over to see if any recent commits. Was within seconds....

cybrnook:
As a side note, one of the recent commits broke the UI displaying versions properly (bottom right of webui):

Pi-hole Version vDev (, ) Web Interface Version vDev (development, vDev-7a74acc) FTL Version

Mcat12:
Thanks, I've found the issue.

PR #2436, opened 2018-09-30
Fix error when reading in new password

Mcat12:
---
Fix crash when setting new password. See #2435 

Revert use of `-r` flag on the `read` command in `webpage.sh`'s `SetWebPassword` function.

None

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -110,7 +110,7 @@ SetWebPassword() {
         # Prevents a bug if the user presses Ctrl+C and it continues to hide the text typed.
         # So we reset the terminal via stty if the user does press Ctrl+C
         trap '{ echo -e "\nNo password will be set" ; stty sane ; exit 1; }' INT
-        read -s -p -r "Enter New Password (Blank for no password): " PASSWORD
+        read -s -p "Enter New Password (Blank for no password): " PASSWORD
         echo ""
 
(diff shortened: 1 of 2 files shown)


Mcat12:
Note: it may be worth looking into why `read` fails instead of dismissing the shellcheck warning.

advanced/Scripts/webpage.sh
```diff
@@ -110,7 +110,8 @@ SetWebPassword() {
         # Prevents a bug if the user presses Ctrl+C and it continues to hide the text typed.
         # So we reset the terminal via stty if the user does press Ctrl+C
         trap '{ echo -e "\nNo password will be set" ; stty sane ; exit 1; }' INT
-        read -s -p -r "Enter New Password (Blank for no password): " PASSWORD
+        # shellcheck disable=SC2162
+        read -s -p "Enter New Password (Blank for no password): " PASSWORD
```
dschaper: The `-r` can still be used, the order of the flags was causing some issues, `-p` needs to be followed by the actual prompt text and in the existing it was followed by `-r` causing issues. 

```suggestion
        read -s -r -p "Enter New Password (Blank for no password): " PASSWORD
```

dschaper: And also needs to be `read -r -a <arrayname>` for the other and that will resolve the `read-only` error.

dschaper:
Fixes #2435

PR #2467, opened 2018-10-17
Update dnsmasq config license

WaLLy3K:

---

Remove reference to GNU GPL and fix "SEPARATE CONFIG" text

### DIFF
--- a/advanced/01-pihole.conf
+++ b/advanced/01-pihole.conf
@@ -1,13 +1,11 @@
 # Pi-hole: A black hole for Internet advertisements
-# (c) 2015, 2016 by Jacob Salmela
-# Network-wide ad blocking via your Raspberry Pi
-# http://pi-hole.net
-# dnsmasq config for Pi-hole
+# (c) 2017 Pi-hole, LLC (https://pi-hole.net)
+# Network-wide ad blocking via your own hardware.
 #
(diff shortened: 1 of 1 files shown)


advanced/01-pihole.conf
```diff
@@ -1,13 +1,11 @@
 # Pi-hole: A black hole for Internet advertisements
-# (c) 2015, 2016 by Jacob Salmela
-# Network-wide ad blocking via your Raspberry Pi
-# http://pi-hole.net
-# dnsmasq config for Pi-hole
+# (c) 2017 Pi-hole, LLC (https://pi-hole.net)
```
Mcat12: Should this be 2018?
```suggestion
# (c) 2017, 2018 Pi-hole, LLC (https://pi-hole.net)
```

WaLLy3K: Hmm, not sure. @jacobsalmela, what do you think? The installer has `2017 - 2018`, and others just have `2017`.

jacobsalmela: I think we can leave it as-is.  Really the copyright statements should all the say year in which they were first publicized and you don't need to keep modifying them year after year.  Copyright law is good for about 70 years based on the publication time, so I suppose we can leave those commits to our predecessors. 😉

PR #2486, opened 2018-10-31
Beginning steps to installer overhaul. Echos to Printf test.

dschaper:

---

In order to move to a non git-repo based install process there will be a number of changes that need to be made. This is an initial change to not use `echo` for terminal communication. All messages to user terminal and log file should be with `printf`. The remaining `echo` commands are for writing heredoc type structures and for the new concept of allowing bash functions to echo a response to calling functions.

Change `echo` to `printf` structures. Any unneeded echo commands (for example echo to a variable) have been removed or changed.

None

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -166,10 +166,9 @@ if command -v apt-get &> /dev/null; then
     fi
     # Check for and determine version number (major and minor) of current php install
     if command -v php &> /dev/null; then
-        phpInsVersion="$(php -v | head -n1 | grep -Po '(?<=PHP )[^ ]+')"
-        printf "  %b Existing PHP installation detected : PHP version %s\\n" "${INFO}" "${phpInsVersion}"
-        printf -v phpInsMajor "%d" "${phpInsVersion%%.*}"
-        printf -v phpInsMinor "%d" "${phpInsVersion##*.}"
+        printf "  %b Existing PHP installation detected : PHP version %s\\n" "${INFO}" "$(php <<< "<?php echo PHP_VERSION ?>")"
+        printf -v phpInsMajor "%d" "$(php <<< "<?php echo PHP_MAJOR_VERSION ?>")"
+        printf -v phpInsMinor "%d" "$(php <<< "<?php echo PHP_MINOR_VERSION ?>")"
         # Is installed php version 7.0 or greater
         if [ "${phpInsMajor}" -ge 7 ]; then
             phpInsNewer=true
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -788,14 +788,14 @@ setStaticIPv4() {
     elif [[ -f "/etc/sysconfig/network-scripts/ifcfg-${PIHOLE_INTERFACE}" ]];then
         # If it exists,
         IFCFG_FILE=/etc/sysconfig/network-scripts/ifcfg-${PIHOLE_INTERFACE}
-        printf -v IPADDR "${IPV4_ADDRESS%%/*}"
+        printf -v IPADDR "%s" "${IPV4_ADDRESS%%/*}"
         # check if the desired IP is already set
         if grep -Eq "${IPADDR}(\\b|\\/)" "${IFCFG_FILE}"; then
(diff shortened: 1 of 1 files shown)


DL6ER:
Is this PR meant to be milestones for v4.1 ?

dschaper:
No real milestone on this or the installer updates. The changes to a new style installer will be dependent on SemVer but this is more internal and prep work. There should be no observable changes to the end user.

automated install/basic-install.sh
```diff
@@ -165,11 +167,11 @@ if command -v apt-get &> /dev/null; then
     # Check for and determine version number (major and minor) of current php install
     if command -v php &> /dev/null; then
         phpInsVersion="$(php -v | head -n1 | grep -Po '(?<=PHP )[^ ]+')"
-        echo -e "  ${INFO} Existing PHP installation detected : PHP version $phpInsVersion"
-        phpInsMajor="$(echo "$phpInsVersion" | cut -d\. -f1)"
-        phpInsMinor="$(echo "$phpInsVersion" | cut -d\. -f2)"
+        printf "  %b Existing PHP installation detected : PHP version %s\\n" "${INFO}" "${phpInsVersion}"
+        printf -v phpInsMajor "%d" "${phpInsVersion%%.*}"
+        printf -v phpInsMinor "%d" "${phpInsVersion##*.}"
```
Mcat12: `phpInsMinor` does not get parsed correctly (the original code took the second match, and this takes from the end):
```
$ php -v | head -n1 | grep -Po '(?<=PHP )[^ ]+'
7.0.32-0ubuntu0.16.04.1
$ phpInsVersion="$(php -v | head -n1 | grep -Po '(?<=PHP )[^ ]+')" && printf "%d" "${phpInsVersion%%.*}"
7
$ phpInsVersion="$(php -v | head -n1 | grep -Po '(?<=PHP )[^ ]+')" && printf "%d" "${phpInsVersion##*.}"
1
```

dschaper: I'm not sure what the minor revision is useful for. It's used to check for a version >= 7.0. I'll fix that grab to be less greedy, but is there a need to know what the minor version is?

Mcat12: I don't think so, at least now. The most it does is print out alongside the major version.

automated install/basic-install.sh
```diff
@@ -165,11 +167,11 @@ if command -v apt-get &> /dev/null; then
     # Check for and determine version number (major and minor) of current php install
     if command -v php &> /dev/null; then
         phpInsVersion="$(php -v | head -n1 | grep -Po '(?<=PHP )[^ ]+')"
-        echo -e "  ${INFO} Existing PHP installation detected : PHP version $phpInsVersion"
-        phpInsMajor="$(echo "$phpInsVersion" | cut -d\. -f1)"
-        phpInsMinor="$(echo "$phpInsVersion" | cut -d\. -f2)"
+        printf "  %b Existing PHP installation detected : PHP version %s\\n" "${INFO}" "${phpInsVersion}"
```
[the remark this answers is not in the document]
dschaper: It can, but there are so many other places that use the double quote that I'd prefer not to have singleton's out there with single quotes. A lot of the `printf` calls will be functionalized anyways to have something like a `warn` `error` and `fatal` function that prints to the terminal and then dumps sanitized text to a log file.

automated install/basic-install.sh
dschaper: Thanks, I see where that comes in now. Digital Ocean's Ubuntu 18.10 is giving me problems right now with their apt mirror missing a Release file so once that is fixed I'll get something working here.

automated install/basic-install.sh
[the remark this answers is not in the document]
dschaper: Marking this as resolved.

automated install/basic-install.sh
dschaper: What do you think of something like `echo "<?php echo PHP_MAJOR_VERSION ?>" | php`

Just asking PHP directly what major and minor version is installed?

Mcat12: That sounds like a much better way :+1:

PR #2504, opened 2018-11-11
configure ifcfg network files by connection name when required

bcambl:
---
Ideally sysconfig network scripts are named `ifcfg-<interface name>`; however, and edge-case was discovered by @DL6ER while testing on Vultr Fedora 29 where the file was named `ifcfg-<nmcli con name>`.

This PR splits the ifcfg file configuration into a separate function which takes a ifcfg file path as an argument. The existing function `setStaticIPv4` will now only determine the configuration file name and pass the filepath to the `setIFCFG` function to be configured.
Debian/dhcpcd configuration is handled the same as before with the exception of the static IP check being moved from `setStaticIPv4` to `setDHCPCD`.

Existing comments retained, additional in-line comments included.

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -857,7 +857,7 @@ setStaticIPv4() {
     fi
     # if an ifcfg config does not exists for the interface name, try the connection name via network manager
     if is_command nmcli && nmcli general status &> /dev/null; then
-        CONNECTION_NAME=$(nmcli dev show "${PIHOLE_INTERFACE}" | grep 'GENERAL.CONNECTION' | cut -d: -f2 | xargs | tr ' ' '_')
+        CONNECTION_NAME=$(nmcli dev show "${PIHOLE_INTERFACE}" | grep 'GENERAL.CONNECTION' | cut -d: -f2 | sed 's/^System//' | xargs | tr ' ' '_')
         if [[ -f "/etc/sysconfig/network-scripts/ifcfg-${CONNECTION_NAME}" ]];then
             # If it exists,
             IFCFG_FILE=/etc/sysconfig/network-scripts/ifcfg-${CONNECTION_NAME}
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -857,7 +857,7 @@ setStaticIPv4() {
     fi
     # if an ifcfg config does not exists for the interface name, try the connection name via network manager
     if is_command nmcli && nmcli general status &> /dev/null; then
-        CONNECTION_NAME=$(nmcli dev show ${PIHOLE_INTERFACE} | grep 'GENERAL.CONNECTION' | cut -d: -f2 | xargs | tr ' ' '_')
+        CONNECTION_NAME=$(nmcli dev show "${PIHOLE_INTERFACE}" | grep 'GENERAL.CONNECTION' | cut -d: -f2 | xargs | tr ' ' '_')
         if [[ -f "/etc/sysconfig/network-scripts/ifcfg-${CONNECTION_NAME}" ]];then
             # If it exists,
             IFCFG_FILE=/etc/sysconfig/network-scripts/ifcfg-${CONNECTION_NAME}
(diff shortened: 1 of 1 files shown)


dschaper:
Is this a fix that should be based to `/release/v4.1`?

DL6ER:
Unfortunately, my test on Vultr's Fedora 29 wasn't successful:
screenshot from 2018-11-25 19-05-05

`-x` output:
```
+ setStaticIPv4
+ local IFCFG_FILE
+ local CONNECTION_NAME
+ [[ -f /etc/dhcpcd.conf ]]
+ [[ -f /etc/sysconfig/network-scripts/ifcfg-ens3 ]]
+ is_command nmcli
+ local check_command=nmcli
+ command -v nmcli
+ nmcli general status
++ nmcli dev show ens3
++ grep GENERAL.CONNECTION
++ cut -d: -f2
++ xargs
++ tr ' ' _
+ CONNECTION_NAME=System_Wired_connection_1
+ [[ -f /etc/sysconfig/network-scripts/ifcfg-System_Wired_connection_1 ]]
+ printf '  %b Warning: Unable to locate configuration file to set static IPv4 address\n' '[i]'
  [i] Warning: Unable to locate configuration file to set static IPv4 address
+ exit 1
[root@vultr .pihole]# 
```

DL6ER:
The `System_` seems wrong:
```
[root@vultr .pihole]# ls /etc/sysconfig/network-scripts/
ifcfg-Wired_connection_1  ifcfg-Wired_connection_1.pihole.orig
```

bcambl:
Ah thanks.
We could likely toss a `sed` between the `xargs` and `tr` to remove `^System`..
Also, I am open to suggestions.. I considered an unwieldy `awk` but the multiple pipe command seemed like it would be more understandable and/or maintainable.

dschaper:
Suggestion, if this is not Raspbian we tell people they need to set their IP addresses per their distribution guidelines and then in the documentation we create a page with the steps for each distro?

bcambl:
@dschaper If we are going to drop static network configurations via the installer for one of the 2 supported distros, then we should drop it for both and clean-up the installer.
To expand on your suggestion, I would not want to key off `raspbian` for anything since we have users on various deb/rpm derivatives. I would rather check to see if a static ip is currently configured (or dhcp is enabled?) and warn/refer the user to documentation.

dschaper:
Agreed that it should be all or nothing. The reason I use `raspbian` as an example is that we know the use `dhcpcd5` and always will. `ubuntu` could be `interfaces`, `netplan`, `connman`, `networkmanager` and what ever else is decided as The One Way. That's where my frustration comes in, it's not just a distribution that we have to consider, it's the specific release of that distribution. But we know `raspbian` will never change.

bcambl:
Understandable.  
I will say the future of networking on Fedora/CentOS/RHEL is consistent via NetworkManager.  What you describe is one of the main reasons I don't use Ubuntu. I understand why people like it because it _'just works'_  but it seems to be at the expense of bloat that eventually results in a dumpster fire.

automated install/basic-install.sh
```diff
@@ -776,77 +776,98 @@ It is also possible to use a DHCP reservation, but if you are going to do that,
...
+        {
+        echo "# Configured via Pi-hole installer"
+        echo "DEVICE=$PIHOLE_INTERFACE"
+        echo "BOOTPROTO=none"
+        echo "ONBOOT=yes"
+        echo "IPADDR=$IPADDR"
+        echo "PREFIX=$CIDR"
+        echo "GATEWAY=$IPv4gw"
+        echo "DNS1=$PIHOLE_DNS_1"
+        echo "DNS2=$PIHOLE_DNS_2"
```
Mcat12: The DHCPCD config uses 127.0.0.1 as the DNS server. Is there a reason that this is not done here?

bcambl: Do we want to force the device to use itself for DNS by default?  
I prefer the ability to stop pi-hole services and not affect the connectivity of my other hosted services. This also helps address when users have a failed install and can no longer update the OS or resolve any url's due to having the loopback set as the nameserver.
Typically pi-hole is setup as a headless service either alone or with other services.. perhaps we should look at adding another option for the user to decide if they would like their server to self-resolve?

bcambl: issue: https://github.com/pi-hole/pi-hole/issues/2535

PR #2574, opened 2019-01-02
Ignore COL_TABLE if the the file does not exists

LudovicRousseau:

---

Ignore COL_TABLE if the the file does not exists

If the file "${PI_HOLE_SCRIPT_DIR}/COL_TABLE" does not exists then pihole displayed an error:
```
$ ./pihole -h
./pihole: line 22: /opt/pihole/COL_TABLE: No such file or directory
```

Now the file is used *only* if it exists.

pihole now first check the file exists before using it.

I reused the same `if [[ -e "file" ]]; then` construction as found in other parts pihole.

No documentation change needed.

### DIFF
--- a/pihole
+++ b/pihole
@@ -19,7 +19,9 @@ readonly blacklist="/etc/pihole/black.list"
 setupVars="/etc/pihole/setupVars.conf"
 
 readonly colfile="${PI_HOLE_SCRIPT_DIR}/COL_TABLE"
-source "${colfile}"
+if [[ -e "${colfile}" ]]; then
+  source "${colfile}"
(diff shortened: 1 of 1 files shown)


DL6ER:
This file should be present. Maybe we should install it when it is absent instead of just ignoring the `source` command?

LudovicRousseau:
I had the "problem" when running pihole without installing it first, just to get the help messages.
I agree this issue can be ignored.

Mcat12:
Pi-hole should be installed before using any of its commands, so errors that occur because it was not installed are not bugs.

PR #2601, opened 2019-01-24
Prevent double disabling/enabling

### LINKED ISSUE #2573: Disabling twice clears blocklist and blacklist

10

---
**Expected behaviour:**

Disabling twice (via `pihole disable`) is the same as disabling once, and enabling afterwards works correctly.

**Actual behaviour:**

Disabling twice clears the blocklist and blacklist, so that after re-enabling, no domains are blocked.

**Steps to reproduce:**

Run these commands:
```
pihole enable
pihole disable
pihole disable
pihole enable
```

See that both `gravity.list`, `black.list`, and their respective `.bck` files are empty. Therefore, no domains are blocked.

N/A

See https://www.reddit.com/r/pihole/comments/aazaqz/using_api_selecting_disable_twice_makes_it/

DL6ER:
…in setupVars.conf

---

Fix #2573 

```
pi@pi-hole:~ $ pihole disable
  [i] Disabling blocking
  [✓] Reloading DNS service
  [✓] Pi-hole Disabled

pi@pi-hole:~ $ pihole disable
  [i] Blocking already disabled, nothing to do

pi@pi-hole:~ $ pihole enable
  [i] Enabling blocking
  [✓] Reloading DNS service
  [✓] Pi-hole Enabled

pi@pi-hole:~ $ pihole enable
  [i] Blocking already enabled, nothing to do
```

Prevent double disabling/enabling by checking `BLOCKING_ENABLED` value in `setupVars.conf`

None

### DIFF
--- a/pihole
+++ b/pihole
@@ -144,6 +144,10 @@ Time:
 
   elif [[ "${1}" == "0" ]]; then
     # Disable Pi-hole
+    if grep -cq "BLOCKING_ENABLED=false" "${setupVars}"; then
+      echo -e "  ${INFO} Blocking already disabled, nothing to do"
+      exit 0
+    fi
     if [[ -e "${gravitylist}" ]]; then
       mv "${gravitylist}" "${gravitylist}.bck"
       echo "" > "${gravitylist}"
@@ -193,6 +197,10 @@ Time:
     fi
   else
     # Enable Pi-hole
+    if grep -cq "BLOCKING_ENABLED=true" "${setupVars}"; then
+      echo -e "  ${INFO} Blocking already enabled, nothing to do"
+      exit 0
+    fi
     echo -e "  ${INFO} Enabling blocking"
     local str="Pi-hole Enabled"
 


pihole
```diff
@@ -144,6 +144,10 @@ Time:
 
   elif [[ "${1}" == "0" ]]; then
     # Disable Pi-hole
+    if grep -cq "BLOCKING_ENABLED=false" "${setupVars}"; then
```
PromoFaux: Do you think maybe we should `source` setupVars here and test the value rather than grepping the file?

Just thinking for the sake of consistency, like we do here

https://github.com/pi-hole/pi-hole/blob/a4ba8d4b8fd0e44e11deeecf33267c82c3d8735e/pihole#L308

DL6ER: I know we source at different places. However, I also seem to recall that we said at some point that sourcing is dangerous as it could lead to executing arbitrary code - which may have entered `setupVars.conf` through an error - with elevated right. Hence, I chose a conservative approach that could be used as a blueprint for a redesign if anyone would want to do this.

PromoFaux: /shrug.

Either method works fine, tbh, I just didn't want to approve without first bringing it up...

dschaper: Good shout though. Both points are good. https://github.com/pruje/libbash.sh/blob/master/docs/libbash.md#lb_read_config may be the way to do this in the future, but for now. 

The grep approach though might need to account for a few different words for unset? `no` | `false` | `off` ? Is it possible for users to edit setupVars and put their own version of the setting?

PromoFaux: I guess they could, but anything other than `true` or `false` is going to be ignored in any case.

e.g if user sets `BLOCKING_ENABLED=unicorn` then it will be ignored!

PR #2611, opened 2019-02-03
Store blocking domains in a database

DL6ER:

---

Store all data generated by `pihole -g`
- whitelisted domains
- blacklisted domains
- gravity domains (domains from selected blocking lists)
in a common SQLite3 database.

This increases flexibility in the future when we, e.g., want to add more details to the individual domains like:
- creating/modification date
- user comments of arbitrary length
- an activated switch to temporarily disable individual entries without having to delete them
- etc.

Modify `gravity.sh` to store the lists in the corresponding database tables instead of in separate files.

None


gravity.sh
```diff
@@ -83,6 +87,65 @@ if [[ -r "${piholeDir}/pihole.conf" ]]; then
...
+  status="$?"
+
+  if [[ "${status}" -ne 0 ]]; then
+    echo -e "\\n  ${CROSS} Unable to create ${table} in database ${gravityDBfile}\\n  ${output}"
+    gravity_Cleanup "error"
+  fi
+
+  if [ "$table" == "whitelist" ] || [ "$table" == "blacklist" ] || [ "$table" == "regex" ]; then
+    # Set disabled to false
+    output=$( { sqlite3 "${gravityDBfile}" <<< "UPDATE ${table} SET disabled = 0 WHERE disabled IS NULL;"; } 2>&1 )
```
Mcat12: This should not be necessary, as `disabled` defaults to false if it is not specified: `disabled BOOLEAN DEFAULT 0`

gravity.sh
```diff
@@ -633,6 +683,21 @@ gravity_Cleanup() {
 
   echo -e "${OVER}  ${TICK} ${str}"
 
+  if ${optimize_database} ; then
+    str="Optimizing domains database"
+    echo -ne "  ${INFO} ${str}..."
+    # Store
```
Mcat12: Is this comment complete?

gravity.sh
```diff
@@ -33,13 +33,17 @@ regexFile="${piholeDir}/regex.list"
 adList="${piholeDir}/gravity.list"
 blackList="${piholeDir}/black.list"
 localList="${piholeDir}/local.list"
+whiteList="${piholeDir}/white.list"
 VPNList="/etc/openvpn/ipp.txt"
 
+piholeGitDir="/etc/.pihole"
+gravityDBfile="${piholeDir}/gravity.db"
+gravityDBschema="${piholeGitDir}/advanced/Templates/gravity.db.schema"
+optimize_database=false
```
Mcat12: Why is optimizing the database not on by default?

advanced/Templates/gravity.db.schema
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE whitelist (domain TEXT UNIQUE NOT NULL, disabled BOOLEAN DEFAULT 0, comment TEXT, DateAdded DATETIME);
```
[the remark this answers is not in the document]
DL6ER: No, the reason for this is somewhat non-intuitive, so let me explain it here.

We use `.import gravity.list` to mass-import the domains into the database. This is be far the fastest way of doing it. The issue is that it will fill all unspecified fields with `NULL`, ignoring the `DEFAULT` option we set there.

Later on, we select the blacklists using
```
WHERE a.domain NOT IN (SELECT domain from whitelist WHERE disabled != 1);
```
This `disabled != 1` matches two things:
- `0` (really enabled), **and**
- `NULL` (never set).

If we would specify an `enabled` field, we would need to test for two things here
```
WHERE a.domain NOT IN (SELECT domain from whitelist WHERE enabled == 1 OR enabled IS NULL);
```

However, I see that this is probably more straightforward and will implement this with a positive (`enabled`) logic.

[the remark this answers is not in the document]
DL6ER: This originates from @PromoFaux's schema, I will make it all lowercase.

advanced/Templates/gravity.db.schema
```diff
@@ -0,0 +1,21 @@
...
+CREATE TABLE blacklist (domain TEXT UNIQUE NOT NULL, disabled BOOLEAN DEFAULT 0, comment TEXT, DateAdded DATETIME);
+CREATE TABLE regex     (domain TEXT UNIQUE NOT NULL, disabled BOOLEAN DEFAULT 0, comment TEXT, DateAdded DATETIME);
```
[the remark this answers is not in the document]
DL6ER: You are right, I will call it `filter`

advanced/Templates/gravity.db.schema
```diff
@@ -0,0 +1,21 @@
...
+CREATE TABLE gravity   (domain TEXT UNIQUE NOT NULL);
+
+CREATE VIEW vw_gravity AS SELECT DISTINCT a.domain
+FROM gravity a
+WHERE a.domain NOT IN (SELECT domain from whitelist WHERE disabled != 1);
```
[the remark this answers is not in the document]
DL6ER: See above, I agree

gravity.sh
DL6ER: See above

gravity.sh
DL6ER: No

gravity.sh
DL6ER: It can take a *very long* time for those who run > 1 mio. domains. @PromoFaux tested it and it was no joy.

advanced/Templates/gravity.db.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE whitelist (domain TEXT UNIQUE NOT NULL, enabled BOOLEAN DEFAULT 1, comment TEXT, dateadded DATETIME);
+CREATE TABLE blacklist (domain TEXT UNIQUE NOT NULL, enabled BOOLEAN DEFAULT 1, comment TEXT, dateadded DATETIME);
+CREATE TABLE regex     (filter TEXT UNIQUE NOT NULL, enabled BOOLEAN DEFAULT 1, comment TEXT, dateadded DATETIME);
+CREATE TABLE gravity   (domain TEXT UNIQUE NOT NULL);
+
```
Mcat12: We should add a table to keep track of the database version. We may want to update the schema at some point, so we need to version it.

mmotti:
Will there be documentation for things like manually removing entries? Some of us currently use scripts to more than half the size of gravity.list depending on use wildcards and/or regexps.

DL6ER:
> Will there be documentation for things like manually removing entries (outside of a Pihole command)?

Sure, we will write an extensive documentation for https://docs.pi-hole.net as soon as the database features converges to an approved solution. As you see, there are still many discussions going on which shuffle around bits and bytes all over the place so starting already now with a documentation would only lead to having to re-write it quite often. Such often re-written documentation tend to unnecessarily complicate things which is why we will wait until at least this and the corresponding FTL PRs are approved.

advanced/Templates/gravity.db.sql
```diff
@@ -1,6 +1,6 @@
-CREATE TABLE whitelist (domain TEXT UNIQUE NOT NULL, enabled BOOLEAN DEFAULT 1, comment TEXT, DateAdded DATETIME);
-CREATE TABLE blacklist (domain TEXT UNIQUE NOT NULL, enabled BOOLEAN DEFAULT 1, comment TEXT, DateAdded DATETIME);
-CREATE TABLE regex     (domain TEXT UNIQUE NOT NULL, enabled BOOLEAN DEFAULT 1, comment TEXT, DateAdded DATETIME);
+CREATE TABLE whitelist (domain TEXT UNIQUE NOT NULL, enabled BOOLEAN DEFAULT 1, comment TEXT, dateadded DATETIME);
+CREATE TABLE blacklist (domain TEXT UNIQUE NOT NULL, enabled BOOLEAN DEFAULT 1, comment TEXT, dateadded DATETIME);
+CREATE TABLE regex     (domain TEXT UNIQUE NOT NULL, enabled BOOLEAN DEFAULT 1, comment TEXT, dateadded DATETIME);
```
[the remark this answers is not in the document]
Mcat12: In SQL statements and code, it may not be obvious if `added` is a boolean, timestamp, date, time, etc.

advanced/Templates/gravity.db.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE whitelist (domain TEXT UNIQUE NOT NULL, enabled BOOLEAN DEFAULT 1, comment TEXT, dateadded DATETIME);
+CREATE TABLE blacklist (domain TEXT UNIQUE NOT NULL, enabled BOOLEAN DEFAULT 1, comment TEXT, dateadded DATETIME);
+CREATE TABLE regex     (filter TEXT UNIQUE NOT NULL, enabled BOOLEAN DEFAULT 1, comment TEXT, dateadded DATETIME);
```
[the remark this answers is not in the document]
DL6ER: I'll use `date --utc +'%s'` as this is the recommended format for storing timestamps in SQLite3. It works as expected, you just have to tell SQLite3 that you want to see a string when querying
```
# sqlite3 /etc/pihole/gravity.db
SQLite version 3.16.2 2017-01-06 16:32:41
Enter ".help" for usage hints.
sqlite> select filter,datetime(date_added, 'unixepoch') from regex;
(^|.)meetrics.net$|2019-02-06 17:57:11
facebook|2019-02-06 17:57:11
sqlite>
```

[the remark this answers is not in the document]
Mcat12: That is actually more useful, because applications using the database can more easily use a numerical timestamp.

gravity.sh
```diff
@@ -113,25 +116,32 @@ gravity_store_in_database() {
     gravity_Cleanup "error"
   fi
 
+  local tmpFile=$(mktemp -p "/tmp" --suffix=".gravity")
+  if [ "$table" == "whitelist" ] || [ "$table" == "blacklist" ] || [ "$table" == "regex" ]; then
+    # Apply format for white-, blacklist, and regex tables, prevent globbing
+    set -f
+    for domain in $(cat < "${source}"); do
+      echo "\"${domain}\",1,$(date --utc +'%s')," >> "${tmpFile}"
```
Mcat12: Will this cause the `comment` column to have an empty string? The importer fills the missing columns with null, so the extra comma is not necessary.

gravity.sh
```diff
@@ -113,25 +116,33 @@ gravity_store_in_database() {
...
   fi
 
+  local tmpFile
+  tmpFile="$(mktemp -p "/tmp" --suffix=".gravity")"
+  if [ "$table" == "whitelist" ] || [ "$table" == "blacklist" ] || [ "$table" == "regex" ]; then
+    # Apply format for white-, blacklist, and regex tables
+    # Read file line by line
+    grep -v '^ *#' < "${source}" | while IFS= read -r domain
+    do
+      echo "\"${domain}\",1,$(date --utc +'%s')," >> "${tmpFile}"
```
Mcat12: The date could be retrieved once, avoiding a call to `date` for every item.

gravity.sh
```diff
@@ -113,25 +116,33 @@ gravity_store_in_database() {
...
+    done
+    inputfile="${tmpFile}"
```
Mcat12: Should this be defined above with `local inputfile`?

gravity.sh
```diff
@@ -113,25 +116,33 @@ gravity_store_in_database() {
...
+  else
+    # No need to modify the input data for the gravity table
+    inputfile="${source}"
+  fi
   # Store domains in gravity database table ${table}
-  output=$( { sqlite3 "${gravityDBfile}" <<< ".import \"${source}\" ${table}"; } 2>&1 )
+  # Use printf as .mode and .import need to be on separate lines
+  # see https://unix.stackexchange.com/a/445615/83260
+  output=$( { printf ".mode csv\\n.import \"%s\" ${table}\\n" "${inputfile}" | sqlite3 "${gravityDBfile}"; } 2>&1 )
```
Mcat12: Why mix different forms of interpolation here? (`%s` and `${table}`)

gravity.sh
```diff
@@ -113,25 +116,33 @@ gravity_store_in_database() {
...
-    output=$( { sqlite3 "${gravityDBfile}" <<< "UPDATE ${table} SET enabled = 1 WHERE enabled IS NULL;"; } 2>&1 )
-    status="$?"
-
-    if [[ "${status}" -ne 0 ]]; then
-      echo -e "\\n  ${CROSS} Unable to set enabled states (${table}) in database ${gravityDBfile}\\n  ${output}"
-      gravity_Cleanup "error"
-    fi
-  fi
+  # Delete tmpfile
+  #rm "$tmpFile" > /dev/null 2>&1 || true
```
Mcat12: Is this intended to be commented out?

gravity.sh
```diff
@@ -83,6 +89,74 @@ if [[ -r "${piholeDir}/pihole.conf" ]]; then
...
+  local source="${2}"
+  local template="${3}"
+
+  # Create database file if not present
+  if [ ! -e "${gravityDBfile}" ]; then
+    generate_gravity_database
+  fi
+
+  # Empty domains
+  output=$( { sqlite3 "${gravityDBfile}" <<< "DELETE FROM ${table};"; } 2>&1 )
```
Mcat12: This makes the source of data the `source` file. If we are moving this data to the database, does it make sense to do this?

advanced/Templates/gravity.db.sql
```diff
@@ -0,0 +1,24 @@
...
+CREATE TABLE blacklist (domain TEXT UNIQUE NOT NULL, enabled BOOLEAN NOT NULL DEFAULT 1, date_added DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP, comment TEXT);
+CREATE TABLE regex     (filter TEXT UNIQUE NOT NULL, enabled BOOLEAN NOT NULL DEFAULT 1, date_added DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP, comment TEXT);
+CREATE TABLE gravity   (domain TEXT UNIQUE NOT NULL);
+CREATE TABLE info      (property TEXT NOT NULL, value TEXT NOT NULL);
+
+INSERT INTO info VALUES("version","1");
+
+CREATE VIEW vw_gravity AS SELECT DISTINCT a.domain
+FROM gravity a
+WHERE a.domain NOT IN (SELECT domain from whitelist WHERE enabled == 1);
```
Mcat12: This does not take into account the blacklisted domains.

gravity.sh
DL6ER: I don't understand what you mean... There is no `source` in this line.

advanced/Templates/gravity.db.sql
DL6ER: This is intended, the blacklist domains have to be loaded separately such that FTL can distinguish between gravity and user blacklisted domains. This is the same as we do it right now. `gravity.list` does not include the domains from `black.list`.

pralor:

https://discourse.pi-hole.net/t/white-blacklist-add-reason-field/1439/11

gravity.sh
Mcat12: This PR moves the storage of domains to the database, so we should not use the files to determine the domains to store.

Currently, the database domains just mirror the domains in the files, instead, the database should be the source of truth.

In order to do this, we will have to update all of the places which read, insert, or update the lists (`list.sh` and the web interface). We will have to decide if we want to update the PHP web interface or start using the database with the new web interface and API.

DL6ER: > Currently, the database domains just mirror the domains in the files, instead, the database should be the source of truth.

The database is already the source of truth. We create a file containing all domains and mass-inject the domains into the database as this is faster than injecting each domain individually.

The corresponding branch of FTL cannot read the list files (it doesn't understand the simple list format we're using any longer) but sources the domains solely from the database file.

Mcat12: It is the source of truth for FTL, but it itself is generated from the list files. In the code that is referenced by this thread, gravity deletes all of the entries from the table and uses the file to fill in the entries. That means that the database entries are sourced from the files. Instead, the database should just do a one-time import from the files (maybe just on upgrade). All future list operations should operate on the database instead of the files.

DL6ER: This function is now only called:
- once for each file during upgrade to Pi-hole v5.0
- to move the generated `preEventHorizon` into the `gravity` table

The source files are always deleted after moving their content to the database.

automated install/basic-install.sh
```diff
@@ -1224,15 +1224,15 @@ version_check_dnsmasq() {
     # Local, named variables
     local dnsmasq_conf="/etc/dnsmasq.conf"
     local dnsmasq_conf_orig="/etc/dnsmasq.conf.orig"
-    local dnsmasq_pihole_id_string="addn-hosts=/etc/pihole/gravity.list"
+    local dnsmasq_pihole_id_string="# Dnsmasq config for Pi-hole's FTLDNS"
```
Mcat12: Older versions would not have this new string, so the old string should still be used.

advanced/Scripts/list.sh
```diff
@@ -88,175 +88,144 @@ HandleOther() {
...
     fi
 
     for dom in "${domList[@]}"; do
-        # Logic: If addmode then add to desired list and remove from the other; if delmode then remove from desired list but do not add to the other
+        # Format domain into regex filter if requested
+        if [[ "${wildcard}" == true ]]; then
+            dom="(^|\\.)${dom//\./\\.}$"
+        fi
+
+        # Logic: If addmode then add to desired list and remove from the other; if delmode then remove from desired list but do not add to the othe
```
Mcat12: Looks like the last character ("r") was cut off.

advanced/Scripts/list.sh
```diff
@@ -88,175 +88,144 @@ HandleOther() {
...
-            fi
-        fi
-    elif [[ "${list}" == "${regexlist}" ]]; then
-        [[ -z "${type}" ]] && type="--wildcard-only"
-        bool=true
-        domain="${1}"
 
-        [[ "${wildcard}" == true ]] && domain="(^|\\.)${domain//\./\\.}$"
+    # Is the domain in the list we want to add it to?
+    num="$(sqlite3 "${gravityDBfile}" "SELECT COUNT(*) FROM ${list} WHERE ${sqlitekey} = \"${domain}\";")"
```
Mcat12: `domain` should be checked to prevent SQL injections. This is a general comment, as it applies to every place we inject a variable that the user has control over into a SQL statement. Usually this is fixed by using a parameterized statement (put `?` where variables go, then supply the variables), but the SQLite CLI does not support this.

advanced/Scripts/list.sh
```diff
@@ -88,175 +88,144 @@ HandleOther() {
...
-                echo -e "  ${INFO} Adding ${domain} to regex list..."
-            fi
-            reload="restart"
-            echo "$domain" >> "${regexlist}"
-        else
-            if [[ "${verbose}" == true ]]; then
-                echo -e "  ${INFO} ${domain} already exists in regex list, no need to add!"
-            fi
+    if [[ "${num}" -eq 0 ]]; then
+        # Domain not found in the file, add it!
```
Mcat12: Comment is out of date ("file" vs "database")

advanced/Scripts/list.sh
```diff
@@ -88,175 +88,144 @@ HandleOther() {
...
+        if [[ "${verbose}" == true ]]; then
+            echo -e "  ${INFO} Adding ${1} to ${listname}..."
+        fi
+        reload=true
+        # Add it to the list we want to add it to
+        local timestamp
+        timestamp="$(date --utc +'%s')"
+        sqlite3 "${gravityDBfile}" "INSERT INTO ${list} (${sqlitekey},enabled,date_added) VALUES (\"${domain}\",1,${timestamp});"
```
Mcat12: The defaults for `enabled` and `date_added` in all of the tables are true and the current timestamp, respectively, so these columns can be omitted from the insert statement. This will also simplify the code.

advanced/Scripts/list.sh
```diff
@@ -88,175 +88,144 @@ HandleOther() {
...
+    else
+        if [[ "${verbose}" == true ]]; then
+            echo -e "  ${INFO} ${1} already exists in ${listname}, no need to add!"
         fi
```
Mcat12: Can this branch be moved earlier? (see `RemoveDomain` as well) This would reduce one level of indentation from the main code.
```bash
if [[ "${num}" -ne 0 ]]; then
    if [[ "${verbose}" == true ]]; then
        echo -e "  ${INFO} ${1} already exists in ${listname}, no need to add!"
    fi
    return
fi

# The rest of the code
```

advanced/Scripts/list.sh
```diff
@@ -88,175 +88,144 @@ HandleOther() {
...
-            string="gravity resistant domains"
-        else
-            string="domains caught in the sinkhole"
-        fi
-        verbose=false
-        echo -e "Displaying $string:\n"
+    local list listname count num_pipes domain enabled status
+
+    if [[ "${listType}" == "regex" ]]; then
+        listname="regex filters list"
```
Mcat12: I'm using the term `regexlist` in the API (although it's not user-facing), would something similar, maybe `regex list`, be usable here?

advanced/Scripts/list.sh
```diff
@@ -88,175 +88,144 @@ HandleOther() {
...
+    else
+        # Whitelist / Blacklist
+        listname="${listType}"
+    fi
+    data="$(sqlite3 "${gravityDBfile}" "SELECT * FROM ${listType};" 2> /dev/null)"
```
Mcat12: If we're only using the domain/filter and enabled columns, then the select statement can specify those columns instead of including the date added column.

advanced/Scripts/list.sh
```diff
@@ -88,175 +88,144 @@ HandleOther() {
...
+
+            echo "  ${count}: ${domain} (${status})"
             count=$((count+1))
-        done < "${listMain}"
-    else
-        echo -e "  ${COL_LIGHT_RED}${listMain} does not exist!${COL_NC}"
+        done <<< "${data}"
+        exit 0;
     fi
-    exit 0;
```
Mcat12: This exit seems to have been fine where it was, is there a reason for moving it?

gravity.sh
```diff
@@ -121,19 +115,19 @@ gravity_store_in_database() {
...
   local inputfile
-  if [ "$table" == "whitelist" ] || [ "$table" == "blacklist" ] || [ "$table" == "regex" ]; then
-    # Apply format for white-, blacklist, and regex tables
+  if [[ "${table}" == "gravity" ]]; then
+    # No need to modify the input data for the gravity table
+    inputfile="${source}"
+  else
+    # Apply format for white-, blacklist, regex, and adlists tables
     # Read file line by line
     grep -v '^ *#' < "${source}" | while IFS= read -r domain
```
Mcat12: Does this properly ignore empty lines?

gravity.sh
```diff
@@ -234,15 +231,9 @@ gravity_CheckDNSResolutionAvailable() {
...
-    rm "${adListDefault}" 2> /dev/null || \
-      echo -e "  ${CROSS} Unable to remove ${adListDefault}"
-  fi
-
-  # Retrieve source URLs from $adListFile
-  # Logic: Remove comments and empty lines
-  mapfile -t sources <<< "$(grep -v -E "^(#|$)" "${adListFile}" 2> /dev/null)"
+  # Retrieve source URLs from gravity database
+  # We source only enabled adlists, sqlite3 stores boolean values as 0 (false) or 1 (true)
+  mapfile -t sources <<< "$(sqlite3 "${gravityDBfile}" "SELECT address FROM adlists WHERE enabled = 1;" 2> /dev/null)"
```
Mcat12: Any reason for not using a view, as done for the whitelist and regex list? Or, should we remove the views from those tables and explicitly search only for enabled rows?

gravity.sh
```diff
@@ -552,58 +541,35 @@ gravity_SortAndFilterConsolidatedList() {
   if [[ "${haveSourceUrls}" == true ]]; then
```
Mcat12: It may be worthwhile to remove the `haveSourceUrls` variable. It seems to indicate if we have any adlists? If we do not have any, then we should exit earlier than this.

advanced/01-pihole.conf
```diff
@@ -18,8 +18,6 @@
 #                    WITHIN /etc/dnsmasq.d/yourname.conf                      #
 ###############################################################################
 
-addn-hosts=/etc/pihole/gravity.list
-addn-hosts=/etc/pihole/black.list
 addn-hosts=/etc/pihole/local.list
```
Mcat12: Perhaps in the future we can also remove this list, as it can be easily determined by FTL (only requires getting the IP addresses and the hostname)

advanced/Scripts/list.sh
DL6ER: Leaving off `enabled` is fine, however, when simplifying this string to
```
domain="abc.de"
"INSERT INTO ${list} (${sqlitekey}) VALUES (\"${domain}\");"
```
it results in an inserted line
```
abc.de|1|2019-04-26 14:28:08|
```
where the `date_added` field is filled with a ISO8601 (UTC) datetime *string* instead of a numeric timestamp. This is allowed by the SQLite specifications (the type of `DATETIME` columns can be `INTEGER`, `TEXT`, or even `REAL`), however, as it consumes extra space, I will keep the injection of a numerical timestamp obtained calling `date` if you have no objections and only omit the `enabled = 1` here.

Note: Even when defining the column as `date_added INTEGER NOT NULL DEFAULT CURRENT_TIMESTAMP` in the `CREATE TABLE` (not the `INTEGER` type!), the injection command will still store a `TEXT` object in the `date_added` column...
```
sqlite> CREATE TABLE www (domain TEXT UNIQUE NOT NULL, enabled BOOLEAN NOT NULL DEFAULT 1, date_added INTEGER NOT NULL DEFAULT CURRENT_TIMESTAMP, comment TEXT);
sqlite> insert into www (domain) values ("abc");
sqlite> select * from www;
abc|1|2019-04-26 14:37:21|
```

advanced/Scripts/list.sh
DL6ER: Sure, I will change it

advanced/Scripts/list.sh
DL6ER: I will also display the date added column. I intentionally left off a precise description to avoid having to differentiate between the name for the lists `domain` and for the regex table `filter`. 

@Mcat12 For the sake of simplicity, I'm more and more inclined to just call the first column of `regex` also `domain` and interested in your opinion.

advanced/Scripts/list.sh
Mcat12: It looks like we can have the best of both worlds, at least for the default:
https://stackoverflow.com/a/11556717

Storing as integer is fine by me, it would narrow the allowed types to what we want

advanced/Scripts/list.sh
Mcat12: That's fine by me. The filters describe a set of domains, so it's not too much of a stretch to use `domain` as the column name.

advanced/Scripts/list.sh
Mcat12: @dl6er it looks like we also need to cast to integer, as that function will return the number as a string: https://stackoverflow.com/a/29420016

gravity.sh
DL6ER: There simply was no view for `adlists` and I didn't think about adding one. I did that now.

advanced/01-pihole.conf
DL6ER: This is a good point. We will surely come back to it when the gravity script is seeing the next overhauling.

gravity.sh
DL6ER: I added some code which will ensure that empty lines won't get added to the import file we pass to the `sqlite3` command.

advanced/Scripts/list.sh
DL6ER: I'm using `printf %q` for this now, however, any help on improving this is appreciated.

advanced/Scripts/list.sh
```diff
@@ -194,19 +196,24 @@ Displaylist() {
             # Extract domain and enabled status based on the obtained number of pipe characters
             domain="$(cut -d'|' -f"-$((num_pipes-2))" <<< "${line}")"
             enabled="$(cut -d'|' -f"$((num_pipes-1))" <<< "${line}")"
+            dateadded="$(cut -d'|' -f"$((num_pipes))" <<< "${line}")"
 
+            echo "${dateadded}"
```
Mcat12: Is this for debugging?

advanced/Scripts/list.sh
```diff
@@ -119,63 +120,64 @@ ProcessDomainList() {
...
+    if [[ "${verbose}" == true ]]; then
+        echo -e "  ${INFO} Adding ${1} to ${listname}..."
+    fi
+    reload=true
+    # Add it to the list we want to add it to
+    local timestamp
+    timestamp="$(date --utc +'%s')"
+    # Insert only domain and date_added here. The enabled fields will be filled
+    # with its default value is true.
+    sqlite3 "${gravityDBfile}" "INSERT INTO ${list} (${sqlitekey},date_added) VALUES (\"${domain}\",${timestamp});"
```
Mcat12: Now that the default date added is in the form we want, we don't have to insert it here.

automated install/basic-install.sh
```diff
@@ -1224,7 +1224,8 @@ version_check_dnsmasq() {
     # Local, named variables
     local dnsmasq_conf="/etc/dnsmasq.conf"
     local dnsmasq_conf_orig="/etc/dnsmasq.conf.orig"
-    local dnsmasq_pihole_id_string="# Dnsmasq config for Pi-hole's FTLDNS"
+    local dnsmasq_pihole_id_string="addn-hosts=/etc/pihole/gravity.list"
+    local dnsmasq_pihole_id_string2="# Dnsmasq config for Pi-hole's FTLDNS"
```
Mcat12: This second string is not needed. There should never be a `/etc/dnsmasq.conf` with this string.

DL6ER: You're right, I misunderstood the purpose of this routine

advanced/Scripts/list.sh
```diff
@@ -125,7 +122,7 @@ AddDomain() {
     list="$2"
 
     # Is the domain in the list we want to add it to?
-    num="$(sqlite3 "${gravityDBfile}" "SELECT COUNT(*) FROM ${list} WHERE ${sqlitekey} = \"${domain}\";")"
+    num="$(sqlite3 "${gravityDBfile}" "SELECT COUNT(*) FROM ${list} WHERE domain = \"${domain}\";")"
```
Mcat12: Double quotes in SQL are usually used to wrap column or table names, whereas single quotes are used for strings (according to the SQL standards). SQLite appears to support using double quotes for strings, but we should try to stick to the standard.

gravity.sh
```diff
@@ -85,8 +82,6 @@ fi
 # Generate new sqlite3 file from schema template
 generate_gravity_database() {
   sqlite3 "${gravityDBfile}" < "${gravityDBschema}"
-  chown ${DATABASE_USER}:${DATABASE_GROUP} "${piholeDir}" "${gravityDBfile}"
-  chmod ${DATABASE_PERMISSIONS} "${piholeDir}" "${gravityDBfile}"
```
Mcat12: Why are the permissions and user ownership removed?

DL6ER: `sqlite3` creates the database with `rw-r--r--`, owned by `root:root`. As only `gravity.sh` and `lists.sh` need write access (they are `root`), this is sufficient. The ownership code slipped in there accidentally and was part of trying to modify the database directly from the web interface. I think it is safe to say that the fewer users have write permissions on this database, the better.

We could still revisit this in a follow-up PR.

Mcat12: These permissions will prevent the (current) web interface and the new API from accessing the database. The web interface directly interacts with the database when reading.

Mcat12: Do we want to allow anyone to read the database? This will make it 644.

advanced/Scripts/list.sh
Mcat12: I think this is mostly resolved now, the domain filter will weed out the `;` and `'` which are the most important characters for doing a SQL injection in this case. We can revisit this when we finally move this functionality into something which supports parameterized queries.

Mcat12:
The query adlists functionality is missing:
```
$ pihole -q test
The file /etc/pihole/adlists.list was not found
```

advanced/Templates/gravity.db.sql
```diff
@@ -1,7 +1,7 @@
...
-CREATE TABLE regex     (domain TEXT UNIQUE NOT NULL, enabled BOOLEAN NOT NULL DEFAULT 1, date_added INTEGER NOT NULL DEFAULT (cast(strftime('%s', 'now') as int)), comment TEXT);
-CREATE TABLE adlists   (address TEXT UNIQUE NOT NULL, enabled BOOLEAN NOT NULL DEFAULT 1, date_added INTEGER NOT NULL DEFAULT (cast(strftime('%s', 'now') as int)), comment TEXT);
+CREATE TABLE whitelist (domain TEXT UNIQUE NOT NULL, enabled BOOLEAN NOT NULL DEFAULT 1, date_added INTEGER NOT NULL DEFAULT (cast(strftime('%s', 'now') as int)), date_modified INTEGER, comment TEXT);
+CREATE TABLE blacklist (domain TEXT UNIQUE NOT NULL, enabled BOOLEAN NOT NULL DEFAULT 1, date_added INTEGER NOT NULL DEFAULT (cast(strftime('%s', 'now') as int)), date_modified INTEGER, comment TEXT);
+CREATE TABLE regex     (domain TEXT UNIQUE NOT NULL, enabled BOOLEAN NOT NULL DEFAULT 1, date_added INTEGER NOT NULL DEFAULT (cast(strftime('%s', 'now') as int)), date_modified INTEGER, comment TEXT);
+CREATE TABLE adlists   (address TEXT UNIQUE NOT NULL, enabled BOOLEAN NOT NULL DEFAULT 1, date_added INTEGER NOT NULL DEFAULT (cast(strftime('%s', 'now') as int)), date_modified INTEGER, comment TEXT);
```
Mcat12: Date modified can have the same properties as date added for now. The difference would be that we would updated it on every modification.

advanced/Scripts/webpage.sh
```diff
@@ -385,19 +387,17 @@ SetWebUILayout() {
...
-        sed -i "\\@${args[3]}@s/^#http/http/g" "${list}"
+        sqlite3 "${gravityDBfile}" "UPDATE adlists SET enabled = 1 WHERE address = '${address}'"
     elif [[ "${args[2]}" == "disable" ]]; then
-        sed -i "\\@${args[3]}@s/^http/#http/g" "${list}"
+        sqlite3 "${gravityDBfile}" "UPDATE adlists SET enabled = 0 WHERE address = '${address}'"
     elif [[ "${args[2]}" == "add" ]]; then
-        if [[ $(grep -c "^${args[3]}$" "${list}") -eq 0 ]] ; then
-            echo "${args[3]}" >> ${list}
-        fi
+        sqlite3 "${gravityDBfile}" "INSERT OR IGNORE INTO adlists (address) VALUES ('${address}')"
```
Mcat12: This should probably be `INSERT OR REPLACE` instead of ignore.

advanced/Scripts/webpage.sh
```diff
@@ -385,19 +387,17 @@ SetWebUILayout() {
 }
 
 CustomizeAdLists() {
-    list="/etc/pihole/adlists.list"
+    local address
+    address="${args[3]}"
 
     if [[ "${args[2]}" == "enable" ]]; then
-        sed -i "\\@${args[3]}@s/^#http/http/g" "${list}"
+        sqlite3 "${gravityDBfile}" "UPDATE adlists SET enabled = 1 WHERE address = '${address}'"
```
Mcat12: So much cleaner :heart_eyes:

gravity.sh
```diff
@@ -140,8 +142,9 @@ database_table_from_file() {
   rm "${tmpFile}" > /dev/null 2>&1 || \
       echo -e "  ${CROSS} Unable to remove ${tmpFile}"
 
-  # Delete source file
-  rm "${source}" 2> /dev/null || \
+  # Move source file to backup directory, create directory if not existing
+  mkdir -p "${backup_path}"
+  mv "${source}" "${backup_file}" 2> /dev/null || \
       echo -e "  ${CROSS} Unable to remove ${source}"
```
Mcat12: Error should be changed to say it was unable to backup the source file.

advanced/Scripts/webpage.sh
PromoFaux: If we replace then it gets a new added date. Surely, by ignoring, we keep the original added date?

Mcat12: Yes. The user wants to insert an adlist, so if the adlist already exists we should either error out or replace the old adlist. It would be better if we threw an error here or checked earlier, but if that is not in place then the adlist should be replaced.

Mcat12: Thinking on this again, unless there is some new data, like a comment, being inserted, then we can ignore the insert as the URL is the same.

advanced/Templates/gravity.db.sql
DL6ER: It depends on your interpretation of `date_modified`, i.e., if adding is a modification; I wouldn't necessarily say so.

advanced/Templates/gravity.db.sql
```diff
@@ -8,21 +8,38 @@ CREATE TABLE info      (property TEXT NOT NULL, value TEXT NOT NULL);
 INSERT INTO info VALUES("version","1");
 
 CREATE VIEW vw_gravity AS SELECT DISTINCT a.domain
```
Mcat12: `DISTINCT` is not necessary, as the domain column holds unique values.

advanced/Templates/gravity.db.sql
```diff
@@ -8,21 +8,38 @@ CREATE TABLE info      (property TEXT NOT NULL, value TEXT NOT NULL);
...
 
 CREATE VIEW vw_whitelist AS SELECT DISTINCT a.domain
-FROM whitelist a
-WHERE a.enabled == 1;
+    FROM whitelist a
+    WHERE a.enabled == 1;
+CREATE TRIGGER tr_whitelist_update AFTER UPDATE ON whitelist
+    BEGIN
+      UPDATE whitelist SET date_modified = (cast(strftime('%s', 'now') as int)) WHERE domain = NEW.domain;
+     END;
```
Mcat12: Are these `END` statements supposed to be indented an extra space?

DL6ER:
@Mcat12 The query adlists function should now be fixed as well

```
# pihole -q -exact doubleclick.net
 Exact match for doubleclick.net found in:
   list.5.s3.amazonaws.com.domains
```
```
# pihole -q -exact -adlist doubleclick.net
 Exact match for doubleclick.net found in:
   https://s3.amazonaws.com/lists.disconnect.me/simple_ad.txt
```

advanced/Templates/gravity.db.sql
```diff
@@ -0,0 +1,45 @@
+CREATE TABLE whitelist (domain TEXT UNIQUE NOT NULL,  enabled BOOLEAN NOT NULL DEFAULT 1, date_added INTEGER NOT NULL DEFAULT (cast(strftime('%s', 'now') as int)), date_modified INTEGER NOT NULL DEFAULT (cast(strftime('%s', 'now') as int)), comment TEXT);
```
Mcat12: Can these `UNIQUE NOT NULL` columns instead be primary keys? A primary key has these constraints asking with additional properties: https://www.w3schools.com/sql/sql_primarykey.asp

Mcat12: Or, we should insert an ID column and make that the primary key.

DL6ER: Yes, that's possible but conflicts with the idea of adding INTEGER ids as primary keys.
Even in the example you referenced they specify the `NOT NULL` in addition to `PRIMARY KEY`...:
```sql
CREATE TABLE Persons (
    ID int NOT NULL PRIMARY KEY,
    LastName varchar(255) NOT NULL,
    FirstName varchar(255),
    Age int
); 
```

Mcat12: > Primary keys must contain UNIQUE values, and cannot contain NULL values.

Because of this, you don't need the `NOT NULL`

advanced/Scripts/list.sh
```diff
@@ -88,175 +87,143 @@ HandleOther() {
...
-            string="gravity resistant domains"
-        else
-            string="domains caught in the sinkhole"
-        fi
-        verbose=false
-        echo -e "Displaying $string:\n"
+    local list listname count num_pipes domain enabled status nicedate
+
+    listname="${listType}"
+    data="$(sqlite3 "${gravityDBfile}" "SELECT * FROM ${listType};" 2> /dev/null)"
```
Mcat12: This will also return the ID, which is not handled. The select should probably be changed to only return the columns we need.
```
$ pihole -w -l
Displaying whitelist:
  1: 1|doubleclick.com (enabled, added Thu, 02 May 2019 03:33:01 +0000)
```

advanced/Scripts/query.sh
```diff
@@ -186,11 +181,8 @@ fi
 
 # Get adlist file content as array
 if [[ -n "${adlist}" ]] || [[ -n "${blockpage}" ]]; then
-    for adlistUrl in $(< "${adListsList}"); do
-        if [[ "${adlistUrl:0:4}" =~ (http|www.) ]]; then
-            adlists+=("${adlistUrl}")
-        fi
-    done
+    # Retrieve source URLs from gravity database
+    mapfile -t adlists <<< "$(sqlite3 "${gravityDBfile}" "SELECT address FROM vw_adlists;" 2> /dev/null)"
```
Mcat12: `query.sh` also searches in the whitelist and blacklist. That code should be updated to work with the database.

Mcat12: This is taken care of in #2734

Th3M3:
@DL6ER:
The pop-up message on the red tile on the Dashboard shows not the correct time, since when the Lists had been updated. 
The shown timespan still gets calculated from the `gravity.list` file.

Mcat12:
@Th3M3 Please make a new issue instead of commenting on this closed PR.

PR #2612, opened 2019-02-03
Remove user/group from dnsmasq config

Mcat12:
This is a hotfix release (just core), and will be merged to `master` and `development`

In the case that FTL is started under root, dnsmasq changes the user to pihole due to this setting. The shared memory is created before this switch, and may cause issues such as failing to delete or reallocate the shared memory.

These config lines were necessary before dnsmasq became part of FTL, but now since dnsmasq runs under FTL, we can configure the user/group through FTL.


pralor:

https://discourse.pi-hole.net/t/crash-after-update-to-4-2/16769/11

PR #2646, opened 2019-02-22
Don't fail when trying to detect OS architecture on RPM-based systems

DL6ER:

---
Fix a crash when trying to detect the FTL architecture on rpm-based x86_64 systems.

There is a check which only works on `deb`-based systems. I modified the code to proceed correctly when `dpkg` is not available.

A crash is seen, e.g., here: https://github.com/pi-hole/FTL/issues/496#issuecomment-466367909

None

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -2294,10 +2294,12 @@ get_binary_name() {
     elif [[ "${machine}" == "x86_64" ]]; then
         # This gives the architecture of packages dpkg installs (for example, "i386")
         local dpkgarch
-        dpkgarch=$(dpkg --print-architecture 2> /dev/null)
+        dpkgarch=$(dpkg --print-architecture 2> /dev/null || true)
 
         # Special case: This is a 32 bit OS, installed on a 64 bit machine
         # -> change machine architecture to download the 32 bit executable
+        # We only check this for Debian-based systems as this has been an issue
+        # in the past (see https://github.com/pi-hole/pi-hole/pull/2004)
         if [[ "${dpkgarch}" == "i386" ]]; then
             printf "%b  %b Detected 32bit (i686) architecture\\n" "${OVER}" "${TICK}"
             binary="pihole-FTL-linux-x86_32"


DL6ER:
@bcambl How'd this look like? In `uname`'s `mna` page,  see
```
       -m, --machine
              print the machine hardware name

       -i, --hardware-platform
              print the hardware platform (non-portable)
```
I'm a bit worried about this "non-portable"

bcambl:
I would say it is more portable than a command that only exists on Debian family.

bcambl:
> Print ‘unknown’ if this information is not available. Note this is non-portable (even across GNU/Linux distributions).

bcambl:
>   -m, --machine            print the machine hardware name
>   -p, --processor          print the processor type (non-portable)
>   -i, --hardware-platform  print the hardware platform (non-portable)
> 
> 

_[bcambl@unknown ~]$ uname -m
x86_64
[bcambl@unknown ~]$ uname -p
x86_64
[bcambl@unknown ~]$ uname -i
x86_64
[bcambl@unknown ~]$_

bcambl:
Also worth noting that the `arch` command ships along with the same package that provides `uname` (`coreutils`) and returns the same data as the _portable_ `uname -m`.

bcambl:
Also we might want to test for both _i386_ and _i686_.

DL6ER:
@bcambl I really don't know how it should look like for `rpm` systems, we've seen `i368` being returned from `dpkg` which is for what this `if` statement was designed for initially. `arch` reported a 64bit CPU for the user but the system was `i386` and even though the system might have been able to run 64bit binaries, the OS was lacking the 64bit libraries.

I think this strange behavior is what you'd expect if you install a 32bit OS on a 64bit CPU machine. We could drop support for this bizarre case but the more users you have, the more uncommon setups you see...

```bash
rpm -q --qf '%{ARCH}\n' glibc | grep x86_64 || echo "i386"
```
seems okay.

bcambl:
Actually.. 32bit should never be installed on Fedora or CentOS any way.
FTL on Fedora 29+ and CentOS 7+ should be x86_64 or ARM only.

DL6ER:
@bcambl So we can go back to my initial idea of
```
dpkg || true
```
to just skip the test for x86_64 without `dpkg`??

bcambl:
I would say so. The last 32bit on the Fedora side was CentOS6 (unsupported).
My apologies for the run-around.. /facepalm

Mcat12:
@DL6ER @bcambl In order to finish off this PR, it looks like the changes should be reverted back to the original `dpkg || true` idea?

DL6ER:
I agree, I will hard-reset to https://github.com/pi-hole/pi-hole/pull/2646/commits/9867b7fccfdd95483680b34f5aa2636674016783 and force-push. This will provide a clean history while the comments in this PR will be preserved.

DL6ER:
@Mcat12 I did as I promised to do. Please re-approve when you find the time to look at it.

PR #2681, opened 2019-03-23
replaced @reboot with @weekly

aoitrix:
Current state is: we rotate logfiles once every reboot... there are several problems
on unstable systems we run lograte once every reboot... (well at least to read in the config file) which causes unneeded io
we rotate logfiles only on reboot, which might cause a garbage on bigger logfiles...

@reboot is a non-standard value, so i picked @weekly instead of a full crontab line

this will allow to run logrotate weekly, still overriding the daily setting in the config file... but i think @weekly should make sense in more cases 

Signed-off-by: Aoi <aoi.klarmann@gmail.com>

- [c] I give this submission freely and claim no ownership.

---

---

### DIFF
--- a/advanced/Templates/pihole.cron
+++ b/advanced/Templates/pihole.cron
@@ -26,7 +26,7 @@
 #          parameter "quiet": don't print messages
 00 00   * * *   root    PATH="$PATH:/usr/local/bin/" pihole flush once quiet
 
-@reboot root /usr/sbin/logrotate /etc/pihole/logrotate
+@weekly root /usr/sbin/logrotate /etc/pihole/logrotate
 
 # Pi-hole: Grab local version and branch every 10 minutes
 */10 *  * * *   root    PATH="$PATH:/usr/local/bin/" pihole updatechecker local


dschaper:
First, please rebase your pull request on Development, not Master as requested in the template.

Logrotate does run on each reboot, but that only reads the configuration for logrotate to determine if the files should actually be rotated. Since the default template shows that logs should be rotate daily, then setting that cronjob to `@weekly` kind of defeats the intent. Would it be better to modify the logrotate script to be a time period other than daily?

```
/var/log/pihole.log {
	# su #
	daily
	copytruncate
	rotate 5
	compress
	delaycompress
	notifempty
	nomail
}

/var/log/pihole-FTL.log {
	# su #
	weekly
	copytruncate
	rotate 3
	compress
	delaycompress
	notifempty
	nomail
}
```
https://github.com/pi-hole/pi-hole/blob/master/advanced/Templates/logrotate#L3

hexathos:
It would be better, to check for a logrotate installation and then copy the logrotate config file to /etc/logrotate.d/ or the proper systems destination, instead of heaving called logrotate multiple times. that defeats the "will not run without reboot" case would also honor the configuration settings.

Mcat12:
Closed due to submitter inactivity.

Issue #975, opened 2016-12-09
Add custom DNS entrees


paulkramme:
**In raising this issue, I confirm the following (please check boxes, eg [X]) Failure to fill the template will close your issue:**

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*
- [X] The issue I'm reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the codebase?:**

- [] 1 (very unfamiliar)
- [X] 2
- [] 3
- [] 4
- [] 5
- [] 6
- [] 7
- [] 8
- [] 9
- [] 10 (very familiar)

---
**FEATURE REQUEST:**
I'd like to add easy DNS entrees, so i can reach my local devices over them. Since hostnames are not working after switching DNS from router to pihole, this would be nice. I'm very sorry if i reposted this, i did not find it under open issues.

_This template was created based on the work of [`udemy-dl`](https://github.com/nishad/udemy-dl/blob/master/LICENSE)._

dschaper:
Are you asking to use a custom upstream provider for your DNS server (in place of Google or OpenDNS for example) or are you asking about entering data into the Pi-hole's HOSTS file so that you can reach local machines by their short names?

paulkramme:
Latter.

Mcat12:
Add your hosts to your Pi-hole's `/etc/hosts` file.

dschaper:
Okay, I will pass that along as a feature request, it's possible to do now, if you use the Pi-hole's `dnsmasq` process as your DHCP server, but we need to work out a way to safely edit the `/etc/hosts` file and determine a process to present those options to the users. If the Pi-hole also acts as the DHCP server, then `dnsmasq` will automatically add the short name to the resolver's lookups but this would require disabling the DHCP server on the router and passing that function over to the Pi. It's a feature we are talking about, but we want to do it in a sane way that will hopefully avert the possibility of a user locking themselves out, :) If you'd like to open a Feature Request over on Discourse.Pi-hole.net then other users will be able to vote and comment on the idea.

jannylund:
[...]

PR #2692, opened 2019-04-02
API and web integration

Mcat12:

---

Note that this is a draft PR for now. It may stay this way until after the beta period is finished.

**How to test**
1. Start a fresh server
2. Clone this repository to `/etc/.pihole`
3. Checkout `feature/api`
4. Make a file at `/etc/pihole/ftlbranch` with the contents: `development`
5. Make a file at `/etc/pihole/apibranch` with the contents: `development`
6. Run `./automated\ install/basic-install.sh`

When the installer is finished, you will have the web interface running on port 80. The API will be able to connect to FTL and serve stats to the web interface. All web interface functions should work, including changing DNS options (it will reflect in `/etc/dnsmasq.d/pihole.conf`).

- Replace lighttpd and PHP with the new API and web interface.
- Remove now-unnecessary files from the repo (lighttpd configs, PHP block page, etc)
- Delegate dnsmasq config generation to the API.

- The lighttpd- and PHP-specific code and files are removed from the repository.
- The API is added in a similar way as FTL, including support for updating and checking out different branches.
- The dnsmasq config generation code runs an API command to generate the config instead of manually modifying it as it did before.

Remove documentation on lighttpd and PHP, replace with API and new web interface documentation.

**TODO Items**
These are some general items, but check the [API and web interface board](https://github.com/orgs/pi-hole/projects/2) for more items.
- [ ] Add an upgrade path from the current Pi-hole version (with lighttpd and PHP) to this one.
- [x] Add support for not hosting the web interface (`INSTALL_WEB_INTERFACE=false`)
- [ ] Write lots of documentation about the API and web interface
- [ ] Start a beta period
- [x] Support CentOS. Currently, the API is compiled with too new of a glibc version. A fix is to change the build system to use CentOS for building. This would make sure any distro that's at least as new as CentOS can run the API.


PromoFaux:
> Add an upgrade path from the current Pi-hole version (with lighttpd and PHP) to this one.

This is going to be an interesting challenge. With the release of `pihole-FTL` we could confidently remove `dnsmasq` without causing any issues, as it uses the same config files in the same locations. How are we going to determine that _we_ actually installed `lighttpd`? What if the user has more than the Pi-hole web interface being served by `lighttpd`?

Mcat12:
We may have to put up a question to the user, "Remove lighttpd and PHP?"

Defaults to yes. If yes, then remove and continue. If no, then change the lighttpd port (or the API's?) and continue.

advanced/Scripts/piholeDebug.sh
```diff
@@ -712,50 +662,6 @@ check_networking() {
     check_required_ports
 }
 
-check_x_headers() {
```
PromoFaux: Are we ditching the idea of a blockpage entirely with the NGweb?

automated install/uninstall.sh
```diff
@@ -131,17 +113,8 @@ removeNoPurge() {
...
-        echo -e "  ${TICK} Removed lighttpd"
-    else
-        if [ -f /etc/lighttpd/lighttpd.conf.orig ]; then
-            ${SUDO} mv /etc/lighttpd/lighttpd.conf.orig /etc/lighttpd/lighttpd.conf
-        fi
-    fi
-
     ${SUDO} rm -f /etc/dnsmasq.d/adList.conf &> /dev/null
-    ${SUDO} rm -f /etc/dnsmasq.d/01-pihole.conf &> /dev/null
+    ${SUDO} rm -f /etc/dnsmasq.d/pihole.conf &> /dev/null
```
PromoFaux: Should this file, despite being generated by the API, not still be named `01-pihole.conf` to ensure it is read in first? I thought files in `/etc/dnsmasq.` were read in alphanumerical order?

Mcat12: The reason for stripping the prefix was that there is only one Pi-hole config file, and the prefix is unnecessary. The order of the config files does not matter much, I don't think dnsmasq really cares. If there is a conflicting option, it would throw an error no matter the order.

advanced/Scripts/piholeDebug.sh
Mcat12: As of now, there's no plans that I know of to add a block page. It is feasible however; the API can support a server-side rendered page. The reason for server-side rather than client-side is that we don't want to alarm the user's browser by running JavaScript on a blocked (possibly malicious) domain.

PromoFaux: I'm just thinking that there are still going to be users that want that functionality, even though we recommend otherwise.

Server side rendered page is probably OK though? Work out all the values in the API and then just send raw HTML back to the browser. I like that idea!

Mcat12: Should this be added to the MVP requirements? @pi-hole/core-approvers

automated install/uninstall.sh
mwoolweaver: i know when you specify multiple dns entries to be distributed it recommended to them in a file that is read as close to last as possible to ensure that they are used. 

https://www.reddit.com/r/pihole/comments/b88ciu/question_im_using_2_piholes_how_to_make_one_hand/

Mcat12: @mwoolweaver I do not see that sentiment in the link you shared. Can you link to the comment with this information?

> close to last as possible to ensure that they are used.

If the config file is read, it will be used. Moving its position in the order of config files does not change if it will be parsed and used.

advanced/Scripts/piholeDebug.sh
Mcat12: A team discussion decided to make this a requirement for release, but not necessarily for alpha/beta release. See pi-hole/api#130

automated install/basic-install.sh
```diff
@@ -2570,49 +2463,40 @@ main() {
     clone_or_update_repos
 
     # Install the Core dependencies
-    local dep_install_list=("${PIHOLE_DEPS[@]}")
-    if [[ "${INSTALL_WEB_SERVER}" == true ]]; then
-        # Install the Web dependencies
-        dep_install_list+=("${PIHOLE_WEB_DEPS[@]}")
-    fi
+    install_dependent_packages PIHOLE_DEPS[@]
```
oakkitten: this fails for me, did you mean `"${PIHOLE_DEPS[@]}"`?

oakkitten:
i was able to install this, but it failed to launch as i already have a web server running on port 80. is there a way to configure the web server? is there a way to turn the server off and use my own web server (which i am doing with the regular pi-hole)? there's an error while adding stuff to the exact blacklist and whitelist, how do i go about debugging that?

PromoFaux:
> is there a way to turn the server off and use my own web server

If you're coming from a standard Pi-hole installation as a base, then `lighttpd` is still running. 

Try `sudo service lighttpd stop`, which will stop it from listening on port 80

oakkitten:
i don't think “Error: Rocket failed to bind network socket to given address/port” is a lighttpd message

PromoFaux:
No, _but_ rocket is unable to bind to the port because `lighttpd` is already bound to it, so if you disable/stop `lighttpd`, then rocket _should_ be able to bind. I think!

oakkitten:
i don't have lighttpd installed. i only have nginx. nginx uses the ports 80 and 443. i can turn it off and it will work, but i don't want to turn it off, as it's serving a website to people

Mcat12:
You can change the port by creating a file at `/etc/pihole/API.toml` and adding the following:
```toml
[general]
port = 80
```

oakkitten:
thanks, that worked! i was also able to bind it to 127.0.0.1, which is nice. what about my other questions?

also, i thought one of the problems this was going to solve was #2729, but it seems that this rocket things uses absolute urls and so every request starts with `/admin`, making it impossible to use pihole on any other location without hacking into rocket's response. which is doable if you replace every `/admin` with `/whatever`, but who knows what it could break.

also, the whitelist/blacklist problem was due to blacklist.txt and whitelist.txt not existing. while it works now, adding stuff to white/blacklists takes around 10-15 seconds. adding to the regex list is instantaneous.

PromoFaux:
> i don't have lighttpd installed. i only have nginx.

Whoops! I misread what you wrote before, and assumed you were on a standard install. My bad.

oakkitten:
i'm trying the following configuration:

```
[general]
port = 8081

[web]
path = "/test"
```
and i'm getting

```
🔧  Configured for production.
    => address: 0.0.0.0
    => port: 8081
    => log: critical
    => workers: 8
    => secret key: generated
    => limits: forms = 32KiB
    => keep-alive: 5s
    => tls: disabled
Warning: environment is 'production', but no `secret_key` is configured
    => [extra] databases: { ftl_database = { url = "/etc/pihole/pihole-FTL.db" } }
🚀  Rocket has launched from http://0.0.0.0:8081
Error: No matching routes for GET /test text/html.
Warning: Responding with 404 Not Found catcher.
```

`localhost:8081/admin` works fine, though

cyberb:
As mentioned above I would like to add a simple ldap authentication option for syncloud app self-hosting project.
Do you plan eventually to merge this so I can start looking at rust implementation otherwise I can do it in php?

Mcat12:
@cyberb we do not have an ETA on when this will be merged. Sorry for the slow progress.

cyberb:
1. Is path to config file (/etc/pihole/API.toml) hardcoded or can be passed as a command line argument?
2. Is it possible to listen on a unix socket instead of a port?

cyberb:
If I understand you are using this web framework which does not have unix socket support yet: https://github.com/SergioBenitez/Rocket/pull/594

Mcat12:
The config location is currently hardcoded, though it should be fine to add a CLI argument to pass a different one (like `-f /some/path/API.toml`):
https://github.com/pi-hole/api/blob/c89c5aba63df266c15d44bdbd1fa7e95230868e1/src/env/config/root_config.rs#L36-L40

You are correct that we use Rocket, and it does not yet support listening on a unix socket. One possible workaround is to use a proxy like nginx which listens on a unix socket and forwards to the API.

cyberb:
I am getting this ftl error:
````
no such table: domain_audit
````
Do you need to rebase this branch on develop as it has gravity db migration scripts?
I am using pihole:feature/api, FTL:development, api:development, web: development

Mcat12:
@cyberb I've updated the branch and fixed some issues due to the new changes. You will need to use the API branch [`fix/connect-missing-db`] until that is merged to development.

[`fix/connect-missing-db`]: https://github.com/pi-hole/api/pull/227

cyberb:
Looks like whitelisting does not work from new web/api and requires FTL to be restarted to include new whitelist entry.
Also could anyone from the core team express an official opinion on the future of new api/web?

PromoFaux:
>Also could anyone from the core team express an official opinion on the future of new api/web?

We're exploring a couple of other paths at the moment, too. For example, we're looking at the possibility of embedding an API directly into FTL (see [this branch](https://github.com/pi-hole/FTL/tree/new/http)), as an alternative to `rust`.

The new web repo will likely remain as it is (react), though obviously depends on us deciding ultimately which route we are going to go down API-wise.

automated install/uninstall.sh
mwoolweaver: https://github.com/pi-hole/pi-hole/issues/975#issuecomment-281074624

https://discourse.pi-hole.net/t/pihole-update-how-to-retain-my-dnsmaq-settings-dhcp-interaface-etc/269/13?u=mwoolweaver1

https://discourse.pi-hole.net/t/allow-more-granular-control-of-dhcp-lease-time/16500/11?u=mwoolweaver1

https://discourse.pi-hole.net/t/dhcp-option-66-tftp/8437/5?u=mwoolweaver1

https://discourse.pi-hole.net/t/ignore-explicit-whitelisted-domains-from-query-log/2335/14?u=mwoolweaver1


It does seem that everyone’s opinion isn’t the same.

Issue #2699, opened 2019-04-11
The Splash page should use relative URLs


obones:
**In raising this issue, I confirm the following:** `{please fill the checkboxes, e.g: [X]}`

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the the source code relevant to this issue?:**

1

---
**Expected behaviour:**
The HTTP server can be configured to serve pihole content in a subfolder like so:
http://hostname/pihole/

When going to that URL, one would expect to see the regular Splash page with the logo, a dark background and a functioning link to the admin panel

**Actual behaviour:**

The showed page has no style, the image is missing and the admin link goes to `/admin` instead of `/pihole/admin`

**Steps to reproduce:**

Configure the web server (lighthttpd, nginx or even Apache) to serve pihole content out of a subfolder and not the root.

**Troubleshooting undertaken, and/or other relevant information:**

The associated pull request addresses the issue by making sure all URLs inside the Splash page are relative.

`{Steps of what you have done to fix this}`

See pull request

Mcat12:
> The HTTP server can be configured to serve pihole content in a subfolder like so:
> http://hostname/pihole/

This is not supported with the PHP (current) web interface.

obones:
I am using it right now on my own network, and it appears to work just fine. What should I check that I have missed to make sure everything is ok?

Mcat12:
When I say not supported, I mean it is not tested or designed to work in that way, and therefore we can not provide support for that use case.

PR #2700, opened 2019-04-11
Relative URLs for when PiHole interface is accessed via a (virtual) subfolder

obones:
Signed-off-by: obones <obones@free.fr>

---
This is a proposed fix for https://github.com/pi-hole/pi-hole/issues/2699

This adds a `.` dot in front of all URLs in the generated Splash page so that URLs are relative to the folder from where the content is served instead of being forced at the root of the webserver.

None that comes to mind

### DIFF
--- a/advanced/index.php
+++ b/advanced/index.php
@@ -58,8 +58,8 @@ if ($serverName === "pi.hole") {
     $splashPage = "
     <html><head>
         $viewPort
-        <link rel='stylesheet' href='/pihole/blockingpage.css' type='text/css'/>
-    </head><body id='splashpage'><img src='/admin/img/logo.svg'/><br/>Pi-<b>hole</b>: Your black hole for Internet advertisements<br><a href='/admin'>Did you mean to go to the admin panel?</a></body></html>
+        <link rel='stylesheet' href='./pihole/blockingpage.css' type='text/css'/>
+    </head><body id='splashpage'><img src='./admin/img/logo.svg'/><br/>Pi-<b>hole</b>: Your black hole for Internet advertisements<br><a href='./admin'>Did you mean to go to the admin panel?</a></body></html>
     ";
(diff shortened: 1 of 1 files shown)


Mcat12:
This change will break using the blockage on any path with at least one full segment, eg. `/test/path`

obones:
> 
> 
> This change will break using the block page on any path with at least one full segment, eg. `/test/path`

Could you elaborate on this? I was under the impression that my change would only ensure the displayed content uses relative URLs and would not interfere with anything else.

Mcat12:
If you load `/test/path`, the relative URL of `./admin` will resolve to `/test/admin` which is incorrect.

obones:
> If you load `/test/path`, the relative URL of `./admin` will resolve to `/test/admin` which is incorrect.

Here, if I load /test/path, I get a 404 error which is what I expect to get with an URL that references something that does not exist.
However, I feel that I am missing something here in the Pi-hole context. Why would the web GUI interface be solicited on URLs that are non existent? Maybe this is a newbie question, but I could not find a reason for this to happen in the current (default) setup that I have.

Mcat12:
The block page is the 404 handler. You changed some paths in the block page to be relative. That means when you load the block page, it will use the relative path. The resolved path will be incorrect.

obones:
Ah, I see.
But I use the standard configuration which returns 0:0:0:0 for any blocked request.
This means that the webserver hosting Pi-hole is never asked for a missing item.

That being said, I understand that Pi-hole is meant to be running alone on its server, which is not the way I want to use it here.

Mcat12:
If you use an IP-based blocking mode, you will see the issue.

Mcat12:
When we release the new web interface, the proxy use case will be supported. However, until then we can't support that use case.

PR #2705, opened 2019-04-14
Fix chronometer refresh incorrectly reading the long argument sleep time

### LINKED ISSUE #2703: pihole chronometer --refresh
**In raising this issue, I confirm the following:** `{please fill the checkboxes, e.g: [X]}`

8

---
**Expected behaviour:**

i would like don't see:
sleep: option not valid -- "-"
Try 'sleep --help' for more information.

**Actual behaviour:**

immagine

**Steps to reproduce:**

when i digit from CLI  pihole chronometer --refresh, there's an error on sleep command

`{Alphanumeric token}`

`{Steps of what you have done to fix this}`

Mcat12:
---
Fixes #2703

`-r` worked, but `--refresh` did not. The parsing of the argument was fixed to support both forms.

None

### DIFF
--- a/advanced/Scripts/chronometer.sh
+++ b/advanced/Scripts/chronometer.sh
@@ -463,8 +463,7 @@ chronoFunc() {
         # Get refresh number
         if [[ "$*" == *"-r"* ]]; then
             num="$*"
-            num="${num/*-r /}"
-            num="${num/ */}"
+            num="${num/*-r* /}"
             num_str="Refresh set for every $num seconds"
         else
             num_str=""
--- a/advanced/Scripts/chronometer.sh
+++ b/advanced/Scripts/chronometer.sh
@@ -444,6 +444,9 @@ get_strings() {
 }
 
 chronoFunc() {
+    local extra_arg="$1"
+    local extra_value="$2"
+
     get_init_stats
 
     for (( ; ; )); do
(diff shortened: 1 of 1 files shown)


advanced/Scripts/chronometer.sh
```diff
@@ -463,8 +463,7 @@ chronoFunc() {
         # Get refresh number
         if [[ "$*" == *"-r"* ]]; then
             num="$*"
-            num="${num/*-r /}"
-            num="${num/ */}"
+            num="${num/*-r* /}"
```
PromoFaux: Will this equally allow for something like `--railway`?

Mcat12: I believe so. The argument parsing needs work overall, and should probably be done in a separate PR. Previously it would have also recognized that argument as a refresh argument.

PromoFaux: I'm happy with this as is for a quick fix., will approve!

But I guess we should be doing it more like this? https://github.com/pi-hole/pi-hole/blob/master/pihole#L446

Mcat12: The issue here seems to be that we allow many different flags, and have a hard time extracting the associated value from the arguments. `pihole` only takes in one sub-command, so it has an easier time. It may be that chronometer should also only take in one command, but it's written to allow multiple.

dschaper: I'll approve as well, but I'll go on record that this is a very bad way of handling flags.

Mcat12: It looks like chronometer actually will prevent those other invalid flags eariler:
```
$ pihole -c --railway 1
Unknown option. Please view 'pihole -c --help' for more information
```

I've pushed a commit to make this explicit here, though.

Issue #2704, opened 2019-04-14
default block list item is accessed over http, allows for DoS


sneak:
https://github.com/pi-hole/pi-hole/blob/995ee41d6bc3a405a0402745b24140ca08c148f3/automated%20install/basic-install.sh#L1184

**Expected behaviour:**

All default block lists are fetched securely via TLS.

**Actual behaviour:**

One of the default block lists is fetched over HTTP.  This allows an attacker to MITM the connection and cause your pi-hole to block any domains the attacker wishes to cause you to block, potentially causing a DoS.

**Steps to reproduce:**

Install pi-hole with default settings.

sneak:
Sadly `sysctl.org` does not support TLS connections, otherwise I would have just sent in a PR.

dschaper:
If there is no TLS support for https://sysctl.org I'm not sure how you propose we access it that way?

sneak:
I did not propose that. My suggestion is to *not* download block lists over unauthenticated connections, because otherwise anyone in between you and that host (in common parlance, “someone with a privileged network position”) can alter the contents in transit and cause you to automatically block anything they want, which is a denial of service.

sneak:
Options:

- contact sysctl.org and get them to update their server to 2010 standards

- re-host it somewhere authenticated (with periodic updates) and use that re-hosted url in the pi-hole software (still somewhat vulnerable to the same problem)

- drop that blacklist source entirely

Mcat12:
If you are suggesting to not download blocklists over HTTP, then you are suggesting to remove that block list. Please be clear about what you are trying to accomplish, and the effects of the change.

If you do not want that block list, then remove it. We do not intend on removing it, as it serves its purpose, the average user does not have such strict security requirements, and if they do have such security requirements, they would have just removed the list or never added it when installing.

dschaper:
Feel free to contact them and ask them about TLS hosting.

It is up to the users to decide what blocklists they would like to use and what method the would like to access those lists by. Requiring that all blocklists must have https endpoints is not something I see as being an option.

sneak:
I don’t use pi-hole. This is not a configuration issue. This is a security issue - you are shipping a default configuration that is vulnerable to a Denial of Service. 

Either the unauthenticated download should be removed, or your project should re-host it elsewhere securely. You are putting your users at risk through your vulnerable default configuration.

sneak:
This is absolutely a security bug in your default configuration, btw.

[...]

PR #2706, opened 2019-04-14
Fix for security bug in #2704

sneak:

This is the fix for the security vulnerability documented in issue #2704.

It removes the vulnerable misconfiguration in the defaults.

N/A.

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1172,7 +1172,6 @@ chooseBlocklists() {
     # In an array, show the options available (all off by default):
     options=(StevenBlack "StevenBlack's Unified Hosts List" on
         MalwareDom "MalwareDomains" on
-        Cameleon "Cameleon" on
         ZeusTracker "ZeusTracker" on
         DisconTrack "Disconnect.me Tracking" on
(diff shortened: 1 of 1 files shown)


Mcat12:
If using HTTP is insecure and can cause a DoS, then using unencrypted DNS is even more insecure and more likely to cause a DoS. Hint: DNS is by default insecure.

Pi-hole is not the place you want to try to solve this.

sneak:
The ISP for sysctl.org can break the web for every single pi-hole user who doesn't change the default blocklists (I would estimate this is 99%+ of users).

If you're okay with this, then fine.  But I doubt that's what your users signed up for.

Mcat12:
The owners of any list can "break the web" for all of their users, even those outside of Pi-hole, if they wished. That is why it is up to the user to decide what blocklists they want to use. These "default" lists are not forced upon anyone, they are recommended at install time.

If you feel so strongly about this issue, then contact the list maintainers instead. They are the ones who control access to their lists. Again, Pi-hole is not the place you want to try to solve such global issues as HTTP or DNS insecurity.

Issue #2387, opened 2018-08-22
Pi-hole installer incorrectly detects IP address on multi-homed server with similar subnets


computergeek125:
**In raising this issue, I confirm the following:** `{please fill the checkboxes, e.g: [X]}`

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the the source code relevant to this issue?:**

`4/10`

---
**Expected behaviour:**

Pi-hole installer successfully detects network interface IP address on a multi-homed server

**Actual behaviour:**

The pi-hole installer can misdetect IP addresses when the string of the intended IP address also appears in other IP addresses

**Steps to reproduce:**

 - Install pi-hole on a server with text-overlapping IP addresses
\~\~OR\~\~
 - Run the snippet of installer code below

**Debug token provided by [uploading `pihole -d` log](https://discourse.pi-hole.net/t/the-pihole-command-with-examples/738#debug):**

`-bash: pihole: command not found` :P Happened in install so I don't have one of these yet.

**Troubleshooting undertaken, and/or other relevant information:**

My target server for installation is also my network DHCP server (plus some more) and has VLAN or VPN interfaces for 6 subnets.  3/6 have the same text IP address beginning, as shown below:
```
ip -o -f inet addr show | grep "${IPv4bare}"
4: ens160.2    inet 10.0.1.1/27 brd 10.0.1.31 scope global noprefixroute ens160.2\       valid_lft forever preferred_lft forever
5: ens160.3    inet 10.0.1.187/26 brd 10.0.1.191 scope global noprefixroute ens160.3\       valid_lft forever preferred_lft forever
7: ens160.5    inet 10.0.1.109/27 brd 10.0.1.127 scope global noprefixroute ens160.5\       valid_lft forever preferred_lft forever
```

[...]

PR #2725, opened 2019-04-27
Fix getting IP address in the presence of similar subnets

Mcat12:

---
Prevent getting multiple IP addresses when subnets are similar.

Uses the change suggested by @computergeek125 in #2387

None


dschaper:
So this would grab the first instance instead of the last instance in a multi-address situation? Or match the most specific? So if the user wanted `192.168.1.1` we would select `192.168.1.1/24` instead of `192.168.1.101/24`?

Mcat12:
It should match the most specific, so the one that is exactly `${ipAddr}/`.

PR #2730, opened 2019-04-30
Adds world readable attribute to files created by Pi-hole to circumve…

### LINKED ISSUE #2724: installation and operation of pi-hole not possible with umask 0027

- [X] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

3

I could program a fix, because it is as easy as adding chmod o+r ... to some source files, but I would possibly not find all the places, where this would be needed.

---
**Expected behaviour:**
Installation and operation (e.g. changing setting in web interface and performing list updates etc.) of pi-hole with umask 0027 in /etc/profile works without errors and need of manal file permission changes.

**Actual behaviour:**
When umask is set to something that does not include world readable, e.g. 0027, in /etc/profile, the installation script fails when starting dnsmasq with a permission denied error, because the configuration files under /etc/pihole/* and /etc/dnsmasq.d/01-pihole.conf as well as the web interface files (/var/www/html/{admin, pihole}) created by the installer are not readable by the pihole or www-data user. When changing the file permissions by hand to world readable, another try for the installation works except for the need for some manual (re-)starting of services.

If a list update is performed after the installation with manually corrected file permissions, the list file chache, black.list and gravity.list are deleted and recreated without respecting the file permissions the files had before deletion, so they are in turn not readable by the pihole user again. The installer and updater should make sure, that the file permissions are not changed or set to the correct value independently of the umask settings of the pihole user/root user.

When opening the settings page in the web interface, even if nothing is changed there, the options /etc/pihole/setupVars.conf is also deleted and recreated and then read again, which causes php to show a permission denied error at the bottom of the page, because the file was in turn created without read access for every user.

The manual entries under /usr/local/share/man/man* are also not world readable and the created folders are not owned by root:staff as all other directories are under Raspbian, which causes apt-get to issue warnings about missing file permissions in the update man-db step when installing or removing any program after running the pi-hole installer.

**Steps to reproduce:**
```
sudo su
echo 'umask 0027' >> /etc/profile
curl -sSL https://install.pi-hole.net | bash
```

not relevant?

I changed the file permissions in /etc/pihole* and the other configuration files under /etc created by the installer to world readable and restarted all the necessary services.
This leads pihole to do its job (except for updates) properly, but without the web interface, which shows 403 responses until I changed the file permissions for all files in the web interface to world readable, which also also made the web interface functionable except for the pages, that change configuration settings or do updates.
A simple fix on the user side would be to not mess file systems's umask, but as mentioned in the [Securing Debian Manual](https://www.debian.org/doc/manuals/securing-debian-howto/ch4.en.html#s4.11.12), this could be useful for some people. 

A simple fix for the pi-hole installer and updater would be to explicitly set file permissions on files, that have to be world readable, to world readable in gravity.sh and basic-install.sh.
A way to track all needed file permissions in the Git repositories could be to use a similar approach as the one used in [etckeeper](http://etckeeper.branchable.com/), where a special file with a list of file permissions is kept in the repository.

pvogt09:
…nt #2724

---
This PR should partially fix #2724 by setting most of the files that Pi-hole creates during install or update are made world readable explicitly instead of relying on the umask being set to something that results in world readable files/directories.
- [x] Maybe `/etc/pihole/setupVars.conf` permissions have to be set in a different place, because there is no `touch` or `echo >` for this file.
- [ ] Maybe not all files have the correct access rights.
- [x] Are the permissions and owner of `/usr/local/share/man/man{5, 8}` correct for all distributions?

`chmod a+r ...` is added to make files world readable and files are copied with `-p` option to keep file permissions (especially for backups).

None.

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1898,13 +1898,14 @@ installPihole() {
             # make the Web directory if necessary
             install -d -m 0755 ${webroot}
         fi
-        chmod a+rx /var/www
-        chmod a+rx /var/www/html
 
         if [[ "${INSTALL_WEB_SERVER}" == true ]]; then
             # Set the owner and permissions
             chown ${LIGHTTPD_USER}:${LIGHTTPD_GROUP} ${webroot}
             chmod 0775 ${webroot}
+            # Repair permissions if /var/www/html is not world readable
+            chmod a+rx /var/www
+            chmod a+rx /var/www/html
             # Give pihole access to the Web server group
             usermod -a -G ${LIGHTTPD_GROUP} pihole
             # If the lighttpd command is executable,
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -414,6 +414,8 @@ make_repo() {
     fi
     # Clone the repo and return the return code from this command
     git clone -q --depth 1 "${remoteRepo}" "${directory}" &> /dev/null || return $?
+    # Data in the repositories is public anyway so we can make it readable by everyone (+r to keep executable permission if already set by git)
+    chmod -R a+r "${directory}"
     # Show a colored message showing it's status
     printf "%b  %b %s\\n" "${OVER}" "${TICK}" "${str}"
     # Always return 0? Not sure this is correct
@@ -447,6 +449,8 @@ update_repo() {
     git pull --quiet &> /dev/null || return $?
     # Show a completion message
     printf "%b  %b %s\\n" "${OVER}" "${TICK}" "${str}"
+    # Data in the repositories is public anyway so we can make it readable by everyone (+r to keep executable permission if already set by git)
+    chmod -R a+r "${directory}"
     # Move back into the original directory
     cd "${curdir}" &> /dev/null || return 1
     return 0
@@ -494,6 +498,8 @@ resetRepo() {
     printf "  %b %s..." "${INFO}" "${str}"
     # Use git to remove the local changes
     git reset --hard &> /dev/null || return $?
+    # Data in the repositories is public anyway so we can make it readable by everyone (+r to keep executable permission if already set by git)
+    chmod -R a+r "${directory}"
     # And show the status
     printf "%b  %b %s\\n" "${OVER}" "${TICK}" "${str}"
     # Returning success anyway?
@@ -1241,8 +1247,7 @@ version_check_dnsmasq() {
             printf "%b  %b Backing up dnsmasq.conf to dnsmasq.conf.orig...\\n" "${OVER}"  "${TICK}"
             printf "  %b Restoring default dnsmasq.conf..." "${INFO}"
             # and replace it with the default
-            cp -p ${dnsmasq_original_config} ${dnsmasq_conf}
-            chmod 644 ${dnsmasq_conf}
+            install -D -m 644 -T ${dnsmasq_original_config} ${dnsmasq_conf}
             printf "%b  %b Restoring default dnsmasq.conf...\\n" "${OVER}"  "${TICK}"
         # Otherwise,
         else
@@ -1253,19 +1258,17 @@ version_check_dnsmasq() {
         # If a file cannot be found,
         printf "  %b No dnsmasq.conf found... restoring default dnsmasq.conf..." "${INFO}"
         # restore the default one
-        cp -p ${dnsmasq_original_config} ${dnsmasq_conf}
+        install -D -m 644 -T ${dnsmasq_original_config} ${dnsmasq_conf}
         printf "%b  %b No dnsmasq.conf found... restoring default dnsmasq.conf...\\n" "${OVER}"  "${TICK}"
     fi
 
     printf "  %b Copying 01-pihole.conf to /etc/dnsmasq.d/01-pihole.conf..." "${INFO}"
     # Check to see if dnsmasq directory exists (it may not due to being a fresh install and dnsmasq no longer being a dependency)
     if [[ ! -d "/etc/dnsmasq.d"  ]];then
-        mkdir "/etc/dnsmasq.d"
-        chmod 755 "/etc/dnsmasq.d"
+        install -d -m 755 "/etc/dnsmasq.d"
     fi
     # Copy the new Pi-hole DNS config file into the dnsmasq.d directory
-    cp ${dnsmasq_pihole_01_snippet} ${dnsmasq_pihole_01_location}
-    chmod 644 ${dnsmasq_pihole_01_location}
+    install -D -m 644 -T ${dnsmasq_pihole_01_snippet} ${dnsmasq_pihole_01_location}
     printf "%b  %b Copying 01-pihole.conf to /etc/dnsmasq.d/01-pihole.conf\\n" "${OVER}"  "${TICK}"
     # Replace our placeholder values with the GLOBAL DNS variables that we populated earlier
     # First, swap in the interface to listen on
@@ -1381,19 +1384,15 @@ installConfigs() {
     if [[ "${INSTALL_WEB_SERVER}" == true ]]; then
         # and if the Web server conf directory does not exist,
         if [[ ! -d "/etc/lighttpd" ]]; then
-            # make it
-            mkdir /etc/lighttpd
-            # and set the owners
-            chown "${USER}":root /etc/lighttpd
-            chmod 755 /etc/lighttpd
+            # make it and set the owners
+            install -d -m 755 -o "${USER}" -g root /etc/lighttpd
         # Otherwise, if the config file already exists
         elif [[ -f "/etc/lighttpd/lighttpd.conf" ]]; then
             # back up the original
             mv /etc/lighttpd/lighttpd.conf /etc/lighttpd/lighttpd.conf.orig
         fi
         # and copy in the config file Pi-hole needs
-        cp ${PI_HOLE_LOCAL_REPO}/advanced/${LIGHTTPD_CFG} /etc/lighttpd/lighttpd.conf
-        chmod 644 /etc/lighttpd/lighttpd.conf
+        install -D -m 644 -T ${PI_HOLE_LOCAL_REPO}/advanced/${LIGHTTPD_CFG} /etc/lighttpd/lighttpd.conf
         # Make sure the external.conf file exists, as lighttpd v1.4.50 crashes without it
         touch /etc/lighttpd/external.conf
         chmod 644 /etc/lighttpd/external.conf
@@ -1427,21 +1426,16 @@ install_manpage() {
     fi
     if [[ ! -d "/usr/local/share/man/man8" ]]; then
         # if not present, create man8 directory
-        mkdir /usr/local/share/man/man8
-        chmod 755 /usr/local/share/man/man8
+        install -d -m 755 /usr/local/share/man/man8
     fi
     if [[ ! -d "/usr/local/share/man/man5" ]]; then
         # if not present, create man5 directory
-        mkdir /usr/local/share/man/man5
-        chmod 755 /usr/local/share/man/man5
+        install -d -m 755 /usr/local/share/man/man5
     fi
     # Testing complete, copy the files & update the man db
-    cp ${PI_HOLE_LOCAL_REPO}/manpages/pihole.8 /usr/local/share/man/man8/pihole.8
-    chmod 644 /usr/local/share/man/man8/pihole.8
-    cp ${PI_HOLE_LOCAL_REPO}/manpages/pihole-FTL.8 /usr/local/share/man/man8/pihole-FTL.8
-    chmod 644 /usr/local/share/man/man8/pihole-FTL.8
-    cp ${PI_HOLE_LOCAL_REPO}/manpages/pihole-FTL.conf.5 /usr/local/share/man/man5/pihole-FTL.conf.5
-    chmod 644 /usr/local/share/man/man5/pihole-FTL.conf.5
+    install -D -m 644 -T ${PI_HOLE_LOCAL_REPO}/manpages/pihole.8 /usr/local/share/man/man8/pihole.8
+    install -D -m 644 -T ${PI_HOLE_LOCAL_REPO}/manpages/pihole-FTL.8 /usr/local/share/man/man8/pihole-FTL.8
+    install -D -m 644 -T ${PI_HOLE_LOCAL_REPO}/manpages/pihole-FTL.conf.5 /usr/local/share/man/man5/pihole-FTL.conf.5
     if mandb -q &>/dev/null; then
         # Updated successfully
         printf "%b  %b man pages installed and database updated\\n" "${OVER}" "${TICK}"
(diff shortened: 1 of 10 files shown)


automated install/basic-install.sh
```diff
@@ -1818,6 +1838,7 @@ finalExports() {
     echo "INSTALL_WEB_INTERFACE=${INSTALL_WEB_INTERFACE}"
     echo "LIGHTTPD_ENABLED=${LIGHTTPD_ENABLED}"
     }>> "${setupVars}"
+    chmod 744 "${setupVars}"
```
Mcat12: This file does not need to be executable.

automated install/basic-install.sh
```diff
@@ -1422,15 +1428,26 @@ install_manpage() {
     if [[ ! -d "/usr/local/share/man/man8" ]]; then
         # if not present, create man8 directory
         mkdir /usr/local/share/man/man8
+        chown root:staff /usr/local/share/man/man8
+        chmod a+r /usr/local/share/man/man8
+        chmod a+x /usr/local/share/man/man8
+        chmod g+s /usr/local/share/man/man8
```
Mcat12: Please run one chmod command to set the permissions instead of multiple.

Mcat12: It should be sufficient to use the following permissions for man folders:
```
drwxr-xr-x. 2 root root 4096 Jul 12  2018 man1
```

pvogt09: In `/usr/share/man` I have the same permissions as you, but in `/usr/local/share/man` the setgid bit is also set. Will this cause any trouble, if it is not set by the installer? (I am running Raspbian GNU/Linux 9 (stretch))
```bash
ls -la /usr/local/share/man
drwxrwsr-x 4 root staff 4096 Mär 30 14:42 .
drwxrwsr-x 8 root staff 4096 Nov 13 14:28 ..
drwxr-sr-x 2 root staff 4096 Mär 30 14:42 man5
drwxr-sr-x 2 root staff 4096 Apr  7 16:26 man8
```

```bash
ls -la /usr/share/man
drwxr-xr-x  34 root root  4096 Nov 13 14:22 .
drwxr-xr-x 227 root root 12288 Mär 30 14:37 ..
drwxr-xr-x   2 root root 69632 Apr  7 16:26 man1
drwxr-xr-x   2 root root 20480 Apr  7 16:26 man5
drwxr-xr-x   2 root root 36864 Mär 30 14:37 man8

```

Mcat12: I don't think it would be an issue. Does it work fine on Raspbian?

pvogt09: `man pihole` works in both cases and I did not find any information about the neccessity of setgid or group `staff` in the man pages of `man`, `mandb` or `manpath`, so I think it will be fine to leave it this way. Since `/usr/local/share/man` already has the setgid, the directories `man5` and `man8` should be created with group `staff` anyway, if I understand setgid correctly (I am not really experienced with linux) and there are no further directories inside anyway, aren't they?

Mcat12: Right, it should only contain files.

Mcat12:
Can instances where a file is copied and assigned permissions be replaced with [`install`](http://man7.org/linux/man-pages/man1/install.1.html)?

automated install/basic-install.sh
```diff
@@ -414,6 +414,8 @@ make_repo() {
     fi
     # Clone the repo and return the return code from this command
     git clone -q --depth 1 "${remoteRepo}" "${directory}" &> /dev/null || return $?
+    # Data in the repositories is public anyway so we can make it readable by everyone (+r to keep executable permission if already set by git)
+    chmod -R a+r "${directory}"
```
Mcat12: Is `umask` affecting git clone? If not, then this is not necessary.

pvogt09: I don't know if `umask` is the reason or some configuration setting in `git` (e.g. [core.sharedRepository ](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coresharedRepository), which I have not set), but when I do:
```
pi@pi-hole:~/Downloads $ git clone --depth 1 https://github.com/pi-hole/pi-hole.git pihole
Klone nach 'pihole' ...
remote: Enumerating objects: 81, done.
remote: Counting objects: 100% (81/81), done.
remote: Compressing objects: 100% (73/73), done.
remote: Total 81 (delta 6), reused 30 (delta 0), pack-reused 0
Entpacke Objekte: 100% (81/81), Fertig.
pi@pi-hole:~/Downloads $ cd pihole
pi@pi-hole:~/Downloads/pihole $ ls -la
insgesamt 160
drwxr-x--- 10 pi pi  4096 Mai  4 11:29 .
drwxr-xr-x  4 pi pi  4096 Mai  4 11:29 ..
drwxr-x---  6 pi pi  4096 Mai  4 11:29 advanced
drwxr-x---  2 pi pi  4096 Mai  4 11:29 automated install
-rwxr-x---  1 pi pi    20 Mai  4 11:29 autotest
drwxr-x---  2 pi pi  4096 Mai  4 11:29 block hulu ads
-rw-r-----  1 pi pi  2489 Mai  4 11:29 CONTRIBUTING.md
-rw-r-----  1 pi pi   777 Mai  4 11:29 .editorconfig
drwxr-x---  8 pi pi  4096 Mai  4 11:29 .git
-rw-r-----  1 pi pi   688 Mai  4 11:29 .gitattributes
drwxr-x---  2 pi pi  4096 Mai  4 11:29 .github
-rw-r-----  1 pi pi  1385 Mai  4 11:29 .gitignore
-rwxr-x---  1 pi pi 25299 Mai  4 11:29 gravity.sh
drwxr-x---  3 pi pi  4096 Mai  4 11:29 .idea
-rw-r-----  1 pi pi 14204 Mai  4 11:29 LICENSE
drwxr-x---  2 pi pi  4096 Mai  4 11:29 manpages
-rwxr-x---  1 pi pi 14637 Mai  4 11:29 pihole
-rw-r-----  1 pi pi 17298 Mai  4 11:29 README.md
-rw-r-----  1 pi pi    60 Mai  4 11:29 requirements.txt
-rw-r-----  1 pi pi   107 Mai  4 11:29 setup.py
-rw-r-----  1 pi pi    69 Mai  4 11:29 .stickler.yml
drwxr-x---  2 pi pi  4096 Mai  4 11:29 test
-rw-r-----  1 pi pi   384 Mai  4 11:29 tox.ini
-rw-r-----  1 pi pi   211 Mai  4 11:29 .travis.yml
```
because the [documentation](https://git-scm.com/docs/git-clone) says

> Clones a repository into a newly created directory, creates remote-tracking branches for each branch in the cloned repository (visible using git branch -r), and creates and **checks out** an initial branch that is forked from the cloned repository’s currently active branch.

So I would suppose, that `git clone` does a `git checkout` in the background and then respects the users `umask` if not configured otherwise.

Mcat12: As long as `git status` doesn't show the files as changed, then this is fine.

Mcat12: After a quick test, it looks like Git is fine with this.

Mcat12:
[Line 2261] of the installer should set the permissions `dnsmasq.conf` again, as it may have been moved a few lines up.

There's also an issue where some web interface folders don't give `www-data` execution permission, such as `/var/www` and the folders in `/var/www/html/admin`. This causes a 403 to be returned by lighttpd.

[Line 2261]: https://github.com/pi-hole/pi-hole/blob/2c3a27e9c7e03ac0464808ae0faf346568f135bc/automated%20install/basic-install.sh#L2261

pvogt09:
- [ ] The file [/etc/sysconfig/network/ifcfg-${PIHOLE_INTERFACE}](blob/2c3a27e9c7e03ac0464808ae0faf346568f135bc/automated%20install/basic-install.sh#L869) may also need other permissions, but I can not test which one is needed, because for Raspbian, the function is not called.

According to [this](https://unix.stackexchange.com/a/90599) answer, it would be possible with:
```
tempvar=$webroot;
while [[ "$tempvar" != "/" && "$tempvar" != "." ]]; do
     chmod o+rx "$tempvar";
     tempvar=$(dirname "$tempvar");
done;
```
What happens if the $webroot variable is changed to something (e.g. depending on the outcome of #2692, if anything remains in `/var/www/html`), where world readable and executable in all parent directories is not wanted by the user or doesn't that matter?

Mcat12:
The `ifcfg` file only needs `root:root` 644, it's a networking file used on RHEL distros like Fedora.

You should be able to do
```
chmod a+r /var/www
chmod a+r /var/www/html
```

And then the `admin` folder is taken care of as part of the git checkout. As we don't allow configuring the location of the web interface, this is fine.

Once we replace the PHP web interface, there will be no files necessary for the web interface. It is embedded in the API's binary.

automated install/basic-install.sh
```diff
@@ -1896,6 +1898,8 @@ installPihole() {
             # make the Web directory if necessary
             install -d -m 0755 ${webroot}
         fi
+        chmod a+rx /var/www
+        chmod a+rx /var/www/html
```
Mcat12: This is partially taken care of below, and should anyways be within the `INSTALL_WEB_SERVER` if block.

Mcat12:
Tested this again. The `/var/www/html/admin` folder and subfolders need to be executable by all users for the web server to be able to serve the files.

Mcat12:
Fails to read the adlists file when you open the settings page:
image

pvogt09:
Is `chooseBlocklists` the only place where this file can be created?
Do the files like `ist.0.raw.githubusercontent.com.domains` have to be changed as well and where are they created?

Mcat12:
Here's a directory listing from a fresh install:
```
total 11556
drwxr-xr-x  2 pihole pihole      4096 May 10 02:55 ./
drwxr-xr-x 98 root   root        4096 May 10 02:54 ../
-rw-r--r--  1 root   root         381 May 10 02:54 adlists.list
-rw-r--r--  1 pihole pihole         0 May 10 02:54 dhcp.leases
-rw-r--r--  1 root   root         592 May 10 02:54 dns-servers.conf
-rw-r--r--  1 root   root          18 May 10 02:54 GitHubVersions
-rw-r--r--  1 root   root     2646365 May 10 02:54 gravity.list
-rw-r--r--  1 root   root        1226 May 10 02:54 install.log
-rw-------  1 root   root     1180890 May 10 02:54 list.0.raw.githubusercontent.com.domains
-rw-------  1 root   root      595973 May 10 02:54 list.1.mirror1.malwaredomains.com.domains
-rw-------  1 root   root      638705 May 10 02:54 list.2.sysctl.org.domains
-rw-------  1 root   root        7305 May 10 02:54 list.3.zeustracker.abuse.ch.domains
-rw-------  1 root   root         613 May 10 02:54 list.4.s3.amazonaws.com.domains
-rw-------  1 root   root       43642 May 10 02:54 list.5.s3.amazonaws.com.domains
-rw-------  1 root   root     1772364 May 10 02:54 list.6.hosts-file.net.domains
-rw-r--r--  1 root   root     2646365 May 10 02:54 list.preEventHorizon
-rw-r--r--  1 root   root          25 May 10 02:54 localbranches
-rw-r--r--  1 root   root          56 May 10 02:54 local.list
-rw-r--r--  1 root   root          26 May 10 02:54 localversions
-rw-r--r--  1 root   root         238 May 10 02:54 logrotate
-rw-r-----  1 root   root     2207744 May 10 02:54 macvendor.db
-rw-rw-r--  1 pihole root          15 May 10 02:54 pihole-FTL.conf
-rw-r--r--  1 pihole pihole     24576 May 10 02:55 pihole-FTL.db
-rw-rw-r--  1 pihole www-data       0 May 10 02:54 regex.list
-rw-r--r--  1 root   root         300 May 10 02:54 setupVars.conf
```

`macvender.db` should be updated with world-read permissions and/or made `pihole:pihole` as it's needed by FTL.

Here, the list file (`destination`) should be `touch`-ed and `chmod`-ed:
https://github.com/pi-hole/pi-hole/blob/54ae9d83b0a4ff09c2a40db079df9952cfd122af/gravity.sh#L330

pvogt09:
The `macvendor.db` file is created before the `pihole` user exists:
``E       assert '[✓] Downloading and Installing FTL' in '  [i] Downloading and Installing FTL...transferred... '
E        +  where '  [i] Downloading and Installing FTL...transferred... ' = CommandResult(command='\n    source /opt/pihole/basic-install.sh\n    binary="...input: Inappropriate ioctl for device\nchown: invalid user: 'pihole:pihole'\n").stdout``
In which way should this be resolved? Should the user be created earlier before installing FTL or the file be created later during `installPihole` or the file not be owned by `pihole`?

Mcat12:
The `pihole` user creation should be moved earlier so that the database can have the correct permissions.

gravity.sh
```diff
@@ -391,6 +394,7 @@ gravity_ParseFileIntoDomains() {
 
       # Remove exceptions
       comm -23 "${destination}" <(sort "${destination}.exceptionsFile.tmp") > "${source}"
+	  chmod 644 "${source}"
```
Mcat12: The `>` operator does not modify the file permissions if the file is already created. `source` is already created, so we do not need to reset the permissions.

If this is not the case, then there is one more issue. Please use spaces instead of tabs.

pvogt09:
> 
> 
> The `pihole` user creation should be moved earlier so that the database can have the correct permissions.

What is the correct place for user creation? I would assume [`main`](pvogt09/pi-hole/blob/225285cb977ced4047de542978749c35a3ee43bf/automated%20install/basic-install.sh#L2595), but then, the tests in [`test_automated_install.py`](pvogt09/pi-hole/blob/225285cb977ced4047de542978749c35a3ee43bf/test/test_automated_install.py#L442) will have to be changed to

```python
def test_FTL_detect_armv6l_no_errors(Pihole):
    '''
    confirms only armv6l package is downloaded for FTL engine
    '''
    # mock uname to return armv6l platform
    mock_command('uname', {'-m': ('armv6l', '0')}, Pihole)
    # mock ldd to respond with aarch64 shared library
    mock_command('ldd', {'/bin/ls': ('/lib/ld-linux-armhf.so.3', '0')}, Pihole)
    detectPlatform = Pihole.run('''
    create_pihole_user
    FTLdetect
    ''')
    expected_stdout = info_box + ' FTL Checks...'
    assert expected_stdout in detectPlatform.stdout
    expected_stdout = tick_box + (' Detected ARM-hf architecture '
                                  '(armv6 or lower)')
    assert expected_stdout in detectPlatform.stdout
    expected_stdout = tick_box + ' Downloading and Installing FTL'
    assert expected_stdout in detectPlatform.stdout
```
to pass, because the user still does not exist in this case for the test case.
And possibly the expected output has to take the messages into account, that are created when the `pihole` user is created or checked for (e.g. `Checking for user 'pihole'`), but I do not know how to archieve this.

gravity.sh
```diff
@@ -377,7 +376,6 @@ gravity_ParseFileIntoDomains() {
       if($0 ~ /^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$/) { $0="" }
       if($0) { print $0 }
     }' "${source}" > "${destination}"
-    chmod 644 "${destination}"
```
Mcat12: This line is probably still needed, as the destination could be created as part of this command.

Mcat12:
The user should be added right before the FTL install in `main`, as you linked. You don't need to add the `pihole` user output to the test.

Mcat12:
FTL is failing to start after the initial install:
```
pihole-FTL[10307]: dnsmasq: cannot read /etc/dnsmasq.conf: Permission denied
```

dnsmasq.conf after install with `umask 0027`:
```
-rw-r----- 1 root root 24 May 12 19:25 /etc/dnsmasq.conf
```
After install without setting `umask`:
```
-rw-r--r-- 1 root root 24 May 12 19:26 /etc/dnsmasq.conf
```

Mcat12:
Thanks for sticking with this PR!

pvogt09:
You're welcome. Can I close #2724 or is it still needed?

Mcat12:
I will mark it as fixed.

PR #2734, opened 2019-05-04
Query black and whitelists database

DL6ER:

---

Implement querying of black- and whitelist when using `pihole -q`.

Implement SQL-statement based searching of the corresponding gravity database tables.

None, the code has been written to maintain pre-v5.0 behavior.

### DIFF
--- a/advanced/Scripts/query.sh
+++ b/advanced/Scripts/query.sh
@@ -103,7 +103,7 @@ if [[ -n "${str:-}" ]]; then
 fi
 
 scanDatabaseTable() {
-    local domain table type querystr result
+    local domain table type querystr result table_prev
     domain="${1}"
     table="${2}"
     type="${3:-}"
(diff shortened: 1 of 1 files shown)


advanced/Scripts/query.sh
```diff
@@ -102,29 +102,52 @@ if [[ -n "${str:-}" ]]; then
...
             exit 0
         elif [[ -n "${exact}" ]]; then
-            echo " ${matchType^} found in ${COL_BOLD}${fileName^}${COL_NC}"
+            echo " ${matchType^} found in ${COL_BOLD}${table^}${COL_NC}"
         else
-            # Only print filename title once per file
-            if [[ ! "${fileName}" == "${fileName_prev:-}" ]]; then
-                echo " ${matchType^} found in ${COL_BOLD}${fileName^}${COL_NC}"
-                fileName_prev="${fileName}"
+            # Only print table name once
```
Mcat12: It is no longer the case that this loop handles both the blacklist and whitelist at the same time, so this check can be simplified to just print the "found in ..." text once before the loop starts.

advanced/Scripts/query.sh
```diff
@@ -126,22 +126,20 @@ scanDatabaseTable() {
 
     # Mark domain as having been white-/blacklist matched (global variable)
     wbMatch=true
-    # Loop through each result
+
+    # Print table name
+    if [[ ! -z "${result}" ]]; then
```
Mcat12: This check is not needed as we already checked for `-z "${result}"` earlier and returned if it was true. Also, Stickler doesn't like it.

PR #2774, opened 2019-06-01
Create FUNDING.yml

dschaper:

Meta


Mcat12:
Are we sure that this will not cause the same Untagged issue as before? I know we now clone with more commits in the history, so we should have the tags, but users will get confused that the version now looks like v4.3-1-gasdfasdf.

Mcat12:
@dschaper Can we change the base branch to dev?

dschaper:
Change base to deb

PR #2796, opened 2019-06-13
Use the filtered IPv6 OpenDNS servers

Mcat12:
---
Use the filtered IPv6 OpenDNS servers instead of the unfiltered ones.

Switch out the unfiltered servers for the filtered ones. See https://support.opendns.com/hc/en-us/articles/227986667-Does-OpenDNS-Support-IPv6-

None


dschaper:
Did we previously state whether or not the OpenDNS upstream was to point to the filtered or unfiltered ODNS source? I would assume that people using that upstream are probably choosing that for the filtration ODNS provides, but I don't know if there should be some kind of notification to the users that this is the case. Or if we should have two entries for this upstream?

dschaper:
And do we know if this new upstream is ECS or non-ECS?

Mcat12:
I checked the IPv4 config, and we are using the "main" DNS servers for OpenDNS, which are filtered. Until now, there was a mismatch between the IPv4 and IPv6 servers, in that they acted differently (filtered vs unfiltered). I didn't find any unfiltered IPv4 OpenDNS servers, so I changed the IPv6 ones to be filtered.

I couldn't find any concrete info on if it supports ECS, besides some general info in this FAQ: https://support.opendns.com/hc/en-us/articles/227987647-EDNS-Client-Subnet-FAQ

dschaper:
Great! Thanks Mcat! I don't think that link gives us much information to go on since it's more EDNS than ECS. I think we may be able to check the git blame and see if there is someone we know or can talk to at OpenDNS that can let us know the definitive yay/nay for ECS. Otherwise it may be better to pull that as we don't want users to think they are having their info scrubbed on the queries and it turns out that ODNS doesn't ECS and filter the client subnet.

Mcat12:
It seems that ECS stands for EDNS Client Subnet, which is not confusing at all :wink:.

https://developers.google.com/speed/public-dns/docs/ecs

Mcat12:
That google link also links to the same OpenDNS FAQ, so it seems that ECS is supported by OpenDNS.

DL6ER:
Merging despite CI errors as they are due to a known issue with the Fedora docker container.

Issue #1959, opened 2018-02-09
Blocking page does not render correctly when system has hostname other than 'pi.hole' due to hard-coded paths


dere:
**In raising this issue, I confirm the following:** `{please fill the checkboxes, e.g: [X]}`

- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [x] The issue I am reporting can be *replicated*.
- [x] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the the source code relevant to this issue?:**

6

---
**Expected behaviour:**

When visiting `http://my-systems-hostname/` the "blocking page" should be rendered correctly, reminding you to log-in to `/admin`

**Actual behaviour:**

![image](https://user-images.githubusercontent.com/602444/36051589-a1a2b858-0db8-11e8-9f3d-ae40e70dd3dd.png)

**Steps to reproduce:**

1. Install pihole on a clean system with a hostname other than `pi.hole`

**Debug token provided by [uploading `pihole -d` log](https://discourse.pi-hole.net/t/the-pihole-command-with-examples/738#debug):**

Not needed (explained below)

**Troubleshooting undertaken, and/or other relevant information:**

This issue is caused by a hardcoding of the hostname `pi.hole` on lines 231 and 232, and 234 of file `advanced/index.php`: https://github.com/pi-hole/pi-hole/blob/master/advanced/index.php#L231. This can be fixed quite easily. For example, line 231 should change `href="<?=$proto ?>://pi.hole/admin/img/favicon.png"` to `href="//<?=$serverName ?>/admin/img/favicon.png"`. Note that the URL scheme does not need to be explicitly defined with `<?=$proto ?>://`, simply `//` accomplishes the same thing.

Mcat12:
How do you reach the block page without being able to also resolve `pi.hole`?

dere:
I point my web browser to the IP address or hostname of the machine. I'm not sure why resolving `pi.hole` would be required.

Mcat12:
[...]

PR #2797, opened 2019-06-14
Removed hardcoded host in index.php

cam8001:

---

In my scenario, my pi.hole is hosted remotely in AWS. The host has a couple of IPs - an external (EIP), and local (in an RFC1918 range) - say `10.0.0.53`.

In this case, `pi.hole` resolves to `10.0.0.53`, but I have my pihole web interface set up with a subdomain under which all my network stuff goes - eg `pihole.example.com`. 

This means that for the `/` page served by index.php, CSS/JS/Favicons are broken, as the paths in the HTML in that file specifically reference `pi.hole` rather than `$serverName`. There's also a long timeout while I wait for my browser to give up waiting to download CSS/JS.

Before:
Screen Shot 2019-06-14 at 13 42 34

After:
Screen Shot 2019-06-14 at 13 43 47

Reported previously here: #1959.

I've just removed the hostname entirely. Now the page renders correctly, and will for any hostname - as long as I can see the page, I should be able to see all assets.

None

---

### DIFF
--- a/advanced/index.php
+++ b/advanced/index.php
@@ -222,10 +222,10 @@ setHeader();
   <?=$viewPort ?>
   <meta name="robots" content="noindex,nofollow"/>
   <meta http-equiv="x-dns-prefetch-control" content="off">
-  <link rel="shortcut icon" href="//pi.hole/admin/img/favicon.png" type="image/x-icon"/>
-  <link rel="stylesheet" href="//pi.hole/pihole/blockingpage.css" type="text/css"/>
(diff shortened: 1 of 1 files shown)


Mcat12:
The reason for using the hostname is to avoid the appearance of downloading scripts from a potentially malicious domain.

`pi.hole` resolves to the IP address you told Pi-hole to use, so you should instead run `pihole -r` to change the IP address to the external IP address. Then no code changes are required.

cam8001:
Sounds good to me, thanks for the response!

PR #2803, opened 2019-06-16
Implement groups for lists

DL6ER:

---

Implement groups for whitelist, blacklist, regex filters, and adlists via a new `group` table.

Add a new table, `group`, which links together domains and adlists from multiple tables.
Views are modified to only show entries corresponding to enabled groups.

Groups are a new feature that needs to be documented.


pralor:

https://discourse.pi-hole.net/t/groups-for-whitelist-blacklist-and-can-enable-disable-those-groups-seperately/20793/3

gravity.sh
```diff
@@ -120,7 +120,7 @@ database_table_from_file() {
     do
       # Only add non-empty lines
       if [[ ! -z "${domain}" ]]; then
-        echo "${rowid},\"${domain}\",1,${timestamp},${timestamp},\"Migrated from ${source}\"" >> "${tmpFile}"
+        echo "${rowid},\"${domain}\",1,${timestamp},${timestamp},0,\"Migrated from ${source}\"" >> "${tmpFile}"
```
DL6ER: `0` is the standard group that is added on database creation.

DL6ER:
If enabled, the foreign key constraint protects the groups from being deleted when there are domains referencing it:
```
sqlite> DELETE FROM domain_groups WHERE id = 0;
Error: FOREIGN KEY constraint failed
```

However, as SQLite's default behavior is to *not* enforce foreign keys, we add triggers that automatically recreate the standard group if it gets deleted for whatever reason.

advanced/Templates/gravity.db.sql
```diff
@@ -48,11 +83,12 @@ INSERT INTO info VALUES("version","1");
 
 CREATE VIEW vw_gravity AS SELECT a.domain
     FROM gravity a
-    WHERE a.domain NOT IN (SELECT domain from whitelist WHERE enabled == 1);
+    WHERE a.domain NOT IN (SELECT domain from vw_whitelist);
 
 CREATE VIEW vw_whitelist AS SELECT a.domain
     FROM whitelist a
-    WHERE a.enabled == 1
+    INNER JOIN domain_groups b ON b.id = a.group_id
```
Mcat12: `INNER` is not necessary here, as it is the default.

Mcat12: For readability it would be best to use better aliases for the tables, or just use the full table names.

advanced/Templates/gravity.db.sql
```diff
@@ -23,16 +41,33 @@ CREATE TABLE regex
 	enabled BOOLEAN NOT NULL DEFAULT 1,
 	date_added INTEGER NOT NULL DEFAULT (cast(strftime('%s', 'now') as int)),
 	date_modified INTEGER NOT NULL DEFAULT (cast(strftime('%s', 'now') as int)),
-	comment TEXT
+	group_id INTEGER NOT NULL DEFAULT 0,
+	comment TEXT,
+	FOREIGN KEY (group_id) REFERENCES domain_groups(id)
 );
+
+CREATE TABLE adlist_groups
```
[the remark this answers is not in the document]
DL6ER: I found it counter intuitive to manage adlists and domains with the same table as they are separate topics. We can surely do this, but it might be harder to understand for users. For now, i think this needs more discussion and I will leave both tables.

advanced/Templates/gravity.db.sql
```diff
@@ -1,11 +1,27 @@
...
+(
+	"id" INTEGER PRIMARY KEY AUTOINCREMENT,
+	"enabled" BOOLEAN NOT NULL DEFAULT 1,
+	"description" TEXT
+);
+INSERT INTO domain_groups ("id","description") VALUES (0,'Standard group');
+CREATE TRIGGER domain_groups_standard_group AFTER DELETE ON domain_groups WHEN OLD.id = 0
+    BEGIN
+      INSERT INTO domain_groups ("id","description") VALUES (0,'Standard group');
+    END;
```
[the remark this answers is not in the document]
DL6ER: I rethought the implementation and will remove this altogether. It makes more sense to allow a list to be not managed by any group (it can still be enabled/disabled manually in this case) then to enforce a group association.

advanced/Templates/gravity.db.sql
DL6ER: My SQL book recommends to explicitly write `INNER JOIN`.

> You might see queries where the `INNER JOIN` is written simply as a `JOIN`. These two are equivalent, but we will continue to refer to these joins as inner-joins because they make the query easier to read once you start using other types of joins, which will be introduced in the following lesson.

DL6ER: I will remove the aliases.

DL6ER:
In my last commit, I added linking tables that conveniently allow domains to be included and managed in multiple groups. Although this sounds trivial, this also ensures that domains can stay individual (not included in a group). Due to the latter, I removes the proposed standard group and edited my comments above, accordingly.

DL6ER:
Exemplary proof of concept:
```sql
select * from vw_whitelist;
'whitelisted.de'

update whitelist set enabled = 0 where id = 1;

select * from vw_whitelist;
'<<< Nothing returned >>>'

update whitelist set enabled = 1 where id = 1;

select * from vw_whitelist;
'whitelisted.de'

insert into domain_groups (description) values ("Test group");

select * from domain_groups;
1|1|Test group

insert into whitelist_by_group (whitelist_id, group_id) values (1,1);

select * from vw_whitelist;
'whitelisted.de'

update domain_groups set enabled = 0 where id = 1;

select * from vw_whitelist;
'<<< Nothing returned >>>'

update domain_groups set enabled = 1 where id = 1;

select * from vw_whitelist;
'whitelisted.de'
```

DL6ER:
Just pointing out again that the database changes can be absorbed using
```bash
cd /etc/pihole/ && rm gravity.db; cp migration_backup/* . && pihole -g
```
You will loose the metadata (date added, last modified, etc.)  but you will not loose the domains (given you didn't add new ones only to the database recently).

advanced/Templates/gravity.db.sql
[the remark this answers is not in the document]
Mcat12: Could you see a user wanting to put both domains and adlists in one group? Say there is an adlist that blocks lots of Microsoft tracking related sites, and the user has also added some custom MS tracking sites to their blacklist. They would want to be able to enable/disable that entire "MS tracking" group, which would require us to allow both domains and adlists into one group.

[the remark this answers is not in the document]
DL6ER: I agree this is a possible use case and will unify the two tables.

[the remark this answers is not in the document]
DL6ER: I pushed a new commit that does this and updated my proof-of-concept comment below with the new results.

advanced/Templates/gravity.db.sql
```diff
@@ -81,48 +94,52 @@ CREATE TABLE info
...
-CREATE VIEW vw_whitelist AS SELECT a.domain
-    FROM whitelist a
-    INNER JOIN domain_groups b ON b.id = a.group_id
-    WHERE a.enabled = 1 AND b.enabled = 1
-    ORDER BY a.id;
+CREATE VIEW vw_whitelist AS SELECT domain
+    FROM whitelist
+    LEFT JOIN whitelist_by_group ON whitelist_by_group.whitelist_id = whitelist.id
+    LEFT JOIN domain_groups ON domain_groups.id = whitelist_by_group.group_id
+    WHERE whitelist.enabled = 1 AND domain_groups.enabled IS NULL OR domain_groups.enabled == 1
```
Mcat12: `AND` has precedence over `OR` (`a AND b OR c` is `(a AND b) OR c`), and `domain_groups.enabled` doesn't make much sense at first. This may be better written as:
```SQL
WHERE whitelist.enabled = 1 AND (whitelist_by_group.group_id IS NULL OR domain_groups.enabled = 1)
```

This applies to the other similar expressions as well. Also note that standard SQL uses `=` instead of `==`, even though SQLite allows the latter.

PromoFaux: agreed.

advanced/Templates/gravity.db.sql
```diff
@@ -1,6 +1,6 @@
 PRAGMA FOREIGN_KEYS=ON;
 
-CREATE TABLE domain_groups
+CREATE TABLE groups
```
Mcat12: We have not been following this practice yet, but usually for tables you want to name them in the singular, ex. "group", "adlist", etc:
https://stackoverflow.com/a/5841297

advanced/Templates/gravity.db.sql
PromoFaux: `WHERE` clause should be:

```suggestion
    WHERE whitelist.enabled = 1 AND domain_groups.enabled != 0
```

advanced/Templates/gravity.db.sql
DL6ER: Also, a table with name `group` is not allowed. We will need to use `groups` or something different as `GROUP BY ...` is an SQL instruction.

Mcat12: You can wrap it in double quotes. The same would happen if you added a `user` table, which is very common.

advanced/Templates/gravity.db.sql
Mcat12: Note we found that the above has an issue because you cannot compare a value with NULL: https://stackoverflow.com/questions/5658457/not-equal-operator-on-null

DL6ER:
Marking as [WIP] as some more testing is needed for the case where multiple groups manage the same domain(s).

advanced/Templates/gravity.db.sql
DL6ER: This does not work as `(NULL != 0)` is always `false` in SQL.

advanced/Templates/gravity.db.sql
```diff
@@ -1,3 +1,12 @@
+PRAGMA FOREIGN_KEYS=ON;
+
+CREATE TABLE "group"
+(
+	id INTEGER PRIMARY KEY AUTOINCREMENT,
+	enabled BOOLEAN NOT NULL DEFAULT 1,
+	description TEXT
```
Mcat12: Does it make sense to have both a required `name` and optional `description` field? Right now, it is possible to create a group without any useful identifier other than the ID.

advanced/Templates/gravity.db.sql
```diff
@@ -25,6 +50,14 @@ CREATE TABLE regex
...
 );
+
+CREATE TABLE regex_by_group
+(
+	regex_id INTEGER NOT NULL REFERENCES regex (id),
+	group_id INTEGER NOT NULL REFERENCES "group" (id),
+	PRIMARY KEY (regex_id, group_id)
+);
+
 CREATE TABLE adlists
```
Mcat12: To keep things consistent, can we rename this table to `adlist`?

advanced/Templates/gravity.db.sql
```diff
@@ -46,44 +87,52 @@ CREATE TABLE info
...
 
-CREATE VIEW vw_whitelist AS SELECT a.domain
-    FROM whitelist a
-    WHERE a.enabled == 1
-    ORDER BY a.id;
+CREATE VIEW vw_whitelist AS SELECT domain
+    FROM whitelist
+    LEFT JOIN whitelist_by_group ON whitelist_by_group.whitelist_id = whitelist.id
+    LEFT JOIN "group" ON "group".id = whitelist_by_group.group_id
+    WHERE whitelist.enabled = 1 AND ("group".enabled IS NULL OR "group".enabled = 1)
```
Mcat12: Instead of checking if `"group".enabled` is null, it would be clearer to check if `whitelist_by_group.group_id` is null.

advanced/Templates/gravity.db.sql
```diff
@@ -1,3 +1,12 @@
+PRAGMA FOREIGN_KEYS=ON;
+
+CREATE TABLE "group"
+(
+	id INTEGER PRIMARY KEY AUTOINCREMENT,
+	enabled BOOLEAN NOT NULL DEFAULT 1,
```
Mcat12: We can use `TRUE` and `FALSE` instead of integers for booleans.

advanced/Templates/gravity.db.sql
```diff
@@ -46,44 +87,52 @@ CREATE TABLE info
...
-    WHERE a.domain NOT IN (SELECT domain from whitelist WHERE enabled == 1);
+CREATE VIEW vw_gravity AS SELECT domain
+    FROM gravity
+    WHERE domain NOT IN (SELECT domain from vw_whitelist);
 
-CREATE VIEW vw_whitelist AS SELECT a.domain
-    FROM whitelist a
-    WHERE a.enabled == 1
-    ORDER BY a.id;
+CREATE VIEW vw_whitelist AS SELECT domain
```
Mcat12: I've tested the case where a domain appears in multiple groups, and with this code it will show up twice in the view. To fix that, we need to add `DISTINCT` after `SELECT` here.

advanced/Templates/gravity.db.sql
DL6ER: > SQLite does not have a separate Boolean storage class. Instead, Boolean values are stored as integers 0 (false) and 1 (true).

https://www.sqlite.org/datatype3.html

I am experiencing issues with using `TRUE` and `FALSE` on older versions of `sqlite3`. While replacing `1` with `TRUE` works with version 3.26.0, it does **not work** with version 3.16.2 (Jan 2017) as shipped, e.g., by recent Armbian Linux.

advanced/Templates/gravity.db.sql
DL6ER: Confirmed:
```sql
insert into "group" (name) values ('Group 1');                                                                                                                                                                        
insert into "group" (name) values ('Group 2');                                                                                                                                                                        
insert into whitelist_by_group (whitelist_id,group_id) values (1,1);                                                                                                                                                  
insert into whitelist_by_group (whitelist_id,group_id) values (1,2);                                                                                                                                              

select * from vw_whitelist;
'whitelisted.de'
```

advanced/Templates/gravity.db.sql
DL6ER: Update: I did some further research on this. SQLite supports true/false literals only since [3.23.0](https://www.sqlite.org/releaselog/3_23_0.html):

> Recognize `TRUE` and `FALSE` as constants. (For compatibility, if there exist columns named "true" or "false", then the identifiers refer to the columns rather than Boolean constants.)

DL6ER: This is even more severe as I though initially, when we changed
```sql
enabled BOOLEAN NOT NULL DEFAULT 1,
```
to
```sql
enabled BOOLEAN NOT NULL DEFAULT TRUE,
```
this actually changed the type to `TEXT` and stored the string causing unexpected behavior: As `FALSE` is **not** an alias for `0` in older versions, it is actually `TRUE`.

Mcat12:
The web interface will have to be updated slightly:
image

Mcat12:
Also, `CustomizeAdLists` in `webpage.sh` needs to be updated with the new adlist table name.

Mcat12:
As well as `query.sh`. I'll stop looking now, there may be other places.

PR #2832, opened 2019-07-05
Migrate audit list to gravity.db database table

DL6ER:

---

Migrate the list of audited domains into the gravity.db database

Well, see above.

The database documentation needs to be extended. This will be done in a follow-up PR.


advanced/Scripts/database_migration/gravity-db.sh
```diff
@@ -0,0 +1,21 @@
...
+# Please see LICENSE file for your rights under this license.
+
+upgrade_gravityDB(){
+	local version=$(sqlite3 "$1" "SELECT "value" FROM "info" WHERE "property" = 'version';")
+
+	case "$version" in
+	1)
+		sqlite3 "$1" < "/etc/.pihole/advanced/Scripts/database_migration/gravity/1_to_2.sql"
+		;;
+	esac
```
Mcat12: We should use a set of if statements like in FTL so we can update multiple versions in one go. e.g.:
```
if version == 1 {
    upgrade_to_2
    version = 2
}

if version == 2 {
    upgrade_to_3
    version = 3
}

etc
```

advanced/Scripts/database_migration/gravity/1_to_2.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE auditlist
```
Mcat12: A clearer name would be `domain_audit` or similar. In the future if we add some other kind of audit table, this name would be confusing.

advanced/Scripts/webpage.sh
```diff
@@ -546,17 +546,20 @@ addAudit()
...
+    shift # skip first domain, as it has already been added
+    for domain in "$@"
     do
-        echo "${var}" >> /etc/pihole/auditlog.list
+      # Insert only the domain here. The date_added field will be
+      # filled with its default value (date_added = current timestamp)
+      domains="${domains},('${domain}')"
     done
-    chmod 644 /etc/pihole/auditlog.list
+    sqlite3 "${gravityDBfile}" "INSERT INTO \"auditlist\" (domain) VALUES ${domains};"
```
Mcat12: This should protect against SQL injections

advanced/Templates/gravity.db.sql
```diff
@@ -80,13 +80,21 @@ CREATE TABLE gravity
 (
 	domain TEXT PRIMARY KEY
 );
+
+CREATE TABLE auditlist
```
Mcat12: Don't add this to this file, create it via the migration script.

gravity.sh
```diff
@@ -17,6 +17,8 @@ coltable="/opt/pihole/COL_TABLE"
 source "${coltable}"
 regexconverter="/opt/pihole/wildcard_regex_converter.sh"
 source "${regexconverter}"
+readonly databaseMigrationScript="/etc/.pihole/advanced/Scripts/database_migration/gravity-db.sh"
+source "${databaseMigrationScript}"
```
Mcat12: This variable can be inlined

gravity.sh
```diff
@@ -116,14 +119,27 @@ database_table_from_file() {
...
     # Read file line by line
-    grep -v '^ *#' < "${source}" | while IFS= read -r domain
-    do
-      # Only add non-empty lines
-      if [[ ! -z "${domain}" ]]; then
-        echo "${rowid},\"${domain}\",1,${timestamp},${timestamp},\"Migrated from ${source}\"" >> "${tmpFile}"
-        rowid+=1
-      fi
-    done
+    if [[ "${table}" == "auditlist" ]]; then
```
Mcat12: This if statement can be moved closer to where the difference is in the two branches (when writing to `tmpFile`).

gravity.sh
```diff
@@ -150,34 +166,40 @@ database_table_from_file() {
...
+    echo -e "  ${INFO} Creating new gravity database"
+    generate_gravity_database
   fi
 
-  echo -e "  ${INFO} Creating new gravity database"
-  generate_gravity_database
+  # Check if gravity database needs to be updated
+  upgrade_gravityDB "${gravityDBfile}"
 
   # Migrate list files to new database
```
Mcat12: These migrations should only be called once. For this audit log migration, the code to migrate the audit log should be contained as part of that specific migration.

advanced/Templates/gravity.db.sql
DL6ER: Why? When we can directly create the correct database, why should we first create an outdated version of the database to then bump it up? I know that we do this in FTL, but this is more because we have a much better error handling in there.

gravity.sh
DL6ER: I did this for consistency with the other lines, we can surely inline it.

gravity.sh
DL6ER: They will check against the version and immediately exit. It seems to be the best solution. When we would make it dependent on the existence of `audtlog.list`, then the migrations would never be triggered for most users.

gravity.sh
DL6ER: My idea was to wrap the `if` around the `do` loop to have it only evaluated once. Otherwise, the same `if` might need to be evaluated hundreds of times if it is checked withing the `do` loop.

advanced/Templates/gravity.db.sql
Mcat12: That is one of the main points of versioning the database scripts. Once you write the script to create the database at version 1, you don't modify it. If you want to modify the database, you write a new script and upgrade to version 2.

If you modified the first script whenever you made a database update, then you would be duplicating code. Also, there are now more than one possible ways that the database could have gotten to version X, which increases the possibility that there is an error in one of the many ways they could have upgraded.

gravity.sh
Mcat12: As long as it isn't used anywhere and inlining doesn't cause any other issues, go ahead with it.

gravity.sh
Mcat12: That is fine, it is worth not duplicating the code. It is a very simple check and premature optimization should generally be avoided.

gravity.sh
Mcat12: Using the example of the audit list, the only case when we want to run the migration is when we are upgrading the database from 1 to 2. Therefore, we should only run the migration right after we upgrade the database to version 2. If we always run the migration (when the file is there), then we are introducing a lot of uncertainty into the upgrade process. Instead of being in one place, the migration happens due to a combination of separate events.

advanced/Templates/gravity.db.sql
DL6ER: Okay, if you think this is the better way than we can do it like this. Just for the records, I still express my doubts about intentionally having a script that generates an outdated version of a database and then use migration script to jump up one by one to the most recent version.

advanced/Scripts/database_migration/gravity/1_to_2.sql
```diff
@@ -1,4 +1,4 @@
-CREATE TABLE auditlist
+CREATE TABLE domain_auditlist
```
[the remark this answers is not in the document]
DL6ER: This is for consistency as well. The other lists are also called `whitelist`, `blacklist`, ... and not `white`, `black`, ... although they serve a similar purpose.

gravity.sh
DL6ER: This produces a lint error now. I guess that's why it's done like this for the other lines as well.

> Not following: /etc/.pihole/advanced/Scripts/database_migration/gravity-db.sh was not specified as input (see shellcheck -x).

DL6ER: I added the corresponding linting instruction that it is okay to not follow this file. This solves the problem. As the exception needs to be added individually for each `source` call, this does not interfere with the checking of the remaining source code in this file.

advanced/Scripts/database_migration/gravity/1_to_2.sql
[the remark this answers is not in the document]
Mcat12: Technically, every table is a "list" of things. We used the terms whitelist and blacklist because that's what those "lists" are conventionally called, but we didn't use "list" for the regex or gravity tables. Also, the whitelist and blacklist are directly viewed by the user, whereas the audited domains are just a way for us to keep track of what domains the user has already audited. The user does not know about any "audit list", just that when they audit a domain they don't see it show up in that view again.

It's not a huge deal either way and won't block the merge, but it is confusing to use the "list" ending for two different concepts.

advanced/Scripts/webpage.sh
```diff
@@ -542,21 +542,38 @@ Teleporter() {
...
+}
+
 addAudit()
 {
     shift # skip "-a"
     shift # skip "audit"
-    for var in "$@"
+    local domains validDomain
+    domains="('$(checkDomain "${1}")')"
+    shift # skip first domain, as it has already been added
```
Mcat12: This assumes that the first domain is valid (not empty after sanitizing).

advanced/Scripts/webpage.sh
```diff
@@ -557,23 +557,27 @@ addAudit()
...
       # filled with its default value (date_added = current timestamp)
       validDomain="$(checkDomain "${domain}")"
       if [[ -n "${validDomain}" ]]; then
-        domains="${domains},('${domain}')"
+        # Put comma in between () when there is
+        # more than one domains to be added
+        if [[ -n "${domains}" ]]; then
+          domains="${domains},"
+        fi
+        domains="${domains}('${domain}')"
```
Mcat12: Should this use `validDomain`?

advanced/Scripts/webpage.sh
```diff
@@ -557,23 +557,27 @@ addAudit()
...
-    shift # skip first domain, as it has already been added
+    domains=""
     for domain in "$@"
     do
       # Insert only the domain here. The date_added field will be
       # filled with its default value (date_added = current timestamp)
       validDomain="$(checkDomain "${domain}")"
       if [[ -n "${validDomain}" ]]; then
-        domains="${domains},('${domain}')"
+        # Put comma in between () when there is
```
Mcat12: What does `()` represent?

advanced/Scripts/webpage.sh
```diff
@@ -557,23 +557,27 @@ addAudit()
...
+        # more than one domains to be added
+        if [[ -n "${domains}" ]]; then
+          domains="${domains},"
+        fi
+        domains="${domains}('${domain}')"
       fi
     done
-    sqlite3 "${gravityDBfile}" "INSERT INTO \"domain_auditlist\" (domain) VALUES ${domains};"
+    sqlite3 "${gravityDBfile}" "INSERT INTO \"domain_audit\" (domain) VALUES ${domains};"
```
Mcat12: The quotes around the table name are not needed.

advanced/Scripts/webpage.sh
DL6ER: I though about this but it seems we should use `${domain}` here to match the user input. `${validDomain}` has been converted to lower-case to ease testing. This is something we can discuss, however.

advanced/Scripts/webpage.sh
DL6ER: It forms a single set of variables. With multiple sets of variables, we can add multiple rows at once.

Adding one row using SQL insertion (as usual):
```sql
INSERT INTO table (domain) VALUES ('abc.de');
```

Adding multiple rows with one insertion:
```sql
INSERT INTO table (domain) VALUES ('abc.de'),('fgh.ij'),('klm.no'),('pqr.st');
```

advanced/Scripts/webpage.sh
DL6ER: But does this mean we should omit them? Now that we need them for `"group"`, I figured it would be better if we keep adding them everywhere.

advanced/Scripts/webpage.sh
Mcat12: Ah, the comment was confusing (sounded like it said `(,)`). Maybe instead say "Put comma in between domains"?

advanced/Scripts/webpage.sh
Mcat12: There is a lot of desugaring that could happen, but we do not need to be more verbose than we need to be. As most of the tables do not need the quotes, we don't need to add them unless they are necessary.

advanced/Scripts/database_migration/gravity-db.sh
```diff
@@ -0,0 +1,32 @@
...
+#
+# Updates gravity.db database
+#
+# This file is copyright under the latest version of the EUPL.
+# Please see LICENSE file for your rights under this license.
+
+upgrade_gravityDB(){
+	local database auditFile version
+	database="${1}"
+	auditFile="${2}"
```
Mcat12: This function should be generic enough for all migrations we do. The audit file location should therefore be in this file, not passed in from gravity.

advanced/Scripts/webpage.sh
```diff
@@ -542,21 +542,44 @@ Teleporter() {
...
     shift # skip "-a"
     shift # skip "audit"
-    for var in "$@"
+    local domains validDomain
+    domains=""
+    for domain in "$@"
     do
-        echo "${var}" >> /etc/pihole/auditlog.list
+      # Insert only the domain here. The date_added field will be
+      # filled with its default value (date_added = current timestamp)
```
Mcat12: This comment should be moved to right above the SQL command.

gravity.sh
```diff
@@ -112,15 +115,21 @@ database_table_from_file() {
...
     declare -i rowid
     rowid=1
-    # Read file line by line
     grep -v '^ *#' < "${source}" | while IFS= read -r domain
     do
       # Only add non-empty lines
-      if [[ ! -z "${domain}" ]]; then
-        echo "${rowid},\"${domain}\",1,${timestamp},${timestamp},\"Migrated from ${source}\"" >> "${tmpFile}"
+      if [[ -n "${domain}" ]]; then
+        if [[ "${table}" == "auditlist" ]]; then
```
Mcat12: The table is now named `domain_audit`

Mcat12:
We should add a timeout to the SQLite commands we run.

```
Error: near line 1: database is locked
Error: near line 8: database is locked
  [i] Migrating content of /etc/pihole/auditlog.list into new database

  [✗] Unable to truncate domain_audit database /etc/pihole/gravity.db
  Error: near line 1: no such table: domain_audit
```

PR #2905, opened 2019-09-01
README.md: added LTT video to coverage

eranb:


dschaper:
Thanks for the submission. Can you open this on the pi-hole/docs repository instead? That is where we are adding new content as the README page here will be removing many sections.

eranb:
Sure np.

PR #2912, opened 2019-09-03
Remove adblock list style support

PromoFaux:

---

Removes support for lists such as easylist / easyprivacy.

We may revisit support in the future, but for now we appear to just be indiscriminently pulling _all_ domains from these lists, causing things to be blocked that should not be. The intention of adding this support may have been well placed, but possibly not as thought through as it could have been

That said, none of these lists are in our defaults, so it only affects users that have manually added these lists to their installs.

See below output for an `adlist.list` only containing easylist/privacy

image

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -353,7 +353,7 @@ gravity_ParseFileIntoDomains() {
   # Determine how to parse individual source file formats
   if [[ "${firstLine,,}" =~ (adblock|ublock|^!) ]]; then
     # Compare $firstLine against lower case words found in Adblock lists
-    echo -ne "  ${CROSS} Format: Adblock (list type not supported)"
+    echo -e "  ${CROSS} Format: Adblock (list type not supported)"
   elif grep -q "^address=/" "${source}" &> /dev/null; then
     # Parse Dnsmasq format lists
     echo -e "  ${CROSS} Format: Dnsmasq (list type not supported)"
--- a/gravity.sh
+++ b/gravity.sh
@@ -353,46 +353,7 @@ gravity_ParseFileIntoDomains() {
   # Determine how to parse individual source file formats
   if [[ "${firstLine,,}" =~ (adblock|ublock|^!) ]]; then
     # Compare $firstLine against lower case words found in Adblock lists
-    echo -ne "  ${INFO} Format: Adblock"
-
-    # Define symbols used as comments: [!
-    # "||.*^" includes the "Example 2" domains we can extract
-    # https://adblockplus.org/filter-cheatsheet
-    abpFilter="/^(\\[|!)|^(\\|\\|.*\\^)/"
-
-    # Parse Adblock lists by extracting "Example 2" domains
-    # Logic: Ignore lines which do not include comments or domain name anchor
-    awk ''"${abpFilter}"' {
-      # Remove valid adblock type options
-      gsub(/\$?~?(important|third-party|popup|subdocument|websocket),?/, "", $0)
-      # Remove starting domain name anchor "||" and ending seperator "^"
-      gsub(/^(\|\|)|(\^)/, "", $0)
-      # Remove invalid characters (*/,=$)
-      if($0 ~ /[*\/,=\$]/) { $0="" }
-      # Remove lines which are only IPv4 addresses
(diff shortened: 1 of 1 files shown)


gravity.sh
```diff
@@ -353,46 +353,7 @@ gravity_ParseFileIntoDomains() {
...
-        if($3) { print $3 }
-      }' "${source}" > "${destination}.exceptionsFile.tmp"
-
-      # Remove exceptions
-      comm -23 "${destination}" <(sort "${destination}.exceptionsFile.tmp") > "${source}"
-      mv "${source}" "${destination}"
-    fi
-
-    echo -e "${OVER}  ${TICK} Format: Adblock"
+    echo -ne "  ${CROSS} Format: Adblock (list type not supported)"
```
dschaper: I think this is supposed to be just an `echo -e`? The output from the screencap doesn't have `(list type not supported)` and it should.

PromoFaux: Two things... 

1) correct! Should be a `-e`
2) I took the screen shot before I added the "(list type not supported)"

Just making sure you're awake.

pralor:

https://discourse.pi-hole.net/t/list-type-not-suppoted-after-upgrade/23827/2

pralor:

https://discourse.pi-hole.net/t/pi-hole-4-3-2-release-notes/23852/1

Malakay666:
this sucks, because of this I had to install back Adblock plugin in my browser :( I was using Pi-hole because there was no necessity to install adblock to all my devices, now I have to install adblock everywhere again. I haven´t ever had problem with displaying any pages with adblock lists -_-

PromoFaux:
The thing is, they're not supposed to be used with DNS style blockers, which is why we removed them. They will not have been doing the same job in Pi-hole as they do in uBlock/Adblock

Trust me, it's better that they're not automatically parsed by Pi-hole.

There are pre-parsed versions of the Easylist and EasyPrivacy lists on https://firebog.net/, but personally I would recommend against using those two lists.

Nowt wrong with running Pi-hole alongside another blocker. I use uBlock Origin as well :)

ryanbr:
Not unexpected, because every url block is considered a first-party. would cause many false positves. Its just the limitations of hosts file format.

foresthus:
Hello, in my opinion the adjustment should be done by the user's decision. The function seems to be desired. Then it would be an easy to install a query with the update "pihole -g" whether the lists are still desired or not. I do not want to install any ad blocking software on any of my clients. This should continue to do pihole. The product is very good and should not be limited in its functionality.

Malakay666:
foresthus : definitely agreed. I just want to use only pihole.not separately install additional adblocker on every client. I have found just one website which pi-hole unwantedly blocks with adblock blocklist,but when I need to visit, I just only stop pi-hole for 10 minutes to work with that web, but that is no big deal, just little minor thing. I can always stop pihole for few minutes.

technicalpyro:
@foresthus  @Malakay666 the point is moot the change has been made.  The reasoning is sound despite the points you make.  

if you would like to continue to use adblock style lists there are ways to parse the data yourself 

Have a nice day.

PromoFaux:
As a final (reiterated) word on this:

Removing the support for this style of list has not made Pi-hole any less capable of blocking ads. 

Pi-hole still maintains the same core functionality as it did before this change, the only difference is we no longer attempt to parse a style of list that was not designed to be used in a DNS blocker like Pi-hole. 

You can still use the now unsupported style of list but not in their original format . You will first have to either:
 - Parse the data yourself, weeding out any false positives
 - Use a pre-parsed list from somewhere like firebog.net, where a list maintainer has put in that work for you

I personally don't understand the argument that this change means people will have to install other ad blockers on their devices. That is no more now than it was before. Just because you use the lists designed for another ad blocker, does not magically mean that you no longer have to use the other ad blocker. I myself maintain a layered approach, i.e using uBlock on my desktop/laptops. Pi-hole then covers areas that are not covered by the browser based blocker. I.e my TV, games consoles, and lightbulbs.

✌️

pralor:

https://discourse.pi-hole.net/t/pihole-does-not-seem-to-be-blocking-ads-after-update-so-ive-downgraded-but-still-having-issues/28689/16

pralor:

https://discourse.pi-hole.net/t/pihole-does-not-seem-to-be-blocking-ads-after-update-so-ive-downgraded-but-still-having-issues/28689/17

PR #2915, opened 2019-09-07
Prevent Firefox from automatically switching over to DNS-over-HTTPS

DL6ER:

---

Prevent Firefox from using DNS-over-HTTPS (DoH) as this would corrupt our efforts to block ads.

We follow the [official Mozilla Support platform](https://support.mozilla.org/en-US/kb/configuring-networks-disable-dns-over-https):
> Network administrators may configure their networks as follows to signal that their local DNS resolver implemented special features that make the network unsuitable for DoH:
> 
> DNS queries for the A and AAAA records for the domain “[use-application-dns.net](http://use-application-dns.net)” must respond with NXDOMAIN rather than the IP address retrieved from the authoritative nameserver.

We we implement such a "special feature", we implement this.

Probably not necessary.

### DIFF
--- a/advanced/01-pihole.conf
+++ b/advanced/01-pihole.conf
@@ -41,3 +41,8 @@ log-facility=/var/log/pihole.log
 local-ttl=2
 
 log-async
+
+# Signal to Firefox that the local network is unsuitable for DNS-over-HTTPS
+# This follows https://support.mozilla.org/en-US/kb/configuring-networks-disable-dns-over-https
+# (sourced 7th September 2019)
+server=/use-application-dns.net/


pralor:

https://discourse.pi-hole.net/t/support-for-returning-nxdomain-for-use-application-dns-net-to-disable-firefox-doh/23243/7

josteink:
This sounds like a no-brainer. As a pi-hole user I’d love to have this working out of the box without a need to manually configure anything.

DL6ER:
@josteink Yes, this pull requests implements it as an always-on feature. No need to flip any switch for it.

advanced/01-pihole.conf
```diff
@@ -41,3 +41,8 @@ log-facility=/var/log/pihole.log
 local-ttl=2
 
 log-async
+
+# Signal to Firefox that the local network is unsuitable for DNS-over-HTTPS
+# This follows https://support.mozilla.org/en-US/kb/configuring-networks-disable-dns-over-https
+# (sourced 7th September 2019)
+server=/use-application-dns.net/
```
Mcat12: This will be deleted whenever the configuration is re-processed: https://github.com/pi-hole/pi-hole/blob/1d43c0a2fa36ddc993db51b1751e26cbc8123fbf/advanced/Scripts/webpage.sh#L141

t0m5k1:
How about Pi-Hole integrate stunnel to provide a way for users/admins to query pihole via DoH?
If you don't want to use stunnel you could move from lighttpd to nginx and then use the streams feature to provide DoH access to pihole.

This way users can just add an address to Firefox and have the Firefox query pihole directly rather than implementing workarounds to mitigate DoH/DoT.

dschaper:
If you want to set up a PR with the above changes outlined then we'll take a look and see if it's something we can work.

t0m5k1:
I'm not a coder in any way, The above was just a suggestion.

DL6ER:
@t0m5k1 Thanks for your suggestion. One thing isn't clear to me, though. What is the benefit of having DoH in your *local* network? It is a massive overhead for a very simple protocol and protecting yourself against man-in-the-middle attacks shouldn't be a thing in your own home network, right?

Even then, DNSSEC is being deployed more and more and protects you as well.
DNSSEC coverage as of yesterday
(https://elists.isoc.org/pipermail/dnssec-maps/)

I don't think the encryption aspect should be necessary in home networks. If you are that concerned (maybe because of your hacker son/daughter), then them just pulling out and messing with the Raspberry's SD card sounds like a much more likely scenario compared to packet interceptions and live rewrites thereof.

t0m5k1:
I know it is an overhead and might seem heavy but with more browsers jumping on the DoH bandwagon it would be good to embrace this rather than just force a user/browser to disable it.

Google have stated that v78 of Chrome will also have DoH add to that the amount of people choosing to run pi-hole in a cloud due to poisoning of DNS by their ISP the more features present in pihole will mean a wider use area.

Additionally with more and more people no longer trusting the LAN I think this feature would be good. I have a few friends who are now using a NAC (Network Access Controller) within their own LAN as they no longer trust it purely due friends arriving on site with phones that have wifi enabled and it connecting.

DL6ER:
You made some point but nothing convinced me/(us) yet.

> run pi-hole in a cloud due to poisoning of DNS by their ISP

I see this but how would an Out-of-The-Box DoH feature change this? Even with Doh it is not safe to open your DNS resolver to the world. DNS amplification attacks might be reduced, however, this also involves some more fine tuned firewall settings to be sure to avoid UDP leaks aside your HTTPS channel.

> with more browsers jumping on the DoH bandwagon it would be good to embrace this rather than just force a user/browser to disable it.

I still don't see why. It adds a lot of overhead with no real gain.

> more and more people no longer trusting the LAN [...] they no longer trust it purely due friends arriving on site with phones that have wifi enabled and it connecting

I see this issue being purely on the network side. If you don't trust the devices of your friends/visitors/whomever (BTW, I can perfectly see why you wouldn't), then you should have a dedicated isolated WiFi for them. Many routers even offer "guest network"s out of the box which is exactly what you should be looking into.

Discussing this in comments on an already implemented pull request on Github seems to be the wrong place as this is not directly related to the code in this PR but rather a request for a new feature. Please open a feature request on our [Discourse Forum](https://discourse.pi-hole.net), as mentioned in our [Issue Templates](https://github.com/pi-hole/pi-hole/blob/master/.github/ISSUE_TEMPLATE.md). This also ensures that a much wider audience will see your request and can give their opinions. After all, my voice is only one out of many and I can surely be outvoted by others.

Thank you for your understanding.

t0m5k1:
> You made some point but nothing convinced me/(us) yet.
> 
Was just a suggestion, and people are already asking on reddit how DoH is going to impact pihole. Knowing Google they will not make it easy disable DoH in Chrome unlike FireFox.

AndreiG6:
@DL6ER your comments are close-
minded and unrealistic to say the least. There's no point in opposing what's to come. Aside from the slight overhead, local DNS poisoning attacks have been "a thing" for decades. You should embrace new security protocols versus stand against them. Period.

AndreiG6:
I long replaced lighttpd/cgi with nginx/fpm (way faster and less resources) and added DoT + DoH support directly in nginx. This helps me secure my kids mobile browsing while they're outside out of my local network. There are also more and more routers which support DNS filtering through these services.

DL6ER:
@AndreiG6 please continue the discussion on Discourse as I mentioned above. If you insist on using GitHub, we would be happy discussing this further based on your PR for this feature.

pralor:

https://discourse.pi-hole.net/t/was-bedeutet-dns-over-https-fur-pi-hole/23762/2

pralor:

https://discourse.pi-hole.net/t/good-news-cloudflared-came-over/27119/12

PR #2935, opened 2019-09-18
Per-client blocking changes

DL6ER:

---

Implement back-end changes for https://github.com/pi-hole/FTL/pull/642

- Update database from 3 to 4 (add client-specific group support, modify existing views)
- Change `gravity.sh`
  - Instead of unifying and sorting domains during gravity, we parse and store the entire adlists in the database. This allows us to assign adlists to groups which would not be possible with unified data.
  - Fix an issue where disabled list entries are still counted towards to total list count

Many changes are required as the entire handling is changed to be group-based.

### DIFF
--- a/advanced/Scripts/database_migration/gravity-db.sh
+++ b/advanced/Scripts/database_migration/gravity-db.sh
@@ -39,4 +39,9 @@ upgrade_gravityDB(){
 		sqlite3 "${database}" < "/etc/.pihole/advanced/Scripts/database_migration/gravity/2_to_3.sql"
 		version=3
 	fi
+	if [[ "$version" == "3" ]]; then
+		# This migration script upgrades ...
+		sqlite3 "${database}" < "/etc/.pihole/advanced/Scripts/database_migration/gravity/3_to_4.sql"
+		version=3
+	fi
 }
--- a/advanced/Scripts/database_migration/gravity-db.sh
+++ b/advanced/Scripts/database_migration/gravity-db.sh
@@ -10,6 +10,8 @@
 # This file is copyright under the latest version of the EUPL.
 # Please see LICENSE file for your rights under this license.
 
+readonly scriptPath="/etc/.pihole/advanced/Scripts/database_migration/gravity"
+
 upgrade_gravityDB(){
 	local database piholeDir auditFile version
 	database="${1}"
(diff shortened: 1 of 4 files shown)


advanced/Scripts/database_migration/gravity-db.sh
```diff
@@ -39,4 +39,10 @@ upgrade_gravityDB(){
 		sqlite3 "${database}" < "/etc/.pihole/advanced/Scripts/database_migration/gravity/2_to_3.sql"
 		version=3
 	fi
+	if [[ "$version" == "3" ]]; then
+		# This migration script upgrades the gravity and adlist views
+		# implementing necessary changes for per-client blocking
+		sqlite3 "${database}" < "/etc/.pihole/advanced/Scripts/database_migration/gravity/3_to_4.sql"
```
Mcat12: Might want to pull `/etc/.pihole/advanced/Scripts/database_migration/gravity` into a constant.

advanced/Scripts/database_migration/gravity-db.sh
```diff
@@ -48,6 +48,20 @@ upgrade_gravityDB(){
...
+		# This migration script upgrades the gravity and list views
+		# implementing necessary changes for per-client blocking
+		echo -e "  ${INFO} Upgrading gravity database from version 4 to 5"
+		sqlite3 "${database}" < "${scriptPath}/4_to_5.sql"
+		version=5
+	fi
+	if [[ "$version" == "5" ]]; then
+		# This migration script upgrades the adlist view
+		# to return an ID used in gravity.sh
+		echo -e "  ${INFO} Upgrading gravity database from version 5 to 6"
```
Mcat12: Do we need two migration scripts, since this is all in one PR?

PromoFaux: One could argue whether we even need 5, seeing as the eventual `release/5.0` -> master will be one PR, also.

I should imagine DL has a sane reason for having it separated

Mcat12: The reason for separating them on development is for those who follow the development branch.

DL6ER: The reason for separating them was that the two changes came with a temporal distance more than one month in between. Even if this is still a more-or-less isolated feature branch, it seems that enough people got interested in it and we know of at least one user actually using and trying it. Because changes to the migration scripts would have needed them to rebuilt their gravity databases from ground up, I felt this would be enough of an issue to warrant to updates.

dschaper: LGTM

PR #2995, opened 2019-11-08
Don't allow repo to go further than latest tag on master

PromoFaux:

---
At present, if we push to the `master` branch after tagging a release, new installs will be left in a state of flux as the installer clones the repo at the latest commit.

After having tried to force a clone of the repo at a specific tag, and failing, I figured it would be much simpler to clone to HEAD and then reset the local repo to the latest tag available.

This is all theory at the moment, though I have tested the commands.. it was not part of the install script. Not sure how best to test it in the script without first releasing... Anyway, open to thoughts and suggestions

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -378,9 +378,9 @@ fi
 is_repo() {
     # Use a named, local variable instead of the vague $1, which is the first argument passed to this function
     # These local variables should always be lowercase
-    local directory="${1}"    
+    local directory="${1}"
     # A variable to store the return code
-    local rc        
+    local rc
     # If the first argument passed to this function is a directory,
     if [[ -d "${directory}" ]]; then
         # move into the directory
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -424,7 +424,7 @@ make_repo() {
     # In case extra commits have been added after tagging/release (i.e in case of metadata updates/README.MD tweaks)    
     curBranch=$(git rev-parse --abbrev-ref HEAD)
     if [[ "${curBranch}" == "master" ]]; then #If we're calling make_repo() then it should always be master, we may not need to check.
-         git reset --hard "$(git describe --abbrev=0)" || return $?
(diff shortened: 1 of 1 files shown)


dschaper:
The mechanics seem sound as does the concept. Playing with something to try here in a bit.

automated install/basic-install.sh
```diff
@@ -435,30 +441,33 @@ update_repo() {
...
     # Stash any local commits as they conflict with our working code
     git stash --all --quiet &> /dev/null || true # Okay for stash failure
     git clean --quiet --force -d || true # Okay for already clean directory
     # Pull the latest commits
     git pull --quiet &> /dev/null || return $?
+    # Check current branch. If it is master, then reset to the latest availible tag.
+    # In case extra commits have been added after tagging/release (i.e in case of metadata updates/README.MD tweaks)    
+    curBranch=$(git rev-parse --abbrev-ref HEAD)
+    if [[ "${curBranch}" == "master" ]]; then
+         git reset --hard "$(git describe --abbrev=0)" || return $?
```
DL6ER: Isn't it dangerous to put whatever `git describe` returns into this call without checking if the returned value is something we want?

I played around and saw this:
```
root@munichpi:~/FTL# git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
root@munichpi:~/FTL# git pull
Already up-to-date.
root@munichpi:~/FTL# git describe --abbrev=0
fatal: No annotated tags can describe 'b60d63f448179cb139755eb8dcfd9d5335df3fd4'.
However, there were unannotated tags: try --tags.
root@munichpi:~/FTL# git describe --abbrev=0 --tags
v4.3.1
```

PromoFaux: So you think we should add the `--tags` to the end?

PromoFaux: Looking at it, adding --tags is at least safer in terms of output. We shouldn't ever be in a situation on `master` (the only branch this code will check) where there are no tags for `describe` to... describe

PR #3052, opened 2019-12-21
Revert "Change the regex used for domain validation"

### LINKED ISSUE #3046: Core domain name validation
@DL6ER You are right, I expect to be able to add the domain via CLI too.

On a quick look I couldn't find the validation check for the web gui, because I wanted to compare them due to different behavior.

Check out  http://regexr.com/3au3g for an alternative validation check for [list.sh](https://github.com/pi-hole/pi-hole/blob/development/advanced/Scripts/list.sh#L95)

_Originally posted by @Th3M3 in https://github.com/pi-hole/AdminLTE/pull/1078#issuecomment-565761008_

DL6ER:

---

Fix #3046

Revert commit 76460f0

None

### DIFF
--- a/advanced/Scripts/list.sh
+++ b/advanced/Scripts/list.sh
@@ -91,8 +91,8 @@ ValidateDomain() {
         if [[ ( "${typeId}" == "${regex_blacklist}" || "${typeId}" == "${regex_whitelist}" ) && "${wildcard}" == false ]]; then
             validDomain="${domain}"
         else
-            # Use regex to check the validity of the passed domain. see https://regexr.com/3abjr
-            validDomain=$(grep -P "^((?!-))(xn--)?[a-z0-9][a-z0-9-_]{0,61}[a-z0-9]{0,1}\.(xn--)?([a-z0-9\-]{1,61}|[a-z0-9-]{1,30}\.[a-z]{2,})$" <<< "${domain}")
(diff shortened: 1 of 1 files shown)


WaLLy3K:
The main purpose of the domain validation logic (as far as memory serves) is to prevent users from adding hostnames/domains that when parsed by FTL (`dnsmasq`), cause unexpected results such as crashing due to the entry being incorrectly formatted.

Basically, the regex causes valid domains from not being added, that's a problem. If it allows `blah123` to be added, then that's fine. 😄

DL6ER:
Background knowledge summary:
Compared to the first days of Pi-hole, we have made some great progress. Pi-hole v5.0 will not load any of the gravity or list domains into its cache. This will also allow for using hundreds of millions of domains without much trouble even on a Pi Zero*.
We store all the domains in a [B-tree](https://en.wikipedia.org/wiki/B-tree) (B as in "balanced") which is also why the run times of `pihole -g` got somewhat longer.

What I actually wanted to write:
We can store whatever we want in the lists. You could even store binary data like images in there if you like*. What will happen is that FTL will look up the queried domain in this B-tree. When a match is found, FTL acts accordingly (block or not). If not, then the domain is on no list and permitted.

**TL;DR**: You can throw whatever you like at the database, you can never corrupt FTL with this.

---

*) Even though it is possible, I don't consider this a good idea!

PR #3066, opened 2020-01-01
Update installer to support CentOS 8

### LINKED ISSUE #2977: Official installation support for CentOS 8/ RHEL 8
Provide official installation support for CentOS 8 and RHEL 8. 
The PHP version in the RHEL 8 app stream is 7.2.11 which will be updated a little in RHEL 8.1.
Other dependencies (if not available) i can request on RHEL bugzilla for EPEL 8. :) +1

bcambl:
- [X] I have squashed any insignificant commits.

---

Fix #2977

Update installer compatibility for:
- Fedora 31
- CentOS 7
- CentOS 8

Package `php-json` was only required for Fedora in the past. The latest CentOS now requires this package as well. I have flipped the logic so that `php-json` is now a main dependency and remove the dependency when installed on CentOS7.

OS support docs will need to be updated to include the distros listed above.

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -287,15 +287,14 @@ elif is_command rpm ; then
     PKG_COUNT="${PKG_MANAGER} check-update | egrep '(.i686|.x86|.noarch|.arm|.src)' | wc -l"
     INSTALLER_DEPS=(dialog git iproute newt procps-ng which chkconfig)
     PIHOLE_DEPS=(bind-utils cronie curl findutils nmap-ncat sudo unzip wget libidn2 psmisc sqlite libcap)
-    PIHOLE_WEB_DEPS=(lighttpd lighttpd-fastcgi php-common php-cli php-pdo php-xml)
+    PIHOLE_WEB_DEPS=(lighttpd lighttpd-fastcgi php-common php-cli php-pdo php-xml php-json)
     LIGHTTPD_USER="lighttpd"
     LIGHTTPD_GROUP="lighttpd"
     LIGHTTPD_CFG="lighttpd.conf.fedora"
     # If the host OS is Fedora,
     if grep -qiE 'fedora|fedberry' /etc/redhat-release; then
         # all required packages should be available by default with the latest fedora release
-        # ensure 'php-json' is installed on Fedora (installed as dependency on CentOS7 + Remi repository)
-        PIHOLE_WEB_DEPS+=('php-json')
+        : # continue
     # or if host OS is CentOS,
     elif grep -qiE 'centos|scientific' /etc/redhat-release; then
         # Pi-Hole currently supports CentOS 7+ with PHP7+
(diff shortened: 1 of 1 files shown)


automated install/basic-install.sh
```diff
@@ -310,7 +309,21 @@ elif is_command rpm ; then
...
+            # create a temporary array as arrays are not designed for use as mutable data structures
+            CENTOS7_PIHOLE_WEB_DEPS=()
+            for i in "${!PIHOLE_WEB_DEPS[@]}"; do
+                if [[ ${PIHOLE_WEB_DEPS[i]} != "php-json" ]]; then
+                    CENTOS7_PIHOLE_WEB_DEPS+=( "${PIHOLE_WEB_DEPS[i]}" )
+                fi
+            done
+            # re-assign the clean dependency array back to PIHOLE_WEB_DEPS
+            PIHOLE_WEB_DEPS=("${CENTOS7_PIHOLE_WEB_DEPS[@]}")
+            unset CLEAN_PIHOLE_WEB_DEPS
```
DL6ER: Shouldn't this be the following?
```suggestion
            unset CENTOS7_PIHOLE_WEB_DEPS
```

bcambl: Ah yes, I had re-named the variable before the pushing and missed that one!   
I am surprised that CI did not catch an unset of an un-used variable.. 
I have updated, squashed, and cleaned up my commit message.

Thanks!

PromoFaux:
Tested working as expected on Fedora 31 and CentOS 7, but as DO do not have an image for CentOS 8.. i'm currently not in a position to test that!

An observation on the CentOS7 part... I noted a pause in the installer output after `[i] Checking for php-xml (will be installed)`. Presumably this is due to rebuilding the dependency array to remove `php-json`... I wonder if it might be faster to do something along the lines of (totally untested probably pseudocode...)

```
PIHOLE_WEB_DEPS=(lighttpd lighttpd-fastcgi php-common php-cli php-pdo php-xml)

if [[  $CURRENT_CENTOS_VERSION -ne 7 ]]; then
   PIHOLE_WEB_DEPS=("${PIHOLE_WEB_DEPS[@]}" php-json)
fi
```

bcambl:
Both Fedora and CentOS8 need `php-json` so we would need to do it twice or have two conditions.  
Since its only CentOS7 that _doesnt_ need it, we only prune it on CentOS7.  
As for the pause, it shouldn't be the array rebuild.. that happens super fast:

```
[testing]$ cat ph.sh 
#!/usr/bin/bash

PIHOLE_WEB_DEPS=(lighttpd lighttpd-fastcgi php-common php-cli php-pdo php-xml php-json)

CENTOS7_PIHOLE_WEB_DEPS=()
for i in "${!PIHOLE_WEB_DEPS[@]}"; do
    if [[ ${PIHOLE_WEB_DEPS[i]} != "php-json" ]]; then
        CENTOS7_PIHOLE_WEB_DEPS+=( "${PIHOLE_WEB_DEPS[i]}" )
    fi
done

PIHOLE_WEB_DEPS=("${CENTOS7_PIHOLE_WEB_DEPS[@]}")
unset CENTOS7_PIHOLE_WEB_DEPS
[testing]$ time ./ph.sh 

real	0m0.005s
user	0m0.003s
sys	0m0.002s
[testing]$

```

PromoFaux:
> As for the pause, it shouldn't be the array rebuild.. that happens super fast:

Hmm, yeah OK that's odd. Able to reproduce the `time` result, so perhaps it was something else. Didn't notice it on Fedora, though it might just be because I was watching that section like a hawk for CentOS 7

bcambl:
I think the pause is due to #2962 not being merged yet.  
The `/dev/null` redirect for the package install is still in place until that PR is merged.

Issue #2720, opened 2019-04-25
Changing port breaks /pihole


dom-madrid:
**In raising this issue, I confirm the following:** `{please fill the checkboxes, e.g: [X]}`

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the the source code relevant to this issue?:**

1

---
**Expected behaviour:**

Install Pihole on a server with an existing web server on port 80/443, change lighttpd port from 80 to 8088 (to restrict access to only lan access - no web), get normal access for both /admin and /pihole

**Actual behaviour:**

/pihole no longer works (it does under port 80).
/admin works

**Steps to reproduce:**

Install Pihole
Change lighttpd.conf port from 80 to 8088
Restart lighttpd and pihole
Goto 192.168.1.xx:8088/pihole
get the following error:
[ERROR]: Unable to parse results from queryads.php: Unhandled error message ()

**Debug token provided by [uploading `pihole -d` log](https://discourse.pi-hole.net/t/the-pihole-command-with-examples/738#debug):**

https://tricorder.pi-hole.net/i5wobdlcel!

**Troubleshooting undertaken, and/or other relevant information:**

Running on existing Ubuntu Server 16.04 LTS with both lighttp and Apache

dschaper:
There is no `/pihole` endpoint in the Pi-hole web server setup. How did you install Pi-hole?

dom-madrid:
I cloned the git repo as per https://github.com/pi-hole/pi-hole/#one-step-automated-install and followed alternate option 1.
Add'al Info
/var/www/html clearly has a /pihole folder
On a straight install to port 80 (on a different machine - VM or PM - Ubuntu or Raspbian), when accessing 192.168.1.xx/pihole you get the Pihole logo and an invitation to redirect to the /admin page.

dschaper:
[...]

PR #3067, opened 2020-01-02
Fixes broken blocking page and landing page when changing server port or host name

mschoettle:

---

Addresses issues #2195 and #2720 to allow changing the server port and/or host name.

* `$_SERVER["SERVER_NAME"]` used instead of `$_SERVER["HOST"] (host contains both host name and port causing the query ads script to result in an invalid domain error)
* uses the `VIRTUAL_HOST` if set in the environment to determine whether to redirect to the admin page (in addition to `pi.hole`). This can be set with the following line in `/etc/lighttpd/external.conf`:
```
setenv.add-environment = ("virtual_host" => "your.fqdn.tld")
```
* it is still assumed that lighttpd runs on localhost but the call to *queryAds* now includes the server port
* replaced use of `pi.hole` prefix in path for assets in the html of the blocking page

I don't know if this addresses the issue when running under docker, maybe someone else can confirm this.

Document setting the VIRTUAL_HOST in the environment when using a host name other than `pi.hole`.

### DIFF
--- a/advanced/index.php
+++ b/advanced/index.php
@@ -6,8 +6,8 @@
 *  This file is copyright under the latest version of the EUPL.
 *  Please see LICENSE file for your rights under this license. */
 
-// Sanitize HTTP_HOST output
-$serverName = htmlspecialchars($_SERVER["HTTP_HOST"]);
+// Sanitize SERVER_NAME output
+$serverName = htmlspecialchars($_SERVER["SERVER_NAME"]);
 // Remove external ipv6 brackets if any
 $serverName = preg_replace('/^\[(.*)\]$/', '${1}', $serverName);
 
@@ -50,7 +50,8 @@ function setHeader($type = "x") {
 }
 
 // Determine block page type
-if ($serverName === "pi.hole") {
+if ($serverName === "pi.hole"
+    || (!empty($_SERVER["VIRTUAL_HOST"]) && $_SERVER["SERVER_NAME"] === $_SERVER["VIRTUAL_HOST"])) {
     // Redirect to Web Interface
     exit(header("Location: /admin"));
 } elseif (filter_var($serverName, FILTER_VALIDATE_IP) || in_array($serverName, $authorizedHosts)) {
@@ -131,7 +132,12 @@ ini_set("default_socket_timeout", 3);
 function queryAds($serverName) {
     // Determine the time it takes while querying adlists
     $preQueryTime = microtime(true)-$_SERVER["REQUEST_TIME_FLOAT"];
-    $queryAds = file("http://127.0.0.1/admin/scripts/pi-hole/php/queryads.php?domain=$serverName&bp", FILE_IGNORE_NEW_LINES);
+    $queryAdsURL = sprintf(
+        "http://127.0.0.1:%s/admin/scripts/pi-hole/php/queryads.php?domain=%s&bp",
+        $_SERVER["SERVER_PORT"],
+        $serverName
+    );
+    $queryAds = file($queryAdsURL, FILE_IGNORE_NEW_LINES);
     $queryAds = array_values(array_filter(preg_replace("/data:\s+/", "", $queryAds)));
     $queryTime = sprintf("%.0f", (microtime(true)-$_SERVER["REQUEST_TIME_FLOAT"]) - $preQueryTime);
 
@@ -226,10 +232,10 @@ setHeader();
   <?=$viewPort ?>
   <meta name="robots" content="noindex,nofollow"/>
   <meta http-equiv="x-dns-prefetch-control" content="off">
-  <link rel="shortcut icon" href="//pi.hole/admin/img/favicon.png" type="image/x-icon"/>
-  <link rel="stylesheet" href="//pi.hole/pihole/blockingpage.css" type="text/css"/>
+  <link rel="shortcut icon" href="admin/img/favicon.png" type="image/x-icon"/>
+  <link rel="stylesheet" href="pihole/blockingpage.css" type="text/css"/>
   <title>● <?=$serverName ?></title>
-  <script src="//pi.hole/admin/scripts/vendor/jquery.min.js"></script>
+  <script src="admin/scripts/vendor/jquery.min.js"></script>
   <script>
     window.onload = function () {
       <?php
--- a/advanced/index.php
+++ b/advanced/index.php
@@ -51,7 +51,7 @@ function setHeader($type = "x") {
 
 // Determine block page type
 if ($serverName === "pi.hole"
-    || (!empty($_SERVER["VIRTUAL_HOST"]) && $_SERVER["SERVER_NAME"] === $_SERVER["VIRTUAL_HOST"])) {
+    || (!empty($_SERVER["VIRTUAL_HOST"]) && $serverName === $_SERVER["VIRTUAL_HOST"])) {
     // Redirect to Web Interface
     exit(header("Location: /admin"));
 } elseif (filter_var($serverName, FILTER_VALIDATE_IP) || in_array($serverName, $authorizedHosts)) {
--- a/advanced/index.php
+++ b/advanced/index.php
@@ -57,10 +57,17 @@ if ($serverName === "pi.hole"
 } elseif (filter_var($serverName, FILTER_VALIDATE_IP) || in_array($serverName, $authorizedHosts)) {
     // Set Splash Page output
     $splashPage = "
-    <html><head>
+    <html>
+      <head>
         $viewPort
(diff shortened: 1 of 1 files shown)


Matthewar:
New to pi-hole, but I'm having a similar issue to the one linked - is `VIRTUAL_HOST` currently documented anywhere? It seems to be used in the [web interface authentication](https://github.com/pi-hole/AdminLTE/blob/38d8e77f999ba5999c013b44f6e8fe808308de7a/scripts/pi-hole/php/auth.php#L39-L42), but I can't see any docs about it.

With regards to the code, would it be better to use the sanitised form of `SERVER_NAME` [here](https://github.com/pi-hole/pi-hole/pull/3067/files#diff-622eb992a761acc873605fcdfbaf1e3bR54)
Also not sure about this, but optionally using `VIRTUAL_HOST` instead of the [loopback address](https://github.com/pi-hole/pi-hole/pull/3067/files#diff-622eb992a761acc873605fcdfbaf1e3bR142) if its available?

mschoettle:
> New to pi-hole, but I'm having a similar issue to the one linked - is `VIRTUAL_HOST` currently documented anywhere? It seems to be used in the [web interface authentication](https://github.com/pi-hole/AdminLTE/blob/38d8e77f999ba5999c013b44f6e8fe808308de7a/scripts/pi-hole/php/auth.php#L39-L42), but I can't see any docs about it.

I couldn't find any documentation. I saw it mentioned in an issue or PR somewhere that this can be set.

It is used here: https://github.com/pi-hole/pi-hole/blob/1d43c0a2fa36ddc993db51b1751e26cbc8123fbf/advanced/index.php#L32

> With regards to the code, would it be better to use the sanitised form of `SERVER_NAME` [here](https://github.com/pi-hole/pi-hole/pull/3067/files#diff-622eb992a761acc873605fcdfbaf1e3bR54)

Good catch, I'll change that.

> Also not sure about this, but optionally using `VIRTUAL_HOST` instead of the [loopback address](https://github.com/pi-hole/pi-hole/pull/3067/files#diff-622eb992a761acc873605fcdfbaf1e3bR142) if its available?

This doesn't work if pi-hole is behind a reverse proxy and you don't open the lighttpd port to the outside.

Matthewar:
Seems to be all working for me with this change and my setup.

> I couldn't find any documentation. I saw it mentioned in an issue or PR somewhere that this can be set.

Yeah it doesn't seem to be well described what either the `FQDN` or `VIRTUAL_HOST` are used for, other than in comments - even then I'm not really sure what they're for.

> This doesn't work if pi-hole is behind a reverse proxy and you don't open the lighttpd port to the outside.

That makes sense, I'm currently using nginx which isn't properly supported, and it's still working with the loopback address (although I'm not entirely sure why as I'm listening to a subdomain address in nginx) - I think ultimately running it on a different port with a reverse proxy is a better approach anyway.

evandanish:
>> Also not sure about this, but optionally using VIRTUAL_HOST instead of the loopback address if its available?

> This doesn't work if pi-hole is behind a reverse proxy and you don't open the lighttpd port to the outside.

I was under the impression that if you were behind a reverse proxy, (i.e. nginx), the `SERVER_NAME` would be re-written by the proxy, so I'm not sure why this wouldn't work.

That being said, it seems fine to assume that lighttpd runs on localhost.

Great work btw.

mschoettle:
> I was under the impression that if you were behind a reverse proxy, (i.e. nginx), the `SERVER_NAME` would be re-written by the proxy, so I'm not sure why this wouldn't work.

The reverse proxy can't rewrite this since it's a server variable.

I pushed the small adjustment suggested by @Matthewar.

mschoettle:
@dschaper @DL6ER Could either of you review this please?

PromoFaux:
I'll try and take a look over this later.. it's been such a long time since I used the block/landing page!

advanced/index.php
```diff
@@ -232,10 +238,10 @@ function queryAds($serverName) {
   <?=$viewPort ?>
   <meta name="robots" content="noindex,nofollow"/>
   <meta http-equiv="x-dns-prefetch-control" content="off">
-  <link rel="shortcut icon" href="//pi.hole/admin/img/favicon.png" type="image/x-icon"/>
-  <link rel="stylesheet" href="//pi.hole/pihole/blockingpage.css" type="text/css"/>
+  <link rel="shortcut icon" href="admin/img/favicon.png" type="image/x-icon"/>
```
XhmikosR: Does this need to be relative? Because there might be more instances of relative to server root URLs

mschoettle:
Thanks, @PromoFaux!

> Not entirely sure how to test the `VIRTUAL_HOST` part.

This is to support host names other than `pi.hole`. Let's say you want to change the hostname where the admin interface is reachable at (use case example: use of reverse proxy). To test, it might be enough to have another hostname resolve to the lighttpd IP, but I am not sure about this.
With the changes of this PR you would still get "Website blocked" page instead of the admin interface.
If you update your `external.conf` of lighttpd as outlined in the PR description, you will be redirected to the admin interface.

> Honestly, I'm amazed there are still folk out there using the block page, considering the rise of `https` content out on the web!

I think that most people actually are not using it. But the two/three pages (block, landing, splash pages...) are coming from the same file (`advanced/index.php`). The issue occurs when changing the port, host name or both of lighttpd.

advanced/index.php
mschoettle: Yes, to support port and host name changes. I replaced all occurrences of this in this file.

XhmikosR: I'm talking about the missing slash at the beginning. Because there is at least one more case that's using `/admin` and not `admin` in this page.

https://github.com/pi-hole/pi-hole/blob/94ec8fdfcade9e7c63b816bdbdc4809c54487dda/advanced/index.php#L335

So, I'm not sure if the missing slash is intentional or not, nor can I confirm this. Just noticed the discrepancy.

amunchet: I think it would be best to have the URL be relative, since you may want URL rewriting.

For example: I want to put the pihole in a docker behind a reverse proxy (and use something like traefik to route via URLs plus HTTPS middleware).  If the URL is relative, I can have it go to https://homelab/pihole/admin/img/favicon.png and things work fine.  If it's /admin, then that won't work.

Where else are you seeing /admin?

XhmikosR: In the line I showed you above and here https://github.com/pi-hole/pi-hole/blob/94ec8fdfcade9e7c63b816bdbdc4809c54487dda/advanced/index.php#L63

mschoettle: > I'm talking about the missing slash at the beginning. Because there is at least one more case that's using `/admin` and not `admin` in this page.

Ah, sorry, misunderstood what you meant.

In the case where it's accessible directly under the root (e.g, `https://pihole.domain.tld`) either of the two works. But there should probably be a consistent use of one of the two.

For the case that @amunchet mentions only the relative URL will work. But in that case I am not sure if the redirect will still work: 

https://github.com/pi-hole/pi-hole/blob/94ec8fdfcade9e7c63b816bdbdc4809c54487dda/advanced/index.php#L56

amunchet: Actually, I think it's fine to only support direct access instead of URL rewriting - i.e. going to http://pi.hole/admin instead of http://homeserver/pihole/admin.  I think the header rewriting won't work with relative paths, so that pretty much settles it.  I don't think there's any harm in using relative paths in the HTML parts though.

XhmikosR: So, why not use in the 2 places too? Just trying to understand why we are not consistent here.

amunchet: I can agree with that.  Up to @mschoettle, as this is his PR.  I don't think it will make any difference functionality-wise, but the consistency will probably "smell" better.

mschoettle: I totally agree. I changed it in the 2 places to use relative URLs as well for consistency.

XhmikosR:
@mschoettle this needs a rebase.

/CC @DL6ER @PromoFaux for review (I don't have access to this repo anyway)

mschoettle:
> @mschoettle this needs a rebase.

Done. I also cleaned up the history a bit by squashing some commits.

ireznik:
@XhmikosR so it can be merged now?

XhmikosR:
@ireznik I don't have push rights myself. You should wait until someone else reviews.

dschaper:
If this goes to `development` then it won't be in version 5.

mschoettle:
> If this goes to `development` then it won't be in version 5.

Is there any other way, so that it lands in version 5?

dschaper:
Rebase on `release/v5.0` and PR against that branch.

mschoettle:
> Rebase on `release/v5.0` and PR against that branch.

See PR #3263

PR #3137, opened 2020-02-16
Change to use API instead of the Location Header

PromoFaux:
[Internal]

Should alleviate some issues we are seeing in CI


dschaper:
I think we need to cherry pick this in to development as well?

PromoFaux:
We could probably do with merging the whole of `release/v5.0` into development, to be honest...

dschaper:
Good point.

PR #3144, opened 2020-02-18
Fix issue with missing newline at the end of adlists

DL6ER:

---

Fix issue with missing newline at the end of adlists. Also, always display warnings, even if they are non-fatal, like:
Screenshot from 2020-02-18 17-18-52
(however, you will **not** see this warning yourself as this PR fixes it in its second commit)

Ensure the last line of what we add to the batch-import file is a newline (this uses `sed`-magic):
This command searches for the end of the file (`$`) and checks if there is anything here (`.$` will ONLY match when the last character is not a newline `\n` as this is the only thing that doesn't match `.`). Only if this is true (newline is not the last character), we append nothing (which will effectively insert a newline in `sed`-world).

None

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -486,13 +486,15 @@ gravity_DownloadBlocklistFromUrl() {
     if [[ "${httpCode}" == "304" ]]; then
       # Add domains to database table file
       #Append ,${arg} to every line and then remove blank lines before import
-      sed -e "s/$/,${adlistID}/;/^$/d" "${saveLocation}" >> "${target}"
+      # /.$/a\\ ensures there is a newline on the last line
+      sed -e "s/$/,${adlistID}/;/^$/d;/.$/a\\" "${saveLocation}" >> "${target}"
     # Check if $patternbuffer is a non-zero length file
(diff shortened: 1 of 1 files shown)


pralor:

https://discourse.pi-hole.net/t/unable-to-fill-gravity-table/28217/27

dschaper:
It looks fine to me but `sed` makes me go blurry. Could use another set of eyes on this.

gravity.sh
```diff
@@ -482,13 +486,15 @@ gravity_DownloadBlocklistFromUrl() {
     if [[ "${httpCode}" == "304" ]]; then
       # Add domains to database table file
       #Append ,${arg} to every line and then remove blank lines before import
-      sed -e "s/$/,${adlistID}/;/^$/d" "${saveLocation}" >> "${target}"
+      # /.$/a\\ ensures there is a newline on the last line
+      sed -e "s/$/,${adlistID}/;/^$/d;/.$/a\\" "${saveLocation}" >> "${target}"
```
jokaorgua: maybe better to move this to function? as it brakes DRY principle

PR #3170, opened 2020-02-26
Add CACHE_SIZE to setupVars

DirkJanIT:

- [ ] I am willing to help maintain this change if there are issues with it later.

---
https://discourse.pi-hole.net/t/add-cache-size-to-setupvars-conf/28667

Adds the necessary variable to setupVars.conf.

Document the addition of CACHE_SIZE in setupVars.conf.

### DIFF
--- a/advanced/01-pihole.conf
+++ b/advanced/01-pihole.conf
@@ -35,7 +35,7 @@ server=@DNS2@
 
 interface=@INT@
 
-cache-size=10000
+cache-size=@CACHE_SIZE@
 
 log-queries
 log-facility=/var/log/pihole.log
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -79,6 +79,7 @@ IPV6_ADDRESS=""
 QUERY_LOGGING=true
 INSTALL_WEB_INTERFACE=true
 PRIVACY_LEVEL=0
+CACHE_SIZE=10000
 
 if [ -z "${USER}" ]; then
   USER="$(id -un)"
@@ -1281,7 +1282,10 @@ version_check_dnsmasq() {
         #
         sed -i '/^server=@DNS2@/d' ${dnsmasq_pihole_01_location}
     fi
-
+	
+	# Set the cache size
+	sed -i "s/@CACHE_SIZE@/$CACHE_SIZE/" ${dnsmasq_pihole_01_location}
+   
     #
     sed -i 's/^#conf-dir=\/etc\/dnsmasq.d$/conf-dir=\/etc\/dnsmasq.d/' ${dnsmasq_conf}
 
@@ -1803,7 +1807,7 @@ finalExports() {
     # If the setup variable file exists,
     if [[ -e "${setupVars}" ]]; then
         # update the variables in the file
-        sed -i.update.bak '/PIHOLE_INTERFACE/d;/IPV4_ADDRESS/d;/IPV6_ADDRESS/d;/PIHOLE_DNS_1/d;/PIHOLE_DNS_2/d;/QUERY_LOGGING/d;/INSTALL_WEB_SERVER/d;/INSTALL_WEB_INTERFACE/d;/LIGHTTPD_ENABLED/d;' "${setupVars}"
+        sed -i.update.bak '/PIHOLE_INTERFACE/d;/IPV4_ADDRESS/d;/IPV6_ADDRESS/d;/PIHOLE_DNS_1/d;/PIHOLE_DNS_2/d;/QUERY_LOGGING/d;/INSTALL_WEB_SERVER/d;/INSTALL_WEB_INTERFACE/d;/LIGHTTPD_ENABLED/d;/CACHE_SIZE/d;' "${setupVars}"
     fi
     # echo the information to the user
     {
@@ -1816,6 +1820,7 @@ finalExports() {
     echo "INSTALL_WEB_SERVER=${INSTALL_WEB_SERVER}"
     echo "INSTALL_WEB_INTERFACE=${INSTALL_WEB_INTERFACE}"
     echo "LIGHTTPD_ENABLED=${LIGHTTPD_ENABLED}"
+    echo "CACHE_SIZE=${CACHE_SIZE}"
     }>> "${setupVars}"
 
     # Set the privacy level


advanced/01-pihole.conf
```diff
@@ -35,7 +35,7 @@ server=@DNS2@
 
 interface=@INT@
 
-cache-size=10000
```
jlagermann: This change is unnecessary.  The end result can be accomplished less code if you leave the default value in place and then use the following to replace only the value if it changed.  The following would go in the basic-install.sh file.
`sed -i 's/\(cache-size=\).*/\1${CACHE_SIZE}/' ${dnsmasq_pihole_01_location}`

automated install/basic-install.sh
```diff
@@ -1281,7 +1282,10 @@ version_check_dnsmasq() {
         #
         sed -i '/^server=@DNS2@/d' ${dnsmasq_pihole_01_location}
     fi
-
+	
+	# Set the cache size
+	sed -i "s/@CACHE_SIZE@/$CACHE_SIZE/" ${dnsmasq_pihole_01_location}
```
jlagermann: If the following is used, 01-pihole.conf will not need to be modified.
`sed -i 's/\(cache-size=\).*/\1${CACHE_SIZE}/' ${dnsmasq_pihole_01_location}`

automated install/basic-install.sh
```diff
@@ -1803,7 +1807,7 @@ finalExports() {
     # If the setup variable file exists,
     if [[ -e "${setupVars}" ]]; then
         # update the variables in the file
-        sed -i.update.bak '/PIHOLE_INTERFACE/d;/IPV4_ADDRESS/d;/IPV6_ADDRESS/d;/PIHOLE_DNS_1/d;/PIHOLE_DNS_2/d;/QUERY_LOGGING/d;/INSTALL_WEB_SERVER/d;/INSTALL_WEB_INTERFACE/d;/LIGHTTPD_ENABLED/d;' "${setupVars}"
+        sed -i.update.bak '/PIHOLE_INTERFACE/d;/IPV4_ADDRESS/d;/IPV6_ADDRESS/d;/PIHOLE_DNS_1/d;/PIHOLE_DNS_2/d;/QUERY_LOGGING/d;/INSTALL_WEB_SERVER/d;/INSTALL_WEB_INTERFACE/d;/LIGHTTPD_ENABLED/d;/CACHE_SIZE/d;' "${setupVars}"
```
jlagermann: This is redundant.  If the default value is used, setupVars does not need to be updated.  If the value is different than the default, it would have been passed in through setupVars, therefor, it's already in the file.  The PHP ini does not need this value so I don't see a good reason to add the default, or delete and then put the non-default right back.

automated install/basic-install.sh
```diff
@@ -1816,6 +1820,7 @@ finalExports() {
     echo "INSTALL_WEB_SERVER=${INSTALL_WEB_SERVER}"
     echo "INSTALL_WEB_INTERFACE=${INSTALL_WEB_INTERFACE}"
     echo "LIGHTTPD_ENABLED=${LIGHTTPD_ENABLED}"
+    echo "CACHE_SIZE=${CACHE_SIZE}"
```
jlagermann: Does this value need to be echo'd back?  The way this is currently implemented, it's appending these settings to setupVars.conf instead of replacing them.  This requires the setting be unnecessarily deleted first.  
The cache size can be confirmed in the settings in the GUI if needed.

dschaper: Seems logical to do so as we output the rest of the values.

automated install/basic-install.sh
dschaper: The PR keeps in the style of the rest of the code as it exists.

advanced/01-pihole.conf
dschaper: I disagree.

automated install/basic-install.sh
DoubleOhmSeven: I indeed followed the rest of the code style, I'm not really that knowledgeable about shell scripts   but can make changes (as long as I understand how :)) if anything is needed.

automated install/basic-install.sh
DoubleOhmSeven: Doesn't the same apply to all the other variables? I only added it because they were all listed in setupVars.

automated install/basic-install.sh
dschaper: It looks good to me.

dschaper:
https://docs.pi-hole.net/ftldns/dns-cache/ will need to be aligned with this PR.

advanced/01-pihole.conf
dschaper: If the default value is left in place then the docker image will always have that value. Removing the configuration is beneficial. @diginc FYI.

jlagermann: The default value would be replaced with the following in the basic-install.sh.  I fixed a typo from my earlier comment.  Bash will not expand variables inside single quotes.  Just by searching for the existing variable name you can replace the value.
`sed -i "s/\(cache-size=\).*/\1${CACHE_SIZE}/" ${dnsmasq_pihole_01_location}`

dschaper: If the docker image creation used basic-install then there'd be a point. It doesn't however.

pralor:

https://discourse.pi-hole.net/t/add-cache-size-to-setupvars-conf/28667/56

DL6ER:
@dschaper Are you okay with overwriting Travis here? It seems to be dead.

dschaper:
No, travis is failed for a reason. Note the branches.

DL6ER (approved):

DL6ER:
@dschaper Are you still okay with this being merged? If so, please go ahead.

dschaper:
Need the `.travis.yml` file from master (or development) to be merged in to the PR so travis will check it.

DL6ER:
The source repository doesn't exist any more

dschaper:
If you think it's a good PR then we can override it.

o5ky:
FYI this breaks when you upgrade from a previous older version using the command below. It inserts the variable instead of the variable value.
`/etc/.pihole/automated\ install/basic-install.sh --repair --unattended`

`cache-size=@CACHE_SIZE@`

FTL failed to start due to error at line 32 of /etc/dnsmasq.d/01-pihole.conf

dschaper:
Is there a reason you didn't want to use the built in upgrade process of `pihole -up` instead of an undocumented and unsupported command line?

o5ky:
I mean that is how to repair and it's a supplied script, it's sometimes required when the system has failed and requires the repair switch.

PR #3191, opened 2020-03-03
unit test for umask problems in #3177 and #2730

pvogt09:

---
Add a unit test for ensuring correct file permissions on `$webroot` files and possibly other important files like requested in #3177.

Adds a test case to test/test_automated_install.py which checks for read (and execute) permissions for folders `/var/www/html/{admin, pihole}` and the blocking page files.

I am not very experienced in python or its unit test framework, so any help with this PR is really appreciated.

None


pvogt09:
@dschaper: It seems like the `sudo` command is not available in the docker image, is there any other way to check for permissions for a different user?

dschaper:
What are you trying to check? (Example of what you're looking for, what is a green and what is a red result would help.)

test/test_automated_install.py
```diff
@@ -293,6 +294,79 @@ def test_installPiholeWeb_fresh_install_no_errors(Pihole):
...
+    # check directories above $webroot for read and execute permission by www-data
+    check_var = 'which sudo'
+    print (Pihole.run(check_var).stdout)
+    check_var = 'which test'
+    print (Pihole.run(check_var).stdout)
+    check_var = 'test -r /var'
+    print (Pihole.run(check_var).stdout)
+    check_var = 'echo $(test -r /var) $?'
+    print (Pihole.run(check_var).stdout)
+    check_var = 'sudo -u www-data test -r /var'
```
dschaper:
```suggestion
    check_var = 'su  -c "test -r /var/" www-data'
```

dschaper: Something along those lines, may need more `su` flags but the man is pretty good.

```sh
root@b5297e950e96:/# su --help
Usage: su [options] [LOGIN]

Options:
  -c, --command COMMAND         pass COMMAND to the invoked shell
  -h, --help                    display this help message and exit
  -, -l, --login                make the shell a login shell
  -m, -p,
  --preserve-environment        do not reset environment variables, and
                                keep the same shell
  -s, --shell SHELL             use SHELL instead of the default in passwd
```

dschaper:
Is there a possibility to do the development on a branch in your forked repo? Every time you push to this PR and the test fails I get a flurry of notifications.

pvogt09:
Sorry for the inconvenience.
How can I activate the travis build in my fork, because I do not have an instance at hand to test it locally?
What is the correct way of setting up the pihole installation in the test to have it in a state where the files in `/var/www/html` are created, the web server user exists and possibly the blockpage can be requested via `curl` to check it for errors?

It seems like some php dependencies are missing in the debian build when installing only the web server dependencies and running the whole installation process does not work at all in the test, so I was not able to setup the test correctly to make it pass all access right checks.
The commands for checking for the correct access rights should be ok.

PromoFaux:
@pvogt09 you should be able to sign up to travis and point it at your fork, then it will (should!) run the tests on your fork.

As for running them locally, I don't really know if it's possible.

corbolais:
> What is the correct way of setting up the pihole installation in the test to have it in a state where the files in `/var/www/html` are created, the web server user exists and possibly the blockpage can be requested via `curl` to check it for errors?
> 
> It seems like some php dependencies are missing in the debian build when installing only the web server dependencies and running the whole installation process does not work at all in the test, so I was not able to setup the test correctly to make it pass all access right checks.
> The commands for checking for the correct access rights should be ok.

@pvogt09 :
I understand you need a way to make automated_install.sh create the stuff your tests are run against in a reproducible way, a way that is optimally the same as in the script itself.

Looking at the relevant source lines, I feel the urge to refactor this. Refactoring would also be the way to go to make this unit-testable, and much more convenient.

RFC:
The code "connects" some very loosely coupled things (from an installer point of view) inside one block: webserver and web UI. Together with the many conditions to be met for how to enforce permissions correctly and where to make this happen, I'd suggest to rework this part of the installer. If "easily" feasible.

pvogt09:
The part where the access rights for `/var/www/html` are set and the webserver user is added to the `pihole` group could be moved to its own function.

It seems like the docker images for centos and fedora are missing `systemd` because lighttpd can not be enabled by `enable_service` and `systemd` gives the error message:
```
systemctl status lighttpd
System has not been booted with systemd as init system (PID 1). Can\'t operate.
Failed to connect to bus: Host is down
```
and after creating `/run/systemd/system`
```
Failed to get D-Bus connection: No such file or directory
```
And `php5-intl` and some other php related packages are missing in the debian image.
Or am I doing something wrong?

pvogt09:
@dschaper: I am having a hard time figuring out how to start `lighttpd` in the test case. It does not work with `systemctl` and mocking that to run `/etc/init.d/lighhtpd` does not work because the file does not exist although it should have been installed during the package installation phase (at least that is printed in the [fedora](https://travis-ci.org/github/pvogt09/pi-hole/builds/664649537?utm_medium=notification&utm_source=github_status) test case).
Can you give me a hint on what I might be missing? The aim is to set up the webpage and then test if a string like `PHP Error(%d+):\s+fopen([^)]+):\s+failed to open stream: Permission denied in` can be found to test for [this](https://github.com/pi-hole/pi-hole/pull/2730#issuecomment-490735172]) not happening.

There might also be a general problem with the `php` dependencies in the docker containers, because I have set up test cases for all [supported](https://docs.pi-hole.net/main/prerequesites/) distributions (I can make a PR for that if needed) and debian and ubuntu are failing with missing or wrong versioned `php` installation.

dschaper:
There are no real init systems inside a Docker container. We kind of work with s6-init but that's generally not how things work. You can't look at the docker images and containers as VMs, they aren't made for that kind of thing. They are process containers, so I don't know if what you are trying to do is going to be actually possible. 

You'll notice a lot of the commands are mocked along with their responses due to the nature of how docker images function.

pvogt09:
With the current implementation of the test it works to start lighttpd but sometimes it fails for some unknown reason and the test times out, so I turned the webpage test off.

Another problem with the current implementation is that FTL can not be started and therefore pi.hole can not be resolved, maybe also a running `resolved` is missing because I did not mock that.

The third problem is that for Debian based images the `php` dependencies can not be installed, so the test always fails for Debian with:
```
debconf: delaying package configuration, since apt-utils is not installed
E: Unable to locate package php5-xml
E: Unable to locate package php-intl
```
How should that be handled?

Maybe someone can look it through and give some hints on how to fix that or if it is enough to test file permissions instead of the web page and not even start the web server and also which files are missing.

Also setting permissions of `/var/cache` seems to be necessary if it did not exist before because during debugging I ran into the problem that `lighttpd` was not able to read the directories below and that is missing in the installer script and might also be the reason for the problem in [this](https://discourse.pi-hole.net/t/lighttpd-permission-denied-var-cache-lighttpd-compress/20553) thread.

How do the braces have to be indented to pass the stickler test?

pvogt09:
In the current version of the test starting lighttpd works, but only for IP 127.0.0.1 and not for domain `pi.hole` because there is still some error when starting FTL:
```
/opt/pihole/updatecheck.sh: line 77: /usr/bin/pihole-FTL: Operation not permitted
/opt/pihole/updatecheck.sh: line 91: /usr/bin/pihole-FTL: Operation not permitted
/opt/pihole/version.sh: line 19: /usr/bin/pihole-FTL: Operation not permitted
```
Also #3311 is obsolete with the latest changes.

pvogt09:
@DL6ER: I am not able to start FTL in the travis-ci build, can you give me some advice on what I am doing wrong?
The binary `pihole-FTL` is executable by every user but does not start with `Operation not permitted` and exit status 126.
`/var/log/pihole.log` is empty after installation.
`ls -la $(which pihole-FTL)` returns `-rwxr-xr-x 1 root root 9252280 May 20 23:57 /usr/bin/pihole-FTL`
A problem might be `Current Pi-hole version is in detached HEAD state at v4.2.2-440-g2828698` because I assume that this should be 5.0.
You can find the failed test at https://travis-ci.org/github/pvogt09/pi-hole/builds/689467390.

Not being able to start FTL also prevents some files from being created and consequently from checking their access rights.

DL6ER:
@pvogt09 I don't think this error is issued by FTL itself, I suspect that it is generated by `su` or whatever is trying to start FTL. One pointer to this is that FTL also does not use any other return code besides 1 (error) or 0 (success). Another one is that there is not a single line of FTL output in your Travis build.

One further idea might be to check if the immutable flag is involved here or if the file system is read-only? I'm only mobile so I cannot really test things, so I'm more thinking out aloud hoping this will trigger finding the solution.

pvogt09:
The problem was that `--add-cap=NET_ADMIN` was missing in the docker command options, I suppose.
`pihole-FTL test` now gives:
```
May 26 23:06:19 dnsmasq[9621]: started, version pi-hole-2.81 cachesize 10000
May 26 23:06:19 dnsmasq[9621]: compile time options: IPv6 GNU-getopt no-DBus no-UBus no-i18n no-IDN DHCP DHCPv6 no-Lua TFTP no-conntrack ipset auth DNSSEC loop-detect inotify dumpfile
May 26 23:06:19 dnsmasq[9621]: warning: interface eth99 does not currently exist
May 26 23:06:19 dnsmasq[9621]: using only locally-known addresses for domain use-application-dns.net
May 26 23:06:19 dnsmasq[9621]: using nameserver 4.2.2.2#53
May 26 23:06:19 dnsmasq[9621]: using nameserver 4.2.2.1#53
May 26 23:06:19 dnsmasq[9621]: read /etc/hosts - 7 addresses
May 26 23:06:19 dnsmasq[9621]: failed to load names from /etc/pihole/custom.list: No such file or directory
May 26 23:06:19 dnsmasq[9621]: failed to load names from /etc/pihole/local.list: No such file or directory
May 26 23:06:21 dnsmasq[9621]: read /etc/hosts - 7 addresses
May 26 23:06:21 dnsmasq[9621]: failed to load names from /etc/pihole/custom.list: No such file or directory
May 26 23:06:21 dnsmasq[9621]: read /etc/pihole/local.list - 4 addresses
May 26 23:06:21 dnsmasq[9621]: read /etc/hosts - 7 addresses
May 26 23:06:21 dnsmasq[9621]: failed to load names from /etc/pihole/custom.list: No such file or directory
May 26 23:06:21 dnsmasq[9621]: read /etc/pihole/local.list - 4 addresses
```
Are the missing files a problem?

I rebased the PR on the current development and it depends on #3213 to pass the test for Debian.
There are still some things to improve and some files to be checked might be missing, but I would remove the draft status, because in principle it is working.

Sometimes the build fails when too many attempts to download FTL were made. I think there was an issue somewhere, where someone proposed to make the install script work with a locally downloaded version of FTL in case no connection to github is possible, but I cannot find it anymore. That might be a way to work around this problem.

DL6ER:
Sorry for the huge delay, we're slowly trying to catch up on things after a lot came into our way...

> Are the missing files a problem?

No, this is just a warning.

> There are still some things to improve and some files to be checked might be missing, but I would remove the draft status, because in principle it is working.

Are you going to work on this more?

pvogt09:
I am willing to work on this more, but I think I need a little more input from the core developers into which direction this PR should go and what it should cover and how the problems that I already came across should/could be solved (e.g. failed tests when github cannot be reached by the travis CI or tests running for more than ten minutes (if the new est routines added by this PR are run for all the supported OSes in all supported versions, the test runs for more than 45 minutes)). One question for example would be whether it is a good idea to login into the admin page and check for possible PHP errors in the output that have been reported frequently when files where missing or not readable.

pvogt09:
With the new Github Actions based test cases the problem arises that the `update_repo` function fails when pulling changes because the repository is in detached HEAD state at something like `* (HEAD detached at pull/6/merge)` and not on a valid branch. (Fails with `Could not update local repository. Contact support.`) How should this be handled? Checking for detached HEAD in `update_repo` and not pulling changes (which should not exist because the merge request are already the newest changes anyway) or overwriting `update_repo` in the test to not pull changes?

DL6ER:
@PromoFaux Can you assist with the GitHub Actions related question?

pvogt09:
I managed to get the test working with Github Actions, but that is only possible with an change in the installation code which might break things for users having a custom branch named `pull/\d+/merge` on updates.
It was also neccessary to add `--add-cap=ALL` to `docker` because FTL needs more permissions than only `NET_ADMIN`, maybe this can be reduced to something more restrictive with the help of @DL6ER.

The test on CentOS 7 fails for some reason, but I do not know why yet.

I do not know what to do about the CodefFactor warnings that are not related to warnings of type fixme.

automated install/basic-install.sh
```diff
@@ -473,7 +473,10 @@ update_repo() {
     git stash --all --quiet &> /dev/null || true # Okay for stash failure
     git clean --quiet --force -d || true # Okay for already clean directory
     # Pull the latest commits
-    git pull --quiet &> /dev/null || return $?
+    if ! git branch | grep --quiet -E '^[*] [(]HEAD detached at pull/[[:digit:]]+/merge[)]$'; then
```
PromoFaux: Just for the hard of thinking (me!) Can you explain the pattern that is being matched here, please?

PromoFaux:
I should check my pings more often, although looks like @pvogt09 has worked around the GHA issues. Just the stickler/codefactor stuff to address now

automated install/basic-install.sh
pvogt09: It is supposed to match `* (HEAD detached at pull/#/merge)` from the start to the end of the line with # being the number (as a free amount of digits, but at least one) of the pull request the test is run for. The `[` and `]` are supposed to create character classes and therefore quote `*`, `(` and `)` in the regex, but there might be a better way to do this.

pvogt09:
I have removed the TODOs and stickler errors and made everything compatible with the lastest development branch and as far as I am concerned, this PR can be merged now after a review by the core reviewers.
The CodeFactor error does not stem from a part of the code this PR touches.

pvogt09:
I have removed the TODOs and stickler errors and made everything compatible with the lastest development branch and as far as I am concerned, this PR can be merged now after a review by the core reviewers.
The CodeFactor error does not stem from a part of the code this PR touches.

PromoFaux:
> The CodeFactor error does not stem from a part of the code this PR touches.

Looks like it had glitched. closing/reopening the PR put the Codefactor message in the right place (it is indeed correct - the new test is a complex method)


Just for sanity's sake - I've merged in latest changes from `development` to make sure things still work. I tried on my own fork to find a way around having the workaround for the repo being in a detached HEAD state, but it's 02:41 and I'm out of ideas...

PromoFaux:
Did a little more research on this, and hopefully have been able to remove the workaround for detached HEAD and handle it in the action script. I feel more comfortable doing it there than in basic-install.sh... lets see what the test run says!

PromoFaux (approved):
Sorry, stepped on this PR a bit, but I still wasn't happy with changes being made in the code to workaround a problem in the tests.

Have added a new function to the test suite which allows us to override specific sub-commands (e.g `git pull`, while leaving `git clone` alone) - this further allows for the us to ignore the fact that the test repo is in a detatched HEAD state.

PromoFaux (approved):

PromoFaux:
Squashmerged

Issue #3196, opened 2020-03-07
If a regexp blocks a site where i need to download lists, PI-Hole doesnt shows the reason.


stefanostatuti:
**In raising this issue, I confirm the following:** `{please fill the checkboxes, e.g: [X]}`

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the the source code relevant to this issue?:**
5
`{Replace this with a number from 1 to 10. 1 being not familiar, and 10 being very familiar}`

---
**Expected behaviour:**
the console shows the correct reason about the block.

   [i] Target: blocklist.site (tracking)
  [✗] Status: blocklist.site is blocked by **regex/wildcard**. Using DNS on 208.67.222.222 to download https://blocklist.site/app/dl/tracking
  [✓] Status: No changes detected

**Actual behaviour:**
no info about the reason for the block, i think is logged like a NULL string.
 [i] Target: blocklist.site (tracking)
  [✗] Status: blocklist.site is blocked by . Using DNS on 208.67.222.222 to download https://blocklist.site/app/dl/tracking
  [✓] Status: No changes detected

**Steps to reproduce:**
add a regEXP to block .site domains like this: 
.site$
now try to download updates 
in query log tab you can see: blocklist.site  -->Blocked (regex/wildcard)

**Debug token provided by [uploading `pihole -d` log](https://discourse.pi-hole.net/t/the-pihole-command-with-examples/738#debug):**

https://tricorder.pi-hole.net/0aiiamtkks
`{Alphanumeric token}`

**Troubleshooting undert

stefanostatuti:
ok, maybe i found the problem.
gravity.sh lines 271-272

 bad_list=$(pihole -q **-adlist** "${domain}" | head -n1 | awk -F 'Match found in ' '{print $2}')
    echo -e "${OVER}  ${CROSS} ${str} ${domain} is blocked by ${bad_list%:}. Using DNS on ${PIHOLE_DNS_1} to download ${url}";

[...]

PR #3198, opened 2020-03-07
added check for https://github.com/pi-hole/pi-hole/issues/3196

stefanostatuti:

- [X ] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md), as well as this entire template.
- [ X] I have commented my proposed changes within the code.
- [ X] I have tested my proposed changes, and have included unit tests where possible.
- [X ] I am willing to help maintain this change if there are issues with it later.
- [X ] I give this submission freely and claim no ownership.
- [X ] It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
- [X ] I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

---
https://github.com/pi-hole/pi-hole/issues/3196

https://github.com/pi-hole/pi-hole/issues/3196#issuecomment-596128381

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -270,7 +270,7 @@ gravity_DownloadBlocklistFromUrl() {
     fi
     bad_list=$(pihole -q -adlist "${domain}" | head -n1 | awk -F 'Match found in ' '{print $2}')
     # added check for https://github.com/pi-hole/pi-hole/issues/3196
-    if [-z "${bad_list}" ]; then
+    if [ -z "${bad_list}" ]; then
 		  bad_list='a Regular Expression'
     fi    
     echo -e "${OVER}  ${CROSS} ${str} ${domain} is blocked by ${bad_list%:}. Using DNS on ${PIHOLE_DNS_1} to download ${url}";
--- a/gravity.sh
+++ b/gravity.sh
@@ -264,7 +264,7 @@ gravity_DownloadBlocklistFromUrl() {
         printf -v port "%s" "${PIHOLE_DNS_1#*#}"
     fi
     ip=$(dig "@${ip_addr}" -p "${port}" +short "${domain}" | tail -1)
-    if [[ $(echo "${url}" | awk -F '://' '{print $1}') = "https" ]]; then 
+    if [[ $(echo "${url}" | awk -F '://' '{print $1}') = "https" ]]; then
       port=443;
     else port=80
     fi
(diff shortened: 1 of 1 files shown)


gravity.sh
```diff
@@ -269,6 +269,10 @@ gravity_DownloadBlocklistFromUrl() {
     else port=80
     fi
     bad_list=$(pihole -q -adlist "${domain}" | head -n1 | awk -F 'Match found in ' '{print $2}')
+    # added check for https://github.com/pi-hole/pi-hole/issues/3196
+    if [ $(echo -n ${bad_list} | wc -m) -eq 0 ]; then
```
dschaper: No need to `echo` and `wc`, just check the `bad_list` variable directly for content. 
```suggestion
    if [-z "${bad_list}" ]; then
```

gravity.sh
```diff
@@ -269,6 +269,10 @@ gravity_DownloadBlocklistFromUrl() {
...
+		  bad_list='a Regular Expression'
```
dschaper: Looks okay, but not sure if an empty `bad_list` is always a regex.

dschaper:
And please see https://github.com/pi-hole/pi-hole/pull/3198/checks?check_run_id=492564435 for the DCO check failure.

dschaper:
Thanks for the submission. Two things:

1.  This PR is based on `master` and that's not our process. This needs to be based on `development` per the Contributor's Guide instructions.
2.  This does solve the output being blank but I think a better approach would be to fix https://github.com/pi-hole/pi-hole/blob/development/advanced/Scripts/query.sh to output the desired string when called with `-adlist`.

gravity.sh
DL6ER: I concur. If it has not been found in an blacklist, it could still have been blocked by entries in `/etc/hosts` or any `/etc/dnsmasq.d` file.

@stefanostatuti The beta version of `pihole -q` does also test regex so it is very likely that what you want to achieve is much easier basing on the most recent code instead of `master`.

stefanostatuti: hello, i still using version 4.X PiHole and for now i don't want upgrade in 5.X so i can't test my code in 5.X branch.
Yes, I agree with You and it's better to fix the gravity.sh from 4.x Branch, ill try to fixit

dschaper: There will be no further v4 releases. The https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md is specific in the requirements for PRs to be approved.

DL6ER:
This change is superseded by `pihole -q` explicitly checking regex filters in Pi-hole v5.x

PR #3236, opened 2020-03-31
Remove hosts-file.net from default lists

PromoFaux:

---

Removes the `hosts-file.net` list from the defaults, as it is no longer active:

https://forums.malwarebytes.com/topic/257401-inquiry-regarding-automated-processing-of-hosts-files/

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1212,8 +1212,7 @@ chooseBlocklists() {
         MalwareDom "MalwareDomains" on
         Cameleon "Cameleon" on
         DisconTrack "Disconnect.me Tracking" on
-        DisconAd "Disconnect.me Ads" on
-        HostsFile "Hosts-file.net Ads" on)
+        DisconAd "Disconnect.me Ads" on)
 
     # In a variable, show the choices available; exit if Cancel is selected
     choices=$("${cmd[@]}" "${options[@]}" 2>&1 >/dev/tty) || { printf "  %bCancel was selected, exiting installer%b\\n" "${COL_LIGHT_RED}" "${COL_NC}"; rm "${adlistFile}" ;exit 1; }
@@ -1235,7 +1234,6 @@ appendToListsFile() {
         Cameleon     )  echo "https://sysctl.org/cameleon/hosts" >> "${adlistFile}";;
         DisconTrack  )  echo "https://s3.amazonaws.com/lists.disconnect.me/simple_tracking.txt" >> "${adlistFile}";;
         DisconAd     )  echo "https://s3.amazonaws.com/lists.disconnect.me/simple_ad.txt" >> "${adlistFile}";;
-        HostsFile    )  echo "https://hosts-file.net/ad_servers.txt" >> "${adlistFile}";;
     esac
 }
 


DL6ER:
Do you have any numbers at hand how many unique (!) domains this removes? The old list is still there, just not being updated following the statement on the forum.

PromoFaux:
It 404's

ref: https://www.reddit.com/r/pihole/comments/fsg11e/hostsfilenet/

jfb-pihole:
From an older gravity update log:
```
  [i] Target: https://hosts-file.net/ad_servers.txt
  [✓] Status: No changes detected
  [i] Received 45736 domains
```

XhmikosR:
Would it make any sense to mirror the list ourselves?

PromoFaux:
Nah, we'd like to stay away from actually hosting lists (even if they are mirrors), as we don't want anyone to have the impression that we are deciding what should and should not be blocked.

Ultimately it is down to the user.

It's a conversation we've had loads over the years. If don't include default blocklists, then some users will complain that it doesn't block anything out of the box, if we do, then some users will complain that we are telling them what should and shouldn't be blocked.

In an effort to please both camps, we made the "default" lists optional from the get-go, so it is up to the user to decide if they want, or do not want, to use them!

pralor:

https://discourse.pi-hole.net/t/is-there-a-problem-with-one-of-the-list-on-pi-hole/30007/2

pralor:

https://discourse.pi-hole.net/t/ftl-offline-and-lost-connection-to-api/30025/4

lightswitch05:
Just out of curiosity I did a diff between the default lists and my last cached version of the ATS list. I got 36,715 hosts that were on the ATS list that are not currently on a default list.

pralor:

https://discourse.pi-hole.net/t/updating-gravity-fails-hosts-file-net/30140/2

thericeking:
Has it been decided which list is now going to replace the old https://hosts-file.net/ad_servers.txt ?
I don't like adding lists that aren't part of the default Pi-hole list because then stuff starts not working.

PromoFaux:
Nothing at the moment. The "default" lists were chosen historically as ones that we had used that didn't cause too many issues, but it is difficult to find "neutral" replacements.

The control of which lists are on an install is always ultimately down to the user. The only reason we include _any_ by "default" (quotes because they're still optional) is because it lowers the barrier to entry for those people that are new to things like this. If there were none, we may get lots of issue tickets saying "My Pi-hole isn't working, I still see ads", only to discover that they have not added any blocklists to their installs.

pralor:

https://discourse.pi-hole.net/t/pihole-5-beta-import-von-einstellungen/31214/5

pralor:

https://discourse.pi-hole.net/t/https-hosts-file-net-ad-servers-txt-down/32718/2

Shadok:
If it helps, the list has been rehosted and is maintained here :
https://raw.githubusercontent.com/evankrob/hosts-filenetrehost/master/ad_servers.txt
or by https://hostsfile.tk : https://hostsfile.tk/ad_servers.txt

yubiuser:
> maintained

Just cloned or actively maintained? It doesn't state when it was updated last. The original list was quite outdated...

Shadok:
hostsfile.tk must just hosts it. evankrob says it will maintain it but for now, there's no change.
It may be better to stop using it indeed.

philshem:
Another option is to use the archive.org snapshot from April?

https://web.archive.org/web/20200308144426/http://hosts-file.net/ad_servers.txt

TheOtherDoctor:
So I understand the status as a current 'pihole -up' making a coffee break at this missing file just due to this change not yet being in the current 5.0 release package?

PromoFaux:
This change is in the 5.0 release. If you install from scratch as a fresh installation then this list will not be included as an option.

If you had it in your previous list, we do not remove it. In this case the list was removed because it is no longer available, but sometimes we might remove a list because we no longer wish to include it in the out of the box experience, but it is otherwise a list a user wishes to use.

From my above comment:

>The control of which lists are on an install is always ultimately down to the user. The only reason we include any by "default" (quotes because they're still optional) is because it lowers the barrier to entry for those people that are new to things like this. If there were none, we may get lots of issue tickets saying "My Pi-hole isn't working, I still see ads", only to discover that they have not added any blocklists to their installs.

Can you explain what you mean by "making a coffee break at this missing file" means? All that should happen if this list is in your adlists table is that you will see a message saying that the list can not be downloaded (i.e it has 404'd)

TheOtherDoctor:
Ok, got it, deactivated it in my web frontend. Maybe that prevents the download attempt next update. I understand (and agree to) the logic of keeping a previously config'd list; just wondered about the "no cached list available" note, as this seems like there has been no previously successful download.

With the coffee break I simply meant the concole output did not move for quite some time (subjectively - did not check objective time); long enough that I worried that the process died. 
Then the output with the red "X" and "Status: Connection refused" note appeared and the upgrade went on smoothly. As it has been major version increase, I might have been just too impatient.

No big thing, just created confusion, and when looking into the reason I simply wondered why a 404 cannot be handled in no time.

PR #3238, opened 2020-03-31
Validate URL by checking against invalid characters.

DL6ER:
Internal


advanced/Scripts/webpage.sh
```diff
@@ -402,23 +402,25 @@ SetWebUILayout() {
...
+    else
+        return 0
+    fi
 }
 
 CustomizeAdLists() {
     local address
     address="${args[3]}"
     local comment
-    comment="${args[4]}"]
```
DL6ER: This also fixes a small typo here

PR #3254, opened 2020-04-05
 Sanitize email address in case of security issues

bash-c:
**By submitting this pull request, I confirm the following:** 
*please fill any appropriate checkboxes, e.g: [X]*

- [√ ] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md), as well as this entire template.
- [√ ] I have made only one major change in my proposed changes.
- [√ ] I have commented my proposed changes within the code.
- [√ ] I have tested my proposed changes, and have included unit tests where possible.
- [√ ] I am willing to help maintain this change if there are issues with it later.
- [√ ] I give this submission freely and claim no ownership.
- [√ ] It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
- [√ ] I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

Please make sure you [Sign Off](https://github.com/pi-hole/pi-hole/wiki/How-to-signoff-your-commits.) all commits. Pi-hole enforces the [DCO](https://github.com/pi-hole/pi-hole/wiki/Contributing-to-the-project).

---
**What does this PR aim to accomplish?:**
*A detailed description, screenshots (if necessary), as well as links to any relevant GitHub issues*
- To sanitize email address before saving it.

**How does this PR accomplish the above?:**
*A detailed description (such as a changelog) and screenshots (if necessary) of the implemented fix*

**What documentation changes (if any) are needed to support this PR?:**
*A detailed list of any necessary changes*
- None, I think

---
* You must follow the template instructions. Failure to do so will result in your pull request being closed.
* Please respect that Pi-hole is developed by volunteers, who can only reply in their spare time.

PR #3255, opened 2020-04-05
Enhanced email validation regex

PromoFaux:

---

Improves upon #3254

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -519,7 +519,9 @@ Options:
     if [[ -n "${args[2]}" ]]; then
 
         # Sanitize email address in case of security issues
-        if [[ ! "${args[2]}" =~ ^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,4}$  ]]; then
+        local regex
+        regex="^[a-z0-9!#\$%&'*+/=?^_\`{|}~-]+(\.[a-z0-9!#$%&'*+/=?^_\`{|}~-]+)*@([a-z0-9]([a-z0-9-]*[a-z0-9])?\.)+[a-z0-9]([a-z0-9-]*[a-z0-9])?\$"
+        if [[ ! "${args[2]}" =~ ${regex}  ]]; then
             echo -e "  ${CROSS} Invalid email address"
             exit 0
         fi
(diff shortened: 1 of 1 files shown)


PromoFaux:
@bash-c / @pvogt09 - please try this branch

bash-c:
I tested and confirm it working

advanced/Scripts/webpage.sh
```diff
@@ -519,7 +519,10 @@ Options:
     if [[ -n "${args[2]}" ]]; then
 
         # Sanitize email address in case of security issues
-        if [[ ! "${args[2]}" =~ ^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,4}$  ]]; then
+        # Regex from https://stackoverflow.com/a/2138832/4065967
+        local regex
+        regex="^[a-z0-9!#\$%&'*+/=?^_\`{|}~-]+(\.[a-z0-9!#$%&'*+/=?^_\`{|}~-]+)*@([a-z0-9]([a-z0-9-]*[a-z0-9])?\.)+[a-z0-9]([a-z0-9-]*[a-z0-9])?\$"
```
DL6ER, line 524: Looking at this, it seems to allow a lot of dangerous things, like `!` or `=` or even `?` and `$` or `/`. Do we really expect these characters to show up in email addresses?... Especially the exclamation mark may give rise to unpleasant behavior in `bash`!

DL6ER: The regex on the SE page you linked goes down all the way into RFC 5322

However, there is also the comment:
> RFC 5322 leaves the domain name part open to implementation-specific choices that won’t work on the Internet today.

DL6ER: Thinking again about it, we could maybe even keep the existing Regex and just remove the maximum length limitation of `4`? We do not enforce any restrictions on the former elements of the address, either.

PromoFaux: Whoops, you are right. This is why all regex I ever submit needs more pairs of eyes on it...

image

bash-c:
Since it's a security issue and everyone can see this pr. I think it's better to settle it down early.

advanced/Scripts/webpage.sh
DL6ER: My proposal: Restore the previous regex and simply remove the maximum TLD length of four restriction:

```suggestion
        regex="^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,}\$"
```

PR #3260, opened 2020-04-09
Improve conditional forwarding

DL6ER:

---

Improve conditional forwarding

Necessary backend changes with automatic conversion of existing legavy settings into the new format. More details may be found in https://github.com/pi-hole/AdminLTE/pull/1208

None


advanced/Scripts/webpage.sh
```diff
@@ -210,8 +210,42 @@ trust-anchor=.,20326,8,2,E06D44B80B8F1D39A95C0B0D7C65D08458E880409BBC68345710423
...
+        if [[ "${REV_SERVER_TARGET}" =~ 10\..* ]]; then
+            # Private network, Class A (RFC 1597 + RFC 1918)
+            REV_SERVER_CIDR="10.0.0.0/8"
+        elif [[ "${REV_SERVER_TARGET}" =~ 192\.168\..* ]]; then
+            # Private network, Class B (RFC 1597 + RFC 1918)
+            REV_SERVER_CIDR="192.168.0.0/16"
+        else
+            # Something else, convert to /24 subnet (preserves legacy behavior)
+            # This sed converts "192.168.1.2" to "192.168.1.0/24"
+            REV_SERVER_CIDR="$(sed "s+\\.[0-9]*$+\\.0/24+" <<< "${REV_SERVER_TARGET}")"
```
dschaper, line 239: I wouldn't even try detecting the intended range. If there's a requirement for CIDR notation then we should error and notify users that they need to include it.

PromoFaux:
What is the status on this please?

advanced/Scripts/webpage.sh
DL6ER: This code is for upgrading the old way of using conditional forwarding to the new way. It is preserving exactly what was used before this update (at least the `else` does that). It will be necessary for an easy/automated update procedure.

DL6ER:
@PromoFaux Merged in the most recent changes, resolved all conflicts. I think this is good to go.

advanced/Scripts/webpage.sh
dschaper: This feels like more handholding when there doesn't need to be. CIDR is the antithesis of natural masks. This would give a `/16` to 192.168.1.0, yes? And it's missing 172.16.0.0. I think we're better off educating users to use the right masking instead of guessing what they may intend and getting it wrong in a lot of cases. Either way we are going to have support questions.

DL6ER: We are currently *implicitly* adding `/24` *always*. This code above is for trying to improve in a few cases. I can remove the `if` and `elif` code and only leave the `else` condition code in there. This will preserve the current behavior without introducing any changes. Is this what you're suggesting?

dschaper: If the `elif` would add `/16` to someone entering `192.168.15.1` then I think it's going to cause a lot of issues. `192.168.0.0` may be Class B but no one uses anything more than a Class C /24 subnet in that space. And really, nobody uses Ten Net as a full class A.

So yes, I would prefer to not change the behavior, it's relatively sane to assume `/24` for the majority of our users, though that's more handholding than I'd like.

DL6ER: @dschaper Handholding removed. The code now simply preserves pre-v5.1 behavior.

pralor:

PR #3263, opened 2020-04-10
Fixes broken blocking page and landing page when changing server port or host name (v5.0)

mschoettle:

---

Addresses issues #2195 and #2720 to allow changing the server port and/or host name.

* `$_SERVER["SERVER_NAME"]` used instead of `$_SERVER["HOST"] (host contains both host name and port causing the query ads script to result in an invalid domain error)
* uses the `VIRTUAL_HOST` if set in the environment to determine whether to redirect to the admin page (in addition to `pi.hole`). This can be set with the following line in `/etc/lighttpd/external.conf`:
```
setenv.add-environment = ("virtual_host" => "your.fqdn.tld")
```
* it is still assumed that lighttpd runs on localhost but the call to *queryAds* now includes the server port
* replaced use of `pi.hole` prefix in path for assets in the html of the blocking page

Document setting the `VIRTUAL_HOST` in the environment when using a host name other than `pi.hole`.

---
*Note:* This is PR #3067 but against `release/v5.0` to land in version 5 ([as suggested here](https://github.com/pi-hole/pi-hole/pull/3067#issuecomment-611205268)).

### DIFF
--- a/advanced/index.php
+++ b/advanced/index.php
@@ -6,8 +6,8 @@
 *  This file is copyright under the latest version of the EUPL.
 *  Please see LICENSE file for your rights under this license. */
 
-// Sanitise HTTP_HOST output
-$serverName = htmlspecialchars($_SERVER["HTTP_HOST"]);
+// Sanitize SERVER_NAME output
+$serverName = htmlspecialchars($_SERVER["SERVER_NAME"]);
 // Remove external ipv6 brackets if any
 $serverName = preg_replace('/^\[(.*)\]$/', '${1}', $serverName);
 
@@ -50,16 +50,24 @@ function setHeader($type = "x") {
 }
 
 // Determine block page type
-if ($serverName === "pi.hole") {
+if ($serverName === "pi.hole"
+    || (!empty($_SERVER["VIRTUAL_HOST"]) && $serverName === $_SERVER["VIRTUAL_HOST"])) {
     // Redirect to Web Interface
     exit(header("Location: /admin"));
 } elseif (filter_var($serverName, FILTER_VALIDATE_IP) || in_array($serverName, $authorizedHosts)) {
     // Set Splash Page output
     $splashPage = "
-    <html><head>
+    <html>
+      <head>
         $viewPort
-        <link rel='stylesheet' href='/pihole/blockingpage.css' type='text/css'/>
-    </head><body id='splashpage'><img src='/admin/img/logo.svg'/><br/>Pi-<b>hole</b>: Your black hole for Internet advertisements<br><a href='/admin'>Did you mean to go to the admin panel?</a></body></html>
+        <link rel='stylesheet' href='pihole/blockingpage.css' type='text/css'/>
+      </head>
+      <body id='splashpage'>
+        <img src='admin/img/logo.svg'/><br/>
+        Pi-<b>hole</b>: Your black hole for Internet advertisements<br/>
+        <a href='/admin'>Did you mean to go to the admin panel?</a>
+      </body>
+    </html>
     ";
 
     // Set splash/landing page based off presence of $landPage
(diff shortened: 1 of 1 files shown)


angrynerd:
Would this mess anything up if you were using apache virtual hosting
instead of lighttpd?

dschaper:
@mschoettle Shoot me an email at dan.schaper@pi-hole.net. I'll get some swag out to you.

PromoFaux:
@angrynerd, unsure to be honest. As it stands, `lighttpd` is the only officially supported webserver with Pi-hole, and whilst it may work with others, we have not tested such a configuration.

There are some changes slated for v6.0 that will make this all a moot point, in any case :)

mschoettle:
@angrynerd, if you are not changing the default hostname (`pi.hole`) it should still work. Otherwise, I believe you can use `SetEnv` to set `virtual_host` to your hostname in your virtual host config.

RenWal:
This PR breaks the Pi-Hole block page on nginx on the Pi and on UbuntuServer Bionic when following this tutorial: https://docs.pi-hole.net/guides/nginx-configuration/

nginx using the distro's defaults is configured to put the statically configured server name (`_` in the tutorial) in `S_SERVER["SERVER_NAME"]`, not the domain which is passed via `S_SERVER["HTTP_HOST"]`.

This will cause the splash page to never show because the domain is always seen as `_` and not whatever you type in the browser, e.g. `pi.hole`.

dschaper:
As the guide linked is marked as "This is an unsupported configuration created by the community", the best option would be to submit an update to the documentation.

mschoettle:
I would suggest to set a proper server name in your config. If you set it to the host name that your pi hole is reachable it should work.

RenWal:
By that you mean changing the _server_name_ directive or overriding the default value that nginx loads into `$_SERVER["SERVER_NAME"]`? The former will cause the splash page to _always_ show.

mschoettle:
Yes, maybe it (`server_name`) can also be removed. In addition, you need to set `VIRTUAL_HOST` in the environment to the host name. This enables the automatic redirect from `your.fqdn.tld` to `your.fqdn.tld/admin`.

RenWal:
I followed your suggestion:

1. define `server_name adblock.fritz.box`
2. define `fastcgi_param VIRTUAL_HOST $server_name` in the `\.php$` location block

This gets it close to working, but as I mentioned, it also causes Pi-hole to try displaying the admin page for every request, leading to a `Failed Host Check` message (because in nginx, `SERVER_NAME` is more equivalent to `VIRTUAL_HOST` than `HTTP_HOST`). To solve this, I had to also override `fastcgi_param SERVER_NAME $host`.

Would this be an appropriate update to submit to the community documentation?

dschaper:
> Would this be an appropriate update to submit to the community documentation?

https://github.com/pi-hole/docs/blob/master/docs/guides/nginx-configuration.md

advanced/index.php
```diff
@@ -50,16 +50,24 @@ function setHeader($type = "x") {
...
-    <html><head>
+    <html>
+      <head>
         $viewPort
-        <link rel='stylesheet' href='/pihole/blockingpage.css' type='text/css'/>
-    </head><body id='splashpage'><img src='/admin/img/logo.svg'/><br/>Pi-<b>hole</b>: Your black hole for Internet advertisements<br><a href='/admin'>Did you mean to go to the admin panel?</a></body></html>
+        <link rel='stylesheet' href='pihole/blockingpage.css' type='text/css'/>
+      </head>
+      <body id='splashpage'>
+        <img src='admin/img/logo.svg'/><br/>
```
MichaIng, line 66: @mschoettle 
I just recognised that the blocking/404 page is broken when accessing a non-existent or blocked URL with a sub directory path, since the absolute asset paths have become relative with this commit. The client is then looking and failing for things like `http://blocked.domain.org/something/pihole/blockingpage.css`

Is there a reason why those need to be relative? Probably not that relevant with the planned removal of the blocking page, but the remaining 404 handler page still contains relative assets starting with our commit: https://github.com/pi-hole/pi-hole/pull/3910/files#diff-90be1003be891f4d6008bf3a167ceedcbc3e06dbf48e490c6bc761eef20ea96bR51

mschoettle: There was a discussion in #3067 about relative paths and keeping it consistent. I guess because the block page is rarely seen (I have never seen it yet) this was not caught at the time. In this particular case it would make sense to not use relative URLs then.

PR #3271, opened 2020-04-13
Add branch name to pihole -v

PromoFaux:

---

Like so:

```
root@raspberrypi:/etc/.pihole# pihole -v
  Pi-hole version is v4.3.5-460-g413fa94 (Latest: v4.4) Branch: release/v5.0
  AdminLTE version is v4.3.2-462-g8e51f754 (Latest: v4.3.3) Branch: release/v5.0
  FTL version is vDev-853261b (Latest: v4.3.1) Branch: release/v5.0
```

### DIFF
--- a/advanced/Scripts/version.sh
+++ b/advanced/Scripts/version.sh
@@ -97,28 +97,52 @@ getRemoteVersion(){
     return 0
 }
 
+getLocalBranch(){
+     # Local FTL btranch is stored in /etc/pihole/ftlbranch
+    if [[ "$1" == "FTL" ]]; then
+        cat /etc/pihole/ftlbranch
+        return 0
+    fi
+
+    # Get the checked out branch of the local directory
+    local directory="${1}"
+    local branch
+
+    cd "${directory}" 2> /dev/null || { echo "${DEFAULT}"; return 1; }
+    branch=$(git rev-parse --abbrev-ref HEAD || echo "$DEFAULT")
+    if [[ ! "${branch}" =~ ^v ]]; then
+        echo "${branch}"
+    elif [[ "${branch}" == "${DEFAULT}" ]]; then
+        echo "ERROR"
+        return 1
+    else
+        echo "No Checked out branch"
+    fi
+    return 0
+}
+
 versionOutput() {
     [[ "$1" == "pi-hole" ]] && GITDIR=$COREGITDIR
     [[ "$1" == "AdminLTE" ]] && GITDIR=$WEBGITDIR
     [[ "$1" == "FTL" ]] && GITDIR="FTL"
 
-    [[ "$2" == "-c" ]] || [[ "$2" == "--current" ]] || [[ -z "$2" ]] && current=$(getLocalVersion $GITDIR)
+    [[ "$2" == "-c" ]] || [[ "$2" == "--current" ]] || [[ -z "$2" ]] && current=$(getLocalVersion $GITDIR) && branch=$(getLocalBranch $GITDIR)
     [[ "$2" == "-l" ]] || [[ "$2" == "--latest" ]] || [[ -z "$2" ]] && latest=$(getRemoteVersion "$1")
     if [[ "$2" == "-h" ]] || [[ "$2" == "--hash" ]]; then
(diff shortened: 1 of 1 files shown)


dschaper:
Do we need to output if the branch is master? 

Would this look better?
`Pi-hole version is v4.3.5-460-g413fa94 release/v5.0 (Latest: v4.4)`
`Pi-hole version is v4.3.5 master (Latest: v4.4)`

I think there's a few things in the new function that could be tweaked, I'll push some bash stuff in a bit.

PromoFaux:
Funnily enough i did it that way around the first time. Then changed my mind, but on reflection I think it looks better to have it directly after the version.

Unsure about `master` or not...

dschaper:
Trust your instincts... 

I'm half and half with master. Dom can break the tie.

PromoFaux:
On reflection, I think branch first, then versions:

```
Pi-hole version is release/v5.0 v4.3.5-460-g413fa94 (Latest: v4.4)
AdminLTE version is release/v5.0 v4.3.2-462-g8e51f754 (Latest: v4.3.3)
FTL version is release/v5.0 vDev-853261b (Latest: v4.3.1)
```

advanced/Scripts/version.sh
```diff
@@ -97,22 +112,48 @@ getRemoteVersion(){
...
 }
 
+getLocalBranch(){
+    # Get the checked out branch of the local directory
+    local directory="${1}"
+    local branch
+
+     # Local FTL btranch is stored in /etc/pihole/ftlbranch
+    if [[ "$1" == "FTL" ]]; then
+        branch=$(cat /etc/pihole/ftlbranch)
```
DL6ER, line 122: This file is not available if the user has never done a local checkout of a custom FTL branch, the output would be
```plain
cat: /etc/pihole/ftlbranch: No such file or directory
  Current FTL version is  v4.3.1-800-g853261bb.
```

Use
```suggestion
        branch="$(pihole-FTL branch)"
```
to get the branch of the installed FTL instance reliably (`pihole-FTL -b` is the short form but it doesn't hurt to be explicit).

advanced/Scripts/version.sh
```diff
@@ -97,22 +112,48 @@ getRemoteVersion(){
...
+        if [[ "${branch}" == "master" ]]; then
+            echo ""
+        else
+            echo "${branch} "
+        fi
+    elif [[ "${branch}" == "${DEFAULT}" ]]; then
+        echo "ERROR"
+        return 1
+    else
+        echo "No Checked out branch"
```
DL6ER, line 137: Could this result in
```
Pi-hole version is No Checked out branchv4.3.5-460-g413fa94 (Latest: v4.4)
```
(also note the missing space between branch and v4.3.5)?

I'm not sure how/if we could reach this `else` condition.

advanced/Scripts/version.sh
DL6ER: I pushed this change already, feel free to resolve this conversation if you agree.

DL6ER:
My last commit will handle detached HEAD mode as well. Example:
```
$ cd /var/www/html/admin

$ git checkout 322d9207 # <------ randomly selected commit hash
Note: checking out '322d9207'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

HEAD is now at 322d9207 Run phpcs to have Pi-hole's PHP code match the PSR-12 standard.                                                                      

$ pihole -v -a
  Current AdminLTE version is in detached HEAD state at v4.3.2-327-g322d9207.

File: advanced/Scripts/version.sh (hunk 2 above)
Comment (PromoFaux) [reply advanced/Scripts/version.sh:139]: We can probably lose the `else` section... the function was a copy-paste, and then tweak of `getLocalVersion()`

**Comment by DL6ER** [comment]
This will print
```
$ pihole -v -a
  Current AdminLTE version is release v4.3.2.
```
if the branch comes out as `v*` (even though I don't think this would happen).

PR #3276, opened 2020-04-15
- Support for Cloudflare's new DNS for Families

Mohammed-Swillam:
Signed-off-by: Mohammed Swillam <[4535397+Mohammed-Swillam@users.noreply.github.com](url)>

**By submitting this pull request, I confirm the following:** 
*please fill any appropriate checkboxes, e.g: [X]*

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md), as well as this entire template.
- [X] I have made only one major change in my proposed changes.
- [ ] I have commented my proposed changes within the code.
- [X] I have tested my proposed changes, and have included unit tests where possible.
- [X] I am willing to help maintain this change if there are issues with it later.
- [X] I give this submission freely and claim no ownership.
- [X] It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
- [X] I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

Please make sure you [Sign Off](https://github.com/pi-hole/pi-hole/wiki/How-to-signoff-your-commits.) all commits. Pi-hole enforces the [DCO](https://github.com/pi-hole/pi-hole/wiki/Contributing-to-the-project).

---
**What does this PR aim to accomplish?:**
Cloudflare has recently launched new DNS entries which aim to provide a more safe DNS service for families.
the new service consists of 2 new DNS addresses:
1- 1.1.1.2 (protection against Malware)
2- 1.1.1.3 (protection against Malware or Adult Content)

this should allow Pihole more family-friendly and allow parents to have a safer option for their families.

**How does this PR accomplish the above?:**
I've extended the DNS entries list inside the `basic-install.sh` file with the suggested entries.

**What documentation changes (if any) are needed to support this PR?:**
You may want to check the official announcement blogpost from here: https://blog.cloudflare.com/introducing-1-1-1-1-for-families/

---
* You must follow the template instructions. Failure to do so will result in your pull request being closed.
* Please respect that Pi-hole is developed by volunteers, who can only reply in their spare time.

PR #3282, opened 2020-04-18
Revert "- Support for Cloudflare's new DNS for Families"

PromoFaux:
Following internal discussion, I had forgotten a conversation about not including any additional pre-configured upstreams, as too much choice can be a bad thing.

Custom upstream still exists as an option, so users will still be able to use these servers if they so wish


dschaper:
https://github.com/pi-hole/pi-hole/pull/3276#issuecomment-615928826

pralor:

https://discourse.pi-hole.net/t/add-cloudflare-family-dns-servers-to-web-interface/30929/3

PR #3434, opened 2020-05-24
Change OS Detection in debug script

PromoFaux:

---

Currently the supported/unsupported OS detection is based on the distro name being one of 5 set values, this is fine, but then we also need to check further. For example Raspbian <9 is not supported.

This PR adds that check.  I'd like to get a similar output into the installer if it fails, however I am starting to think that maybe the installer should prompt user to run and _read_ the debug log if an update fails.  Though obviously that won't necessarily work on a fresh install.

Examples using a Debian 10 (Buster) system:
`supportedos.txt` contains:
```
Raspbian=9,10
Ubuntu=16,18
Debian=9,10
Fedora=31,32
CentOS=7,8
```
Debug output:
image

`supportedos.txt` contains:
```
Raspbian=9,10
Ubuntu=16,18
Debian=9
Fedora=31,32
CentOS=7,8
```
Debug Output:
image

`supportedos.txt` contains:
```
Raspbian=9,10
Ubuntu=16,18
Fedora=31,32
CentOS=7,8
```
Debug Output:
image

### DIFF
--- a/advanced/Scripts/piholeDebug.sh
+++ b/advanced/Scripts/piholeDebug.sh
@@ -405,7 +405,7 @@ os_check() {
     detected_os="${detected_os_pretty%% *}"
     detected_version=$(cat /etc/*release | grep VERSION_ID | cut -d '=' -f2- | tr -d '"')
 
-    mapfile -t supportedOS < <(dig +short -t txt ${remote_os_domain} | tr -d '"' | tr ' ' '\n')
+    IFS=" "; read -r -a supportedOS < <(dig +short -t txt ${remote_os_domain} | tr -d '"')
 
     for i in "${supportedOS[@]}"
     do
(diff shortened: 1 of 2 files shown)


yubiuser:
From you example
> Debian=9
Distro is supported but version is not - please upgrade

Wording doesn't make sense if version > supported version

PromoFaux:
`Debian=9`

The intention of this is not to say "Version 9 and up are supported", but rather "_only_ version 9 is supported", hence it fails the check when the detected version is `10`

Obviously in real life we support Debian 10, but this is just an demonstration of what the output would look like if the detected version is not listed.

yubiuser:
Better but not perfect ;-)
You could check if ${os_version} is bigger or smaller than the biggest/smallest supported version and change the error accordingly (not yet supported/not supported - please upgrade)

PromoFaux:
That's not too bad an idea. I'll have a think on it later

PromoFaux:
I may change up some of the logic in this PR, actually, based on #3441

pralor:

PR #3489, opened 2020-06-13
Improve IP validation function

PromoFaux:

---

This PR is two-fold:

My first aim was to fix a potential exploit reported by validating the IP used in `pihole -a setdns`. This is done by sourcing the install script in `webpage.sh` and using the `valid_ip()` function.

Secondly, a "whilst I'm in here" - I have expanded the valid_ip() function to accept custom IP addresses with a port number (i.e 1.1.1.1#5353)

Exemplary output from testing the function:

```
[adam@ADAM-PC ~]# ./valid_ip.sh 
4.2.2.2             : good
a.b.c.d             : bad
192.168.1.1         : good
0.0.0.0             : good
255.255.255.255     : good
255.255.255.256     : bad
192.168.0.1         : good
192.168.0           : bad
1234.123.123.123    : bad
192.168.0.119#5353  : good
1.1.1.1#65536       : good
1.1.1.1#65537       : bad
1.1.1.1#12345       : good
1.1.1.1#00000       : bad
1.1.1.1#0           : bad
1.1.1.1#            : bad
```

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -10,17 +10,20 @@
 # This file is copyright under the latest version of the EUPL.
 # Please see LICENSE file for your rights under this license.
 
-readonly setupVars="/etc/pihole/setupVars.conf"
 readonly dnsmasqconfig="/etc/dnsmasq.d/01-pihole.conf"
 readonly dhcpconfig="/etc/dnsmasq.d/02-pihole-dhcp.conf"
 readonly FTLconf="/etc/pihole/pihole-FTL.conf"
 # 03 -> wildcards
 readonly dhcpstaticconfig="/etc/dnsmasq.d/04-pihole-static-dhcp.conf"
-readonly PI_HOLE_BIN_DIR="/usr/local/bin"
 readonly dnscustomfile="/etc/pihole/custom.list"
 readonly dnscustomcnamefile="/etc/dnsmasq.d/05-pihole-custom-cname.conf"
 
 readonly gravityDBfile="/etc/pihole/gravity.db"
+
+# Source install script for ${setupVars}, ${PI_HOLE_BIN_DIR} and valid_ip()
+readonly PI_HOLE_FILES_DIR="/etc/.pihole"
+PH_TEST="true"
+source "${PI_HOLE_FILES_DIR}/automated install/basic-install.sh"
 
 coltable="/opt/pihole/COL_TABLE"
 if [[ -f ${coltable} ]]; then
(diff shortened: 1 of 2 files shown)


pvogt09:
Since this PR sources `automated_install.sh` for `valid_ip()` it might be worth rethinking about using this function in the validation of email addresses like proposed in #3254.

In that case a second parameter of the `valid_ip()` function would be helpful to indicate if specifying a port is allowed.

I can set up a PR if that sounds useful.

PromoFaux:
I'm not sure I follow... I'd rather not get into the territory of having one function that tried to validate either an IP or an email address... Would you mind explaining a bit more what you mean?

This should also not need to have a second parameter because the port detection is already built in

pvogt09:
I did not intend to use the function for combined IP and email validation as well.
As proposed [here](https://github.com/pi-hole/pi-hole/pull/3254#issuecomment-609503548), I would just call the `valid_ip()` function on the domain part of the email address to check if it is valid. The other parts concering the IDN domain and IPv6 can be left out from the code that I proposed in this comment because I agree that this would make it too complex.

When `valid_ip()` is called from `valid_email()` then it might be neccessary to be able to explicitly forbid the specification of a port because that would result in an invalid email address like `test@127.0.0.1#53` considered as valid. And a second parameter to indicate that ports are forbidden would make it possible to handle that case as well.

PromoFaux:
OK, I understand a bit better now what you are getting at.

To be honest, my point still stands about being overly complex for the sake of something that i'm sure is used by a very few number of users in the wild. Personally, I actually think we should rather remove the functionality to include the email address all together. As said, the only place it is used is on the block page, which is less and less relevant these days with the rise of `https`. We even set the default blocking mode for new installs to be `NULL`. For a user to enable the block page, they would need to be making manual to the configs anyway, so perhaps it is not that much of a stretch to expect them to also manually set the admin email in `setupVars.conf`

pvogt09:
OK, I am fine with that. And if the email will be removed anyway sooner or later it might not be worth the effort anyway.

DL6ER:
Also note that with Pi-hole v6.0 much of the current code will go away and implemented in the API. Validation will be easier in there plus things cannot open security holes like this as the API is written in C no there will be no interpreter getting into our way here.

DL6ER:
@PromoFaux Now that I looked at the code, I'm afraid that we have this already since a long time. Check out [IPv4 regex](https://github.com/pi-hole/AdminLTE/blob/d187219f5a901bd3a1bc9825d0a6155e4c284f59/scripts/pi-hole/js/utils.js#L131-L141) and [IPv6 regex](https://github.com/pi-hole/AdminLTE/blob/d187219f5a901bd3a1bc9825d0a6155e4c284f59/scripts/pi-hole/js/utils.js#L143-L167). The changes needed for supporting a port instead of a CIDR are marginal...

pralor:

PR #3490, opened 2020-06-13
Fix new os_check

PromoFaux:

---

The newly added os_check works fine if `dig` is already installed... I guess one thing we did not test was a fresh installation!

This PR moves the call to the  `os_check` function to _after_ the installer dependencies are installed, and also moves `bind-utils` (`yum`) and `dnsutils` (`apt`) to the installer dependencies

If we want to have this at the top of the installer, then we will not be able to use the `TXT` record on `versions.pi-hole.net`, and instead will need to `curl` the same information from somewhere else.


pralor:

PR #3524, opened 2020-07-01
Fix pihole status to not rely on a TCP port test

DL6ER:

---

The current test can fail even when there is no error i case the max. number of TCP workers is reached.

Actually test if there is a `pihole*` process running on port 53. If so, analyze if we bound to IPv4/6 and UDP/TCP. If all four sockets are accepting/listening, we report that the DNS service is running. If some are missing, we report this, too. If no `pihole*` process is listening on port 53, then this is the new (better) criterion for saying: it is not running

Example for when everything is fine:
```
$ pihole status
  [✓] DNS service is listening
  [✓] Pi-hole blocking is enabled
```

When, e.g., IPv6 is not available:
```
$ pihole status
  [✗] DNS service is partially listening
     [✓] UDP (IPv4)
     [✓] TCP (IPv4)
     [✗] UDP (IPv6)
     [✗] TCP (IPv6)

  [✓] Pi-hole blocking is enabled
```

Or, when FTL is offline:
```
$ pihole status
  [✗] DNS service is NOT listening
```

None

### DIFF
--- a/pihole
+++ b/pihole
@@ -253,35 +253,30 @@ Options:
 analyze_ports() {
   # FTL is listening at least on at least one port when this
   # function is getting called
-  if [[ $(grep -c "IPv4" <<< "${1}") -gt 1 ]] && \
-     [[ $(grep -c "IPv6" <<< "${1}") -gt 1 ]]; then
-    echo -e "  ${TICK} DNS service is listening"
+  echo -e "  ${TICK} DNS service is listening"
+  # Check individual address family/protocol combinations
+  # For a healthy Pi-hole, they should all be up (nothing printed)
+  if grep -q "IPv4.*UDP" <<< "${1}"; then
+      echo -e "     ${TICK} UDP (IPv4)"
   else
-    echo -e "  ${CROSS} DNS service is partially listening"
-    # Check individual address family/protocol combinations
-    # For a healthy Pi-hole, they should all be up (nothing printed)
-    if grep -q "IPv4.*UDP" <<< "${1}"; then
-        echo -e "     ${TICK} UDP (IPv4)"
-    else
-        echo -e "     ${CROSS} UDP (IPv4)"
-    fi
-    if grep -q "IPv4.*TCP" <<< "${1}"; then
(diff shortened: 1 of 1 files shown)


DL6ER:
This fixed the error described in https://github.com/pi-hole/FTL/issues/816#issuecomment-652285584 among probably many more seen on Discourse over time.

dschaper:
I think I'd like the expanded status to be the default status. (The second image). I'm not sure if that would cause more confusion with people thinking they need IPv6 to get a 100% A+++ score though.

DL6ER:
I added the more verbose output as default. There is no harm in getting it to listen at least on `lo ::1` which suffices to get the IP6 ticks

pihole
```diff
@@ -253,35 +253,30 @@ Options:
...
   # FTL is listening at least on at least one port when this
   # function is getting called
-  if [[ $(grep -c "IPv4" <<< "${1}") -gt 1 ]] && \
-     [[ $(grep -c "IPv6" <<< "${1}") -gt 1 ]]; then
-    echo -e "  ${TICK} DNS service is listening"
+  echo -e "  ${TICK} DNS service is listening"
+  # Check individual address family/protocol combinations
+  # For a healthy Pi-hole, they should all be up (nothing printed)
+  if grep -q "IPv4.*UDP" <<< "${1}"; then
+      echo -e "     ${TICK} UDP (IPv4)"
```
dschaper, line 260: Because whitespace is a pain to deal with, how about `printf` instead?

```suggestion
      printf "\\t%s UDP (IPv4)\\n" "${TICK}"
```

dschaper: That may need to be `%b` instead of `%s` for the tick/cross.

DL6ER: I followed what is already there, but no harm in improving things. I will check after dinner

pralor:

https://discourse.pi-hole.net/t/multiple-instances-of-pihole-ftl-running/34865/16

dschaper:
Did I forget this or did we pause it?

DL6ER:
This PR is still valid and awaiting review+approval+merge

PR #3564, opened 2020-07-16
Pi-hole Core v5.1.2

PromoFaux:
- Allow port number on IPV6 validator 643d2c0f3ea7871b64f9eaa2c9a5a2060d9ad14e (@PromoFaux)
- Fix for gravity cronjob #3599 (@DL6ER)
- Wrap OS Check function in gate and add ns1 resolver for dig #3596 (@PromoFaux)
- Add OS and version to output if unsupported OS is detected #3597 (@yubiuser)
- Report and exit if dig supportedOS returns nothing #3620 (@yubiuser)


dschaper:
Making release 5.2 from branch release 5.1.2 is going to confuse more people than just calling it release 5.1.2. (And everything added is bugfixes so there's no actual justification to using 5.2 instead.)

PromoFaux:
Agreed

dschaper:
Trying to figure out the push and PR build conditions for Travis still. https://github.com/pi-hole/.github/pull/8

automated install/basic-install.sh
```diff
@@ -1054,7 +1054,7 @@ valid_ip6() {
     # optional port number starting '#' with range of 1-65536
     local portelem="(#([1-9]|[1-8][0-9]|9[0-9]|[1-8][0-9]{2}|9[0-8][0-9]|99[0-9]|[1-8][0-9]{3}|9[0-8][0-9]{2}|99[0-8][0-9]|999[0-9]|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-6]))?"
     # build a full regex string from the above parts
-    local regex="^(((${ipv6elem}))((:${ipv6elem}))*::((${ipv6elem}))*((:${ipv6elem}))*|((${ipv6elem}))((:${ipv6elem})){7})${v6cidr}${portelem}$"
+    local regex="^(((${ipv6elem}))*((:${ipv6elem}))*::((${ipv6elem}))*((:${ipv6elem}))*|((${ipv6elem}))((:${ipv6elem})){7})${v6cidr}${portelem}$"
```
dschaper, line 1057: Wouldn't it be easier to say `^:{0,2}${ipv6elem}` ... ? Not sure what language the regex is in or needs to be in.

DL6ER: No, the validator matching `::1` exists separately and is a little bit later, it's the
```
::((${ipv6elem}))*
```
element in
```
^(((${ipv6elem}))*((:${ipv6elem}))*::((${ipv6elem}))*((:${ipv6elem}))*
```

dschaper: Yeah, I'm just asking if it's simpler across the board to do 0 or 2 leading, either one is valid.

dschaper: `::a:a` is a valid IPv6 address, checking specifically for `::1` is just a single, special case.

DL6ER:
```
^:{0,2}${ipv6elem} (and then the original regex)
```
would also allow
```
::2004::a
```
which is clearly wrong as two `::` would be allowed. The current change ensures, at the very least, that only one `::` can appear. Otherwise, we'd have to restructure the whole validator. Doable, but I find it at least questionable if it is worth the effort. It may introduce new unwanted side-effects if we redesign the whole thing. Looking over Stackoverflow and similar, it seems Pi-hole already seems to have the most advanced validator when it comes to IPv6 validation. I haven't seen something better.

dschaper: My suggestion may not have been clear. I didn't mean to just change the first portion of the check, and leave the rest as is. I meant to convey that it would be checking for zero or two :: at the start and then normal separators for the rest. 

The `...` was meant to just not have to type `${ipv6elem}` multiple times again.

dschaper: I still think this can be cleaner but it's not a blocker.

Issue #3578, opened 2020-07-19
Pihole install fails silently with MS WS2008R2 upstream DNS


Pyrroc:
**In raising this issue, I confirm the following:** `{please fill the checkboxes, e.g: [X]}`

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the the source code relevant to this issue?:**

3

---
**Expected behavior:**

Execute "curl -sSL https://install.pi-hole.net | bash" as root and completes successfully

**Actual behavior:**

Silent install script failure after prerequisites check, during the OS check.

**Steps to reproduce:**

    Microsoft DNS upstream (experienced with WS2008R2 - unknown if newer versions have the same issue)
    vmware esxi 6.7, 1 vCPU, 1GB RAM, 8GB HDD
    Debian 10.4 (both netinstall and DVD isos)
    Minimum install: SSH Server, Standard System Utilities... (no desktop at all)
    Had to add: sudo, curl
    sudo -i
    Executed: curl -sSL https://install.pi-hole.net | bash

**Debug token provided by [uploading `pihole -d` log](https://discourse.pi-hole.net/t/the-pihole-command-with-examples/738#debug):**

pihole does not install successfully so no token possible

**Troubleshooting undertaken, and/or other relevant information:**

Worked with @dschaper to discover that the MS DNS server (at least WS2008R2) does not handle DNS cookie options correctly, returning a status of FORMERR and no upstream results.

**Workaround: (could be permanent fix if warranted)**

1. Download script:  curl -sSL https://install.pi-hole.net > install-script-sh
2. Add +nocookie to the dig command: dig +short -t txt +nocookie ${remote_os_domain}
3. chmod the script to executable and execute as root 

**Possible script change for non-silent failure**

Replace current line 190:

    IFS=" " read -r -a supportedOS < <(dig +short -t txt ${remote_os_domain} | tr -d '"')

with:

[...]

PR #3582, opened 2020-07-19
Tweaks to compatibility check output

PromoFaux:

---

 - Allows entire check to be bypassed with the ENV var, rather than doing the check first, and then reading in the ENV var.
 - Include information about the detected OS and Version in the error output, e.g if Debian 10 were unsupported it would look like this:

image

This allows for the check to be bypassed entirely in cases like #3578

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -175,59 +175,77 @@ is_command() {
 }
 
 os_check() {
-    # This function gets a list of supported OS versions from a TXT record at versions.pi-hole.net
-    # and determines whether or not the script is running on one of those systems
-    local remote_os_domain valid_os valid_version detected_os_pretty detected_os detected_version display_warning
-    remote_os_domain="versions.pi-hole.net"
-    valid_os=false
-    valid_version=false
-    display_warning=true
 
-    detected_os_pretty=$(cat /etc/*release | grep PRETTY_NAME | cut -d '=' -f2- | tr -d '"')
-    detected_os="${detected_os_pretty%% *}"
-    detected_version=$(cat /etc/*release | grep VERSION_ID | cut -d '=' -f2- | tr -d '"')
+    if [ "$PIHOLE_SKIP_OS_CHECK" != true ]; then
+        # This function gets a list of supported OS versions from a TXT record at versions.pi-hole.net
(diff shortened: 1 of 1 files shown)


automated install/basic-install.sh
```diff
@@ -175,59 +175,77 @@ is_command() {
...
+                    printf "  %b Error: %b%b is supported but version %b is currently unsupported%b\\n" "${CROSS}" "${COL_RED}" "${detected_os}" "${detected_version}" "${COL_NC}"
+                fi
+            else
+                printf "  %b Distro: %b%b%b\\n" "${CROSS}" "${COL_RED}" "${detected_os}" "${COL_NC}"
+                printf "  %b Error: %b%b is not a supported distro%b\\n" "${CROSS}" "${COL_RED}" "${detected_os}" "${COL_NC}"
+            fi
+            printf "\\n"
+            printf "      https://docs.pi-hole.net/main/prerequesites/#supported-operating-systems\\n\\n"
+            printf "      This check can be skipped by setting the environment variable %bPIHOLE_SKIP_OS_CHECK%b to %btrue%b\\n" "${COL_LIGHT_RED}" "${COL_NC}" "${COL_LIGHT_RED}" "${COL_NC}"
+            printf "      e.g: 'sudo PIHOLE_SKIP_OS_CHECK=true curl -sSL https://install.pi-hole.net | bash'\\n"
```
dschaper, line 236:
```suggestion
            printf "      e.g: 'curl -sSL https://install.pi-hole.net | PIHOLE_SKIP_OS_CHECK=true sudo bash'\\n"
```
The ENVVAR needs to be set in the bash environment executing the script.

Possibly `export PIHOLE_SKIP_OS_CHECK=true; curl -sSL https://install.pi-hole.net | bash` but I'd have to play around with it more.

automated install/basic-install.sh
```diff
@@ -175,59 +175,77 @@ is_command() {
...
+            printf "      or   'sudo PIHOLE_SKIP_OS_CHECK=true pihole -up'\\n\\n"
```
dschaper, line 237: I don't think this actually works. You'd need something like:

```bash
export PIHOLE_SKIP_OS_CHECK=true; sudo pihole -up
```

PR #3589, opened 2020-07-20
Validate setup vars

jlagermann:

---
Using the --unattended flag to install requires the use of setupVars.conf. If this configuration file is missing required field or not formatted correctly, the installation will fail, without displaying errors.

This PR request adds validation for setupVars.conf and adds an example setupVars.conf.

If useUpdateVars=true, it will first make sure setupVars.conf exists and has the correct permissions, then run accountForRefactor, then validate_setupVars.

It checks to make sure INSTALL_WEB_SERVER is enabled if INSTALL_WEB_INTERFACE is enabled is (part of the validate_setupVars).

If the undocumented flag "--disable-install-webserver" is used, both INSTALL_WEB_INTERFACE and INSTALL_WEB_SERVER will be set to false.

Initial setting of INSTALL_WEB_SERVER=true was moved up with other variables.

added pihole-setupVars.conf example
added validate_setupVars
validate_setupVars runs immediately after accountForRefactor

Installation instruction may need to be updated for unattended installs.  However, the process did not change.

---


jlagermann:
The failing L2245 and L2250 are preexisting and not part of this change.

automated install/basic-install.sh
```diff
@@ -109,14 +110,13 @@ c=$(( c < 70 ? 70 : c ))
...
 runUnattended=false
-INSTALL_WEB_SERVER=true
 # Check arguments for the undocumented flags
 for var in "$@"; do
     case "$var" in
         "--reconfigure" ) reconfigure=true;;
         "--i_do_not_follow_recommendations" ) skipSpaceCheck=true;;
         "--unattended" ) runUnattended=true;;
-        "--disable-install-webserver" ) INSTALL_WEB_SERVER=false;;
+        "--disable-install-webserver" ) INSTALL_WEB_SERVER=false INSTALL_WEB_INTERFACE=false;;
```
MichaIng, line 119: As I mentioned in the other PR, this breaks most use cases of `--disable-install-webserver`. It is not used to skip the web interface install but to allow other webservers or things like Lighttpd with PHP-FPM (instead of PHP-CGI).

But the other way round it makes sense: If the web interface is not wanted (extremely rare case), then the web server is definitely not required. Related PR: #3225

automated install/pihole-setupVars.conf.example
```diff
@@ -0,0 +1,35 @@
...
+# ; *  IMPORTANT NOTE:  This file is used as a variable file for the Bash install script and
+# ;    as a PHP ini file for /var/www/html/admin/settings.php.
+# ; *  Bash reads comments as "#"
+# ; *  PHP reads comments as ";" (As of PHP 7.0.0, "#" are no longer recognized as PHP comments)
+# ; *  All lines must end with a LF, not CRLF
+# ; *  There must be a blank line at the end of the file.
+# ; *  Options set during or after installation will be appended to the last line.
+# ;
+# ; *  To create a random Web GUI password, delete "WEBPASSWORD=".  If it is left as is, no password will
+# ;    be required to login.  The password can be changed after installation with 'pihole -a -p'
```
MichaIng, line 11: Probably it is best to comment `WEBPASSWORD=` by default, to preserve the current default behaviour (random generated password) when this line is not touched?

automated install/basic-install.sh
PromoFaux: See my comments on https://github.com/pi-hole/pi-hole/pull/3590 (Which, looking at it, includes the same changesets...)

automated install/basic-install.sh
```diff
@@ -1977,6 +1977,102 @@ accountForRefactor() {
...
+            echo "${var} is not set"
+        else
+            echo "${var} is ${!var}"
+            echo "If ${var} is set in setupVars.conf, valid options are: true or false."
+            exit 1
+        fi
+    done
+    # INSTALL_WEB_SERVER must be set to true if INSTALL_WEB_INTERFACE is set to true
+    if [[ "${INSTALL_WEB_INTERFACE}" == true ]] && [[ "${INSTALL_WEB_SERVER}" == false ]]; then
+        echo "You can install the Web Server without the web interface, but you cannot install the web interface without the web server."
```
MichaIng, line 2012: As of above, installing a web server without the web interface does not make any sense, but installing the web interface but allowing to install a different web server or Lighttpd with custom PHP handler is a common use case and the reason why this variable + installer option has been added.

automated install/basic-install.sh
jlagermann: That makes sense.  I'll have to make a change to the validate_setupVars function also.

automated install/pihole-setupVars.conf.example
```diff
@@ -0,0 +1,35 @@
...
+INSTALL_WEB_SERVER=true
+INSTALL_WEB_INTERFACE=true
+LIGHTTPD_ENABLED=true
+# ; WEBPASSWORD=
+TEMPERATUREUNIT=F
+WEBUIBOXEDLAYOUT=boxed
+BLOCKING_ENABLED=true
+DNSMASQ_LISTENING=single
+PIHOLE_DNS_1=x.x.x.x#5353
+PIHOLE_DNS_2=x.x.x.x
```
MichaIng, line 28: `PIHOLE_DNS_1` and `PIHOLE_DNS_2` are doubled, I guess this is not intended.

jlagermann:
thanks, I'll fix it.


Thank you,

*James Lagermann* | Systems Engineer
c: 636-575-7249
james.lagermann@corelight.com

[image: Corelight]
<https://app.salesforceiq.com/r?target=5e7232c4c9e77c0076f38131&t=AFwhZf1GBKQ82RfBl_ysPyfYtIb4-WUzgM9l0iNVWBIqb_rgJA6A_i_PiheVB7MySqXaf4jk7R3BN_1qqCJwESAmpTGqB3USK-m7p13jpRaw7yrTkQE7BefEBLoHXx0GG29WYKwQp0eW&url=https%3A%2F%2Fwww.corelight.com%2F>

DL6ER (approved):
@jlagermann Could you merge in the latest `development` branch? There are no merge conflicts so this should work right away. We're good to merge afterwards (if the tests also all go green).

automated install/pihole-setupVars.conf.example
```diff
@@ -0,0 +1,33 @@
+# ;
```
DL6ER, line 1: That is the motivation for adding this template/example file?

jlagermann:
To provide an example of most of the configurable variables with the proper
formatting for /etc/pihole/setupVars.conf.  Without this example, you end
up configuring some basic information, if any.  Then you have to configure
the rest from the GUI.  That defeats the purpose for automated installs.


Thank you,

*James Lagermann* | Systems Engineer
c: 636-575-7249
james.lagermann@corelight.com

[image: Corelight]
<https://app.salesforceiq.com/r?target=5e7232c4c9e77c0076f38131&t=AFwhZf1GBKQ82RfBl_ysPyfYtIb4-WUzgM9l0iNVWBIqb_rgJA6A_i_PiheVB7MySqXaf4jk7R3BN_1qqCJwESAmpTGqB3USK-m7p13jpRaw7yrTkQE7BefEBLoHXx0GG29WYKwQp0eW&url=https%3A%2F%2Fwww.corelight.com%2F>

MichaIng:
```
mv pihole-setupVars.conf.example setupVars.conf
nano setupVars.conf
./basic-install.sh --unattended
```
☕ ...
```
systemctl status pihole-FTL
```
👍

dschaper:
Now, skip step 2, which most of our users will do.

MichaIng:
> Now, skip step 2, which most of our users will do.

This is why the validation function has been implemented. All invalid values need to be re-entered manually then.

dschaper:
> The risk for wrong entries is much higher when no example config exists that shows all settings and example values that show how they should look like.

Understood, for me the file is wrong. (And it is, there's values that are incorrect and configurations that have changed.) This adds more moving parts and places that need to be watched and maintained for a small audience that would benefit. The chances for things to go wrong spike with something like this.

`setupVars.conf` was initially designed to be just a setup file for overriding defaults. It wasn't meant to be a runtime configuration file. If you want to solve that issue then fix that issue and create a proper runtime configuration. Patching in an overly complex system that adds to the developer and support burden to maintain and update/upgrade is not the proper solution.

jlagermann:
> Understood, for me the file is wrong. (And it is, there's values that are incorrect and configurations that have changed.) This adds more moving parts and places that need to be watched and maintained for a small audience that would benefit. The chances for things to go wrong spike with something like this.
> 
> `setupVars.conf` was initially designed to be just a setup file for overriding defaults. It wasn't meant to be a runtime configuration file. If you want to solve that issue then fix that issue and create a proper runtime configuration. Patching in an overly complex system that adds to the developer and support burden to maintain and update/upgrade is not the proper solution.

I agree it does add more moving parts to watch, however, this needs to be documented somewhere.  I figured it out with trial and error until I gave up and just configured it via the GUI.  All of the examples here were captured after making changes in the GUI when I originally wrote this.  My assumption at the time was this .conf file served two purposes, enable unattended installations and preserve 'changes from defaults' so they are not lost during a reinstall or upgrade.

MichaIng:
> `setupVars.conf` was initially designed to be just a setup file for overriding defaults.

That has not changed, as far as I understood the PR? The aim is to make it easier and safer to override setup defaults, even all of them to allow reasonable use of the `--unattended` install option (aside of `pihole -r/up`, I know it's used internally).

> complex system that adds to the developer and support burden to maintain

Jep, instead of changing a setting at one place, it now needs to be changed at a second place, so while it makes life for some users easier, it adds tasks to developers. But what is "complex" about it? The validation is reasonable on its own, as the `setupVars.conf` could be malformed for other reasons.

dschaper:
My point is if you want a system that does what you are requesting then fix the system to do that. Patching this current system isn't the proper fix.

MichaIng:
> My point is if you want a system that does what you are requesting then fix the system to do that. Patching this current system isn't the proper fix.

You mean adding a new config file to override setup variables aside of `setupVars.conf` which then shall only store chosen setup variables for updates and reinstalls? That sounds like a lot of duplication and complicated to me. Really everything that applies to setup variable overrides applies to preserved variables as well.

What I can think of is another installer command option that allows to pre-create a `setupVars.conf` without doing the actual install. Like a dummy install that only implies the whiptail menus and ends after that interactive part. While this could reduce the double maintenance of the values in code, maintaining such an additional command and implementing it correctly adds much work as well 🤔.

Another idea: Add command line options for each setup variable. But while an example config is not required then, again the same variables are one more time present in the installer script => a bid increased maintenance effort.
____________
Whatever your choice will be in the idea to enable unattended installs, the `validate_setupVars` would still be a stability enhancement on its own.

automated install/basic-install.sh
```diff
@@ -1990,6 +1990,100 @@ accountForRefactor() {
...
+    else
+            echo "${CONDITIONAL_FORWARDING_REVERSE} is not a valid reverse address"
+            exit 1
+    fi
+    # LIGHTTPD_ENABLED should not be written to the setupVars.conf file.  It's a temporary variable and not used for installation or updates.
+    sed -i '/LIGHTTPD_ENABLED=/ d' "${setupVars}"
+    # DNSMASQ_LISTENING is not used during setup or updates.
+    sed -i '/DNSMASQ_LISTENING=/ d' "${setupVars}"
+    # Ensure setupVars ends with a newline
+    sed -i -e '$a\ ' "${setupVars}"
```
MichaIng, line 2084: If I am not mistaken, since the file is never fully removed or overwritten, this would add an additional newline on every update/reinstall. AFAIK bash has no problem to source files with missing trailing newline, does it?

jlagermann: It will make sure there is exactly one blank line at the end of the file.  Off the top of my head, I don't remember if it was bash that had the problem or php (because the same file is used for both).  I do remember it will fail if the blank line is not there.

jlagermann: As a side note, I've have now enabled the options in VS Code to automatically remove trailing white spaces and ensure each file ends with exactly one blank line.

PromoFaux:
We've spoken quite extensively (albeit internally) about the bash script coming to the end of its usefulness. It's really grown organically over the last 5 years to the point where we are doing things just because they work, rather than because we should.

The direction we want to head in, really, needs to be a single binary (FTL) - eventually we could probably port a lot of the scripts into native functionality rather than hacking together bits of bash to make it work (plus side of this is reducing the number of dependencies)

We've also (quietly) included a lua interpreter in FTL - which opens up a whole heap of possibilities for quick scripting where we don't port things.

In an ideal world the install script would just be a simple bootstrapper to download the correct version of FTL, which then goes on to either generate a proper (yaml/toml/JSON) config file.

At this point I know the install script like the back of my hand, but even I can admit it's becoming difficult to maintain

automated install/basic-install.sh
MichaIng: > It will make sure there is exactly one blank line at the end of the file.

That is not true, the sed will always add an additional newline, regardless if there is already one. Just tested in bash shell to be sure, but the syntax is clear on it's own. It would require a test for an existing character.
Ah it even adds a space instead of an empty like.

An empty new line would be `sed -i -e '$a\\n' "${setupVars}"` or better `echo >> "${setupVars}"`. I'm currently failing on the syntax for "if last line an it contains a character" with sed 😄.

automated install/basic-install.sh
```diff
@@ -1990,6 +1990,100 @@ accountForRefactor() {
...
+            exit 1
+        fi
+    if [[ "${CONDITIONAL_FORWARDING_REVERSE}" =~ ${addressReverse} ]]; then
+            echo "CONDITIONAL_FORWARDING_REVERSE is ${CONDITIONAL_FORWARDING_REVERSE}"
+    elif [[ -z "${CONDITIONAL_FORWARDING_REVERSE}" ]]; then
+            echo "CONDITIONAL_FORWARDING_REVERSE is not set"
+    else
+            echo "${CONDITIONAL_FORWARDING_REVERSE} is not a valid reverse address"
+            exit 1
+    fi
```
dschaper, line 2078: CONDITIONAL_FORWARDING has been changed to REV_SERVER. 

`arpa` notation was also replaced with CIDR in that same change.

https://github.com/pi-hole/pi-hole/pull/3882

dschaper (changes requested):

automated install/basic-install.sh
PromoFaux: Yeah, nice catch. Although the install script (per the linked pull) will reconfigure `CONDITIONAL_FORWARDING*` vars into `REV_SERVER*` vars - they are [deprecated](https://github.com/pi-hole/docker-pi-hole#deprecated-environment-variables), and we are/should be encouraging folk to use the correct variables, and not let the script do the work of changing the old ones.

MichaIng:
@PromoFaux 
> In an ideal world the install script would just be a simple bootstrapper to download the correct version of FTL, which then goes on to either generate a proper (yaml/toml/JSON) config file, or read one that is provided by a user seeking to automate installation

_Probably it's only that I like bash too much, as I can't really write much other languages, also a bit OOT, but to defend the installer bash script:_
While PH7 implementation and others will clearly simplify the installer, major parts will stay as long as you don't want to move away from the one-line read-to-run install method: The binary will always require a service wrapped around it, a user that runs the service, config, data and log directories is requires write access to (although log+logrotate could be removed in favour of syslog/stdout only), cron job, man page, correct OS and architecture detection to pull the correct binary, remaining at least dnsmasq dependencies, i.e. detecting and using package manager robustly, not to forget the interactive parts: dialogs, static IP warning, upstream DNS and blocklist choices, network interface detection and/or choice. Some of the interactive parts could be moved into the web UI as first access wizard or with the need to manually configure away from standards, but that doesn't eliminate the need to validate inputs, maintain those parts but only moves them from bash to another language, which is then probably more a question of personal preference than more or less development/maintenance effort?

automated install/basic-install.sh
MichaIng: There it is:
```
sed -i '$s/^..*$/\0\n/' "${setupVars}"
```
or
```
sed -Ei '$s/^.+$/\0\n/' "${setupVars}"
```
but I like basic regex where possible.

dschaper:
Bash just isn't meant for what is being done with it. There's no way to expand upon it anymore, it needs to be replaced with something that can actually do what needs to be done. The plan is a simple posix compliant script that downloads FTL (which has in it everything it needs to run a proper installation script now) and the installation script.

MichaIng:
> Bash just isn't meant for what is being done with it.

It doesn't matter what whoever meant it for, as long as it works. But probably with C it is easier, that is what I cannot evaluate. I guess stings will move step-by-step after PH7 implementation, so will be exciting to watch how far it goes 😃.

dschaper:
> It doesn't matter what whoever meant it for, as long as it works.

That's not our philosophy at Pi-hole.



> But probably with C it is easier, that is what I cannot evaluate.

`c` isn't a scripting language so it's not going to be used for scripting.



> I guess stings will move step-by-step after PH7 implementation

PH7 is a small part of things, and a lot of what's needed is already in the `pihole-FTL` binary that is being downloaded right now.

MichaIng:
> `c` isn't a scripting language so it's not going to be used for scripting.

> FTL (which has in it everything it needs to run a proper installation script now) and the installation script.

I missed the "and the installation script". So you want to use one script to download and run another script? Regardless, you'll have a proper plan and we are OOT.

@jlagermann 
For going forward with the PR, I think it would make sense to remove the example setupVars and clearly reduce the aim to validating the variables as a step to detect issues with them as early as possible and exit with a meaningful error message. The time to officially support a fully automated setup will certainly be later when there are fewer things to setup.

yubiuser:
I'll close this PR as there was no activity for over a year. Glimpsing over the discussion it's not clear if it would have been merged. Opinions might change over time. If you feel now is a good time to re-try solve the merge conflicts and re-open.

Thanks for your contribution.

PR #3590, opened 2020-07-20
Preserve setup vars

jlagermann:

---
What does this PR aim to accomplish?:
This PR preserves the setupVars.conf file during the installation.  The existing basic-install.sh will delete and rewrite variables in the setupVars.conf putting them back in a random order, possibly breaking any comments that may be included.

This PR also sets up future ability to use the setupVars.conf file to manage any variable.  For example, DNS Cache Size or Logging to External Syslog Collectors in addition to the local log files.

How does this PR accomplish the above?:
Added a function to replace, add or create variables in files, to include the variable file itself.  
The function, replaceAddCreate takes two variables
 - 1 - the name of the local variable that represents the new value (without the "${ }")
 - 2 - the path and name of the target file to modify
For example,
replaceAddCreate PIHOLE_INTERFACE ${setupVars}

It also adds a function to replace, add, create or rename variables in files.  
The function, replaceAddCreateRename takes three variables
 - 1 - the name of the local variable that represents the new value (without the "${ }")
 - 2 - the name of the variable in the target file (without the "${ }")
 - 3 - the path and name of the target file to modify
For example, 
replaceAddCreateRename PRIVACY_LEVEL PRIVACYLEVEL "${PI_HOLE_CONFIG_DIR}/pihole-FTL.conf"

What documentation changes (if any) are needed to support this PR?:
The installation process did not change.

---


jlagermann:
This PR requires PR #3589

automated install/basic-install.sh
```diff
@@ -2008,8 +2169,12 @@ installPihole() {
         fi
     fi
     # For updates and unattended install.
-    if [[ "${useUpdateVars}" == true ]]; then
-        accountForRefactor
+    if [[ "${runUnattended}" == true ]]; then
+        if [[ -f "${setupVars}" ]]; then
+            chmod 644 "${setupVars}"
+            accountForRefactor
```
PromoFaux, line 2175: note for myself more than anything else. : this function can probably go away, now. I don't even remember the version that it was introduced in, but I very very much doubt that anyone is still updating from that version!

automated install/basic-install.sh
```diff
@@ -1973,8 +2038,104 @@ accountForRefactor() {
...
+            exit 1
+        fi
+    done
+    # INSTALL_WEB_SERVER must be set to true if INSTALL_WEB_INTERFACE is set to true
+    if [[ "${INSTALL_WEB_INTERFACE}" == true ]] && [[ "${INSTALL_WEB_SERVER}" == false ]]; then
+        echo "You can install the Web Server without the web interface, but you cannot install the web interface without the web server."
+        echo "INSTALL_WEB_SERVER is set to ${INSTALL_WEB_SERVER} and INSTALL_WEB_INTERFACE is set to ${INSTALL_WEB_INTERFACE}."
+        exit 1
+    fi
+    # Regex patterns to match
```
PromoFaux, line 2081: No need to reinvent the wheel here, please see (and re-use) the existing functions `valid_ip` and `valid_ip6` (there is a change to `valid_ip6` in the following [commit](https://github.com/pi-hole/pi-hole/pull/3564/commits/643d2c0f3ea7871b64f9eaa2c9a5a2060d9ad14e) in #3564)

automated install/basic-install.sh
```diff
@@ -1885,7 +1885,81 @@ create_pihole_user() {
     fi
 }
 
-#
+# replaceAddCreateRename takes three varialbles
+# 1 - the name of the local variable that represents the new value
+# 2 - the name of the variable in the target file
+# 3 - the path and name of the target file to modify
+replaceAddCreateRename() {
```
PromoFaux, line 1892: maybe a better name for this function would be something like `saveSetting` or the slightly more verbose`saveSettingToFile`

automated install/basic-install.sh
jlagermann: The IP address has to include the subnet mask.  I looked at those functions but they do not check for subnet in bits "/xx".  
I can remove line 2084, 'addressPort' and replace any uses with valid_ip.
Line 2082, 'address' is not used currently so I can remove it.

Line 2083 and 2085, 'addressMask' and 'addressReverse' need to stay.  The existing functions won't validate them.

automated install/basic-install.sh
jlagermann: I agree, it needs a better name.  I could not think of anything good so I just put what it did.  It can be used for more than just saving setting to a file though.  It's main purpose is to create and/or update variable files.

How about updateVarsFile (I've been working with Ansible a lot lately, it uses vars_files for everything.)

automated install/basic-install.sh
jlagermann: It was a good spot to call the validate_setupVars() function.

automated install/basic-install.sh
jlagermann: Also, there are two different functions, replaceAddCreateRename and replaceAddCreate.  Both have the same purpose but they take different arguments.

replaceAddCreate only has two arguments, the name of the variable and the file to modify.  It assumes the name of the variable in the file is the same as the variable in the basic-install script.

replaceAddCreateRename has three arguments, the name of the variable in the basic-install script, the name of the varialbe in the vars file, and the file to modify.

I will remove 'replaceAddCreate' completely and then just modify the one time it's called to use 'replaceAddCreateRename', renamed to updateVarsFile

DL6ER:
I'm good with this given two tings:
1. #3589 is updated and merged 
2. This PR is updated (there are currently merge conflicts)

yubiuser:
I'll close this PR as there was no activity for over two years. Part of the idea was implemented later with `addKey()` and `addKeyValuePair()` in `utils.sh`

Thanks for your contribution.

PR #3625, opened 2020-07-28
Update Test suite

PromoFaux:

---

- Updates Python version used in test scripts to `3.7` from `3.6`
- Updates the base images to versions of the OS that we actually support (thanks @bcambl  for the Centos stuff)
- Adds two new tests to prove the OS compatibility check is working/falls over on an invalid OS

### DIFF
--- a/test/test_automated_install.py
+++ b/test/test_automated_install.py
@@ -562,6 +562,7 @@ def test_validate_ip_invalid_letters(Pihole):
 
     assert output.rc == 1
 
+
 def test_os_check_passes(Pihole):
     ''' Confirms OS meets the requirements '''
     Pihole.run('''
--- a/test/test_automated_install.py
+++ b/test/test_automated_install.py
@@ -563,6 +563,24 @@ def test_validate_ip_invalid_letters(Pihole):
     assert output.rc == 1
 
 
+def test_os_check_fails(Pihole):
+    ''' Confirms install fails on unsupported OS '''
+    Pihole.run('''
+    source /opt/pihole/basic-install.sh
+    distro_check
+    install_dependent_packages ${INSTALLER_DEPS[@]}
+    cat <<EOT > /etc/os-release
+    ID=UnsupportedOS
+    VERSION_ID="2"
+    EOT
+    ''')
+    detectOS = Pihole.run('''t
+    source /opt/pihole/basic-install.sh
+    os_check
+    ''')
+    expected_stdout = 'Unsupported OS detected: UnsupportedOS'
+    assert expected_stdout in detectOS.stdout
+
 def test_os_check_passes(Pihole):
     ''' Confirms OS meets the requirements '''
     Pihole.run('''
--- a/test/test_automated_install.py
+++ b/test/test_automated_install.py
@@ -562,7 +562,7 @@ def test_validate_ip_invalid_letters(Pihole):
 
     assert output.rc == 1
 
-def test_os_check_pases(Pihole):
+def test_os_check_passes(Pihole):
     ''' Confirms OS meets the requirements '''
     Pihole.run('''
     source /opt/pihole/basic-install.sh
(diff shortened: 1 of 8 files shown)


test/test_automated_install.py
```diff
@@ -488,3 +488,17 @@ def test_validate_ip_invalid_letters(Pihole):
     ''')
 
     assert output.rc == 1
+
+def test_os_check_pases(Pihole):
```
pvogt09, line 492:
```suggestion
def test_os_check_passes(Pihole):
```

pvogt09:
I tried something similiar in https://github.com/pvogt09/pi-hole/tree/feature/test_install_deps, but the installation of the web dependencies would never pass the test because of missing/wrong `php` versions in the images. In case you find anything useful therein, feel free to use it.

This PR will also makes #3213 obsolete.

I also have a branch with all supported OS (at the time of the creation of that branch) in the test cases, in case that should be needed in https://github.com/pvogt09/pi-hole/tree/feature/test_all_supported. I have to admit that this increases the time for the travis build to sometimes more than 15 minutes and might not be desired.

PromoFaux:
I've never really played about with the tests before, it was only yesterday where I figured we should probably have a test for the new functionality of the supported version checker that I had a look!

I hit a bit of a brick wall, though. I'd like to try and get a test that mocks the output of PRETTY_NAME/Version so that we can test that the os check function will fail in case of an unsupported OS. But my head hurts

test/test_automated_install.py
```diff
@@ -488,3 +488,17 @@ def test_validate_ip_invalid_letters(Pihole):
...
+def test_os_check_passes(Pihole):
+    ''' Confirms OS meets the requirements '''
+    Pihole.run('''
+    source /opt/pihole/basic-install.sh
+    distro_check
+    install_dependent_packages ${INSTALLER_DEPS[@]}
+    ''')
+    detectOS = Pihole.run('''
+    source /opt/pihole/basic-install.sh
+    os_check
```
dschaper, line 501:
```suggestion
    source /opt/pihole/basic-install.sh
    pretty_name = 'PRETTY_NAME="Debian GNU/Linux 10 (buster)"'
    with open('/etc/os-release','w') as f:
        f.write(pretty_name)
    os_check
```

Or something in that same vein.

dschaper: Might have to break out the python code to a line before `.run`, I don't remember if you can do python inside the `run` or not.

PromoFaux: The only thing was the file `*-release` file having different names in different distros...

dschaper: Okay, just change the name of the file you're writing to when you change the `pretty_name` string.

PromoFaux:
OK, I think this is now ready for a re-review. 

As noted by @pvogt09, this will render #3213 obsolete. One thing I've noted from that one, however, is that the target is `stretch`, and here I have targeted `buster`. Someone else can make the decision which one we use, it's all the same to me :), provided it's not `jessie` or `wheezy` ;)

pvogt09:
I chose the oldest supported version in #3213. As mentioned there this seems legit, but I would vote for testing all supported versions. I can set up a PR for this as soon as this one is merged, since I already have done something in this direction.

dschaper:
I prefer lowest supported if we only have one version. My thinking is newer versions would still support the packages in the previous version, it's very rare for an OS version release to have changes that diverge it from a previous version.

bcambl:
> 
> 
> I prefer lowest supported if we only have one version. My thinking is newer versions would still support the packages in the previous version, it's very rare for an OS version release to have changes that diverge it from a previous version.

CentOS falls into this catagory. CentOS8 is very similar to Fedora (new php, dnf package manager, etc).  
CentOS7 ships with an old PHP version and requires the additional image for PHP logic testing. Once CentOS7 support is dropped, then we can go back to a single container images for both `centos` and `fedora`.

Adam-SAI (dismissed):

diginc (dismissed):

PromoFaux:
Sorry, @diginc, had to dismiss your review to make a change.

> I can set up a PR for this as soon as this one is merged, since I already have done something in this direction.

Sounds good to me.

diginc (approved):

PR #3650, opened 2020-08-04
Vacuum DB to reduce disk space usage

TheGroundZero:

---

Reduce disk space usage of the database file.
Note: this temporarily increases disk usage as VACUUM create a temp copy of the DB

Use SQLite's VACUUM command

One extra command to be executed

---

### DIFF
--- a/advanced/Scripts/piholeLogFlush.sh
+++ b/advanced/Scripts/piholeLogFlush.sh
@@ -59,7 +59,9 @@ else
     fi
     # Delete most recent 24 hours from FTL's database, leave even older data intact (don't wipe out all history)
     deleted=$(sqlite3 "${DBFILE}" "DELETE FROM queries WHERE timestamp >= strftime('%s','now')-86400; select changes() from queries limit 1")
-
+    # Reduce disk space usage
+    sqlite3 "${DBFILE}" "VACUUM"
+    
     # Restart pihole-FTL to force reloading history
     sudo pihole restartdns
 fi


DL6ER:
Just to set the context: The `VACUUM` command rebuilds the database file, repacking it into a minimal amount of disk space.

> The VACUUM command works by copying the contents of the database into a temporary database file and then overwriting the original with the contents of the temporary file. When overwriting the original, a rollback journal or write-ahead log WAL file is used just as it would be for any other database transaction. This means that when VACUUMing a database, **as much as twice the size of the original database file is required in free disk space**.

Okay, you mention this. I just wanted to repeat it, again, for context.

So let's do an experiment to see what is happening here:
```
$ ls -lh /etc/pihole/pihole-FTL.db 
-rw-r--r-- 1 pihole pihole 561M Aug 17 19:40 /etc/pihole/pihole-FTL.db

$ cp /etc/pihole/pihole-FTL.db ~/pihole-FTL.db

$ time sqlite3 ~/pihole-FTL.db "VACUUM"

real	7m24.679s
user	2m8.873s
sys	0m36.006s
```
(ran on a Raspberry Pi 3)

I have a mildly active network with only 2-3 real clients doing requests. I assume databases in the wild will be at least similar in size, if not substantially larger. Yet, a whooping 7.5 minutes of VACUUMing on the database seems to much. Imagine how this looks like on a Raspberry Pi Zero (or similar).

There are good reasons why we do not run a VACUUM, mainly because:

> When content is deleted from an SQLite database, the content is not usually erased but rather the space used to hold the content is marked as being available for reuse.

As the database is using indices with pointers to the individual rows, fragmentation does not mean speed-penalties. Furthermore, the reduction of space would be there, I agree, however, the database will still continue to grow afterwards. I do not really see the issue with the database being large with a lot of free space to put additional records into.

In the end, this means the database reaches a specific maximum size (can be controlled through the [MAXDBSIZE](https://docs.pi-hole.net/ftldns/configfile/#maxdbdays) auto-cleaning setting) and doesn't continue to grow any further.

TheGroundZero:
The command `MAXDBDAYS` does indeed limit somewhat the size the DB may grow, but in some situations this may still be considerably big.

I had an issue with my setup, where my firewall (UniFi Security Gateway) would make thousands of SOA requests for one specific system in my network.  
DB would reach 20GB in a matter of days.

Flushing the DB wouldn't help (contrary to what I expected, this [only deletes past 24h, not before](https://github.com/pi-hole/pi-hole/blob/6b536b7428a1f57ff34ddc444ded6d3a62b00a38/advanced/Scripts/piholeLogFlush.sh#L61))?  
Since `VACUUM` wasn't in place, this wouldn't decrease disk usage.  
However, since `VACUUM` creates a copy, you wouldn't be able to even run this command on an already full disk.

dschaper:
If you want to VACUUM your database then I think you can do that, using what ever method you'd like.

Adding that to every flush run would render flushing inoperable. I'm not going to give my approval to adding 8 minutes to each flush command.

dschaper:
> I had an issue with my setup, where my firewall (UniFi Security Gateway) would make thousands of SOA requests for one specific system in my network.
> DB would reach 20GB in a matter of days.

That sounds like a problem that should be handled by the hardware vendor. I don't see this PR providing any useful changes to Pi-hole.

TheGroundZero:
You're right.

I was trying to solve a problem that wasn't related to PiHole.  
Since I installed PiHole on a more powerful system, I didn't think about potential impact for RPi-users.

PR #3684, opened 2020-08-12
Tweaks to os-check

dschaper:
A few tweaks to the os-check function.

1. Reduce subshell use where possible.
2. Use descriptive variable names.
3. Remove boolean variables. Bash doesn't have booleans, they are just strings.
4. Check using `/etc/os-release` since all supported distro's use that file.

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -179,6 +179,7 @@ is_command() {
 }
 
 os_check() {
+<<<<<<< HEAD
     if [ "$PIHOLE_SKIP_OS_CHECK" != true ]; then
         # This function gets a list of supported OS versions from a TXT record at versions.pi-hole.net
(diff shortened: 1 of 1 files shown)


automated install/basic-install.sh
```diff
@@ -183,27 +183,23 @@ os_check() {
     # and determines whether or not the script is running on one of those systems
     local remote_os_domain valid_os valid_version detected_os_pretty detected_os detected_version display_warning
```
zebulon501, line 184: This should be cleaned up as some variables don't exist anymore

dschaper: Thanks, good catch!

dschaper:
Sure, can do.

dschaper:
Yeah, screw git.

PR #3688, opened 2020-08-14
Tweaks to os_check() Redux

dschaper:
A few tweaks to the os-check function.

    Reduce subshell use where possible.
    Use descriptive variable names.
    Remove boolean variables. Bash doesn't have booleans, they are just strings.
    Check using /etc/os-release since all supported distro's use that file.

See #3864 for detail.

### DIFF
--- a/advanced/Scripts/piholeDebug.sh
+++ b/advanced/Scripts/piholeDebug.sh
@@ -425,17 +425,17 @@ os_check() {
 
     # Display findings back to the user
     if [ "$valid_os" = true ]; then
-        log_write "${TICK} Distro:  ${COL_GREEN}${detected_os}${COL_NC}"
+        log_write "${TICK} Distro:  ${COL_GREEN}${detected_os^}${COL_NC}"
 
         if [ "$valid_version" = true ]; then
             log_write "${TICK} Version: ${COL_GREEN}${detected_version}${COL_NC}"
         else
             log_write "${CROSS} Version: ${COL_RED}${detected_version}${COL_NC}"
-            log_write "${CROSS} Error: ${COL_RED}${detected_os} is supported but version ${detected_version} is currently unsupported (${FAQ_HARDWARE_REQUIREMENTS})${COL_NC}"
+            log_write "${CROSS} Error: ${COL_RED}${detected_os^} is supported but version ${detected_version} is currently unsupported (${FAQ_HARDWARE_REQUIREMENTS})${COL_NC}"
         fi
     else
-        log_write "${CROSS} Distro:  ${COL_RED}${detected_os}${COL_NC}"
-        log_write "${CROSS} Error: ${COL_RED}${detected_os} is not a supported distro (${FAQ_HARDWARE_REQUIREMENTS})${COL_NC}"
+        log_write "${CROSS} Distro:  ${COL_RED}${detected_os^}${COL_NC}"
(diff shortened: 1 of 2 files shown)


automated install/basic-install.sh
```diff
@@ -182,47 +182,43 @@ os_check() {
...
-          fi
+        fi
 
         if [ "$valid_os" = true ] && [ "$valid_version" = true ]; then
             display_warning=false
         fi
 
-        if [ "$display_warning" = true ]; then
+        if [ "$display_warning" != false ]; then
             printf "  %b %bUnsupported OS detected: %s%b\\n" "${CROSS}" "${COL_LIGHT_RED}" "${detected_os_pretty}" "${COL_NC}"
```
yubiuser, line 222: You removed `detected_os_pretty`, so best would be to replace it here with `detected_os + `detected_version`

dschaper: Check latest commit and resolve conversation if it is the correct fix.

PromoFaux: Confirmed working

image

advanced/Scripts/piholeDebug.sh
```diff
@@ -396,49 +396,46 @@ check_critical_program_versions() {
...
-    local remote_os_domain valid_os valid_version detected_os_pretty detected_os detected_version
+    local remote_os_domain valid_os valid_version detected_os detected_version
     remote_os_domain="versions.pi-hole.net"
-    valid_os=false
-    valid_version=false
 
-    detected_os_pretty=$(cat /etc/*release | grep PRETTY_NAME | cut -d '=' -f2- | tr -d '"')
-    detected_os="${detected_os_pretty%% *}"
-    detected_version=$(cat /etc/*release | grep VERSION_ID | cut -d '=' -f2- | tr -d '"')
+    detected_os=$(grep "\bID\b" /etc/os-release | cut -d '=' -f2 | tr -d '"')
```
PromoFaux, line 402: I wonder if we should add in a check for the existence of `/etc/os-release`, and if it is not there, show a message saying we are unable to detect the OS (similarly, if we get a blank response on the `dig` command?)

dschaper: Sure, if you do `grep -s` then you won't get the "file not found" error and the output will be empty.

```plain
dan@Pi-hole:~$ det=$(grep "\bID\b" /etc/os-release)
dan@Pi-hole:~$ echo $det
ID=debian
dan@Pi-hole:~$ det=$(grep -s "\bID\b" /etc/no-release)
dan@Pi-hole:~$ echo $det

dan@Pi-hole:~$
```

dschaper:
```plain
dan@Pi-hole:~$ det=$(grep "\bID\b" /etc/no-release)
grep: /etc/no-release: No such file or directory
dan@Pi-hole:~$ echo $det

dan@Pi-hole:~$
```

dschaper: #3695 addresses the `dig`, may want to comment on that.

PromoFaux: I'll give it a whirl in a bit, It's probably going to need a rebase, that said.

dschaper: Sounds good, I'll resolve this convo then and it can be taken up on #3695

PR #3698, opened 2020-08-18
Remove check for free disk space and associated variables

yubiuser:
Removes the check for free disk space  (~52MB was required)  and associated variables.
The check was introduced here: https://github.com/pi-hole/pi-hole/pull/301.

Removal discussion here https://github.com/pi-hole/docs/pull/364#discussion_r471727209


PromoFaux (approved):
I'm largely OK with this PR in principle. It also gets rid of a nearly-useless undocumented flag.

Requesting secondary review from @dschaper or @DL6ER before merging

dschaper:
I haven't tested or checked for actual function but the concept is okay by me.

PromoFaux:
(I tested before merge, btw!)

PR #3665, opened 2020-08-09
Change default value of temp_unit to 'C'

scboucher:
Signed-off-by: Samuel Boucher <scboucher@users.noreply.github.com>

**By submitting this pull request, I confirm the following:** 
*please fill any appropriate checkboxes, e.g: [X]*

- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md), as well as this entire template.
- [x] I have made only one major change in my proposed changes.
- [ ] I have commented my proposed changes within the code.
- [ ] I have tested my proposed changes, and have included unit tests where possible.
- [x] I am willing to help maintain this change if there are issues with it later.
- [x] I give this submission freely and claim no ownership.
- [ ] It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
- [x] I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

Please make sure you [Sign Off](https://docs.pi-hole.net/guides/github/how-to-signoff/) all commits. Pi-hole enforces the [DCO](https://docs.pi-hole.net/guides/github/contributing/).

---
**What does this PR aim to accomplish?:**
*A detailed description, screenshots (if necessary), as well as links to any relevant GitHub issues*
See issue https://github.com/pi-hole/pi-hole/issues/3664

**How does this PR accomplish the above?:**
*A detailed description (such as a changelog) and screenshots (if necessary) of the implemented fix*
Change the default value to match the  value it is checked against

**What documentation changes (if any) are needed to support this PR?:**
*A detailed list of any necessary changes*

---
* You must follow the template instructions. Failure to do so will result in your pull request being closed.
* Please respect that Pi-hole is developed by volunteers, who can only reply in their spare time.

PR #3716, opened 2020-08-30
Fix chronometer default temperature

SamuelHill:
default temperature should be Celcius (was already, but this wasn't being actually happening as the 'C' was a 'c')


yubiuser:
Thanks for the PR.

This has been fixed (https://github.com/pi-hole/pi-hole/pull/3665) in the development branch already and will be included in the next version of pihole.

P.S. All PRs should be made against the `development` branch, not `master`

Issue #3728, opened 2020-09-04
Unused log dir /var/log/pihole


moranaga:
### Versions

* Pi-hole: 5.1.2
* AdminLTE: 5.1.1
* FTL: 5.2

### Platform

* OS and version: Debian 10.5 (Buster) x86_64
* Platform: VMware ESXi 7 VM

### Expected behavior

Log files are placed in a dedicated log dir below /var/log

### Actual behavior / bug

On start of service pihole-FTL a dedicated log dir is created, but apparently never used. Log files are placed on the same level:

**/etc/init.d/pihole-FTL**
```
    # Touch files to ensure they exist (create if non-existing, preserve if existing)
    touch /var/log/pihole-FTL.log /var/log/pihole.log
    [...]
    mkdir -p /var/log/pihole
    chown pihole:pihole /run/pihole /var/log/pihole
    [...]
    chown pihole:pihole /var/log/pihole-FTL.log /var/log/pihole.log
```

### Steps to reproduce

Start / restart of service pihole-FTL.

### Additional context

Ist that on purpose? If not, the suggestion would be to move these log files into the dedicated directory.

dschaper:
It's just how it's always been.

As for a feature request: 

We ask that you open all Feature Requests at our [Discourse Forum](https://discourse.pi-hole.net/c/feature-requests/8). 

Thank you for your understanding.

Issue #3629, opened 2020-07-29
pi-hole no longer logging to /var/log/pihole/


phrostbit:
PiHole is no longer logging to /var/log/pihole/ but I can see logs being generated when looking at them in the web GUI. I looked through the release notes and didn't see anything indicating if this was modified.

Recently updated to 
![image](https://user-images.githubusercontent.com/27188998/88832793-2aac9e00-d186-11ea-8ce2-0f5c608910b6.png)

![image](https://user-images.githubusercontent.com/27188998/88832675-0486fe00-d186-11ea-8b0f-67b2e4b3a4dc.png)

![image](https://user-images.githubusercontent.com/27188998/88832842-3dbf6e00-d186-11ea-9c5c-3371ae852ab6.png)

It appears the log files were moved back one directory. Was this intentional? 

![image](https://user-images.githubusercontent.com/27188998/88833099-9a228d80-d186-11ea-954b-13888a5596c0.png)

dschaper:
Is there any space left?

phrostbit:
Yes, over 30GB of space free.

dschaper:
`df -h` please.

dschaper:
Have you changed the user that `pihole-FTL` is running under? If it's still the `pihole` user then it won't be able to write to a file that has no write permissions for it.

phrostbit:
No user or permissions were changed. I updated to the latest release last week and was looking around my SIEM today and noticed there were no events. Compared the file monitor for my SIEM that was set from previous version and the log files from pihole moved back one directory in comparison. I made no modifications other than the upgrade to either pihole or my SIEM.

![image](https://user-images.githubusercontent.com/27188998/88834781-2c2b9580-d189-11ea-89a2-aea04fda20c3.png)

phrostbit:
I apologize. I'm unable to provide a debug token at this time. Working from home logs data I don't want to be seen.

dschaper:
We haven't change logfile locations in years. It's always been files in the `/var/log` directory itself.

dschaper:
You can still fill out the template, typically we autoclose issues that don't follow the basic directions.

phrostbit:
Well, that's what mine started with from initial install and not something I did manually. That's even what the technical add-on provided on Splunkbase references for the logging directory in their most recent release which was May 2020. I was just making sure you guys were aware if something has changed. I can easily modify my SIEM configuration to reflect the change I experienced. If this is expected behavior, then please disregard.

[...]

PR #3729, opened 2020-09-04
Remove creation of unused dir /var/log/pihole

yubiuser:
Don't create directory `/var/log/pihole` on start of `pihole-FTL.service` as it is never used for logging (all happens on `/var/log`).

Fixes https://github.com/pi-hole/pi-hole/issues/3728 and https://github.com/pi-hole/pi-hole/issues/3629

### DIFF
--- a/advanced/Templates/pihole-FTL.service
+++ b/advanced/Templates/pihole-FTL.service
@@ -27,8 +27,7 @@ start() {
     touch /run/pihole-FTL.pid /run/pihole-FTL.port
     touch /etc/pihole/dhcp.leases
     mkdir -p /run/pihole
-    mkdir -p /var/log/pihole
-    chown pihole:pihole /run/pihole /var/log/pihole
+    chown pihole:pihole /run/pihole
     # Remove possible leftovers from previous pihole-FTL processes
     rm -f /dev/shm/FTL-* 2> /dev/null
     rm /run/pihole/FTL.sock 2> /dev/null


dschaper:
Better solution would be to actually use the directory.

yubiuser:
True. But there might be a reason why this has never happend till now?

The dir was introduced 3 years ago... https://github.com/pi-hole/pi-hole/commit/4efb533d2eea6ea2b40729a6c4d59be411b1dc9e

dschaper:
No reason, no one's written the code to move things to be under the directory.

yubiuser:
Yeah... that's a lot more work than this PR.

dschaper:
Which is the proper way to resolve this.

dschaper:
Take it out to put it back in later? I think it's fine as it is, it's not affecting anything.

yubiuser:
> Take it out to put it back in later?

Yes. Later or ....never..?

Ok, then I'll close the PR.

dschaper:
> Yes. Later or ....never..?

Feel free to write the code now if we're not moving fast enough for your liking.

PromoFaux:
https://github.com/pi-hole/pi-hole/pull/3732
https://github.com/pi-hole/AdminLTE/pull/1577
https://github.com/pi-hole/FTL/pull/882
https://github.com/pi-hole/docs/pull/379

PR #3736, opened 2020-09-06
Update index.php

michi-zuri:
Why? When forwarding the web interface of pihole through ssh tunnel to localhost:xx I didn't get the splash page. With this change I can access the splash page correctly and don't get told, that localhost was blocked because "it wasn't found on any blacklist". Whitelisting localhost did not work, but this small change does the trick.


yubiuser:
All PRs should be done against development branch. Please PR again.

PR #3737, opened 2020-09-06
Update index.php

michi-zuri:
As requested.
See #3736

### DIFF
--- a/advanced/index.php
+++ b/advanced/index.php
@@ -24,7 +24,7 @@ unset($setupVars);
 $landPage = "../landing.php";
 
 // Define array for hostnames to be accepted as self address for splash page
-$authorizedHosts = ["localhost"];
+$authorizedHosts = [ "localhost" ];
 if (!empty($_SERVER["FQDN"])) {
     // If setenv.add-environment = ("fqdn" => "true") is configured in lighttpd,
     // append $serverName to $authorizedHosts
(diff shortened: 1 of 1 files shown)


PromoFaux:
Seems sane, but please sign off your commit so that the DCO bot passes... see here for instructions:

https://github.com/pi-hole/pi-hole/pull/3737/checks?check_run_id=1077567936

PromoFaux (approved):

PromoFaux:
Thanks. I'll add this in because it is an innocuous change, but FYI, this can be avoided by using the built in `pi.hole` or typing the IP address directly. In either of those cases, the request will be redirected to `/admin` automagically. No harm in including `localhost` in that, though!

michi-zuri:
Cool, thanks for approving. Since I'm tunneling port 80 of my pi to localhost:4653 (on my laptop) I didn't really have the option to use an ip address or pi.hole for the pi, otherwise I wouldn't need the tunnel in the first place. Next project is to figure out how to tunnel port 53 to localhost, but that seems more complicated with udp and all... Goal is to use pihole behind a NAT while on the road.

PromoFaux:
VPN probably easier! https://docs.pi-hole.net/guides/vpn/overview/

If you want to get really fancy, something like this: https://www.reddit.com/r/pihole/comments/fy8zmu/pihole_traefik_dnsovertls/

PR #3751, opened 2020-09-15
Allow configuration of dnsmasq log path.

Smarre:

---
In some setups, dnsmasq may be needed to log out somewhere else than hardcoded /var/log/pihole.log. 

This commit parses the path from dnsmasq’s configuration file, or defaults to /var/log/pihole.log if no path has been set.

- None.

---

### DIFF
--- a/advanced/Scripts/piholeLogFlush.sh
+++ b/advanced/Scripts/piholeLogFlush.sh
@@ -25,8 +25,15 @@ if [ -z "$DBFILE" ]; then
     DBFILE="/etc/pihole/pihole-FTL.db"
 fi
 
+# Find out dnsmasq log file.
+dnsmasq_log=$(awk 'match($0, /log-facility=(.*)/, a) {print a[1]}' /etc/dnsmasq.d/01-pihole.conf)
+if [ -z "$dnsmasq_log" ]; then
+  # If no config set, use default log location.
+  dnsmasq_log=/var/log/pihole.log
+fi
+
 if [[ "$@" != *"quiet"* ]]; then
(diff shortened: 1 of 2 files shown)


DL6ER:
There are some more locations where the file is hard-coded, especially the service script is critical as starting `pihole-FTL` may fail if it cannot write to the log file:

https://github.com/pi-hole/pi-hole/blob/6b536b7428a1f57ff34ddc444ded6d3a62b00a38/advanced/Templates/pihole-FTL.service#L38

Note that there is also https://github.com/pi-hole/pi-hole/pull/3732 which is currently under review. It is directly related to your suggested change.

PromoFaux:
Just out of curiosity, what would be the use case for having the log file be somewhere else?

Smarre:
My rationale was probably same as in the linked PR, that is, to make some sense in the log paths. I’ve had this change (along with other supporting changes it seems) locally for some time now, and finally got around to upstream it.

Pi-hole probably shouldn’t be used so, but there is a faint possibility that there is some managemental baggage why a log should be saved to some specific location. I don’t really see that justifying this change though, so if you think there is no need for this configurability, I’m completely fine with closing this one in favor of #3732.

PromoFaux:
I think I would be more comfortable (for now) fixing the paths so that they're defaulted to `/var/log/pihole/*.log`. 

I'm not 100% opposed to making the log location configurable in future, but it would need to be reflected across the three repos (this one, adminLTE, and FTL) - see linked PRs in #3732, there are quite a few places where the log file is referred to outside of the core repo.

DL6ER:
When they are all inside `/var/log/pihole`, you are free to `mount` this directory from anywhere

PR #3768, opened 2020-10-01
Use non-github link for StevenBlack/hosts

lakshayg:

---
Prevent downloading StevenBlack/hosts when there are no updates to the list

The github link for StevenBlack/hosts does not provide a `Last-Modified` header,
which requires the file to be downloaded everytime gravity is updated even if
there's no change to the file since the last update. The new link, taken from
the StevenBlack/hosts repository provides that header

None

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1298,7 +1298,7 @@ chooseBlocklists() {
 # in installDefaultBlocklists
 appendToListsFile() {
     case $1 in
-        StevenBlack  )  echo "https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts" >> "${adlistFile}";;
+        StevenBlack  )  echo "http://sbc.io/hosts/hosts" >> "${adlistFile}";;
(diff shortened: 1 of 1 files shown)


PromoFaux:
My concern with this is that the non-github link is just plain `http`, which I feel is a step backwards.

lakshayg:
I agree, that is a valid concern https://github.com/StevenBlack/hosts/issues/878

PromoFaux:
Ah, OK. I've shared a link with him in that issue thread that may help move things along...!

yyyopio (approved):

yyyopio (approved):

DL6ER:
Unfortunately, still no movement to HTTPS. Also, I'll rather not merge this PR as the new adlist target has been mentioned as an "aging Windows server" the maintainer would rather like to see going silent forever.

It gives me the feeling that if the server fails for some reason (e.g., hardware failure), it will not return into operation - or at least not quickly. This would leave the majority of users with an forever outdated list. We can make a new PR once HTTPS is available upstream (we should all get notified by Github if he writes something in there).

PR #3825, opened 2020-10-29
Change icon from cross to [i] for root user check

yubiuser:

---
Change the icon from cross to [i] when performing root user check. As discussed [here](https://discourse.pi-hole.net/t/pi-hole-5-probleme-in-kombination-mit-fritzbox-7490/32665/142?u=yubiuser)

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -2613,7 +2613,7 @@ main() {
     # Otherwise,
     else
         # They do not have enough privileges, so let the user know
-        printf "  %b %s\\n" "${CROSS}" "${str}"
+        printf "  %b %s\\n" "${INFO}" "${str}"
         printf "  %b %bScript called with non-root privileges%b\\n" "${INFO}" "${COL_LIGHT_RED}" "${COL_NC}"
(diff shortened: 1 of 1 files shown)


dschaper:
I think this needs to stay a cross. You can not run the script as non-sudo so it is indeed an error.

yubiuser:
At this points it just informs the user that it needs root privileges, the check for sudo comes a few lines later and will fail if sudo does not exist.

dschaper:
Yes, it's still an error.

DL6ER:
I agree with @yubiuser about this change. If the user runs this as non-root (the typical scenario), it will just re-`curl` and `sudo bash` the installer. If this does not work (because the user doesn't know their password, etc.), `sudo` will give them a much more useful output. If `sudo` is not available, the user will be informed about this fact (and a cross is shown).

PromoFaux (approved):

PromoFaux (approved):
See my comments here:
https://discourse.pi-hole.net/t/pi-hole-5-probleme-in-kombination-mit-fritzbox-7490/32665/143?u=promofaux

PR #3827, opened 2020-10-30
Re-run the script as root instead of piping to bash

hvnsweeting:

---
Stop piping bash command when install it via alternative method.

```
Piping to bash is controversial, as it prevents you from reading code that is about to run on your system. Therefore, we provide these alternative installation methods which allow code review before installation:
```

Re-invoke the script with sudo instead of pipe curl output to sudo bash.

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -2623,8 +2623,16 @@ main() {
         # If the sudo command exists,
         if is_command sudo ; then
             printf "%b  %b Sudo utility check\\n" "${OVER}"  "${TICK}"
-            # Download the install script and run it with admin rights
-            exec curl -sSL https://raw.githubusercontent.com/pi-hole/pi-hole/master/automated%20install/basic-install.sh | sudo bash "$@"
+
+            # when run via curl piping
+            if [[ "$0" == "bash" ]]; then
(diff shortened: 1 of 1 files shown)


automated install/basic-install.sh
```diff
@@ -2624,7 +2624,7 @@ main() {
         if is_command sudo ; then
             printf "%b  %b Sudo utility check\\n" "${OVER}"  "${TICK}"
             # Download the install script and run it with admin rights
-            exec curl -sSL https://raw.githubusercontent.com/pi-hole/pi-hole/master/automated%20install/basic-install.sh | sudo bash "$@"
+            exec sudo bash "$0" "$@"
```
dschaper, line 2627: Does this work when running the default install method of `curl | bash`?

PromoFaux:
Your latest push appears to work in both cases:

non-root user `curl`ing:
image

non-root user `wget`ing and executing locally:
image

Worth noting, that it should be moot anyway, as the alternative install method instructs the user to run `sudo bash basic-install.sh` after `wget`ing, or `git clone`ing

image

But always good to have an extra hardener

PromoFaux (approved):

PR #3830, opened 2020-11-01
Adding YouTube filter blocker (video ads blocker)

DanielAbsi:
For a next update, I recommended to make a blocker for video on youtube or /and implementing uBlock Origin filter to block even more unwanted content.


PromoFaux:
I'm not sure if you're just using this to practice using Github, but you're going about it the wrong way. Random pull requests from branches on the repo are not the way to get feature requests noticed.

Also, DNS blocking does not work like in-browser blockers. Google is your friend.

PR #3846, opened 2020-11-16
Push three IPv4 DNS server options via DHCP

deHakkelaar:
Push three IPv4 DNS servers via DHCP instead of a single one which is currently implemented.
This because some devices automatically populate a secondary or tertiary DNS if not supplied via DHCP.
Below example auto configures 8.8.8.8 for the secondary DNS on a OnePlus device:

https://discourse.pi-hole.net/t/removing-google-dns-after-using-pi-hole-dhcp/40435

Nmap output after the change:

```
$ sudo nmap -sU -p67 --script dhcp-discover 10.0.0.4
Starting Nmap 7.70 ( https://nmap.org ) at 2020-11-16 14:25 CET
Nmap scan report for ph5.dehakkelaar.nl (10.0.0.4)
Host is up (0.0012s latency).

PORT   STATE SERVICE
67/udp open  dhcps
| dhcp-discover:
|   DHCP Message Type: DHCPACK
|   Server Identifier: 10.0.0.4
|   Subnet Mask: 255.255.255.0
|   Broadcast Address: 10.0.0.255
|   Domain Name: dehakkelaar.nl
|   Router: 10.0.0.1
|_  Domain Name Server: 10.0.0.4, 10.0.0.4, 10.0.0.4
MAC Address: B8:27:EB:xx:xx:xx (Raspberry Pi Foundation)

Nmap done: 1 IP address (1 host up) scanned in 0.96 seconds
```

- [ X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md), as well as this entire template.
- [X ] I have tested my proposed changes, and have included unit tests where possible.
- [X ] I am willing to help maintain this change if there are issues with it later.
- [X ] I give this submission freely and claim no ownership.
- [X ] It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
- [X ] I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

---

---

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -364,6 +364,7 @@ ProcessDHCPSettings() {
 ###############################################################################
 dhcp-authoritative
 dhcp-range=${DHCP_START},${DHCP_END},${leasetime}
+dhcp-option=option:dns-server,0.0.0.0,0.0.0.0,0.0.0.0
 dhcp-option=option:router,${DHCP_ROUTER}
 dhcp-leasefile=/etc/pihole/dhcp.leases
 #quiet-dhcp


DL6ER:
@deHakkelaar Is there any particular reason for adding *three* DNS servers (all point to ourselves)?

---

It's also worth citing the special role of `0.0.0.0` in this context:
>  The special address 0.0.0.0 is taken to mean "the address of the machine running dnsmasq". 

(`man dnsmasq`)

deHakkelaar:
No not particular.
I guess two should suffice.
Its just to make absolutely sure no device has a third DNS server configured automatically.

I also have doubts about it being the default.
This will most likely conflict with users having configured DNS servers in other config files.
I'll have looksee if I can make it into an option.

DL6ER:
@deHakkelaar Do you have anything new for us?

deHakkelaar:
@DL6ER, no I haven't had time to look into it yet to make it an option.
I'll create another pull for that if succeed.
Will it be a useful option or should we expect users to configure this themselves in separate config files ?
What do you think ?

DL6ER:
It is definitely a useful thing and it should also be enabled by default.

You are right, the only conflict may be users defining other DNS servers in other config files. However, why would they do this? The only valid use-case for this seems to be a form of high-availability setup with multiple Pi-holes in your network. This keeping in mind that the Pi-hole DHCP server is also meant for if your router cannot be changed, so (a) anyone with a complex network setup will likely use another DHCP server and (b) anything else than other Pi-hole wouldn't make any sense here as this would defeat the purpose.

So my suggestion could be to provide a text-field on the web interface like `IP addresses of other Pi-holes to be handed out via DHCP` plus some explanatory text why leaving this field empty will announce the local Pi-hole multiple times. This rather than a simple checkbox.

jfb-pihole:
> a form of high-availability setup with multiple Pi-holes in your network.

This is not uncommon, where users have a router that doesn't allow DNS assignment (ISP routers in particular).  They use Pi-hole for DHCP and in the case of multiple Pi-holes it is not uncommon to see both as DHCP servers on non-overlapping ranges, each providing the DNS of itself and the opposite Pi-hole.  We should try to not lose the ability to do this.

DL6ER:
>  We should try to not lose the ability to do this.

Sure, hence the suggestion of making this a text box where you can put the IP of the other and, in absence of such a setup, broadcast the single pi-hole multiple times are proposed by @deHakkelaar 

This would even raise this kind of setup to be one that is supported via the UI instead of needing manual config file tinkering.

deHakkelaar:
What about three DNS fields in the DHCP section already default configured with 0.0.0.0 ?
One could alter any three to configure your own.

DL6ER:
If we add three, there will be users asking for four. I'd suggest a comma-separated list, prefilled with `0.0.0.0,0.0.0.0,0.0.0.0`

deHakkelaar:
That makes sense and sounds not to hard to implement.
Ps. I'm in the middle of two big migrations that need my attention first.
Guess I can spend time on this by the end of next month.

DL6ER:
Just to make your life a bit easier when you find the time to look at this. Your core changes should go here: https://github.com/pi-hole/pi-hole/blob/master/advanced/Scripts/webpage.sh

and should be something like

``` bash
AddDHCPDNSServers() {
    serverlist="${args[2]}"
    change_setting "DHCP_MULTIDNS" "${serverlist}"
}
```
and then something like
``` bash
    if [ -z ${DHCP_MULTIDNS} ]; then
        add_setting "DHCP_MUTLIPLE" "0.0.0.0,0.0.0.0,0.0.0.0"
    fi
    echo "dhcp-option=option:dns-server,${DHCP_MULTIDNS}" >> "${dhcpconfig}"
```
right in between these two lines `fi`: https://github.com/pi-hole/pi-hole/blob/41bdb741b7e1a70184ecd7dbfc23f3ebcaa5e3bc/advanced/Scripts/webpage.sh#L431-L432

Then, add `"adddhcpdnsservers"   ) AddDHCPDNSServers;;` at the end of this list (before the `*` case):
https://github.com/pi-hole/pi-hole/blob/41bdb741b7e1a70184ecd7dbfc23f3ebcaa5e3bc/advanced/Scripts/webpage.sh#L734

---

On the web interface, add the text field + short description to `settings.php` and then add the backend call to https://github.com/pi-hole/AdminLTE/blob/master/scripts/pi-hole/php/savesettings.php

Please verify the list to defend against malicious input by exploding it at `,` and then checking if each component using [`validIP($ip)`](https://github.com/pi-hole/AdminLTE/blob/780dff0e956e3cfa9d625f3b6e1024c049408b0e/scripts/pi-hole/php/savesettings.php#L14) which supports both IPv4 and IPv6.

deHakkelaar:
Much appreciated!

deHakkelaar:
Hi @DL6ER & @jfb-pihole , I have below two brewing that I would like to ask your input for:

https://github.com/pi-hole/pi-hole/compare/development...deHakkelaar:patch-7

https://github.com/pi-hole/AdminLTE/compare/devel...deHakkelaar:patch-1

image

```
pi@ph5:~ $ pihole-FTL dhcp-discover
[..]
   dns-server: 10.0.0.4
   dns-server: 10.0.0.4
   dns-server: 10.0.0.1
   dns-server: 192.168.1.1
   dns-server: 127.0.0.1
[..]
```

Works with a fresh `setupVars.conf` and also when already populated with the newly `DHCP_DNS=` variable
Default while enabling the IPv4 DHCP service via the GUI is `0.0.0.0,0.0.0.0,0.0.0.0` and can be altered.
I wasnt able to check for IPv4 only with the `validIP()` function so had to make up my own.
This because I found out that you cant advertise an IPv6 address with the IPv4 `dhcp-option=option:dns-server` directive.
For that, you need the IPv6 `dhcp-option=option6:dns-server` directive.

Also I cut some corners like not creating a `AddDHCPDNSServers()` function because I believe the way I did it, this function isnt necessary ... not sure though ?
The `ProcessDHCPSettings()` always poops out a `dhcp-option=option:dns-server` line to simplify.

Now the tricky bit, I am not sure how to superimpose this to the IPv6 side.
I have no IPv6 setup on my LAN for testing and am still too unfamiliar with it to be sure.
I see the DHCPv6 DNS is set below:

https://github.com/pi-hole/pi-hole/blob/debc9a69c50ce9d91f102792a10ae39a67414152/advanced/Scripts/webpage.sh#L425-L428

Should we create a separate IPv6 DHCP DNS servers field on the GUI so you can edit them similar as for IPv4 ?
And where should we position this new IPv6 field ? Somewhere under the "Advanced DHCP settings" section ... idk ?
I welcome your input ?

deHakkelaar:
I also added a hint:
image

@DL6ER & @jfb-pihole  , did you find time to have looksee ?
Shall I just proceed pulling the IPv4 part first as is ?

JJS777 (approved):

JJS777 (approved):

JJS777 (approved):

DL6ER:
I don't think the hint is needed in its current (long) form - or actually not at all. Do we need to be able to change the setting for DHCPv6 as well? If we don't see a need for this, we shouldn't add an option (only because of symmetry, etc.). More input fields give more surface for possibly brickingthe configuration.

DL6ER:
Actually, thinking again about, it: Do we need the input field at all? For the purpose of pushing out the DNS server multiple times, we could also have a checkbox ("push DNS server multiple times" or similar). In my network, this seems not to be needed, you mentioned a specific phone where you observed the strange behavior when there is only DNS server being pushed.
No input field would mean a simpler (and maybe more robust) implementation.

This is merely thinking out aloud, just to mention the idea. No need to give up on the input field if you still think this should be the way to go. I'm trying to estimate what the general user might think when seeing this (may feel "complicated").

deHakkelaar:
I agree, every added setting creates a larger puzzle for users to mess up.
Maybe this option belongs under the "Advanced DHCP settings" section to better distinct that this setting is not one that needs changing OOTB if dont have a HA setup ?
But I do believe the three DNS servers will solve allot of those mysteries where users are complaining their IoT devices etc are bypassing Pi-hole.
Allot are blaming Google/Android of purposely adding their DNS servers contributing to plot theories of Google wanting to track etc.
I believe stock Android doesnt do this (I have two Android devices).
But some other brands do so it seems.
I believe the sole purpose of this is to maintain redundancy if only one DNS is supplied via DHCP.

And I believe it should not be an option/checkbox but should be the default to push three.
Currently some users redirect all port 53 traffic to their preferred DNS servers (Pi-hole) using the firewall.
This would become less of an issue/redundant if three DNS servers were pushed thus simplifying things considerably.

About DHCPv6, we need someones opinion/knowledge who actually is setup with IPv6 on their LAN.
I was asking because I dont have IPv6 on my LAN and am a bit reluctant to change my home setup for testing purposes.
Maybe we should get a larger audience and ask opinions/advice on Discourse ?

Also I will look into some more things to familiarise with how all those DHCP settings (IPv4 + IPv6) are tied together.
I discover something new every time I check the man pages ;)

yubiuser:
I would agree with @deHakkelaar to make it an input field rather than a checkbox if this feature is going to implemented. Users who are so advanced to need/use/want this feature will immediately ask to make this configurable if it would be a checkbox only. 

I agree, this should be moved to "Advanced DHCP settings". 
Is there a way to make the default not "0.0.0.0" but the actual IP Pi-hole is listing on?

deHakkelaar:
Actually the way ```pihole-FTL``` is configured now, its not listening to any IP in particular.
Its listening on all IP's ```0.0.0.0``` with a default ```dnsmasq``` configuration:

```
pi@ph5:~ $ man dnsmasq
[..]
       The dnsmasq DHCP server supports static address  assignments  and
       multiple  networks. It automatically sends a sensible default set
       of DHCP options, and can be configured to send any desired set of
       DHCP  options .....
```

If the GUI were to default display the own IP three times for DNS, the only source I can think of is the IP thats in the ```setupVars.conf``` file:

```
pi@ph5:~ $ cat /etc/pihole/setupVars.conf
[..]
IPV4_ADDRESS=10.0.0.4/24
IPV6_ADDRESS=
```

But that would not be an actual representation of how ```pihole-FTL``` is actually listening on all ```0.0.0.0```.
I prefer to keep ```0.0.0.0``` as it simplifies things and I like simple ;)

Bucking-Horn:
As far as Stateful DHCPv6 is concerned, the options should be set equivalent, using `[::]` instead of  `0.0.0.0`, e.g. `dhcp-option=option6:dns-server,[::]` (and that should cover Stateless DHCPv6 as well).

The unknown here is that I am not aware of clients that would misbehave for IPv6 in the same way they would for IPv4, i.e. inventing a secondary DNS if none is provided. 
So we (or at least I) don't know if there is an actual demand for IPv6.

I'd definitely support `0.0.0.0` and `[::]` over actual IPs, as _`dnsmasq`_ would handle that transparently:
> The special address 0.0.0.0 is taken to mean "the address of the machine running dnsmasq"
(from _`dnsmasq`_ man)

What's not clear by that sentence: 
Would _`dnsmasq`_ even do so in multi subnet configurations? 
Perhaps someone with such a network configuration could check that?

If that would be the case, using those "wildcard" addresses would be clearly superior.

deHakkelaar:
Yeah, good one.
I assigned a second aliased IP ```10.0.0.7``` to my Pi-hole setup:

```
pi@ph5:~ $ ip -br -4 a
lo               UNKNOWN        127.0.0.1/8
eth0             UP             10.0.0.4/24
eth0.0@eth0      UP             10.0.0.7/24
```

And run the discovery:

```
pi@ph5:~ $ pihole-FTL dhcp-discover
Scanning all your interfaces for DHCP servers
Timeout: 10 seconds
[..]
* Received 311 bytes from eth0.0:10.0.0.7
[..]
   dns-server: 10.0.0.7
   dns-server: 10.0.0.7
   dns-server: 10.0.0.7
   router: 10.0.0.1
   --- end of options ---
[..]
* Received 311 bytes from eth0:10.0.0.4
[..]
   dns-server: 10.0.0.4
   dns-server: 10.0.0.4
   dns-server: 10.0.0.4
   router: 10.0.0.1
   --- end of options ---
```

So it seems embedded ```dnsmasq``` is aware on which interface the DHCP request was received and advertises the associated IP for DNS accordingly.
Thats decided then, ```0.0.0.0``` or ```[::]``` three times will be preferred default instead of an actual IP which would complicate allot ?

The DNS entry field should be under the advanced section as to better distinct that this setting doesnt need changing if dont run a HA setup.

The DNS option will need to be two separate fields for IPv4 and IPv6 because of the ```option:dns-server``` directive not being able to advertise IPv6 addresses and vice versa for the ```option6:dns-server``` directive.

And I'll drop the hint.

If no more advice/hints/requests, I'll cook up something new ?

DL6ER:
@deHakkelaar Sorry, I forgot about this change. Can you give us a quick update on the current status?

deHakkelaar:
@DL6ER , I got lost in setting up a sandbox Raspi with btrfs for snaps.
I'll renew my effort and update asap.

deHakkelaar:
Hi folks, I saw the Discourse mention in my mail but want to say that currently I dont have the resources to continue.
I dont want to tinker with my live setup and dont have time, budget and energy to setup a sandbox.
Plus I have a medical condition which restricts my daily activities considerably.
It doesnt mean that I might pick it up if circumstances changes.
Sorry for the inconvenience.

I have little experience with Github, should I just close this one for now or should I leave it open?

dunxd:
This would be very helpful, since many Android implementations (e.g. OnePlus) add 8.8.8.8 as a second DNS server if only one is configured by DHCP. Currently, those that figure out why their android devices are sometimes bypassing pihole have to create a new dnsmasq.d config file.

Note that anyone that is running two pihole servers in the same network may need to do the same thing if one of the pihole servers is acting as DHCP server. Would the approach in this pull request conflict with that, or simply pre-populate the dnsmasq settings that could then be manually overridden for the purposes of second pihole?

dschaper:
We can close this and then either reopen or open a new PR when it can be addressed again.

DL6ER:
Implemented for Pi-hole v6.0 in

https://github.com/pi-hole/FTL/commit/6a199427b43fdc54d4a1ed9fa6d76e08338024e6

Issue #3845, opened 2020-11-15
Add an Response for "pihole blacklist --nuke"


DjCrays:
### Versions
  Pi-hole version is v5.1.2 (Latest: v5.1.2)
  AdminLTE version is v5.1.1 (Latest: v5.1.1)
  FTL version is v5.2 (Latest: v5.2)

### Platform

- OS and version:  Raspbian 10 buster
- Platform:  Rasberry Pi

### Expected behavior

_A clear and concise description of what you expected to happen._

When using pihole blacklist --nuke i should get something like "done - Black list is now empty" or something like this

### Actual behavior / bug

Just shows just the help screen for pihole blacklist -h.

### Steps to reproduce

1. Put something in the blacklist.
2. In an shell just type :" pihole blacklist --nuke"
3. See the response

## Debug Token

- URL:   https://tricorder.pi-hole.net/s1rulqs8gp

dschaper:
Are you sure you're not overlooking a warning about readonly databases?

What is the exact output including your input?

DjCrays:
Yes i am sure. :) But here is the exact output:

```
pi@raspberrypi:~ $ pihole blacklist --nuke
Usage: pihole -b [options] <domain> <domain2 ...>
Example: 'pihole -b site.com', or 'pihole -b site1.com site2.com'
Blacklist one or more domains

Options:
  -d, --delmode       Remove domain(s) from the blacklist
  -nr, --noreload     Update blacklist without reloading the DNS server
  -q, --quiet         Make output less verbose
  -h, --help          Show this help dialog
  -l, --list          Display all your blacklistlisted domains
  --nuke              Removes all entries in a list
pi@raspberrypi:~ $ 

```

dschaper:
Okay, try `sudo bash -x /opt/pihole/list.sh blacklist --nuke`

DjCrays:
okay, done: 
```
pi@raspberrypi:~ $ sudo bash -x /opt/pihole/list.sh blacklist --nuke
+ basename=pihole
+ piholeDir=/etc/pihole
+ gravityDBfile=/etc/pihole/gravity.db
+ reload=false
+ addmode=true
+ verbose=true
+ wildcard=false
+ web=false
+ domList=()
+ typeId=
+ comment=
+ declare -i domaincount
+ domaincount=0
+ colfile=/opt/pihole/COL_TABLE
+ source /opt/pihole/COL_TABLE
++ [[ -t 1 ]]
+++ tput colors
[...]

PR #3847, opened 2020-11-16
Add some output to the --nuke command

PromoFaux:

---

Addresses #3845

output after PR:

image

### DIFF
--- a/advanced/Scripts/list.sh
+++ b/advanced/Scripts/list.sh
@@ -231,7 +231,14 @@ Displaylist() {
 }
 
 NukeList() {
-    sqlite3 "${gravityDBfile}" "DELETE FROM domainlist WHERE type = ${typeId};"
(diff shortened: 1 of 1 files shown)


DL6ER (commented):
Shouldn't it be

> Removed ${count} domain(s) from **the** ${listname}

?

Also, why printing the number two times? Maybe

> Removed **all** domains from ...

PromoFaux:
>**the** ${listname}

I agree with that, oversight on my part.

What I was going for with the latter was to say "there are `n` number of domains in total on the ${listname}, and now I have removed `n` domains", rather than just "I removed some domains" I think it would probably make more sense with some if conditions in there and a check after the delete command has run.. but likely overkill

PromoFaux:
Force pushed some changes

image

DL6ER (approved):

PR #990, opened 2016-12-16
Don't kill setupVars.conf on update/fresh install

DL6ER:
**By submitting this pull request, I confirm the following (please check boxes, eg [X])Failure to fill the template will close your PR:**

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] I have checked that [another pull request](https://github.com/pi-hole/pi-hole/pulls) for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:**

- [X] 10 (very familiar)

---
Previous behavior was to remove `setupVars.conf` and write it from scratch. This didn't preserve any settings.

Test: I create the file `/etc/pihole/setupVars.conf` before running the installer:
<pre>
root@ubuntu-512mb-fra1-01: ~# cat /etc/pihole/setupVars.conf
<b>SOME_SETTING=abc</b>
<b>PIHOLE_INTERFACE=something_very_bad</b>
root@ubuntu-512mb-fra1-01: ~#
</pre>

Without this PR after the install:
<pre>
root@ubuntu-512mb-fra1-01: ~# cat /etc/pihole/setupVars.conf
<b>PIHOLE_INTERFACE=eth0</b>
IPV4_ADDRESS=...
IPV6_ADDRESS=...
PIHOLE_DNS_1=8.8.8.8
PIHOLE_DNS_2=8.8.4.4
QUERY_LOGGING=true
root@ubuntu-512mb-fra1-01: ~#
</pre>
------> Interface has been corrected, but other settings are lost

With this PR after the install:
<pre>
root@ubuntu-512mb-fra1-01: ~# cat /etc/pihole/setupVars.conf
<b>SOME_SETTING=abc</b>
<b>PIHOLE_INTERFACE=eth0</b>
IPV4_ADDRESS=...
IPV6_ADDRESS=...
PIHOLE_DNS_1=8.8.8.8
PIHOLE_DNS_2=8.8.4.4
QUERY_LOGGING=true
root@ubuntu-512mb-fra1-01: ~#
</pre>
------> Interface has been corrected and other settings are preserved

_This template was created based on the work of [`udemy-dl`](https://github.com/nishad/udemy-dl/blob/master/LICENSE)._

PR #3874, opened 2020-11-29
Don't delete entries including and greater than `PIHOLE_DNS_10` on update/repair

### LINKED ISSUE #3873: DNS servers get lost upon repair/update
### Versions
- Pi-hole: 5.2
- AdminLTE: 5.2
- FTL: 5.3.1

### Platform
- OS and version: Ubuntu 20.04
- Platform: Rapsberry Pi

### Expected behavior

All 12 checked DNS servers in settings stay checked after an repair or update.

### Actual behavior / bug

After a `pihole -r` (repair mode) or `pihole -up` the DNS servers are trimmed to the first 9 servers. You can also verify that with the entries in `setupVars.conf`.

### Steps to reproduce

1. Go to `/admin/settings.php?tab=dns`
2. Select 12 DNS servers of your choice
3. Save the settings (also check `setupVars.conf`)
4. Call `pihole -r` with "repair" or `pihole -up` (if there's an update)

## Debug Token

- URL: https://tricorder.pi-hole.net/ykw0i771l1

### Screenshots

Before:
Screenshot_20201129_151106

After steps to reproduce:
Screenshot_20201129_151005

PromoFaux:

---

Fixes #3873 

Currently the `sed` has no word boundry in place, so the `PIHOLE_DNS_1/d` is matching `PIHOLE_DNS_10`, `PIHOLE_DNS_11` etc.

I guess this has always been the case for at least the last 5 years when the function was introduced in #990 - so it's probably an edge case. I don't fully understand why one would select _all_ the DNS servers, but there is no harm in this, and I don't see why we should kill them off on every update/repair

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1913,7 +1913,7 @@ finalExports() {
     # If the setup variable file exists,
     if [[ -e "${setupVars}" ]]; then
         # update the variables in the file
-        sed -i.update.bak '/PIHOLE_INTERFACE/d;/IPV4_ADDRESS/d;/IPV6_ADDRESS/d;/PIHOLE_DNS_1/d;/PIHOLE_DNS_2/d;/QUERY_LOGGING/d;/INSTALL_WEB_SERVER/d;/INSTALL_WEB_INTERFACE/d;/LIGHTTPD_ENABLED/d;/CACHE_SIZE/d;' "${setupVars}"
+        sed -i.update.bak '/PIHOLE_INTERFACE/d;/IPV4_ADDRESS/d;/IPV6_ADDRESS/d;/PIHOLE_DNS_1\b/d;/PIHOLE_DNS_2\b/d;/QUERY_LOGGING/d;/INSTALL_WEB_SERVER/d;/INSTALL_WEB_INTERFACE/d;/LIGHTTPD_ENABLED/d;/CACHE_SIZE/d;' "${setupVars}"
     fi
     # echo the information to the user
     {


DL6ER (approved):

midoo844 (commented):
Thanks

dandjo:
> I don't fully understand why one would select _all_ the DNS servers, ...

Because there are a lot of reasons spreading DNS queries (redundancy [routing problems], "special" entries the "big" provider do not resolve, ...).
Screenshot_20201213_154445

dschaper:
`pihole-FTL` doesn't spread DNS queries. It picks the one with the lowest latency. (And checks that latency every `n` queries to see if there is a better one.)

https://docs.pi-hole.net/ftldns/dns-resolver/#improve-detection-algorithm-for-determining-the-best-forward-destination

dandjo:
Okay, this implementation is new to me, thanks for sharing. But also in this case multiple fallbacks make sense if (in some circumstances) the route to the currently used forward destination is slower than the others.

Do I get it right, that FTL does not ask the other defined servers if it receives NXDOMAIN from the selected one?

dschaper:
No, an NXDOMAIN response is an active NOERROR response from the queried DNS server.

dschaper:
As an aside, this has been the behavior from the start. It's how `dnsmasq` works, we've just made the time periods to be longer so the re-rank of upstream DNS servers happens less frequently.

DL6ER:
> so the re-rank of upstream DNS servers happens less frequently

Just to add that we did this for two reasons: Privacy and bandwidth saving. The original `dnsmasq` code does this re-ranking EVERY 10 seconds or EVERY 50 queries (whatever comes first). This seemed *way* too aggressive for us.

PR #3887, opened 2020-12-01
Fix "Never forward non-FQDNs" checkbox

DL6ER:

---

When there is a Pi-hole domain set and "Never forward non-FQDNs" is ticked, we add `local=/$domain/` to tell FTL that this domain is purely local and FTL may answer queries from `/etc/hosts` or DHCP but should never forward queries on that domain to any upstream servers.

Saving the changed checkbox can currently be seen as broken because it requires you to change the box, save the page and then go to the DHCP settings page and also save this one (w/ or w/o changes doesn't matter).

This bug has been reported on Discourse (see link by Pralor below)

Always re-process the DHCP settings when saving the DNS settings to avoid users having to click twice.

None


PromoFaux (approved):
Ignore CircleCI, I was playing with something last night

Issue #3911, opened 2020-12-06
basic auth is not working


mdujava:
If I set up adlist address as: https://username:password@fqdn/dir/file.list, `pihole -g` or update on the web will always fail with `Invalid Target` for this adlist. On lines mentioned below `@` is not considered as usable character.

https://github.com/pi-hole/pi-hole/blob/0d8ece1be249fdc5993f01a45f8d0464257df819/gravity.sh#L397-L403

My local workaround was to add `@` to regex.

dschaper:
Because of the number of support requests and the very limited resources we have as a free open-source project run by volunteers, we will be closing all Issues that do not follow all instruction in the Issue Template and do not provide debug tokens if applicable. Thank you for your understanding.

Please search [open](https://github.com/pi-hole/pi-hole/issues) and [closed](https://github.com/pi-hole/pi-hole/issues?q=is:issue+is:closed) issues for others that may have your same question. Also, visit our [Discourse](https://discourse.pi-hole.net) site for additional assistance.

PR #3912, opened 2020-12-06
Fix validation of adlist url

mdujava:

---
This PR is fixing https://github.com/pi-hole/pi-hole/issues/3911 as last verification of url, before downloading adlist from it, has missing good character, therefor valid url with basicauth user:pass (eg. `https://user:pass@fqdn/dir/file.list`) is evaluated as invalid.

Regex which is used as validator is updated to accept `@` as valid character. As char `@` is missing, adding it to the regex will fix the issue.

This is fixing back-end URL validation and there is no change in user interaction, therefor no documentation change needed.

---

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -486,10 +486,15 @@ SetWebUITheme() {
 }
 
 CheckUrl(){
-    local regex
+    local regex check_url
     # Check for characters NOT allowed in URLs
-    regex="[^a-zA-Z0-9:/?&%=~._-]"
-    if [[ "${1}" =~ ${regex} ]]; then
+    regex="[^a-zA-Z0-9:/?&%=~._()-;]"
+
+    # this will remove first @ that is after schema and before domain
+    # \1 is optional schema, \2 is userinfo
+    check_url="$( sed -re 's#([^:/]*://)?([^/]+)@#\1\2#' <<< "$1" )"
+
+    if [[ "${check_url}" =~ ${regex} ]]; then
         return 1
     else
         return 0
--- a/gravity.sh
+++ b/gravity.sh
@@ -393,10 +393,15 @@ gravity_DownloadBlocklists() {
     esac
 
     echo -e "  ${INFO} Target: ${url}"
-    local regex
+    local regex check_url
     # Check for characters NOT allowed in URLs
     regex="[^a-zA-Z0-9:/?&%=~._()-;]"
-    if [[ "${url}" =~ ${regex} ]]; then
+
+    # this will remove first @ that is after schema and before domain
+    # \1 is optional schema, \2 is userinfo
+    check_url="$( sed -re 's#([^:/]*://)?([^/]+)@#\1\2#' <<< "$url" )"
+
+    if [[ "${check_url}" =~ ${regex} ]]; then
         echo -e "  ${CROSS} Invalid Target"
     else
        gravity_DownloadBlocklistFromUrl "${url}" "${cmd_ext}" "${agent}" "${sourceIDs[$i]}" "${saveLocation}" "${target}" "${compression}"


dschaper:
It's not a good character.

https://stackoverflow.com/a/19511469

You need to percent encode in your configuration, not allow an invalid character in the URL.

mdujava:
Hi, I rewrote the fix inspired from: https://github.com/pi-hole/pi-hole/blob/4a7ef7bb18a923b51cc3708d56de5bafbeb94a4f/gravity.sh#L634-L639 for removal of schema and bascauth before validating against unchanged regex for valid chars.

dschaper:
If you want `@` in the URI then you need to percent encode it yourself. Adding the `@` character is a no from me.

mdujava:
`curl https://user:password@fqdn.com/path` is working fine, lets encode it: `https://user:password%40fqdn.com/path` this will return `curl: (6)  Could not resolve host: user; Unknown error`

3.2.(1.)? section of RFC 3986 is allowing @ right before host for including User Information.

> authority   = [ userinfo "@" ] host [ ":" port ]

If Pi-hole should not support this feature why https://github.com/pi-hole/pi-hole/blob/4a7ef7bb18a923b51cc3708d56de5bafbeb94a4f/gravity.sh#L634 is acknowledging that it's valid part or URL.

dschaper:
> curl: (6) Could not resolve host: user;

That's probably because you didn't quote the URL? 

`curl "https://user:password%40fqdn.com/path"` ?

dschaper:
> If Pi-hole should not support this feature why

That tells me that code has been there for over 3 years, so either you're the first person in 3 years to find it doesn't work, or there's something else that needs to be addressed before changing the code.

mdujava:
still the same error code if using `curl 7.29.0 (armv7hl-redhat-linux-gnu) libcurl/7.29.0 NSS/3.53.1 zlib/1.2.7 libidn/1.28 libssh2/1.8.0`

on `curl 7.68.0` is different error: `curl: (3) URL using bad/illegal format or missing URL`

---

Yeah I am probably the first one. I want to host mine adlist, but I don't want it to be public, so for authentication I am using basicauth.

There should not be url encoding of @: 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#Access_using_credentials_in_the_URL
https://en.wikipedia.org/wiki/Basic_access_authentication#URL_encoding
https://en.wikipedia.org/wiki/URL#Syntax

@ is reserved char per https://tools.ietf.org/html/rfc3986#section-2.2, you are right that you should not have @ in the URL path/parameters part.

Per 2.2 @ is reserved char, that can be used in URL at most one time and few more restrictions. Therefor checking existing goodchars only in `fqdn/path?args#fragment` was easier/quicker to implement.

I think I can not provide more information than this. If this is not sufficient, feel free to close this PR.

DL6ER:
An alternative solution would be to check for `@` only within the allowed URL path, i.e., between `://` and the next `/`. We can just replace it by a valud char (say `X`) it if it appears only once. I'm thinking about something like
``` bash
url="https://user:pass@fqdn/dir/file.list"
check_url="$(sed -r "s#(.*)@#\1X#" <<< "${url}")"
```
the result is `https://user:pass`**`X`**`fqdn/dir/file.list` which would pass the original regex we have in place (test the output of this, not `$url` itself!).

Note that it will only replace one `@` so `https://user:pass@@fqdn/dir/file.list` would be rejected by the regex as one `@` would remain in the output. If `$url` does not contain any `@`, the `sed` will just echo what you put in.

What do you think?

mdujava:
> ```shell
> url="https://user:pass@fqdn/dir/file.list"
> check_url="$(sed -r "s#(.*)@#\1X#" <<< "${url}")"
> ```
This is not consicering the placement of /.

> What do you think?

This looks way better. I will play with it to ensure the place of @.

mdujava:
> It's not a good character.
> 
> https://stackoverflow.com/a/19511469

both ways are possible as first answer shows, you can encode it as %40 or not even in path section.

Both of those URLs will redirect you to wikipedia page "@".

- http://en.wikipedia.org/wiki/%22@%22
- http://en.wikipedia.org/wiki/%22%40%22

dschaper:
And does `curl` accept both forms?

mdujava:
Yes
```
$ curl -O "https://en.wikipedia.org/wiki/%22@%22" -O "https://en.wikipedia.org/wiki/%22%40%22"
100 78120  100 78120    0     0   286k      0 --:--:-- --:--:-- --:--:--  286k
100 78120  100 78120    0     0  1556k      0 --:--:-- --:--:-- --:--:-- 1556k
```
```
$ md5sum %22*
5abc9c22c81f3f12deb5aeb001600950  %22@%22
5abc9c22c81f3f12deb5aeb001600950  %22%40%22
```

DL6ER (approved):

PR #3946, opened 2020-12-26
Log output of pihole -g to pihole_updateGravity.log

yubiuser:

---
Logs the output of `pihole -g` to `/var/log/pihole_updateGravity.log`.
The cron job already [does log to that file](https://github.com/pi-hole/pi-hole/blob/41bdb741b7e1a70184ecd7dbfc23f3ebcaa5e3bc/advanced/Templates/pihole.cron#L21), but manually running gravity (or via web interface) did not log so far.

### DIFF
--- a/pihole
+++ b/pihole
@@ -70,7 +70,7 @@ reconfigurePiholeFunc() {
 }
 
 updateGravityFunc() {
-  "${PI_HOLE_SCRIPT_DIR}"/gravity.sh "$@"
+  "${PI_HOLE_SCRIPT_DIR}"/gravity.sh "$@" | tee /var/log/pihole_updateGravity.log
   exit $?
 }
 


dschaper:
I don't see the use in this, but if you all do then you'll need to add that file to `logrotate`. Dumping the output to the file every time gravity is run will increase the size of that file way beyond what we intended.

https://github.com/pi-hole/pi-hole/blob/master/advanced/Templates/logrotate

pihole
```diff
@@ -70,7 +70,7 @@ reconfigurePiholeFunc() {
 }
 
 updateGravityFunc() {
-  "${PI_HOLE_SCRIPT_DIR}"/gravity.sh "$@"
+  "${PI_HOLE_SCRIPT_DIR}"/gravity.sh "$@" | tee /var/log/pihole_updateGravity.log
```
dschaper, line 73: I think you may want `tee -a` to append. Just doing `tee` is going to overwrite the file every time.

yubiuser: That's intended. I don't care about gravity's history but only the last run.

dschaper: Then how will you differentiate between the output from a manual run (that displays the output to you) and the output from the cronjob that really is only meant to be displayed when there is an error. 

If you could explain the purpose of this then we may be able to figure out something that works for both situations. I just don't see the use in logging what is already displayed, either on console or on web interface.

jfb-pihole: I don't see a use for this either. The output is clearly shown to the user when gravity is updated manually. That's what the user should read.

jfb-pihole: It's only when gravity is updated unattended that a capture of the output is needed.

yubiuser: The idea was inspired by [this discourse post](https://discourse.pi-hole.net/t/info-uber-tote-listen/42098?u=yubiuser). If the download of an list fails there is currently no way to investigate this after I close the terminal. Of course it is displayed right away, but this would give the opportunity to see the result of the last gravity run also later. 

You're right, the cronjob would need to be modified as well to not only log in case of an error.
I can't (and need) to differentiate between the manually run and the cronjob.

dschaper: You can manually run `pihole -g | tee` or redirect from the command line if this is a rarely used function. Gravity shouldn't be run very frequently.

dschaper: And from the discourse link, it seems that a log wouldn't be useful anyways, you'd still need to go to the command line and that was reported as not being a solution.

yubiuser: The idea further down the road would be that failures of gravity could be incorporated into pihole's diagnosis system. It could warn the users if a list download failed in the web interface.

jfb-pihole: If that's the case, any changes should be made when needed to support a new feature. Overwriting the unattended log file on each gravity update is highly undesirable. It adds no value and removes a troubleshooting asset.

yubiuser: > removes a troubleshooting asset.

Why does it remove troubleshooting asset? Errors could be logged as well. If there are recurred errors, they would be loggend. If it is a one-time error and is working on the next try it will be overwritten but the error is gone anyway.

jfb-pihole: The purpose of that specif log file is to store the record of the unattended gravity rebuild.  This record is available until the next unattended rebuild.  This helps us answer the "all of a sudden, my domains on blocklist is different" or "suddenly some domains won't load" support questions.  Without that log, the record of the activity is lost.

If we overwrite this log with a subsequent manual gravity rebuild, we lose the data we sometimes need to help a user.  As I noted, since a manual gravity rebuild shows the output clearly on the screen, there is no compelling reason to store this in any log, much less overwriting an existing log that was created for another purpose.

Even if we append manual gravity rebuilds to this log (and then the next unattended update overwrites the file as intended), if a user has disable their cron (or modified it), this log could continue to grow as it is not part of a log rotation scheme.

yubiuser:
Closing this PR as there is a more comprehensive PR https://github.com/pi-hole/pi-hole/pull/3951

PR #3954, opened 2020-12-29
Fix connection mechanism between Chronometer and FTL

DL6ER:

---

Fix connection mechanism between Chronometer and FTL

The existing interaction script between Chronometer (`pihole -c`) and FTL had a few drawbacks:

2. The connection was closed "suddenly" while FTL was waiting for the next query, resulting in a warning in FTL's log.
1. `read` is a `bash`-internal and does not populate. As a result, the existing check for `$?` didn't work and chronometer would have stalled forever if the connection was interrupted midway.

The two points are addressed with two changes:

1. Send `>whatevercommand > quit` to signal FTL that the pipe is to be closed when done sending the results of the query. This resolves the seen warnings properly.
2. Check for an empty returned value by `read` *and* a closed file descriptor `3`. This is the proper way to check for a terminated connection. Relying on the (non-existing) return value of `read` is removed as well as the check for a string `EOM` in the payload.

None

---

These fixes should be applied to PADD as well after they have been evaluated and approved.

### DIFF
--- a/advanced/Scripts/chronometer.sh
+++ b/advanced/Scripts/chronometer.sh
@@ -13,6 +13,7 @@ LC_NUMERIC=C
 
 # Retrieve stats from FTL engine
 pihole-FTL() {
+    local ftl_port LINE
     ftl_port=$(cat /run/pihole-FTL.port 2> /dev/null)
     if [[ -n "$ftl_port" ]]; then
         # Open connection to FTL
@@ -20,12 +21,13 @@ pihole-FTL() {
 
         # Test if connection is open
         if { "true" >&3; } 2> /dev/null; then
-            # Send command to FTL
-            echo -e ">$1" >&3
+            # Send command to FTL and ask to quit when finished
+            echo -e ">$1 >quit" >&3
 
-            # Read input
+            # Read input until we received an empty string and the connection is
+            # closed
             read -r -t 1 LINE <&3
-            until [[ ! $? ]] || [[ "$LINE" == *"EOM"* ]]; do
+            until [[ -z "${LINE}" ]] && [[ ! -t 3 ]]; do
                 echo "$LINE" >&1
                 read -r -t 1 LINE <&3
             done


dschaper (approved):

David1sv:
Unsubscribe

Sent from my iPhone

On Dec 30, 2020, at 5:13 AM, DL6ER <notifications@github.com> wrote:

﻿

Merged #3954<https://github.com/pi-hole/pi-hole/pull/3954> into development.

advanced/Scripts/chronometer.sh
```diff
@@ -13,19 +13,21 @@ LC_NUMERIC=C
...
-            echo -e ">$1" >&3
+            # Send command to FTL and ask to quit when finished
+            echo -e ">$1 >quit" >&3
 
-            # Read input
+            # Read input until we received an empty string and the connection is
+            # closed
             read -r -t 1 LINE <&3
-            until [[ ! $? ]] || [[ "$LINE" == *"EOM"* ]]; do
+            until [[ -z "${LINE}" ]] && [[ ! -t 3 ]]; do
```
MichaIng, line 30: @DL6ER 
One question since you chose to use _and_ here: Is closing the connection an atomic step or is it theoretically possible now to have a longer or endless empty line loop, if for some reason closing the connection (terminal) takes a little longer or fails for some reason? If so, probably keeping either the `EOM` as _or_ loop exit condition would be more robust, or not sure if en empty line could serve for this alone, when it is never part of a valid output?

DL6ER: An empty line may be dangerous as I'm unsure if it is guaranteed that received lines always arrive in one piece. So we may receive just the newline character at the end of a line and may interpret this incorrectly as empty line. (I don't think this will happen, but still may be worth mentioning)

Neither do I think closing a connection can fail.

MichaIng: If there was an unexpected newline, at least when being in the middle of a word, the output would be broken anyway as it is parsed by specific word/field counts and `echo` without `-n` appends a newline, hence separates a word/field.

As you say "dangerous", if the output is terminated too early, only a single chronometer print misses some or all values, but if the script hangs in an endless loop, the whole server might be hanging, which undoubtedly is a much higher danger 😉. But until there is any user actually reporting such a behaviour, I think you're right that practically that connection close can not fail.

It's kinda mystery to me how those /dev/tcp interfaces work in general, e.g. why does `cat` not work, how does the server know when a line was read, to output the next line or EOM. But very interesting and obviously very efficient as well 😃.

PR #3982, opened 2021-01-11
Fix list: User same wildcard regex that the UI use

andras-tim:

---
Same and working whitelist/blacklist handling that UI has.

* As you can see in the code, I swapped the parts of the regex group to fit with the UI's.

I don't think this is necessary now.

---

### DIFF
--- a/advanced/Scripts/list.sh
+++ b/advanced/Scripts/list.sh
@@ -112,7 +112,7 @@ ProcessDomainList() {
     for dom in "${domList[@]}"; do
         # Format domain into regex filter if requested
         if [[ "${wildcard}" == true ]]; then
-            dom="(^|\\.)${dom//\./\\.}$"
+            dom="(\\.|^)${dom//\./\\.}$"
         fi
 
         # Logic: If addmode then add to desired list and remove from the other;


andras-tim:
**With the latest docker release** (`pihole/pihole:v5.3.4`) you can add the same domain twice if you use the **wildcard** option on the UI and the `--wild` or `--white-wild` option w/ CLI.

1. Exactly, if I want to add **foo.bar** to whitelist as wildcarded domain...

    * the UI will create: `(\.|^)foo\.bar$`
    * the CLI (`pihole --white-wild foo.bar`) will create: `(^|\.)foo\.bar$`

    _(so, I will have two regexes for one purpose)_

1. And of course the CLI is consistent w/ itself, so (`pihole --white-wild -d foo.bar`) will remove the `(^|\.)foo\.bar$` only.

**With my fix**, the CLI add wildcard records as the UI do, so in `(\.|^)<DOMAIN>$` form.

_(of course, my first idea was I make the `list.sh` be smarter to be able to handle both formats, but the current code requires more refactor that I wanted to do now)_

dschaper:
That's an OR grouping, so the order really doesn't matter?

andras-tim:
> That's an OR grouping, so the order really doesn't matter?

Both interfaces extend the same SQLite table with different regexes. These regexes are in OR condition (as you exclude two different domains, e.g. foo.com and bar.com)

dschaper (approved):

andras-tim:
Thank you!

dschaper:
Thank you for making the submission!

Issue #4048, opened 2021-02-11
Running Teleporter from the CLI does not include hostname


benemon:
### Versions

```
pi@green:~/pihole $ pihole version
  Pi-hole version is v5.2.4 (Latest: v5.2.4)
  AdminLTE version is v5.3.2 (Latest: v5.3.2)
  FTL version is v5.6 (Latest: v5.6)
```
### Platform

- OS and version:  

- Platform:  
```
pi@green:~/pihole $ cat /etc/*release
PRETTY_NAME="Raspbian GNU/Linux 10 (buster)"
NAME="Raspbian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=raspbian
ID_LIKE=debian
HOME_URL="http://www.raspbian.org/"
SUPPORT_URL="http://www.raspbian.org/RaspbianForums"
BUG_REPORT_URL="http://www.raspbian.org/RaspbianBugs"
```

### Expected behavior

running `pihole -a -t` results in a tar.gz containing the hostname of the current host, in the same manner that running teleporter from the web ui does.

### Actual behavior / bug

running `pihole -a -t` results in a tar.gz that does not contain the hostname of the current host. This is problematic for automating backups across several pihole instances.

### Steps to reproduce

```
pi@green:~/pihole $ pihole -a -t
pi@green:~/pihole $ ls -la
total 20
drwxr-xr-x 3 pi   pi   4096 Feb 11 11:27 .
drwxr-xr-x 7 pi   pi   4096 Feb 11 11:13 ..
-rw-r--r-- 1 root root 3002 Feb 11 11:27 pi-hole-teleporter_2021-02-11_11-27-27.tar.gz
```

## Debug Token

- URL:  

### Screenshots

_If applicable, add screenshots to help explain your problem._

### Additional context

If the command line tool cannot automatically add the hostname as per the web ui, perhaps the user could be allowed to provide a custom filename in place of the current default filename.

dschaper:
> This is problematic for automating backups across several pihole instances.

How do you see this as problematic?

You should be setting the hostname on the node by what ever means you are creating the OS.

PromoFaux:
> How do you see this as problematic?

I guess because the filename will always be `pi-hole-teleporter_YYYY-MM-DD_HH-mm-ss.tar.gz` it's more difficult to ascertain where it came from.

I guess we should update it to be consistent with the web interface

[...]

PR #4049, opened 2021-02-11
Add hostname to teleporter backup file if called from cli

yubiuser:

---
Add the `hostname` to the teleporter backup file if called via cli (`pihole -a -t`). Same behavior as the web interface

https://github.com/pi-hole/AdminLTE/pull/1417

Include the hostname in the filename. Replaces all `.` with `_` as in https://github.com/pi-hole/AdminLTE/pull/1517

None.

Fixes https://github.com/pi-hole/pi-hole/issues/4048

Before
`-rw-r--r-- 1 root   root   4.8K Feb 11 17:20 pi-hole-teleporter_2021-02-11_17-20-00.tar.gz`

After

`-rw-r--r-- 1 root   root   4.8K Feb 11 19:44 pi-hole-rockpi-4b-teleporter_2021-02-11_19-44-48.tar.gz`

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -640,7 +640,7 @@ Teleporter() {
     datetimestamp=$(date "+%Y-%m-%d_%H-%M-%S")
     host=$(hostname)
     host="${host//./_}"
-    php /var/www/html/admin/scripts/pi-hole/php/teleporter.php > "pi-hole-${host}-teleporter_${datetimestamp}.tar.gz"
+    php /var/www/html/admin/scripts/pi-hole/php/teleporter.php > "pi-hole-${host:-noname}-teleporter_${datetimestamp}.tar.gz"
 }
 
 checkDomain()
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -238,18 +238,18 @@ trust-anchor=.,20326,8,2,E06D44B80B8F1D39A95C0B0D7C65D08458E880409BBC68345710423
         #          168.192.in-addr.arpa to 192.168.0.0/16
         #          192.in-addr.arpa to 192.0.0.0/8
         if [[ "${CONDITIONAL_FORWARDING_REVERSE}" == *"in-addr.arpa" ]];then
-            arrRev=("${CONDITIONAL_FORWARDING_REVERSE//./ }")        
-            case ${#arrRev[@]} in 
+            arrRev=("${CONDITIONAL_FORWARDING_REVERSE//./ }")
+            case ${#arrRev[@]} in
                 6   )   REV_SERVER_CIDR="${arrRev[3]}.${arrRev[2]}.${arrRev[1]}.${arrRev[0]}/32";;
                 5   )   REV_SERVER_CIDR="${arrRev[2]}.${arrRev[1]}.${arrRev[0]}.0/24";;
                 4   )   REV_SERVER_CIDR="${arrRev[1]}.${arrRev[0]}.0.0/16";;
-                3   )   REV_SERVER_CIDR="${arrRev[0]}.0.0.0/8";; 
+                3   )   REV_SERVER_CIDR="${arrRev[0]}.0.0.0/8";;
             esac
         else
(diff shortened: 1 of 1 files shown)


dschaper:
I don't see this as any use. The hostname is purely what is set up by the user when they install the OS. 

And it's going to make it so that backup can only be used for a single restore.

advanced/Scripts/webpage.sh
```diff
@@ -238,18 +238,18 @@ trust-anchor=.,20326,8,2,E06D44B80B8F1D39A95C0B0D7C65D08458E880409BBC68345710423
         #          168.192.in-addr.arpa to 192.168.0.0/16
         #          192.in-addr.arpa to 192.0.0.0/8
         if [[ "${CONDITIONAL_FORWARDING_REVERSE}" == *"in-addr.arpa" ]];then
-            arrRev=("${CONDITIONAL_FORWARDING_REVERSE//./ }")        
-            case ${#arrRev[@]} in 
+            arrRev=("${CONDITIONAL_FORWARDING_REVERSE//./ }")
```
PromoFaux, line 241: Thanks, I've been meaning to PR this whitespace away since I [introduced it](https://github.com/pi-hole/pi-hole/commit/e004d27b28bf3da43101948903c56a7cf66be4eb) 😉

advanced/Scripts/webpage.sh
```diff
@@ -636,8 +636,11 @@ Interfaces:
 
 Teleporter() {
     local datetimestamp
+    local host
     datetimestamp=$(date "+%Y-%m-%d_%H-%M-%S")
-    php /var/www/html/admin/scripts/pi-hole/php/teleporter.php > "pi-hole-teleporter_${datetimestamp}.tar.gz"
+    host=$(hostname)
+    host="${host//./_}"
+    php /var/www/html/admin/scripts/pi-hole/php/teleporter.php > "pi-hole-${host}-teleporter_${datetimestamp}.tar.gz"
```
PromoFaux, line 643: presumably if `${host}` is blank here (I can't see why it would be), we would end up with a `pi-hole--teleporter...` file?

advanced/Scripts/webpage.sh
yubiuser: I did not intend to change the whitespaces and was surprised they appear to be removed here. Must be a default setting in `atom` editor.

advanced/Scripts/webpage.sh
yubiuser: I think so. Have not tried it myself, but that's what I expect to see. Does not look beautiful but would not be a good reason to me to introduce a function to prevent this edge case.

dschaper:
```suggestion
    hostname=$(hostname)
    host="${hostname//./_}"
    php /var/www/html/admin/scripts/pi-hole/php/teleporter.php > "pi-hole-${host:-noname}-teleporter_${datetimestamp}.tar.gz"
```

PromoFaux: I do like a cheeky bash-ism when I see one. Does that say "if `host` is empty then state `noname`"?

dschaper: Yeah, and really it could be done without changing the variable names:

```suggestion
    host=$(hostname)
    host="${host//./_}"
    php /var/www/html/admin/scripts/pi-hole/php/teleporter.php > "pi-hole-${host:-noname}-teleporter_${datetimestamp}.tar.gz"
```

PromoFaux: Or even: ?

```suggestion
    host="${hostname//./_}"
    php /var/www/html/admin/scripts/pi-hole/php/teleporter.php > "pi-hole-${host:-noname}-teleporter_${datetimestamp}.tar.gz"
```

dschaper: That doesn't actually call `hostname` though. You still need the first line to execute the `hostname` call.

yubiuser: The `:-noname` is really nice. Didn't know that you can put in default values.

dschaper (approved):

PR #4081, opened 2021-03-14
Update pihole-FTL.conf.5 - specify comment formats

jgrisham:
; Use a semi-colon as a comment character
; FTL is a C program and accepts # and ; comment markers
;  ... but the web interface requires ; as it uses the php 'parse-ini-file' command
; /var/www/html/admin/scripts/pi-hole/php/FTL.php:23

---
Update man page (pihole-FTL.conf.5) for the FTL configuration file to document the valid way(s) to comment a line or part of a line.

This does NOT alter any code files.

I added a sentence or two near the beginning of the man file

Only this, but putting this same information somewhere on the documentation website might be helpful as well.

---

Signed-off-by: Jim Grisham <pi-hole_github@jimgrisham.com>


PromoFaux:
I'm kinda OK with this change, but will defer to @DL6ER for an approval/decision.

My only reservation is that we are working in the background on changes to the web interface - and php _should_ become a thing of the past.

DL6ER:
> FTL is a C program and accepts # and ; comment markers

In fact, FTL is smart enough to detect comments without the need for any indicators of a comment. Comments in-line are not explicitly supported. Did you test that this works? I think something like
```
BLOCKINGMODE=IP ; We do this for fun
```
is invalid.

However,
```
We do this for checking stuff
DEBUG_DATABASE=true

???? What is this doing ????
WHATEVER_SETTING=whatever-value
```

and so on are all valid examples. Due to the mentioned limitation of the (current / still) web interface, I'd phrase it differently, don't even point to the web interface but just say `;` at line beginning are fine. They surely don't hurt.

jgrisham:
@DL6ER I'm not sure if I tried inline comments as your example. _(Had I done so and then reloaded FTL with the web interface, PHP may have stripped those inline comments before FTL ever actually had a chance to see them)._

The problem I ran into as an end-user _(before digging into the weeds here)_ was that I initially used '#' for comments _(since I'm used to that working in shell scripts)_.  Apparently it wasn't just me:
- The default main config "[01-pihole.conf](https://github.com/pi-hole/pi-hole/blob/master/advanced/01-pihole.conf)" that pi-hole installs for _dnsmasq_ uses '#' for comments.
- [This ](https://aur.archlinux.org/cgit/aur.git/tree/pi-hole-ftl.conf?h=pi-hole-ftl)is 3rd party, but it's still out there.
- [This old commit](https://github.com/pi-hole/pi-hole/commit/f2c675cff64a99bd66f913fbeff4bec5d816a21d#diff-fdb44a50fa29f8f6f825d130de9fb4611a64c6ab2104ef0c2782ce3652beebb1) comes up as one of the top results when you search this repo for "pihole-FTL.conf".
- For those installing _unbound_ alongside _pi-hole_, unbound allows "#" comment lines.

That worked fine with FTL during system startup and when running / reloading from the command line.

A few weeks later, I discovered it was not possible to reload FTL using the web interface. I hadn't used PHP _(or any .ini files)_ in probably 20 years, so didn't even know / remember that ';' might have been an option. 

------------

**Summary**: While I personally understand how this works now*, we can prevent others' future pain with documentation. 

The [text ](https://github.com/pi-hole/pi-hole/pull/4081/files)inside my commit says _(formatted for man)_:
```
/etc/pihole/pihole-FTL.conf will be read by \fBpihole-FTL(8)\fR on startup.
.br
Comments may start with ';' or '#' (anywhere in a line), but as the Pi-hole web front-end currently relies on the PHP 'parse-ini-file' routine to read this file, use ';' unless you do not plan on installing the web interface.
.br
For each setting the option shown first is the default.
.br
```

If I understand you correctly, we can make that more technically correct by changing the middle sentences to:

> Comments may start with ';' or '#' (at the beginning of a line), but as the Pi-hole web front-end currently relies on the PHP 'parse-ini-file' routine to read this file, use ';' unless you do not plan on installing the web interface.

Once the web interface no longer uses PHP, that part can be removed. _(The reason I mentioned that at all is so that if in 5 years, if we forget to take it out, do one will have to wonder "why does this say we need to use semicolons for comments?")_

If we just say...

> To add comments to the configuration file, use a semicolon at the beginning of every line that FTL should ignore. pihole-FTL will ignore any lines starting with a semicolon.

... or ...

> Comments may be included as long as the line starts with ';' or '#'.

... will that make sense after we remove the PHP dependency if someone forgets to change the documentation again?


If you want simplicity:

>  "**Do not use # for comments** or you may break your ability to reload FTL with the web interface. Configuration lines beginning with a semicolon will be ignored by FTL."

--------------

That covers us in case there are some other unknown characters (e.g. tabs, non-breaking spaces, etc.) that will make FTL.php unhappy while it is still with us. Rephrasing your example:
```
暁による最終変更：

Мы делаем это для проверки
DEBUG_DATABASE=true

WHATEVER_SETTING=whatever-value
Don't use WHATEVER_SETTING=different-value
```

---
_   * For reference, my pihole-FTL.conf now begins with:

```
; This is a php-style .INI file used by
;   /var/www/html/admin/scripts/pi-hole/php/FTL.php:23
;      ... and ...
;  /usr/bin/pihole-FTL
;
; Use a semi-colon as a comment character
; FTL is a C program and accepts # and ; comment markers
;  ... but the web interface requires ; as it uses the php 'parse-ini-file' command
;
; Details: man pihole-FTL.conf
;          https://docs.pi-hole.net/ftldns/configfile/
;          https://github.com/pi-hole/pi-hole/commit/b2fb3fd35aa8858269d865d5fbc5c2074b0fe04a
;
; Reload: sudo killall -SIGHUP pihole-FTL
```

DL6ER:
I'm just trying to make things as easy as possible for users. Some have no clue what PHP even is. So if we simply say comments should start with `;` then this shouldn't cause any issues.

yubiuser:
Are you going to continue to work on that PR?

jgrisham:
> Are you going to continue to work on that PR?

Oops.

I’ll try to look at it when I’m back in the office on Tuesday; honestly after responding to the last comment I had forgotten it was still open _(I incorrectly assumed* that if the approvers thought this was a good idea they would have tweaked it to their liking prior to merging/‘pulling’ … or just closed it without action long ago)_.

Thanks for the poke!

 \- Jim

- - - -
\* _(I’m far more of an end user / data modeler / tech writer than a programmer - and my limited programming experience involved mostly hardware drivers and ‘old-timey’ languages; at the time this was probably one of my first ever PR-style contribution attempts)_

yubiuser:
> So if we simply say comments should start with ; then this shouldn't cause any issues.

Unfortunately, there is a place where issues occur: gravity update
https://github.com/pi-hole/pi-hole/pull/4081

jgrisham:
@yubiuser Good catch.

Well, that’s awkward … 

So, perhaps no comment delimiters can be allowed in `pihole-FTL.conf` until either

1. the new web interface (eliminating php) is eliminated?
2. we modify `update_gravity.sh` to preprocess `pihole-FTL.conf` _(e.g. with `sed`)_ to make any input `bash`-safe?
3. have one process that parses configuration files, so they aren’t read directly by `php` nor by any shell scripts _(but as long as the web interface can **write** to the configuration file, that doesn’t seem workable)_
4. perhaps _(I haven’t tested this)_, it would be possible to escape the offending character from the attention of either `php` _(e.g. use something like `\#`)_ or `bash` _(e.g. something like `\;`)_
5. … ?

yubiuser:
My vote would be either to recommend
> no comment delimiters can be allowed in pihole-FTL.conf

or

> the combination of #;

(in case this works for the OP in #4444)

Tuxdude:
For the issue reported in #4444, I can confirm that using `#;` to begin the comment lines work with both the UI (`php`) and with the `updateGravity` command line.

jgrisham:
I’ve updated `pihole-FTL.conf.5`to state that any comments lines in `pihole-FTL.conf` files should begin with the two-character delimiter `#;`.

As soon this can be validated as suggested below\*, I’ll `git rebase` the branch to both _squash_ the commits as well as add the missing _DCO signoffs_.

\* _(I probably won’t be back at a computer until sometime next week, so if someone feels like testing this in the meantime, please report what specifically was tested and if `FTL` started back up without error afterwards!)_
- - - -

<details><summary>Validation testing suggestions</summary>

Since `#;` does indeed appear to work, we should test it followed by otherwise _potentially illegal_ characters _(i.e. those likely to be used in a comment; e.g. `@$&()%-+//!{}<>~_|…§€£^`)_ in the following circumstances:

1. indented by spaces
2. indented by tabs
3. indented by spaces followed by tabs, and the opposite
4. following a valid configuration string _(e.g. `MAXDBDAYS=90 #; five fortnights is too long for us!`)_

… in each of the following modes, checking logs for any errors or warnings and making sure that in each case `FTL` is running afterwards:
A. `pihole updateGravity`
B. restart `FTL` using the web interface _(ideally after changing any FTL setting that exists in that interface)_
C. reboot / cold boot _(or at least cleanly exit `pihole-FTL` and restart it directly from the command-line _(probably redundant with ‘B’)_
</details>

yubiuser:
@jgrisham 

We removed `pihole-FTL.conf.5` completely (https://github.com/pi-hole/pi-hole/pull/4489#pullrequestreview-841221345)  because it was horribly outdated. We decided to keep the documentation online only need to maintain it in one place.

However, I think adding a note how comments work in `pi-hole.conf` is a good and necessary thing. We do see bug reports from time to time about it (https://discourse.pi-hole.net/t/syntax-error-after-pihole-up/51929).

My first idea was to add it to `pihole-FTL.conf` directly, but this requires us to be careful as we don't want to overwrite existing configuration option. And/or we could add a `pihole-FTL.conf` template which gets installed on every **fresh** install.
But a not to the documentation right at the top should be made anyway: 
https://docs.pi-hole.net/ftldns/configfile/

jgrisham:
@yubiuser Instead of removing `pihole-FTL.conf.5` completely, but to avoid maintaining documentation in two places, perhaps the `man` page(s) could contain just one line: 
- the URL to the versioned _(hopefully)_ documentation. 

_(This would be especially helpful to those who only use text-based terminals*. e.g. `man pihole-FTL.conf | sed … | curl … | <some html-text-converter> | less` or `man pihole-FTL.conf | sed … | lynx / links2`)_

Thanks for the update!

- - - -
_\* As a very specific example, most of my semi-annual _Pi-Hole_ updates occur on a headless RPi via a ssh app on an iPad. (Switching between browsers and the terminal emulator on mobile devices often causes one or both to reload a connection.)_
- - - -
_(Once nice thing about `man` or `info` pages is that they travel with the source code, and so are by default version-controlled. Whether someone is running an old version or a bleeding-edge alpha, they can always hypothetically be kept synchronized. The same can’t always be said for web-based documentation, especially following a redesign.)_

Long-term, perhaps both `man`, `info`, and web-based documentation can all be automatically be generated from a ‘single source of truth’.

DL6ER:
> Long-term, perhaps both man, info, and web-based documentation can all be automatically be generated from a ‘single source of truth’.

This is planned for v6.0, albeit directly in the config file.
See a config file example of the current development version here: https://github.com/pi-hole/FTL/blob/new/http/test/pihole-FTL.toml

yubiuser:
I added a link to the online resource in the manpage of pihole-FTL in the above linked PR 😉

yubiuser:
I'm going to close the PR as we removed the man page entirely (https://github.com/pi-hole/pi-hole/pull/4489) added a `pihole-FTL.conf` file on new installations with comment syntax information (https://github.com/pi-hole/pi-hole/pull/4496) and extended the online documentation (https://github.com/pi-hole/docs/pull/630)

PR #4103, opened 2021-04-02
Obtain FTL's PID from the PID file

DL6ER:

---

Obtain FTL's PID from the PID file instead of relying on `pkill` finding the right process by itself. This allows the script to work in even when FTL is running inside the memory checker `valgrind`.

Try to obtain FTL's PID from the PID file. If this fails, try to identify the main process using `pgrep --oldest`. If this fails as well, we can safely assume FTL is not running and should show a hint instead of throwing an error.

None

### DIFF
--- a/pihole
+++ b/pihole
@@ -108,25 +108,14 @@ getFTLPID() {
     pid="$(<"$FTL_PID_FILE")"
     # Verify that the value read from the file is numeric
     if [[ "$pid" =~ [^[:digit:]] ]]; then
-      # If no, unset the variable so we use pgrep in the next step
+      # Exploit prevention: unset the variable if there is malicious content
       pid=
     fi
   fi
 
-  # If no PID file is available, try to get a PID from pgrep
-  if [ -z "$pid" ]; then
-    # Use --oldest to get the main process of FTL (all TCP workers are created
-    # after the main process has been started)
-    pid="$(pgrep --oldest pihole-FTL)"
-  fi
-
-  # If we still have no PID, FTL is not running, substitute negative PID to
-  # signal this to the caller
-  if [ -z "$pid" ]; then
-    pid="-1"
-  fi
-
-  echo "$pid"
+  # If FTL is not running, or the PID file contains malicious stuff, substitute
+  # negative PID to signal this to the caller
+  echo "${pid:=-1}"
 }
 
 restartDNS() {
--- a/pihole
+++ b/pihole
@@ -16,6 +16,7 @@ readonly PI_HOLE_SCRIPT_DIR="/opt/pihole"
 # error due to modifying a readonly variable.
 setupVars="/etc/pihole/setupVars.conf"
 PI_HOLE_BIN_DIR="/usr/local/bin"
+readonly FTL_PID="/run/pihole-FTL.pid"
 
 readonly colfile="${PI_HOLE_SCRIPT_DIR}/COL_TABLE"
 source "${colfile}"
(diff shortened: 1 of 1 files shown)


DL6ER:
This change is needed as found in https://github.com/pi-hole/FTL/issues/960#issuecomment-812091473

pihole
```diff
@@ -16,6 +16,7 @@ readonly PI_HOLE_SCRIPT_DIR="/opt/pihole"
 # error due to modifying a readonly variable.
 setupVars="/etc/pihole/setupVars.conf"
 PI_HOLE_BIN_DIR="/usr/local/bin"
+readonly FTL_PID="/run/pihole-FTL.pid"
```
dschaper, line 19: Be specific that this is a file.
```suggestion
readonly FTL_PID_FILE="/run/pihole-FTL.pid"
```

pihole
```diff
@@ -98,8 +99,38 @@ versionFunc() {
...
+
+  # If we still have no PID, FTL is not running, substitute negative PID to
+  # signal this to the caller
+  if [ -z "$pid" ]; then
+    pid="-1"
+  fi
+
+  echo "$pid"
+}
+
```
dschaper, line 131: This uses some bashisms and really should be POSIX, but you know....

```suggestion
# Get PID of main pihole-FTL process
getFTLPID() {
  local ftl_pid
 
  if [ -s "${FTL_PID_FILE}" ]; then
    # -s: FILE exists and has a size greater than zero
    ftl_pid="$(<"$FTL_PID_FILE")"
    # Verify that the value read from the file is numeric
    if [[ "$ftl_pid" =~ [^[:digit:]] ]]; then
      # If pid value is not numeric then try to get a PID from pgrep
      # Use --oldest to get the main process of FTL (all TCP workers are created
      # after the main process has been started) OR use sentinel value to show pgrep failure
      ftl_pid="$(pgrep --oldest pihole-FTL)" || ftl_pid="-1"
    fi
  fi
  
  # echo the pid value found or -1 if variable is still unset
   echo "${ftl_pid:=-1}"
}

```

dschaper (changes requested):

pihole
DL6ER: This has been borrowed from
https://github.com/pi-hole/pi-hole/blob/cbfb58f7a283c2a3e7aad95a834a0287175ccb24/advanced/Scripts/piholeDebug.sh#L133

I'm happy to change this.

pihole
DL6ER: The proposed change would change behavior as we'd never try to get the PID through `pgrep` if the PID file is absent (or empty).

pihole
DL6ER: Pushed as a separate commit instead of amending to ease the review process. We can squash the entire thing in the end.

pihole
DL6ER: I pushed yet another commit addressing the issue that a system with a missing PID file is problematic.

pihole
```diff
@@ -98,8 +99,25 @@ versionFunc() {
...
+# Get PID of main pihole-FTL process
+getFTLPID() {
+  local pid
+
+  if [ -s "${FTL_PID_FILE}" ]; then
+    # -s: FILE exists and has a size greater than zero
+    pid="$(<"$FTL_PID_FILE")"
+    # Exploit prevention: unset the variable if there is malicious content
+    # Verify that the value read from the file is numeric
+    [[ "$pid" =~ [^[:digit:]] ]] && pid=
```
dschaper, line 111: Do you want to set `pid` to empty or actually `unset pid`?

DL6ER: I don't care as long as it trips `${FOO:=val}`. Should happen both for unset and empty.

dschaper: `bash` style would be to use `unset`. `c` style would be to set it to nothing and let garbage collection handle it. 

Can we use `unset` for a purely stylistic reason?

dschaper:
```suggestion
    [[ "$pid" =~ [^[:digit:]] ]] && unset pid
```

dschaper (approved):

dschaper:
Dev or 5.3?

PromoFaux:
5.3 plz

dschaper (approved):

PR #4122, opened 2021-04-14
HotFix release v5.3.1

dschaper:
---

Hotfixes for release v5.3.


dschaper:
Looks like the splashpage is now complete again with href links and you can use `#` as the port separator for custom DNS.

Checked with `pihole checkout core release/v5.3.1` for those two issues.

DL6ER (commented):
Approved (better later than never!)

PR #4130, opened 2021-04-16
Read local.list IPs from "ip a" not setupVars.conf

DL6ER:

---

Fix a bug reported on Discourse (see below).

Instead of relying on the correct IP address being in `setupVars.conf`, we should add all available addresses to `local.list`. FTL uses the option `localise-queries` that will pick the most appropriate one (arriving interface dependent) for the requesting client.

None

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -47,16 +47,6 @@ domainsExtension="domains"
 setupVars="${piholeDir}/setupVars.conf"
 if [[ -f "${setupVars}" ]];then
   source "${setupVars}"
-
-  # Remove CIDR mask from IPv4/6 addresses
-  IPV4_ADDRESS="${IPV4_ADDRESS%/*}"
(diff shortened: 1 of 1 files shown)


DL6ER:
This PR still has a bit of a drawback by relying on the IP addresses at `pihole -g` run time (not accounting for frequently changing GUA addresses, etc.). The alternative solution would be to remove the handling of the hostname and `pi.hole` *entirely* from gravity and move into FTL to be done at run time. This would always pick up the correct IP address for all the interfaces.

@dschaper @PromoFaux What do you say?

PromoFaux:
For the latter part, as long as performance wasn't affected - I'd personally say that's a grand idea! Although I can't really comment on GUA stuff, because IPv6 & me don't get on very well...

dschaper:
I'll have to think about it and that's going to need a few days to do.

DL6ER:
Superseded by #4131

PR #4131, opened 2021-04-16
Handle pi.hole and hostname in FTL instead of local.list

DL6ER:

---

Fix an issue with gravity not working correctly when the (now optional) `IPV4_ADDRESS` and `IPV6_ADDRESS` variables are removed from `setupVars.conf`.

Do not put the hostname and `pi.hole` into `local.list` - this will be handled by FTL (https://github.com/pi-hole/FTL/pull/1111)

Also remove some other places where we expect the variable to be present in the system and modify the debugger to test all interfaces (on all addresses) instead of one from the one address in `setupVars.conf`:

Screenshot from 2021-04-16 12-21-08
(default, `NULL` blocking mode)

Screenshot from 2021-04-16 12-15-30
(`IP` blocking mode)

### This PR supersedes #4130

None

### DIFF
--- a/advanced/Scripts/chronometer.sh
+++ b/advanced/Scripts/chronometer.sh
@@ -498,10 +498,6 @@ chronoFunc() {
         printFunc " RAM usage: " "$ram_perc%" "$ram_info"
         printFunc " HDD usage: " "$disk_perc" "$disk_info"
 
-        if [[ "$scr_lines" -gt 17 ]] && [[ "$chrono_width" != "small" ]]; then
-            printFunc "  LAN addr: " "${IPV4_ADDRESS/\/*/}" "$lan_info"
-        fi
-
         if [[ "$DHCP_ACTIVE" == "true" ]]; then
             printFunc "DHCP usage: " "$ph_dhcp_percent%" "$dhcp_info"
         fi
--- a/advanced/Scripts/piholeDebug.sh
+++ b/advanced/Scripts/piholeDebug.sh
@@ -863,8 +863,9 @@ dig_at() {
     # Set more local variables
     # We need to test name resolution locally, via Pi-hole, and via a public resolver
     local local_dig
-    local pihole_dig
     local remote_dig
+    local interfaces
+    local addresses
     # Use a static domain that we know has IPv4 and IPv6 to avoid false positives
     # Sometimes the randomly chosen domains don't use IPv6, or something else is wrong with them
     local remote_url="doubleclick.com"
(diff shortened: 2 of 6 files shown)


advanced/Scripts/piholeDebug.sh
```diff
@@ -1032,7 +1051,7 @@ parse_file() {
     local file_lines
     # For each line in the file,
     for file_lines in "${file_info[@]}"; do
-        if [[ ! -z "${file_lines}" ]]; then
+        if [[ -n "${file_lines}" ]]; then
```
DL6ER, line 1054: Stickler complaint I had to address

PromoFaux:
Don't forget the tests ;)

dschaper:
There's something strange with the debug output. It's doubling up the dig's on `localhost` and `lo` and it looks like there's a color change missing somewhere since the `lo` opening bracket is the same blue color as the `localhost` IP address. 

Not sure if that's just the screenshot or if it's in the rewritten `dig` testing. I'll check that over.

DL6ER:
Ah yes, you had me when you said

> doubling up the `dig`'s on `localhost` and `lo`

because I forgot to renew the screenshot after 3b57409 

Screenshot from 2021-04-29 07-33-31

DL6ER:
Ah yes, you had me when you said

> doubling up the `dig`'s on `localhost` and `lo`

because I forgot to renew the screenshot after 3b57409 

Screenshot from 2021-04-29 07-33-31

jasi232 (approved):
E

advanced/Scripts/chronometer.sh
```diff
@@ -498,10 +498,6 @@ chronoFunc() {
         printFunc " RAM usage: " "$ram_perc%" "$ram_info"
         printFunc " HDD usage: " "$disk_perc" "$disk_info"
 
-        if [[ "$scr_lines" -gt 17 ]] && [[ "$chrono_width" != "small" ]]; then
-            printFunc "  LAN addr: " "${IPV4_ADDRESS/\/*/}" "$lan_info"
```
PromoFaux, line 502: I'm fine with the answer being "yes" here, by the way...

Is the `Lan addr` line being removed because it's simpler to not work out what it is in chronometer?

dschaper: And will this need changes to PADD as well?

PromoFaux: Eventually `IPV4_ADDRESS` will not be stored in `setupVars.conf` at all, so if PADD is sourcing from there...

DL6ER: Yes. With automatically determined reply addresses depending on the interface a query arrives, displaying one single address will not be suitable in enough cases to be concerned. When changing IPv6 prefix makes a static address stored in setupVars rather useless, too.

PromoFaux: > And will this need changes to PADD as well?

image

I don't know if we want to just remove the IP address from PADD or use some other method..?

PromoFaux (approved):

advanced/Scripts/chronometer.sh
DL6ER: I guess we should just remove it

Otherwise, we could run something like
```
ip a | grep inet  | awk  '{print $2}'
```
which results in something like
```
127.0.0.1/8
::1/128
192.168.2.15/24
fd00::5858:711c:8625:34c4/64
fd00::8678:7fd8:8857:a6f9/64
2a02::3ab4:5555:2eb2:1368/64
2a02::a003:906e:9865:51cd/64
fe80::e7c6:5562:5687:8958/64
```
and we'd *just* have to pick the one the user wants to see. I bet this will be an issue most of the time.

PromoFaux: Have it display a random one from this list each time the variable is called. 😏

DL6ER: Actually, we already have this in PADD:

Screenshot from 2021-07-04 14-50-57

DL6ER: See https://github.com/pi-hole/PADD/pull/161

PR #4172, opened 2021-05-15
Keep old database

jpgpi250:

---
These changes will allow the user to keep, use and consult the old gravity database

option in setupVars.conf: KEEPOLDGRAVITY=true|false (default true)
pihole -g -s will switch the databases and restartdns. This will allow the user to check if things worked before the last gravity run
pihole -q -countnew will report the number of new entries, since the last gravity run
pihole -q -blocked will report the domains, new since the last gravity run, and already blocked a query

image

image

image

setupVars.conf: KEEPOLDGRAVITY=true|false (default true)
new option for gravity (pihole -g -s OR pihole -g -switch)
new option for query (pihole -q -blocked)
new option for query 'pihole -q -countnew)

---

### DIFF
--- a/advanced/Scripts/query.sh
+++ b/advanced/Scripts/query.sh
@@ -105,7 +105,7 @@ EOSQL
   done
   if [[ -z "${all}" ]] && [[ "${#newdomainarray[*]}" -ge 100 ]]; then
     echo -e "  ${INFO} Over 100 new domains found.
-        This can be overridden using the -all option"
+        If you want to see them all, append \"-all\" to the command"
   fi
   exit 0
 }
--- a/advanced/Scripts/query.sh
+++ b/advanced/Scripts/query.sh
@@ -25,6 +25,17 @@ if [[ -f "${pihole_FTL}" ]]; then
   source "${pihole_FTL}"
 fi
 
+# Source setupVars from install script
+setupVars="${piholeDir}/setupVars.conf"
+if [[ -f "${setupVars}" ]];then
+  source "${setupVars}"
+
+  # The variable KEEPOLDGRAVITY in /etc/pihole/setupVars.conf must be used to enable additional query features (countnew, listnew, blockednew).
+  if [[ -z "${KEEPOLDGRAVITY}" ]] ; then
+    KEEPOLDGRAVITY="false"
+  fi
+fi
+
 # Set this only after sourcing pihole-FTL.conf as the gravity database path may
 # have changed
 gravityDBfile="${GRAVITYDB}"
@@ -62,8 +73,28 @@ scanList(){
     esac
 }
 
+CheckDatabaseDates() {
+  # Check if the OLD database exists.
+  if [[ ! -f "${gravityOLDfile}" ]]; then
+    echo -e "  ${CROSS}  Old database ${gravityOLDfile} doesn't exist."
+    exit 1
+  else
+    # Check if the current database is more recent.
+    if [[ $(sqlite3 "${gravityOLDfile}" "SELECT value FROM info \
+              WHERE property IS 'updated';") \
+        > \
+          $(sqlite3 "${gravityDBfile}" "SELECT value FROM info \
+              WHERE property IS 'updated';") ]]; then
+      echo -e "  ${CROSS}  Old database ${gravityOLDfile} is more recent."
+      exit 1
+    fi
+  fi
+}
+
 # function to count new domain entries since last gravity run
 CountNewDomains() {
+  # Execute only if the most recent database is active
+  CheckDatabaseDates
   printf %s "  ${INFO} Running query, please be patient..."
   # store query result (count) in variable
   newdomainentries="$(
@@ -78,6 +109,8 @@ EOSQL
 
 # function to list new domain entries since last gravity run
 ListNewDomains() {
+  # Execute only if the most recent database is active
+  CheckDatabaseDates
   QueryLimit=""
   if [ -z "${all}" ]; then
     QueryLimit="Limit 100"
@@ -115,80 +148,69 @@ EOSQL
 # present (new) in the latest (active) gravity database,
 # and NOT present in the old gravity database.
 FindNewBlockedDomains() {
-  # Check if the OLD database exists.
-  if [[ ! -f "${gravityOLDfile}" ]]; then
-    echo -e "  ${CROSS}  Old database ${gravityOLDfile} doesn't exist."
-    exit 1
-  else
-    # Check if the current database is more recent.
-    if [[ $(sqlite3 "${gravityOLDfile}" "SELECT value FROM info \
-              WHERE property IS 'updated';") \
-        > \
-          $(sqlite3 "${gravityDBfile}" "SELECT value FROM info \
-              WHERE property IS 'updated';") ]]; then
-      echo -e "  ${CROSS}  Old database ${gravityOLDfile} is more recent."
-      exit 1
-    fi
-    # Run the queries
-    printf %s "  ${INFO} Running queries, please be patient..."
-    # Find the status 1 queries first
-    mapfile -t domainarray < <( sqlite3 "${ftlDBfile}" << EOSQL
-      ATTACH '${gravityDBfile}' AS new;
-      ATTACH '${gravityOLDfile}' AS old;
-      SELECT DISTINCT queries.domain, '' || adlist.address || '' FROM queries
-      JOIN gravity ON new.gravity.domain = queries.domain
-      JOIN adlist ON adlist.id = new.gravity.adlist_id
-      WHERE queries.domain NOT IN (SELECT domain FROM old.gravity)
-      AND queries.domain IN (SELECT domain FROM new.gravity)
-      AND queries.status IN (1)
-      AND timestamp > (SELECT value FROM new.info WHERE property = 'updated');
+  # Execute only if the most recent database is active
+  CheckDatabaseDates
+  # Run the queries
+  printf %s "  ${INFO} Running queries, please be patient..."
+  # Find the status 1 queries first
+  mapfile -t domainarray < <( sqlite3 "${ftlDBfile}" << EOSQL
+    ATTACH '${gravityDBfile}' AS new;
+    ATTACH '${gravityOLDfile}' AS old;
+    SELECT DISTINCT queries.domain, '' || adlist.address || '' FROM queries
+    JOIN gravity ON new.gravity.domain = queries.domain
+    JOIN adlist ON adlist.id = new.gravity.adlist_id
+    WHERE queries.domain NOT IN (SELECT domain FROM old.gravity)
+    AND queries.domain IN (SELECT domain FROM new.gravity)
+    AND queries.status IN (1)
+    AND timestamp > (SELECT value FROM new.info WHERE property = 'updated');
 EOSQL
-      )
-    if [ ! "${#domainarray[@]}" -eq 0 ]; then
-      for (( j=0; j<"${#domainarray[@]}"; j++ )); do
-        IFS='|'
-        read -r domain urllist <<< "${domainarray[j]}"
-        printf '\r%s\n' "$(tput el)  ${INFO} Blocked: ${domain} (${COL_GREEN}${urllist}${COL_NC})"
-      done
-    fi
-    # Now find the CNAME matches (status 9)
-    mapfile -t cnamearray < <( sqlite3 "${ftlDBfile}" << EOSQL
-      ATTACH '${gravityDBfile}' AS new;
-      ATTACH '${gravityOLDfile}' AS old;
-      SELECT DISTINCT queries.domain, '' || adlist.address || '', queries.additional_info FROM queries
-      JOIN gravity ON new.gravity.domain = queries.additional_info
-      JOIN adlist ON adlist.id = new.gravity.adlist_id
-      WHERE queries.additional_info NOT IN (SELECT domain FROM old.gravity)
-      AND queries.additional_info IN (SELECT domain FROM new.gravity)
-      AND queries.status IN (9)
-      AND timestamp > (SELECT value FROM new.info WHERE property = 'updated');
+    )
+  if [ ! "${#domainarray[@]}" -eq 0 ]; then
+    for (( j=0; j<"${#domainarray[@]}"; j++ )); do
+      IFS='|'
+      read -r domain urllist <<< "${domainarray[j]}"
+      printf '\r%s\n' "$(tput el)  ${INFO} Blocked: ${domain} (${COL_GREEN}${urllist}${COL_NC})"
+    done
+  fi
+  # Now find the CNAME matches (status 9)
+  mapfile -t cnamearray < <( sqlite3 "${ftlDBfile}" << EOSQL
+    ATTACH '${gravityDBfile}' AS new;
+    ATTACH '${gravityOLDfile}' AS old;
+    SELECT DISTINCT queries.domain, '' || adlist.address || '', queries.additional_info FROM queries
+    JOIN gravity ON new.gravity.domain = queries.additional_info
+    JOIN adlist ON adlist.id = new.gravity.adlist_id
+    WHERE queries.additional_info NOT IN (SELECT domain FROM old.gravity)
+    AND queries.additional_info IN (SELECT domain FROM new.gravity)
+    AND queries.status IN (9)
+    AND timestamp > (SELECT value FROM new.info WHERE property = 'updated');
 EOSQL
-      )
-    if [ ! "${#cnamearray[@]}" -eq 0 ]; then
-      for (( j=0; j<"${#cnamearray[@]}"; j++ )); do
-        IFS='|'
-        read -r domain urllist additional_info <<< "${cnamearray[j]}"
-        printf '\r%s\n' "$(tput el)  ${INFO} Blocked: ${domain}. (${COL_BLUE}CNAME${COL_NC})"
-        echo -e "        Domain: ${additional_info} (${COL_GREEN}${urllist}${COL_NC})"
-      done
-    fi
-    if [[ "${#domainarray[@]}" -eq 0 && "${#cnamearray[@]}" -eq 0 ]]; then
-      printf '\r%s\n' "$(tput el)  ${INFO} No new domains have been blocked yet."
-    fi
+    )
+  if [ ! "${#cnamearray[@]}" -eq 0 ]; then
+    for (( j=0; j<"${#cnamearray[@]}"; j++ )); do
+      IFS='|'
+      read -r domain urllist additional_info <<< "${cnamearray[j]}"
+      printf '\r%s\n' "$(tput el)  ${INFO} Blocked: ${domain}. (${COL_BLUE}CNAME${COL_NC})"
+      echo -e "        Domain: ${additional_info} (${COL_GREEN}${urllist}${COL_NC})"
+    done
+  fi
+  if [[ "${#domainarray[@]}" -eq 0 && "${#cnamearray[@]}" -eq 0 ]]; then
+    printf '\r%s\n' "$(tput el)  ${INFO} No new domains have been blocked yet."
   fi
   exit 0
 }
 
 if [[ "${options}" == "-h" ]] || [[ "${options}" == "--help" ]]; then
-    echo "Usage: pihole -q [option] <domain>
+  echo "Usage: pihole -q [option] <domain>
 Example: 'pihole -q -exact domain.com'
 Query the adlists for a specified domain
 
-Options:
-  -blockednew         List blocked domains since last gravity run
+Options:"
+  if ${KEEPOLDGRAVITY,,}; then
+    echo "  -blockednew         List blocked domains since last gravity run
   -countnew           Count new domain enties since last gravity run
-  -listnew            List new domain entries since last gravity run
-  -exact              Search the block lists for exact domain matches
+  -listnew            List new domain entries since last gravity run"
+  fi
+  echo "  -exact              Search the block lists for exact domain matches
   -all                Return all query matches within a block list
   -h, --help          Show this help dialog"
   exit 0
@@ -202,12 +224,14 @@ else
     if [[ "${options}" == *"-exact"* ]]; then
         exact="exact"; matchType="exact ${matchType}"
     fi
-    # Count new domain enties since last gravity run
-    [[ "${options}" == *"-countnew"* ]] && CountNewDomains
-    # List new domain enties since last gravity run
-    [[ "${options}" == *"-listnew"* ]] && ListNewDomains
-    # List blocked domains since last gravity run
-    [[ "${options}" == *"-blockednew"* ]] && FindNewBlockedDomains
+    if ${KEEPOLDGRAVITY,,}; then
+      # Count new domain enties since last gravity run
+      [[ "${options}" == *"-countnew"* ]] && CountNewDomains
+      # List new domain enties since last gravity run
+      [[ "${options}" == *"-listnew"* ]] && ListNewDomains
+      # List blocked domains since last gravity run
+      [[ "${options}" == *"-blockednew"* ]] && FindNewBlockedDomains
+    fi
 fi
 
 # Strip valid options, leaving only the domain and invalid options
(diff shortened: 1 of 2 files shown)


jpgpi250:
looks like it's not entirely correct again, DCO warning. I'm using github desktop, is there any way to achieve a correct signoff, using github desktop?
At least, I managed to solve the stickler warnings, using shellcheck...

dschaper:
Your signoff is good for me. I'll manually override that. You've included the line, DCObot wants a certain name and email pair but that's a bit picky.

jpgpi250:
Thank you for fixing the DCO error, I was already preparing to close the PR and start again.

I've been at it for over 12 hours yesterday, thank you for supporting me in my first attempt to create a valid PR, thus hopefully contributing to pi-hole.

Unfortunately, I noticed an error in one of my echo commands, one to many ')' characters (at the end):
`  [i] Blocked: tdbtrk.com (file:///home/pi/quidsup/notrack-blocklist.txt))`

Now what do I do to correct this?
- start over again?
- a new commit of the file?
- an additional PR to correct this?

Sorry for the inconvenience, newbie, still learning...

advanced/Scripts/query.sh
```diff
@@ -58,12 +61,68 @@ scanList(){
...
+      AND queries.domain IN (SELECT domain FROM new.gravity)
+      AND queries.status IN (1,9)
+      AND timestamp > (SELECT value FROM new.info WHERE property = 'updated');
+EOSQL
+      )
+    if [ "${#resultarray[@]}" -eq 0 ]; then
+      echo -e "  ${INFO} No new domains have been blocked yet."
+    else
+      for (( j=0; j<"${#resultarray[@]}"; j++ )); do
+        echo -e "  ${INFO} Blocked: ${resultarray[j]})"
```
DL6ER, line 111:
```suggestion
        echo -e "  ${INFO} Blocked: ${resultarray[j]}"
```

This is the source of the extra `)`. You can even commit the change here on Github.

jpgpi250: Thank you, committed and signed (hopefully) the suggested change.

What would of happened if I edited the file, using github desktop, and committed? Would the PR automatically been updated?

yubiuser: Yes. When you commit locally on your desktop and push the changes upstream to your github repo this PR will be updated automatically.

advanced/Scripts/query.sh
```diff
@@ -58,12 +61,68 @@ scanList(){
     esac
 }
 
+# function to count new domain entries since last gravity run
+CountNewDomains() {
+  newdomainentries=$(
+    sqlite3 "${gravityDBfile}" << EOSQL
+      ATTACH '${gravityOLDfile}' AS old;
+      SELECT count(DISTINCT domain) FROM gravity WHERE domain NOT IN (SELECT domain FROM old.gravity);
```
yubiuser, line 69: No tried, but it might be faster with an additional `distinct`

`SELECT count(DISTINCT domain) FROM gravity WHERE domain NOT IN (SELECT  DISTINCT domain FROM old.gravity);`

advanced/Scripts/query.sh
```diff
@@ -58,12 +61,68 @@ scanList(){
...
+    #Run the query
+    mapfile -t resultarray < <( sqlite3 -separator ' ' "${ftlDBfile}" << EOSQL
+      ATTACH '${gravityDBfile}' AS new;
+      ATTACH '${gravityOLDfile}' AS old;
+      SELECT DISTINCT queries.domain, '(' || adlist.address || ')' FROM queries
+      JOIN gravity ON new.gravity.domain = queries.domain
+      JOIN adlist ON adlist.id = new.gravity.adlist_id
+      WHERE queries.domain NOT IN (SELECT domain FROM old.gravity)
+      AND queries.domain IN (SELECT domain FROM new.gravity)
+      AND queries.status IN (1,9)
```
yubiuser, line 103: In case of type 9, I think you need to take the domain from `additional_info` column because this is the actual blocked domain.

You might want to separate the CNAME stuff as it might get a bit to complexe otherwise.

gravity.sh
```diff
@@ -872,12 +889,44 @@ gravity_Cleanup() {
...
+    exit 1
+  fi
+  if [[ -f "${gravityDBfile}" ]]; then
+    echo -e "  ${TICK}  ${gravityDBfile} exists."
+    if [[ ${gravityOLDversion} -eq $(pihole-FTL "${gravityDBfile}" "SELECT value FROM info WHERE property IS 'version';") ]]; then
+      echo -e "  ${INFO}  Switching database."
+      echo "         current database, updated $(date -d @"$(pihole-FTL "${gravityDBfile}" "SELECT value FROM info WHERE property IS 'updated';")")"
+      oldDBdate=$(pihole-FTL "${gravityOLDfile}" "SELECT value FROM info WHERE property IS 'updated';")
+      echo "         old database, updated $(date -d @"${oldDBdate}")"
+      mv "${gravityOLDfile}" "$(dirname -- "${gravityOLDfile}")/gravity_mv.db"
```
yubiuser, line 907: This might be a good place to add a check for disk space again?

advanced/Scripts/query.sh
jpgpi250: without the additional distinct:
`time pihole -q -countnew
  [i] 44 new domain(s) found.

real    0m25.470s
user    0m16.324s
sys     0m5.009s
`

with the additional distinct:
` time pihole -q -countnew
  [i] 44 new domain(s) found.

real    1m13.115s
user    0m51.230s
sys     0m4.820s
`
I appreciate that you are reviewing this.

yubiuser: Okay, let's forget that.

gravity.sh
jpgpi250: I'm open to all suggestions, but why should I check the diskspace here? No additional diskspace is required...
What happens is:
- rename gravity_old.db -> gravity_mv.db
- rename gravity.db -> gravity_old.db
- rename gravity_mv.db -> gravity.db
End result is a name switch, nothing more.
Are you suggesting to simply remove gravity_old.db, if there isn't enough diskspace, and thus, not perform a switch?

DL6ER (commented):
How long do these queries run overall? It may be worth adding a message in the sense of "be patient, this may take some time" (better wording is welcome) if otherwise the user would sit like half minute asking themselves if it is still running or already dead.

gravity.sh
yubiuser: You're right. I mixed it up.

advanced/Scripts/query.sh
jpgpi250: Would this be acceptable for you?
`pi@raspberrypi:~ $ pihole -q -blocked
  [i] Blocked: amzonse.xyz (https://phishing.army/download/phishing_army_blocklist_extended.txt)
  [i] Blocked: aod.cm.cn.miaozhen.com (https://www.github.developerdan.com/hosts/lists/ads-and-tracking-extended.txt)
  [i] Blocked: www.dnsimple.com. (CNAME)
        Domain: dnsimple.com (file:///home/pi/test.txt)
`
image

jpgpi250:
> How long do these queries run overall? It may be worth adding a message in the sense of "be patient, this may take some time" (better wording is welcome) if otherwise the user would sit like half minute asking themselves if it is still running or already dead.

Would this be acceptable to you?

Animation

The query actually takes about 25 seconds (raspberry pi 3B).

advanced/Scripts/query.sh
yubiuser: Which domains is which? `www.dnsimple.com` is CNAME for `dnsimple.com`?

It wasn't meant as a hint for how to display things, but actually to include  domains from `additional_info` into your comparison. (You might have addressed this already and not just the output)

dschaper:
If you dig around the code and history you might find a function we had called a spinner. @PromoFaux you think this may be a legitimate use of that spinner?

advanced/Scripts/query.sh
jpgpi250: I added a test list (could be a domain in a regular list) dnsimple.com. I than ran a dig www.dnsimple.com (not in any list). Since this query triggers CNAME detection (status 9), the separate query (as you suggested) finds www.dnsimple.com with status 9, so the user is informed the dns query for this domain, which is a CNAME was blocked. the second line (one for each CNAME) lists the gravity entry (dnsimple.com) and the list.

I think this should be very clear to the user, the output matches the info in the query log

image

yubiuser: I've not checked the query in detail, but from a glance it looked like what I had in mind.

jpgpi250:
Just pushed an update with the suggested changes, ready for review.

Thanks to all for the suggestions, feedback, time and effort.
Thanks for tolerating a newbie.

yubiuser:
> If you dig around the code and history you might find a function we had called a spinner.

(I was curious and found it)
https://github.com/pi-hole/pi-hole/pull/459

jpgpi250:
at dschaper, PromoFaux, yubiuser.

Do I need to reintroduce the spinner function, or do you accept the solution I used.

May I suggest, If the spinner function is a requirement, to put this code in /opt/pihole/COL_TABLE (don't even know this will work, since it is a function - correct me if I'm wrong). It might be usable in other scripts...

DL6ER:
#459 said:

> It doesn't work when run by `cron`

It filled the log with `/\-/\-/\-/\-/\-/\-/\-/\-/\-/\-/\-/\-/\-/\- [...]` or something similar but this new feature should not be part of a `cron` job, anyway. I'm okay with the hint variant without any moving spinner element.

I yet have to find the time for a proper review, I only glanced through so far.

PromoFaux:
Yeah, no don't reintroduce that. It looked pretty, but it wasn't very reliable.

I'm good with a simple:

`  [i] Please wait...` 

or something like that

jpgpi250:
- added 'please be patient' messages for all new queries I added
- added a new option listnew (pihole -q -listnew [-all]) the result is limited to 100 domains, unless '-all' is a command line option.

For my tests, in order to get results, with a lot of changes, I've added the list https://dbl.oisd.nl (yes I know, not your favorite list). This list changes every couple of hours, ensuring you get results (e.g. changes between the old and the new gravity database).

To test the CNAME code, I created a file, containing 'dnsimple.com', added it (file:///home/pi.test.txt) and ran 'pihole -g'. Than, dig 'www.dnsimple.com', this will cause a CNAME block, and the correct result, when running 'pihole -q - blocked'

jpgpi250:
idea (NOT part of this PR)

The results of comparing the new and old gravity database appear to be correct.

I was testing the following:

add a field 'new_entry' to the gravity table of gravity.db, default 0
`ALTER TABLE gravity ADD COLUMN new_entry INTEGER DEFAULT 0;`

by running the following update on the database, while both old (temp) and new still exist, this field would be updated:
`( sudo sqlite3 "${gravityDBfile}" << EOSQL
      ATTACH '${gravityOLDfile}' AS old;
      UPDATE gravity SET new_entry = '1'
	  WHERE domain IN 
	    (SELECT domain FROM gravity WHERE domain NOT IN (SELECT domain FROM old.gravity));
EOSQL
      )` 
On a database with 4.087.232 entries and 4.223 new entries this takes only 35 seconds (raspberry pi 3B).
`real    0m34.983s
user    0m14.937s
sys     0m0.767s`

This database field could than be used to inform the user in the web interface, a new gravity entry caused a blocked query, for example add ' (NEW') behind the domain name.
This could be the user friendly way to inform the user blocking behavior has changed, due to a recent gravity update.

Unfortunately, I'm not smart / confident enough to update all the required components of pi-hole (API, web), thus asking for help.

DL6ER:
We should discuss the addition of a new database column somewhere else. Adding it to the `gravity` table does have some impact because it is the largest (or second largest behind `queries`) database table users will have. 35 seconds for 4 mio queries doesn't sound that dramatic. TBH, I'd have expected worse.

Points for the discussion somewhere else (likely Discourse):
- How long does gravity run overall with such a high number of domains? I'm wondering if we're talking about an increase of 1% or 10% or 50% in total runtime.
- How large is the same database once with and once without the extra column? It'll be surely some megabytes of difference.

This would also need a change in FTL. Currently, we're are just using the tree index of gravity when checking if a domain is blocked. This index is loaded into memory and super-fast even when there are dozens of million of domains in gravity. However, when this change is made, FTL needs to actually look up the database record to get the value of `new`. Surely doable but it'd remove one major speed benefit of the database. I'm neither convinced that it worth it nor do I have any actual numbers for comparison at hand.

jpgpi250:
Where do you want to discuss it than? [Here](https://discourse.pi-hole.net/t/gravity-database/46182)?

I can already answer some of your question regarding this, I've been experimenting already with this.

remember, my gravity database has 4.087.232 entries, there were 4.223 unique new entries, compared to the previous gravity database (I keep an old version of the gravity database, this to increase the diff between the databases every time I run 'pihole -g' (manual copy required)

image

How long does gravity run? As I see it, this wouldn't affect the gravity run at all (it would require new entries to be added with 3 fields, currently 2, but the new field would always be '0').
The gravity build would't change at all. The idea is to modify the 'new_entry' column (run the update query) just before the gravity.sh instruction 'mv "${gravityTEMPfile}" "${gravityDBfile}"' (apply the change on the 'TEMPfile'). This would than add a minute, give or take to the gravity run, but since that happens at night (sunday morning), you would only notice this if you forced a run, using 'pihole -g'

advanced/Scripts/query.sh
```diff
@@ -58,12 +61,131 @@ scanList(){
     esac
 }
 
+# function to count new domain entries since last gravity run
+CountNewDomains() {
+  printf %s "  ${INFO} Running query, please be patient..."
+  # store query result (count) in variable
+  newdomainentries=$(
```
DL6ER, line 68:
```suggestion
  newdomainentries="$(
```

advanced/Scripts/query.sh
```diff
@@ -58,12 +61,131 @@ scanList(){
...
+    sqlite3 "${gravityDBfile}" << EOSQL
+      ATTACH '${gravityOLDfile}' AS old;
+      SELECT count(DISTINCT domain) FROM gravity WHERE domain NOT IN (SELECT domain FROM old.gravity);
+EOSQL
+      )
```
DL6ER, line 73:
```suggestion
      )"
```

advanced/Scripts/query.sh
```diff
@@ -58,12 +61,131 @@ scanList(){
...
+  if [ -z "${all}" ]; then
+    QueryLimit="Limit 100"
+  fi
+  printf %s "  ${INFO} Running query, please be patient..."
+  # store query result (list) in array
+  # order by adlist.address makes the query slower, but the result is more useful
+  mapfile -t newdomainarray < <( sqlite3 "${gravityDBfile}" << EOSQL
+      ATTACH '${gravityOLDfile}' AS old;
+      SELECT domain, '' || adlist.address || '' FROM gravity
+      JOIN adlist ON adlist.id = gravity.adlist_id
```
DL6ER, line 89:
```suggestion
      JOIN adlist ON adlist.id = gravity.adlist_id
```

advanced/Scripts/query.sh
```diff
@@ -58,12 +61,131 @@ scanList(){
...
+      )
+  printf '\r%s\n' "$(tput el)  ${INFO} ${newdomainentries} new domain(s) found."
+  exit 0
+}
+
+# function to list new domain entries since last gravity run
+ListNewDomains() {
+  if [ -z "${all}" ]; then
+    QueryLimit="Limit 100"
+  fi
```
DL6ER, line 82:
```suggestion
  QueryLimit=""
  if [ -z "${all}" ]; then
    QueryLimit="Limit 100"
  fi
```

To prevent a possible SQL injection from outside. This is easily demonstrated by:
``` bash
echo "QueryLimit=\"ABC\"" >> /etc/pihole/pihole-FTL.conf
pihole -q -listnew -all
```
and getting the error:
``` plain
  [i] Running query, please be patient...Error: near line 2: near "ABC": syntax error
  [i] No new domains have been found.
```

The injection could have been something worse, like disabling all adlists or injecting some (unwanted) whitelist domains.

advanced/Scripts/query.sh
```diff
@@ -58,12 +61,131 @@ scanList(){
...
+    printf '\r%s\n' "$(tput el)  ${INFO} ${#newdomainarray[*]} new domain(s) found."
+  fi
+  for (( j=0; j<"${#newdomainarray[@]}"; j++ )); do
+    IFS='|'
+    read -r domain urllist <<< "${newdomainarray[j]}"
+    echo -e "  ${domain} (${COL_GREEN}${urllist}${COL_NC})"
+  done
+  if [[ -z "${all}" ]] && [[ "${#newdomainarray[*]}" -ge 100 ]]; then
+    echo -e "  ${INFO} Over 100 new domains found.
+        This can be overridden using the -all option"
```
DL6ER, line 107: Clarify what can be overwritten: the limit.

```suggestion
    echo -e "  ${INFO} Over 100 new domains found.
        If you want to see them all, append \"-all\" to the command"
```

advanced/Scripts/query.sh
```diff
@@ -58,12 +61,131 @@ scanList(){
...
+  exit 0
+}
+
 if [[ "${options}" == "-h" ]] || [[ "${options}" == "--help" ]]; then
     echo "Usage: pihole -q [option] <domain>
 Example: 'pihole -q -exact domain.com'
 Query the adlists for a specified domain
 
 Options:
+  -blocked            List blocked domains since last gravity run
```
DL6ER, line 186: I wonder if the command should also have a trailing `new` as the other have. Just for consistency. This is open for debate, of course.
```suggestion
  -blockednew         List blocked domains since last gravity run
```

gravity.sh
```diff
@@ -73,6 +73,7 @@ fi
 # have changed
 gravityDBfile="${GRAVITYDB}"
 gravityTEMPfile="${GRAVITYDB}_temp"
+gravityOLDfile="$(dirname -- "${gravityDBfile}")/gravity_old.db"
```
DL6ER, line 76:
```suggestion
gravityDIR="$(dirname -- "${gravityDBfile}")"
gravityOLDfile="${gravityDIR}/gravity_old.db"
```

gravity.sh
```diff
@@ -123,8 +124,24 @@ gravity_swap_databases() {
...
 
-  # Swap databases and remove old database
-  rm "${gravityDBfile}"
+  if [[ -z "${KEEPOLDGRAVITY}" ]] ; then
+    KEEPOLDGRAVITY="true"
+  fi
+
+  # Swap databases and remove or conditionally rename old database
+  # Number of available blocks on disk
+  availableBlocks=$(stat -f --format "%a" "$(dirname -- "${gravityDBfile}")")
```
DL6ER, line 133:
```suggestion
  availableBlocks=$(stat -f --format "%a" "${gravityDIR}")
```

gravity.sh
```diff
@@ -123,8 +124,24 @@ gravity_swap_databases() {
...
+  # Number of blocks, used by gravity.db
+  gravityBlocks=$(stat --format "%b" ${gravityDBfile})
+  # Only keep the old database if available disk space is at least twice the size of the existing gravity.db.
+  # Better be safe than sorry...
+  # The variable KEEPOLDGRAVITY in /etc/pihole-FTL.conf can be used to disable this feature.
```
DL6ER, line 138:
```suggestion
  # The variable KEEPOLDGRAVITY in /etc/pihole/pihole-FTL.conf can be used to disable this feature.
```

gravity.sh
DL6ER, line 907:
```suggestion
      mv "${gravityOLDfile}" "${gravityDIR}/gravity_mv.db"
```

gravity.sh
```diff
@@ -872,12 +889,44 @@ gravity_Cleanup() {
...
+  if [[ -f "${gravityDBfile}" ]]; then
+    echo -e "  ${TICK}  ${gravityDBfile} exists."
+    if [[ ${gravityOLDversion} -eq $(pihole-FTL "${gravityDBfile}" "SELECT value FROM info WHERE property IS 'version';") ]]; then
+      echo -e "  ${INFO}  Switching database."
+      echo "         current database, updated $(date -d @"$(pihole-FTL "${gravityDBfile}" "SELECT value FROM info WHERE property IS 'updated';")")"
+      oldDBdate=$(pihole-FTL "${gravityOLDfile}" "SELECT value FROM info WHERE property IS 'updated';")
+      echo "         old database, updated $(date -d @"${oldDBdate}")"
+      mv "${gravityOLDfile}" "$(dirname -- "${gravityOLDfile}")/gravity_mv.db"
+      mv "${gravityDBfile}" "${gravityOLDfile}"
+      mv "$(dirname -- "${gravityOLDfile}")/gravity_mv.db" "${gravityDBfile}"
```
DL6ER, line 909:
```suggestion
      mv "${gravityDIR}/gravity_mv.db" "${gravityDBfile}"
```

advanced/Scripts/query.sh
```diff
@@ -27,6 +28,8 @@ fi
 # Set this only after sourcing pihole-FTL.conf as the gravity database path may
 # have changed
 gravityDBfile="${GRAVITYDB}"
+gravityOLDfile="$(dirname -- "${gravityDBfile}")/gravity_old.db"
```
DL6ER, line 31:
```suggestion
gravityDIR="$(dirname -- "${gravityDBfile}")"
gravityOLDfile="${gravityDIR}/gravity_old.db"
```

DL6ER (changes requested):
Thanks for your submission and sorry for the review having taken so long. This is almost good to go, there are a few things I've found, including one SQL code injection vulnerability that needs to be fixed.

yubiuser:
> setupVars.conf: KEEPOLDGRAVITY=true|false (default true)

I don't think this should be `true` by default. If users are so advanced that they know/want/need this possibility they can easily switch it on.

jpgpi250:
I applied all suggested changes and some...
- query.sh : **blocked** needed to be changed into **blockednew** twice
- gravity.sh: eliminating excessive use of **dirname --** required adding **gravityDIR="$(dirname -- "${gravityDBfile}")"**. This was only suggested in query.sh

I don't understand why **rm "${gravityDBfile}"** should be removed. It's part of the original code. What I changed is:
- rename, using mv, DB to DB_old (if the conditions are ok, e.g. diskcheck and KEEPOLDGRAVITY=true)
OR
- erase, using rm, the DB (this is required because the next step is rename, using mv, DBtemp to DB.
The only other place where the identical rm command is used, is in the recreate_database code.

jpgpi250:
> `I don't think this should be true by default. If users are so advanced that they know/want/need this possibility they can easily switch it on.`

This will require some extra code to hide the additional options (help), and block processing of the command line options, if the value is false.

Fine by me, if everybody thinks this is necessary, just let me know...

jfb-pihole:
I agree that the default should be false.

jpgpi250:
> I agree that the default should be false.

> I don't think this should be true by default. If users are so advanced that they know/want/need this possibility they can easily switch it on.

done, default is false, additional help is hidden, new switches will not work, however, if KEEPOLDDATABASE is unset or false, the switches (-countnew, -listnew, -blockednew) will be interpreted as domain search parameter for pihole -q (searches for domain matches).

dschaper:
Was there a reason you closed and deleted the branch?

jpgpi250:
Last Saturday, I used 'pihole -g -s' (switch databases) to check something, and forgot to switch them back. As a result, the cron gravityupdate, running on sunday morning, deleted the most recent database (not active at that time, but renamed to old), removing all changes to clients, whitelist, blacklist, regex, url's, I made last week.
My bad, unlucky, but since this might happen to other users, once this would be merged, it didn't feel like a good idea anymore, so I decided to remove everything.

I will be starting over, submit a new PR, keeping only the code to keep the old database, nothing more (no switching, no additional queries). Since the general opinion appears to be this is for advanced users only (default false), they can write / execute queries for themselves, if required.

dschaper:
Oh, completely understandable. Thanks for letting us know what is going on. We were concerned that maybe you closed it because we hadn't been active with it. We've all been swamped with non Pi-hole things so we weren't intentionally ignoring it.

Please do submit a new PR when you have something and we can talk it over.

jpgpi250:
Sorry for the troubles I'm causing, thanks to all for tolerating me and my poor attempts to contribute.

dschaper:
Honestly those are not things that even came to mind. No troubles and your contributions are not poor.

PR #4206, opened 2021-06-20
Be explicit in IPv6 RA values.

### LINKED ISSUE #4142: Wrong IPv6 dhcp-range parameters for FTL/dnsmasq
If option `DHCP_IPv6` is true, then Pi-hole adds this [code to `/etc/dnsmasq.d/02-pihole-dhcp.conf`](
https://github.com/pi-hole/pi-hole/blob/4736e03108763cc2d5659f48d8a1e8a64d9b2608/advanced/Scripts/webpage.sh#L429):

```
dhcp-range=::100,::1ff,constructor:${interface},ra-names,slaac,${leasetime}
```

But according to `dnsmasq`'s manpage in [Ubuntu 18.04 LTS (Bionic Beaver)](
https://manpages.ubuntu.com/manpages/bionic/en/man8/dnsmasq.8.html) and [Debian 10 (Buster)](
https://manpages.debian.org/buster/dnsmasq-base/dnsmasq.8.en.html) the value of `${leasetime}` will not set `<lease time>` but `<prefix-len>` instead because `dhcp-range` option of `dnsmasq` looks like that:

```
-F, --dhcp-range=[tag:<tag>[,tag:<tag>],][set:<tag>,]<start-IPv6addr>[,<end-IPv6addr>|constructor:<interface>][,<mode>][,<prefix-len>][,<lease time>]
```

When using Pi-hole (e.g. with [`Docker Pi-hole`](https://github.com/pi-hole/docker-pi-hole/)) without defining environment variable `DHCP_LEASETIME` or defining it with a value less than `64` will result in an error from dnsmasq causing the container to fail with message:
```
dnsmasq: prefix length must be at least 64 at line 16 of /etc/dnsmasq.d/02-pihole-dhcp.conf
```

Pi-hole uses [FTL](https://github.com/pi-hole/FTL), but the relevant code for options parsing (in [`src/dnsmasq/option.c`](https://github.com/pi-hole/FTL/blob/b90ab8b1e78d7aece11c8fee1c3c00817761de52/src/dnsmasq/option.c#L3006)) is taken from [dnsmasq](https://thekelleys.org.uk/dnsmasq/) and thus the above still applies to FTL.

dschaper:

---

Fixes #4142 

Be explicit in the values for the IPv6 RA packets. 

The format for the range statement is: (from https://thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html)

`--dhcp-range=[tag:<tag>[,tag:<tag>],][set:<tag>,]<start-IPv6addr>[,<end-IPv6addr>|constructor:<interface>][,<mode>][,<prefix-len>][,<lease time>]`

We set the prefix length and hard code in a lease time. Previously we used the lease time variable that was used for IPv4 addressing. Because of other issues it was possible for a misinterpreted short lease time to be seen as a prefix length and cause `pihole-FTL`/`dnsmasq` to fail to start. 

The intention of the RA is to advertise the DNS capabilities. We do not intend to be a stateful IPv6 DHCP server. I think the hard code value for lease time is okay since there shouldn't ever be an actual lease generated. 

Be as explicit as possible for all values that should not vary.

None

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -426,7 +426,7 @@ dhcp-leasefile=/etc/pihole/dhcp.leases
         echo "#quiet-dhcp6
 #enable-ra
 dhcp-option=option6:dns-server,[::]
-dhcp-range=::100,::1ff,constructor:${interface},ra-names,slaac,${leasetime}
+dhcp-range=::100,::1ff,constructor:${interface},ra-names,slaac,64,3600
 ra-param=*,0,0
 " >> "${dhcpconfig}"
     fi


dschaper:
@JM1 Does this look like it's fixing your issue?

JM1:
@dschaper: Tested your patch and it works. Thanks!

DL6ER (approved):

PR #4207, opened 2021-06-20
Don't overwrite existing logrotate

### LINKED ISSUE #3822: /etc/pihole/logrotate is overwritten on upgrade
### Versions

- Pi-hole:  
- AdminLTE:  
- FTL:  

### Platform

- OS and version:  Ubuntu 20.04

- Platform:  Linux LXC

### Expected behavior

_A clear and concise description of what you expected to happen._
When pihole is updated, /etc/pihole/logrotate is overwritten.  Ideally this would be handled similarly to how package managers handle existing configuration files.
### Actual behavior / bug

_A clear and concise description of what the bug is._

### Steps to reproduce
customize /etc/pihole/logrotate and then upgrade

1. edit /etc/pihole/logrotate
2. upgrade pihole
3. review /etc/pihole/logrotate
4. See edits are gone
## Debug Token

- URL:  

### Screenshots

_If applicable, add screenshots to help explain your problem._

### Additional context

_Add any other context about the problem here._

see https://www.reddit.com/r/pihole/comments/jjasvs/logrotate_config_file_gets_overwritten_on_each/ for details

dschaper:

---

Fixes #3822 

Longstanding bug that would overwrite a users modified logrotate configuration. This would happen on update and is not intended.

Conditionally checks for file existence before installing the file.

```bash
root@debian-s-1vcpu-1gb-nyc1-01:~# ls -la /etc/pihole/logrotate
ls: cannot access '/etc/pihole/logrotate': No such file or directory

--- Run install script

  [✓] Installing latest logrotate script
root@debian-s-1vcpu-1gb-nyc1-01:~# ls -la /etc/pihole/logrotate
-rw-r--r-- 1 root root 12 Jun 20 18:17 /etc/pihole/logrotate

--- Run install script

  [i] Installing latest logrotate script...
        [i] Existing logrotate file found. No changes made.
```

None

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1954,7 +1954,7 @@ installLogrotate() {
         return 2
     fi
     # Copy the file over from the local repo
-    install -D -m 644 -T ${PI_HOLE_LOCAL_REPO}/advanced/Templates/logrotate ${target}
+    install -D -m 644 -T "${PI_HOLE_LOCAL_REPO}"/advanced/Templates/logrotate ${target}
     # Different operating systems have different user / group
     # settings for logrotate that makes it impossible to create
     # a static logrotate file that will work with e.g.
@@ -1963,7 +1963,7 @@ installLogrotate() {
     # the local properties of the /var/log directory
     logusergroup="$(stat -c '%U %G' /var/log)"
     # If there is a usergroup for log rotation,
-    if [[ ! -z "${logusergroup}" ]]; then
+    if [[ -n "${logusergroup}" ]]; then
         # replace the line in the logrotate script with that usergroup.
         sed -i "s/# su #/su ${logusergroup}/g;" ${target}
     fi
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1944,9 +1944,17 @@ finalExports() {
 # Install the logrotate script
 installLogrotate() {
     local str="Installing latest logrotate script"
+    local target=/etc/pihole/logrotate
+
     printf "\\n  %b %s..." "${INFO}" "${str}"
+    if [[ -f ${target} ]]; then
+        printf "\\n\\t%b Existing logrotate file found. No changes made.\\n" "${INFO}"
+        # Return value isn't that important, using 2 to indicate that it's not a fatal error but
+        # the function did not complete. 
+        return 2
+    fi
     # Copy the file over from the local repo
-    install -D -m 644 -T ${PI_HOLE_LOCAL_REPO}/advanced/Templates/logrotate /etc/pihole/logrotate
+    install -D -m 644 -T ${PI_HOLE_LOCAL_REPO}/advanced/Templates/logrotate ${target}
     # Different operating systems have different user / group
     # settings for logrotate that makes it impossible to create
     # a static logrotate file that will work with e.g.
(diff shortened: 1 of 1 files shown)


yubiuser:
Doesn't that mean, pihole itself can never update its own logrotate file? Just image you want, for whatever reason, change the default weekly update from Sunday to Wednesday

dschaper:
Yes, it does mean that. I don't think that's going to be an issue since we've never changed it in the many years it's been in existence. It's a good question but I don't see it as a problem. It's something the user should have full control over. The people that know why that file should be changed for their system also know they don't want us overriding their changes. The ones that don't know what the file is for probably don't care. 

I don't see a way for us to easily determine if a file is our initial one or if it's one that a user has modified for their system. The only ways I can see would be to do a line by line comparison of the file contents.

yubiuser:
> The only ways I can see would be to do a line by line comparison of the file contents.

Or use a hash function?

> I don't see a way for us to easily determine if a file is our initial one or if it's one that a user has modified for their system. 

I think there is no easy way. You would need to compare the current file against the previous version found in `{PI_HOLE_LOCAL_REPO}`: If they match, the file delivered with Pi-hole changed, if not, the user made the modification.

dschaper:
I guess we could do a hash similar to how we check for FTL's binary to match? 

So you're saying that we should

Check if TARGET doesn't exist or exists and hash matches SOURCE then copy; else return?

yubiuser:
```
if target exists
    if hash source matches taget //file unchanges, do nothing
      print "nothing changed"
    else
      if hash source-1 (previous version) matches taget // file delivered by Pi-hole changed
        copy
      else //user modified the file
        print "user modification, don't touching the file
      fi
    fi
else //create non-existing file
  copy 
fi
```

If you only compare TARGET and SOURCE (line by line or hash) you can not distinguish between user modification or Pi-hole modification. That's why you need to compare against SOURCE-1 as well. There is still the slim chance that a user modified the file in a way that it matches the SOURCE-1 version, but chances are rather low.

I don't know how FTL binaries are compared, but for the adlists DL used `sha1sum` to create a `.sha1` file. See https://github.com/pi-hole/pi-hole/pull/3951

dschaper:
How do we find SOURCE-1? It copies the file from `/etc/.pihole` which is the git repo cloned in. 

In the case of a fresh install there will never be a pre-existing file. In the case of an update then the `/etc/.pihole` directory will be the update version already so I'm not sure how we could find the hash of the previous version of the file? (Or I might not be understanding the concept.)

dschaper:
Ah, I think I see what you mean.

We install Pi-hole fresh, with a file that has hash value of 1.
We update the file in the repo to change something. Now it has a hash value of 2.

We update, the process compares hashes and says "Okay, 2 isn't 1 so copy". Which is true however we can't know that the user made a change and has hash value 3. We'd end up with "Okay, 2 isn't 3 so copy" and we blow away their changes. 

We'd need a way to say "Okay, 2 isn't 1 but I know that 1 was the previous version so it's okay to copy." Along with "2 isn't 3 but I've never seen 3 so I'm going to leave it alone."

I don't know how we'd do that.

DL6ER (approved):
> since we've never changed it in the many years

We will if we ever move our logs into the directory `/var/log/pihole/...`

Your suggestion seems fine for now, we can revisit this later when we really want to update the `logrotate` file.

The final solution will likely be including a directory with previous versions of the script. If the local script matches any of the files in this folder, we know it's a Pi-hole provided file and overwrite the `logrotate` script. If no match is found it is already the newest one or (more likely) a user-modified one which we want to leave alone (but still tell them that something changed and that this should check the CHANGELOG to see what is going on).

yubiuser:
I thought there might be a git way of "recreating" old versions of files from all changes applied to it backwards. But including the previous versions in a separate directory might be the easier way :-D

dschaper:
I guess you could do diff files. Which would also be an interesting idea to research for applying the changes. Instead of copying in a file completely we could apply a diff with the changes. I don't know if that would be helpful in this case and eventually I'd like to not rely on `git` anymore or have a clone of the repo on the Pi-hole devices as install. 

For now I think this works and the conversation brought up some good ideas to think about.

PR #4221, opened 2021-07-07
Remove mod_compress (mod_deflate) from lighttpd to provide backwards compatible support for Debian Bullseye (11)

lightswitch05:

---

Support Debian Bullseye without breaking support for older versions of Debian.

Remove mod_compress (mod_deflate) from Debian's lighttpd config to provide backwards compatible support for Debian Bullseye (11). I did testing of this on Bullseye through docker. Another benefit of this change is that we no longer get this message on boot:

> Warning: "mod_compress" is DEPRECATED and has been replaced with "mod_deflate".  A future release of lighttpd 1.4.x will not contain mod_compress and lighttpd may fail to start up

* `mod_compress` is deprecated
* `mod_deflate` requires a new package on Debian Bullseye which leads to package complexity with older Debian versions. See:
  * https://github.com/pi-hole/pi-hole/pull/4212
  * https://github.com/pi-hole/pi-hole/pull/4218

**Please note, that I left [Fedora's configuration](https://github.com/pi-hole/pi-hole/blob/a52a5e7ef2e202e386ae8362788c4e47ea308278/advanced/lighttpd.conf.fedora) alone since this issue only affects Debian.**

None?

---

### DIFF
--- a/advanced/lighttpd.conf.debian
+++ b/advanced/lighttpd.conf.debian
@@ -20,7 +20,6 @@ server.modules = (
     "mod_accesslog",
     "mod_auth",
     "mod_expire",
-    "mod_compress",
     "mod_redirect",
     "mod_setenv",
     "mod_rewrite"
@@ -40,26 +39,6 @@ accesslog.format            = "%{%s}t|%V|%r|%s|%b"
 index-file.names            = ( "index.php", "index.html", "index.lighttpd.html" )
 url.access-deny             = ( "~", ".inc", ".md", ".yml", ".ini" )
 static-file.exclude-extensions = ( ".php", ".pl", ".fcgi" )
-
-compress.cache-dir = "/var/cache/lighttpd/compress/"
-compress.filetype  = (
-    "application/json",
-    "application/vnd.ms-fontobject",
-    "application/xml",
-    "font/eot",
-    "font/opentype",
-    "font/otf",
-    "font/ttf",
-    "image/bmp",
-    "image/svg+xml",
-    "image/vnd.microsoft.icon",
-    "image/x-icon",
-    "text/css",
-    "text/html",
-    "text/javascript",
-    "text/plain",
-    "text/xml"
-)
 
 mimetype.assign = (
     ".ico"   => "image/x-icon",


PromoFaux (approved):

PromoFaux:
Actually, for consistency, I think it makes sense to remove from the fedora config, too. I'll line up a PR now

If people really want it it can always be manually added to `external.conf`

XhmikosR:
@PromoFaux Just to clarify, isn't `mod_deflate` included anymore in Debian 11?

Using compression results in like 65-70% savings so we should keep it on IMHO.

PromoFaux:
It's not worth the [added complexity](https://github.com/pi-hole/pi-hole/pull/4218) to the installer, especially as V6 will remove `lighttpd` entirely

Ultimately the web interface should only really be served on a local network, where bandwidth really shouldn't be an issue, ok.. maybe a VPN at a push. Are you able to provide real world figures of the difference between compression being on and off on the dashboard?


> If people really want it it can always be manually added to external.conf

XhmikosR:
Well, I know people can enable it manually, but the default should have compression on IMHO. When v6 is out, it can be removed if it's impossible to keep it, but for v5 it should stay in my opinion.

Locally, it might not make a huge difference, but it's a matter of promoting good practices. Nothing should be server uncompressed when the gains are so big...

lightswitch05:
Hey @XhmikosR - sadly, I'm the one who removed support for this. I tried to preserve support in the PR that @PromoFaux linked to above, but unfortunately my efforts failed. After my initial attempt failed, it looked like the only solution left is adding a bunch of OS-specific packages and dealing with back-ports, which was undesirable from a support and complexity point of view - especially considering `lighttpd` will be removed entirely in the future. Anyways, [this is a good comment to read](lighttpd) for a breakdown. Just being straightforward with you, I don't have a lot of systems experience, so perhaps there is a clean solution to this problem which won't become a maintenance nightmare.

dschaper:
I appreciate the comments but I think this is a case of needing to do what we can at the best of our ability. I don't want to add more complexity for the minimal gains that it may bring. The interface is light, the data we send is minimal, putting a turbo engine on a tricycle may be impressive but it doesn't really do much for anyone.

PR #1130, opened 2017-01-12
Use logrotate for pihole.log

DL6ER:
**By submitting this pull request, I confirm the following (please check boxes, eg [X]) _Failure to fill the template will close your PR_:**

***Please submit all pull requests against the `development` branch. Failure to do so will delay or deny your request***

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] I have checked that [another pull request](https://github.com/pi-hole/pi-hole/pulls) for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:** 10

---

Add logrotate for pihole.log (see also [this thread](https://discourse.pi-hole.net/t/implement-logrotation-for-var-log-pihole-log/1266/2) on Discourse)

Selected options:

- `daily` (rotate the log once per day)
- `copytruncate` (copy the existing file to the new place and truncate the original file afterwards)
- `rotate 5` (keep 5 days/manual flush cycles)
- `compress` + `delaycompress` (compress files older than 1 day)
- `notifyempty` (do not rotate the log if it is empty)
- `nomail` (don't mail old log files to any address)

As fallback the old flushing routine will still be used if `logrotate` is not found.

_This template was created based on the work of [`udemy-dl`](https://github.com/nishad/udemy-dl/blob/master/LICENSE)._

PR #4225, opened 2021-07-09
Give pihole its own logrotate state file

duxovni:

---
In the pihole cron file, `pihole flush` is run daily at midnight. On many systems, this occasionally happens at the same time as a default system-installed logrotate cronjob. When multiple invocations of logrotate are running at the same time, they may attempt to operate on the default logrotate state file `/var/lib/logrotate/status` simultaneously, causing errors. For this reason, pihole's invocation of logrotate needs to use its own separate state file.

This PR modifies the `piholeLogFlush.sh` script and the `pihole.cron` cronjob, providing all invocations of logrotate with the additional flag `--state /var/lib/logrotate/pihole` so that they won't run into conflicts with other logrotate processes.

None.

---

### DIFF
--- a/advanced/Scripts/piholeLogFlush.sh
+++ b/advanced/Scripts/piholeLogFlush.sh
@@ -11,6 +11,11 @@
 colfile="/opt/pihole/COL_TABLE"
 source ${colfile}
 
+# In case we're running at the same time as a system logrotate, use a
+# separate logrotate state file to prevent stepping on each other's
+# toes.
+STATEFILE="/var/lib/logrotate/pihole"
+
 # Determine database location
 # Obtain DBFILE=... setting from pihole-FTL.db
 # Constructed to return nothing when
@@ -32,7 +37,7 @@ if [[ "$@" == *"once"* ]]; then
     # Nightly logrotation
     if command -v /usr/sbin/logrotate >/dev/null; then
         # Logrotate once
-        /usr/sbin/logrotate --force /etc/pihole/logrotate
+        /usr/sbin/logrotate --force --state "${STATEFILE}" /etc/pihole/logrotate
     else
         # Copy pihole.log over to pihole.log.1
         # and empty out pihole.log
@@ -47,8 +52,8 @@ else
     # Manual flushing
     if command -v /usr/sbin/logrotate >/dev/null; then
         # Logrotate twice to move all data out of sight of FTL
-        /usr/sbin/logrotate --force /etc/pihole/logrotate; sleep 3
-        /usr/sbin/logrotate --force /etc/pihole/logrotate
+        /usr/sbin/logrotate --force --state "${STATEFILE}" /etc/pihole/logrotate; sleep 3
+        /usr/sbin/logrotate --force --state "${STATEFILE}" /etc/pihole/logrotate
     else
         # Flush both pihole.log and pihole.log.1 (if existing)
         echo " " > /var/log/pihole.log
--- a/advanced/Templates/pihole.cron
+++ b/advanced/Templates/pihole.cron
@@ -26,7 +26,7 @@
 #          parameter "quiet": don't print messages
 00 00   * * *   root    PATH="$PATH:/usr/sbin:/usr/local/bin/" pihole flush once quiet
 
-@reboot root /usr/sbin/logrotate /etc/pihole/logrotate
+@reboot root /usr/sbin/logrotate --state /var/lib/logrotate/pihole /etc/pihole/logrotate
 
 # Pi-hole: Grab local version and branch every 10 minutes
 */10 *  * * *   root    PATH="$PATH:/usr/sbin:/usr/local/bin/" pihole updatechecker local


yubiuser:
Thanks for the PR.
Wouldn't be a simpler solution to rotate 5 min past midnight?

duxovni:
No, that won't work. On my Debian 10 server, for example, the system 
logrotate is in cron.daily rather than being scheduled with a specific time. The conflict only happens intermittently, about once every two weeks but not quite regularly as far as I can tell. Having a separate state file is a simple bulletproof way to squash this bug regardless of system configuration, and is recommended by the logrotate manpage.

advanced/Scripts/piholeLogFlush.sh
```diff
@@ -32,7 +37,7 @@ if [[ "$@" == *"once"* ]]; then
     # Nightly logrotation
     if command -v /usr/sbin/logrotate >/dev/null; then
         # Logrotate once
-        /usr/sbin/logrotate --force /etc/pihole/logrotate
+        /usr/sbin/logrotate --force --state $STATEFILE /etc/pihole/logrotate
```
dschaper, line 40:
```suggestion
        /usr/sbin/logrotate --force --state "${STATEFILE}" /etc/pihole/logrotate
```

https://github.com/koalaman/shellcheck/wiki/SC2086

Needs to be quoted, and the `{}` are a style that we use.

advanced/Scripts/piholeLogFlush.sh
```diff
@@ -47,8 +52,8 @@ else
     # Manual flushing
     if command -v /usr/sbin/logrotate >/dev/null; then
         # Logrotate twice to move all data out of sight of FTL
-        /usr/sbin/logrotate --force /etc/pihole/logrotate; sleep 3
-        /usr/sbin/logrotate --force /etc/pihole/logrotate
+        /usr/sbin/logrotate --force --state $STATEFILE /etc/pihole/logrotate; sleep 3
+        /usr/sbin/logrotate --force --state $STATEFILE /etc/pihole/logrotate
```
dschaper, line 56:
```suggestion
        /usr/sbin/logrotate --force --state "${STATEFILE}" /etc/pihole/logrotate; sleep 3
        /usr/sbin/logrotate --force --state "${STATEFILE}" /etc/pihole/logrotate
```

dschaper (changes requested):
Shellcheck/Style fixes.

dschaper:
Can you copy/paste in the recommendations so we can have them as part of the PR history/docs?

duxovni:
From the Debian 10 logrotate manpage:
> -s, --state statefile
> Tells logrotate to use an alternate state file.  This is useful if logrotate is being run as a different user for various sets of log files.  The default state file is /var/lib/logrotate/status.

In this case it's not a matter of different users, but it's the same issue of multiple separate logrotate invocations that can't be merged into a single config with a single cronjob.

MichaIng:
What is the reason that Pi-hole even uses its own logrotate instance instead of creating a config file for the system logrotate instance, i.e. `/etc/logrotate.d/pihole`? Are the implementations so different across distros?

dschaper:
That's a good question. I don't know if there was a specific reason for the current setup. It's about 5 years old at this point.

The PR that introduced it is https://github.com/pi-hole/pi-hole/pull/1130

It's linked to discussion at https://discourse.pi-hole.net/t/implement-logrotation-for-var-log-pihole-log/1266 

@DL6ER may have an idea or it may have been implementation of what was discussed on Discourse and nothing more.

MichaIng:
Also pinging @jpgpi250 which had the idea initially 🙂.

jpgpi250:
I was asked by Dan [here](https://discourse.pi-hole.net/t/failure-in-logrotate-installation-function/48533/5) to take a look at this. The observation regarding the state file is correct, I already noticed this a long time ago, see [here](https://discourse.pi-hole.net/t/make-location-of-pihole-log-files-configurable/20636/8), simply adding **--state /file-location** solves the problem.

A long time ago, the pihole installation script created /var/log/pihole, clearly with the intention to keep the logs in a separate directory, however, the logfiles didn't follow that intention... My installation (one instance only) modifies this behavior (pihole logs in /var/log/pihole), thus making it possible to keep the state file in the same folder.
As a result, a small file is created / updated (var/log/pihole/status), content:
`logrotate state -- version 2`
`"/var/log/pihole/pihole-FTL.log" 2021-7-28-0:0:1`
`"/var/log/pihole/pihole.log" 2021-7-28-0:0:1`

My cron job looks like this:
`0 0 * * * root /usr/sbin/logrotate --force --state /var/log/pihole/status /etc/pihole/logrotate`

- using TMPFS for logging, thus no cron job for @reboot.
- I haven't changed the log file configuration in the pihole configuration files, simply replaced the log files with symlinks to achieve the move to /var/log/pihole

dschaper:
How would it affect you if we moved to using the systems `/etc/logrotate*` directories instead?

jpgpi250:
I assume you want to move /etc/pihole/logrotate to /etc/logrotate.d/pihole, without changing the actual rotation instructions.
I don't see any problem (for me) with this, it would make the separate status file obsolete.

however

It will require changes (if even possible) to /opt/pihole/piholeLogFlush.sh. Looking at that code, **pihole flush** will also work on systems without logrotate, which is probably the reason why the rotation isn't handled by the system (/etc/logrotate.d).
Do you want to make piholeLogFlush.sh more complicated, this to check if rotation actually took place?

I remember discussing logrotation (the feature request) with DL6ER (the initial problem was it requires copytruncate to make things work with dnsmasq - pihole-FTL didn't exist yet). You should ask him, if he remembers why he moved the rotation out of the systems control and opted for /opt/pihole/piholeLogFlush.sh. The code indicates it ensures rotation, even without logrotate.

dschaper:
> It will require changes (if even possible) to /opt/pihole/piholeLogFlush.sh.

Good point that I didn't consider. We'll have to look at that closely.

dschaper:
Okay, I think this is a good start to get going with. 

We can look at moving to a system directory for logrotate in the future but it appears (I have not tested) that this works now.

dschaper (approved):

MichaIng:
Should be fine, and indeed moving to system's logrotate can be done separately.

Checking on reasons:
> Do you want to make piholeLogFlush.sh more complicated, this to check if rotation actually took place?

I don't see a reason why the script would need to do an additional check. It currently checks whether the logrotate executable exists, and that should be sufficient: Only when **no** logrotate executable exist, the script needs to do it manually. If the logrotate executable does exist, but its system cron job (or is it a systemd timer?) is disabled, then obviously the admin does not want it to happen automatically, so Pi-hole should not do it either? This would simplify the code instead of making it more complicated 😎.

PR #4246, opened 2021-07-28
Remove comparison of IP addresses with setupVars.conf

DL6ER:

---

Setting `IPV4_ADDRESS` and `IPV6_ADDRESS` in `setupVars.conf` is deprecated in Core v5.4, the debugger still checked if the address in this file is the same as found on the interfaces.

Remove the check to avoid false-negative errors.

None

### DIFF
--- a/advanced/Scripts/piholeDebug.sh
+++ b/advanced/Scripts/piholeDebug.sh
@@ -605,38 +605,6 @@ parse_locale() {
     parse_file "${pihole_locale}"
 }
 
-does_ip_match_setup_vars() {
-    # Check for IPv4 or 6
-    local protocol="${1}"
-    # IP address to check for
-    local ip_address="${2}"
-    # See what IP is in the setupVars.conf file
-    local setup_vars_ip
-    setup_vars_ip=$(< ${PIHOLE_SETUP_VARS_FILE} grep IPV"${protocol}"_ADDRESS | cut -d '=' -f2)
-    # If it's an IPv6 address
-    if [[ "${protocol}" == "6" ]]; then
-        # Strip off the / (CIDR notation)
-        if [[ "${ip_address%/*}" == "${setup_vars_ip%/*}" ]]; then
-            # if it matches, show it in green
-            log_write "   ${COL_GREEN}${ip_address%/*}${COL_NC} matches the IP found in ${PIHOLE_SETUP_VARS_FILE}"
-        else
-            # otherwise show it in red with an FAQ URL
-            log_write "   ${COL_RED}${ip_address%/*}${COL_NC} does not match the IP found in ${PIHOLE_SETUP_VARS_FILE} (${FAQ_ULA})"
-        fi
-
-    else
-        # if the protocol isn't 6, it's 4 so no need to strip the CIDR notation
-        # since it exists in the setupVars.conf that way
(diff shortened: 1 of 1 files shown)


advanced/Scripts/piholeDebug.sh
```diff
@@ -653,8 +621,7 @@ detect_ip_addresses() {
         log_write "${TICK} IPv${protocol} address(es) bound to the ${PIHOLE_INTERFACE} interface:"
         # Since there may be more than one IP address, store them in an array
         for i in "${!ip_addr_list[@]}"; do
-            # For each one in the list, print it out
-            does_ip_match_setup_vars "${protocol}" "${ip_addr_list[$i]}"
+            log_write "    ${ip_addr_list[$i]}"
```
dschaper, line 624: Is there any calls to `pihole-FTL` that will expose what IP addresses it will respond to? Can we add them here if there is that functionality?

DL6ER: No, this does not exist. `dnsmasq` determines this at run-time. We could rebuild this for ourselves but it would be quite involved.

dschaper: Okay, might be something to put on the list to consider.

Would be nice to have `pihole-FTL --list-address` or similar that would output a list of the interface and the address it is using for that interface.

yubiuser: I guess one has to deduct from `PIHOLE_INTERFACE=XXXX` and the listening behavior based on `/etc/dnsmasq.d/01-pihole.conf` and `Name resolution (IPv4/IPv6) using a random blocked domain and a known ad-serving domain`?

DL6ER: The main issue I see right now is that it would launch a dedicated engine (instead of interfacing with the existing one). So the new engine would need to read and parse the config and find out itself how the currently running engine *might* work (if the config wasn't changed).

We may rather add an API endpoint for this to query this. We'd need a JSON deserializer for bash here, but that shouldn't be an issue (PADD / chronometer needs this as well as dependency).

dschaper: `jq`

DL6ER: @yubiuser I didn't see your reply. Mind there may be additional lines in any user-provided config files that may contain lines like `listen-address=...`

dschaper: Would we be able to set a CHAOS or TXT record that emits that information when queried?

DL6ER: Yes, that would maybe possible. `dnsmasq` has a rather complex determination if it should listen to an address. There are various modes, the default being listen on wildcards and possibly ignore some when we should not listen to them. This can only be determined at runtime and I wouldn't know how to, e.g., specify "listen to all but a few subnets" as CHAOS string. Encoding and parsing would be quite a complex task.

It might be easier to ask for a specific address and let `dnsmasq` check if it would listen to this one. But is this what we need?

PromoFaux (approved):
Tested working (https://tricorder.pi-hole.net/vBskL3YX/)

Leaving unmerged in case @dschaper has anything to add (if not, just press the button)

PR #4257, opened 2021-08-03
Remove dhcpcd5 dependency

PromoFaux:

---

An alternative to https://github.com/pi-hole/pi-hole/pull/3715 

Removes the `dhcpcd5` dependency - which causes issues when installed on any OS other than Raspbian.

This PR removes all functions that (attempt to) set a static IP on the device that Pi-hole is being installed on. In the other PR, some dhcpcd5 config stuff is left in for setting the IP to static on Raspbian, but I personally (and quite strongly) believe that we should make a clean break here and leave it with the users to configure their network themselves. For us, as developers/maintainers it decreases the complexity of the install script (the more stuff we can rip out, the easier it will be to make it a simple bootstrap script for installing FTL)

I've also introduced a change to the third welcome dialog about static IPs that explains the need for a static IP, and asks the user to choose yes to indicate that they understand. This is defaulted to no, so button mashers will at least get tripped up.

youmustagreetocontinue

Something else to note, too: since there is no longer anything that sources the IP addresses from `setupVars.conf`, the installer doesn't even need to know the IP address of the device that Pi-hole is being installed on. FTL is able to determine the IP address that it has been queried on dynamically, which is nice. So that's some more stuff that can be ripped out of this install script in due course.

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -288,21 +288,6 @@ if is_command apt-get ; then
     PKG_INSTALL=("${PKG_MANAGER}" -qq --no-install-recommends install)
     # grep -c will return 1 if there are no matches. This is an acceptable condition, so we OR TRUE to prevent set -e exiting the script.
     PKG_COUNT="${PKG_MANAGER} -s -o Debug::NoLocking=true upgrade | grep -c ^Inst || true"
-    # Some distros vary slightly so these fixes for dependencies may apply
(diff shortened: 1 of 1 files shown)


yubiuser:
I prefer this PR over #3715 . Get rid of all the `dhcp5` stuff. Users running their own DNS servers should be able to grasp the concept of static IPs (or DHCP reservations)

PR #368, opened 2016-03-07
uptime -p doesn't work on older versions

Morlok8k:
Fixes pull request #328 

Changes proposed in this pull request:

uptime -p doesn't work in version 3.3.3 (which is in wheezy)

this new command works with 3.3.3 and was taken from here: http://stackoverflow.com/questions/28353409/bash-format-uptime-to-show-days-hours-minutes

PR #4265, opened 2021-08-07
Add uptime to debug log

yubiuser:

---
Adds the system's uptime to the debug log. Sometimes it is unclear for how long a system was running or if it had an unnoticed reboot in between.

Bildschirmfoto zu 2021-08-07 06-31-09

Re-use the uptime code from PADD

### DIFF
--- a/advanced/Scripts/piholeDebug.sh
+++ b/advanced/Scripts/piholeDebug.sh
@@ -230,6 +230,7 @@ copy_to_debug_log() {
 }
 
 initialize_debug() {
+    local system_uptime
     # Clear the screen so the debug log is readable
     clear
     show_disclaimer
@@ -237,6 +238,9 @@ initialize_debug() {
     log_write "${COL_PURPLE}*** [ INITIALIZING ]${COL_NC}"
     # Timestamp the start of the log
     log_write "${INFO} $(date "+%Y-%m-%d:%H:%M:%S") debug log has been initialized."
+    # Uptime of the system
+    system_uptime=$(uptime | awk -F'( |,|:)+' '{if ($7=="min") m=$6; else {if ($7~/^day/){if ($9=="min") {d=$6;m=$8} else {d=$6;h=$8;m=$9}} else {h=$6;m=$7}}} {print d+0,"days,",h+0,"hours,",m+0,"minutes"}')
+    log_write "${INFO} System is running for ${system_uptime}"
 }
 
 # This is a function for visually displaying the current test that is being run.
--- a/advanced/Scripts/piholeDebug.sh
+++ b/advanced/Scripts/piholeDebug.sh
@@ -239,8 +239,9 @@ initialize_debug() {
     # Timestamp the start of the log
     log_write "${INFO} $(date "+%Y-%m-%d:%H:%M:%S") debug log has been initialized."
     # Uptime of the system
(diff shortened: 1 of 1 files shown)


yubiuser:
`uptime` is part of `procps` (at least on Debian based systems). Can we assume that is available by default or should I add it to the installer dependencies?

advanced/Scripts/piholeDebug.sh
```diff
@@ -230,13 +230,17 @@ copy_to_debug_log() {
...
+    local system_uptime
     # Clear the screen so the debug log is readable
     clear
     show_disclaimer
     # Display that the debug process is beginning
     log_write "${COL_PURPLE}*** [ INITIALIZING ]${COL_NC}"
     # Timestamp the start of the log
     log_write "${INFO} $(date "+%Y-%m-%d:%H:%M:%S") debug log has been initialized."
+    # Uptime of the system
+    system_uptime=$(uptime | awk -F'( |,|:)+' '{if ($7=="min") m=$6; else {if ($7~/^day/){if ($9=="min") {d=$6;m=$8} else {d=$6;h=$8;m=$9}} else {h=$6;m=$7}}} {print d+0,"days,",h+0,"hours,",m+0,"minutes"}')
```
DL6ER, line 242: Why do we need such a complex processing of the output?
Anyway, please add a comment about where you got this from. I guess one of the answers from

https://stackoverflow.com/questions/56669767/escaping-complicated-command-line-arguments-in-bash-and-ssh

?

DL6ER (commented):
Yes, `procps` is always there:
```
$ aptitude search ~prequired -F"%p" | grep "procps"
libprocps8
libprocps8:i386
procps
procps:i386
```

see [Debian Policy Manual » 3. Binary packages](https://www.debian.org/doc/debian-policy/ch-binary.html#s3.7):

> ## 3.7. Base system
> 
> The base system is a minimum subset of the Debian system that is installed before everything else on a new system. Only very few packages are allowed to form part of the base system, in order to keep the required disk usage very small.
> 
> The base system consists of **all** those **packages with priority `required`** or `important`. Many of them will be tagged essential (see below).

advanced/Scripts/piholeDebug.sh
yubiuser: > Re-use the uptime code from PADD

yubiuser: Tracked it down to
https://github.com/pi-hole/pi-hole/pull/368

Their source was 
https://stackoverflow.com/questions/28353409/bash-format-uptime-to-show-days-hours-minutes

Can add a note to the comment

DL6ER: Yeah, it may just being nit-picky, but it's always nice to respect (original) authorship.

advanced/Scripts/piholeDebug.sh
```diff
@@ -230,13 +230,17 @@ copy_to_debug_log() {
...
+    log_write "${INFO} System is running for ${system_uptime}"
```
PromoFaux, line 243: Talking of nit-picky...

```suggestion
    log_write "${INFO} System has been running for ${system_uptime}"
```

DL6ER (approved):

PR #4269, opened 2021-08-10
Configure stale action

dschaper:

Configure Stale action for Core.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,4 +1,4 @@
-name: Mark stale issues and pull requests
+name: Mark stale issues
 
 on:
   schedule:
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -18,7 +18,7 @@ jobs:
         debug: true
         days-before-stale: 30
         days-before-close: 5
-        stale-issue-message: 'Stale issue message'
+        stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Please comment or update this issue or it will be closed in 5 days.'
         stale-issue-label: 'Submitter Attention Required'
         exempt-issue-labels: 'pinned, Fixed in next release'
         exempt-all-issue-assignees: true
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -20,5 +20,5 @@ jobs:
         days-before-close: 5
         stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Please comment or update this issue or it will be closed in 5 days.'
         stale-issue-label: 'Submitter Attention Required'
-        exempt-issue-labels: 'pinned, Fixed in next release'
+        exempt-issue-labels: 'pinned, Fixed in next release', 'Bug: Confirmed'
         exempt-all-issue-assignees: true
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,24 @@
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+  - cron: '30 * * * *'
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        debug: true
+        days-before-stale: 30
+        days-before-close: 5
+        stale-issue-message: 'Stale issue message'
+        stale-issue-label: 'Submitter Attention Required'
+        exempt-issue-labels: 'pinned, Fixed in next release'
+        exempt-all-issue-assignees: true
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -20,5 +20,5 @@ jobs:
         days-before-close: 5
         stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Please comment or update this issue or it will be closed in 5 days.'
         stale-issue-label: 'Submitter Attention Required'
-        exempt-issue-labels: 'pinned, Fixed in next release', 'Bug: Confirmed'
+        exempt-issue-labels: 'pinned, Fixed in next release, Bug: Confirmed'
         exempt-all-issue-assignees: true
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,24 @@
...
+      issues: write
+
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        debug: true
+        days-before-stale: 30
+        days-before-close: 5
+        stale-issue-message: 'Stale issue message'
```
DL6ER, line 21: Is this the message that will get posted?

dschaper: Yes, that was a placeholder that I didn't change. Will update.

dschaper:
```suggestion
        stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Please comment or update this issue or it will be closed in 5 days.'
```

.github/workflows/stale.yml
```diff
@@ -0,0 +1,24 @@
...
+  stale:
+
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
```
PromoFaux, line 17: Should this be set up as an org secret instead of repo secret?

dschaper: It's not a token we actually set. Any token that starts with `GITHUB_` is reserved by GitHub for their use.

See https://docs.github.com/en/actions/reference/authentication-in-a-workflow

I had the same confusion and tried to set a PAT for Pralor with that value. GitHub won't let you.

PromoFaux: AH, right ok that makes sense!

.github/workflows/stale.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+
+    steps:
+    - uses: actions/stale@v3
```
PromoFaux, line 15: Just fyi looks like there is a `v4`, now.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,24 @@
...
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        debug: true
```
PromoFaux, line 18: @yubiuser my understanding is that it would be in the log output of the Action run, which will show up here: https://github.com/pi-hole/pi-hole/actions

.github/workflows/stale.yml
```diff
@@ -0,0 +1,24 @@
+name: Mark stale issues and pull requests
```
PromoFaux, line 1: Have you missed PR's or do we just want to do it issues first to test?

.github/workflows/stale.yml
PromoFaux: See also https://github.com/actions/stale#debugging

DL6ER:
Can we revive this?

.github/workflows/stale.yml
dschaper: Thanks, will check.

.github/workflows/stale.yml
dschaper: Intentionally set to debug so that we have an idea on the first run what the result would be when it goes to live mode.

.github/workflows/stale.yml
dschaper: I was just going to to Issues first since we have a relatively good handle on PRs.

.github/workflows/stale.yml
dschaper:
```suggestion
    - uses: actions/stale@v4
```

.github/workflows/stale.yml
yubiuser: `debug` or `debug-only`?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+    - uses: actions/stale@v4
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        debug: true
+        days-before-stale: 30
+        days-before-close: 5
+        stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Please comment or update this issue or it will be closed in 5 days.'
+        stale-issue-label: 'Submitter Attention Required'
+        exempt-issue-labels: 'pinned, Fixed in next release'
```
yubiuser, line 23:
```suggestion
        exempt-issue-labels: 'pinned, Fixed in next release', 'Bug: Confirmed'
```

.github/workflows/stale.yml
dschaper:
```suggestion
        debug-only: true
```

I think that may have changed between v3 and v4.

.github/workflows/stale.yml
PromoFaux: In which case (I know I know, I'm a stickler....)

```suggestion
name: Mark stale issues
```

.github/workflows/stale.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+    - uses: actions/stale@v4
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        debug-only: true
+        days-before-stale: 30
+        days-before-close: 5
+        stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Please comment or update this issue or it will be closed in 5 days.'
+        stale-issue-label: 'Submitter Attention Required'
+        exempt-issue-labels: 'pinned, Fixed in next release', 'Bug: Confirmed'
```
PromoFaux, line 23: typo?

```suggestion
        exempt-issue-labels: 'pinned, Fixed in next release, Bug: Confirmed'
```

PromoFaux: Other than that, good to go!

yubiuser: My typo :-|

yubiuser:
One last question before approving: does it need to go directly to `master` or is `development` sufficient?

dschaper:
It needs to be on the default branch in order for it to trigger.

yubiuser (approved):

DL6ER (approved):
Engage

Issue #2945, opened 2019-09-26
piholeDebug.sh wrong bind analysis


Somebodyisnobody:
**In raising this issue, I confirm the following:** `{please fill the checkboxes, e.g: [X]}`

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the the source code relevant to this issue?:** 6-7

---
**Current behaviour:**

Following code in the `pihole -d` command doesn't match the description and is in result unusable to make a safe analysis:
https://github.com/pi-hole/pi-hole/blob/1d43c0a2fa36ddc993db51b1751e26cbc8123fbf/advanced/Scripts/piholeDebug.sh#L545-L570

Currently `ip -"${protocol}" addr show dev "${PIHOLE_INTERFACE}" | awk -F ' ' '{ for(i=1;i<=NF;i++) if ($i ~ '/^inet/') print $(i+1) }'` gives you just the output of available IP-addresses on a specified interface. But the log says these addresses are bound to pihole-FTL.
`"${TICK} IPv${protocol} address(es) bound to the ${PIHOLE_INTERFACE} interface:"`

I read this as the addresses that are also bound to FTL, cause that's the only reason for the data collection. Thats wrong in a case when you get IP addresses after pihole-FTL has already been started **and** pihole-FTL isn't bound to `::` or `0.0.0.0` but on the addresses that were available when the daemon started.

**Expected behaviour:**
Solution A:
Use something like `netstat -tulpen | grep :53` instead or supplementary

Solution B:
Rename line 558 to:
`"${TICK} IPv${protocol} address(es) available to the ${PIHOLE_INTERFACE} interface:"`

Solution C:
Make sure that pihole-FTL is bound to `::` and `0.0.0.0` and warn if it's not.

[...]

PR #4273, opened 2021-08-13
Change wording of available interfaces in debug script

yubiuser:

---
Fixes https://github.com/pi-hole/pi-hole/issues/2945

Only change wording

### DIFF
--- a/advanced/Scripts/piholeDebug.sh
+++ b/advanced/Scripts/piholeDebug.sh
@@ -618,7 +618,7 @@ detect_ip_addresses() {
         # Local iterator
         local i
         # Display the protocol and interface
-        log_write "${TICK} IPv${protocol} address(es) bound to the ${PIHOLE_INTERFACE} interface:"
+        log_write "${TICK} IPv${protocol} address(es) available to the ${PIHOLE_INTERFACE} interface:"
         # Since there may be more than one IP address, store them in an array
         for i in "${!ip_addr_list[@]}"; do
             log_write "    ${ip_addr_list[$i]}"


advanced/Scripts/piholeDebug.sh
```diff
@@ -618,7 +618,7 @@ detect_ip_addresses() {
         # Local iterator
         local i
         # Display the protocol and interface
-        log_write "${TICK} IPv${protocol} address(es) bound to the ${PIHOLE_INTERFACE} interface:"
+        log_write "${TICK} IPv${protocol} address(es) available to the ${PIHOLE_INTERFACE} interface:"
```
DL6ER, line 621:
```suggestion
        log_write "${TICK} IPv${protocol} address(es) attached to the ${PIHOLE_INTERFACE} interface:"
```

DL6ER (changes requested):
Well, it's not a mistake as such. The OP of the references issue said:

> But the log says these addresses are bound to `pihole-FTL`.

No, it doesn't. This is just their own interpretation as they admit a bit further down:

> I read this as the addresses that are also bound to FTL, cause that's the only reason for the data collection.

---

I'm not convinced an address can be "available" to an interface. It really is "bound" to the interface. Note: this has nothing to do with any process possibly bound to the addresses as well.

`man ip address` says:
``` plain
DESCRIPTION
       The address is a protocol (IPv4 or IPv6) address **attached** to a network
       device. Each device must have at least one address to use the corre‐
       sponding protocol. It is possible to have several different addresses
       attached to one device. These addresses are not discriminated, so that
       the term alias is not quite appropriate for them and we do not use it
       in this document.

       The ip address command displays addresses and their properties, adds
       new addresses and deletes old ones.
```

Hence, if we make this change at all, I'd vote for "attached" instead of "available". However, "attached" is that close to "bound" that the change may not be needed but I'll not vote against it.

Feedback by others wanted, too.

advanced/Scripts/piholeDebug.sh
jfb-pihole: Bound is a common term.  Apps bind to ports and interfaces.

dschaper:
I have no opinion on it. How ever you'd like to see it. `bound` was fine for me.

yubiuser:
I'm fine with leaving it as `bond` too.  Going to close this PR and the issue.

PR #4277, opened 2021-08-14
Replace whiptails in RPM part of package_manager_detect

yubiuser:

---
In the `RPM` block of `detect_package_manager` two whiptail are present, despite we can not assume this package is already installed at this point, because installation of `install_dependent_packages "${INSTALLER_DEPS[@]}` happens later in the script.
This PR replaces both whiptails with `read -p`

See the discussion here https://github.com/pi-hole/pi-hole/issues/3297

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -449,8 +449,8 @@ elif is_command rpm ; then
             # The PHP version available via default repositories is older than version 7
             printf "  %b PHP 7 Update (recommended)\\n" "${INFO}"
             printf "  %b PHP 7.x is recommended for both security and language features.\\n" "${INFO}"
-            printf "  %b Would you like to install PHP7 via Remi's RPM repository?\\n" ${INFO}"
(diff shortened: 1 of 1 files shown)


dschaper:
I think a better approach is to not include Remi or `PHP` in the `package_manager_detect`. We should determine if `apt` or `yum|dnf` needs to be called, then install the `installer-deps` and then after we know the installer has all the tools it needs we can go ahead and do the extra repo additions.

yubiuser:
I can do that for the `PHP` part, but what's about the `Unsupported RPM based distribution` whiptail? I think we should not even install `installer-deps` if the user don't wish to continue at this point. So this whiptail need to be replaced...

yubiuser:
Continue here https://github.com/pi-hole/pi-hole/pull/4278

dschaper:
Any unsupported RPM based distribution would be caught in the DNS check for supported versions. Maybe we don't need to keep that `Unsupported` line anymore?

yubiuser:
I've been thinking about this too.

PR #4282, opened 2021-08-15
Install RFC6761 config file

DL6ER:

---

Implement what a caching DNS server should do following RFC 6761

Add additional config similar to what OpenWRT does as well

None

### DIFF
--- a/advanced/06-rfc6761.conf
+++ b/advanced/06-rfc6761.conf
@@ -1,0 +1,41 @@
+# Pi-hole: A black hole for Internet advertisements
+# (c) 2021 Pi-hole, LLC (https://pi-hole.net)
+# Network-wide ad blocking via your own hardware.
+#
+# RFC 6761 config file for Pi-hole
+#
+# This file is copyright under the latest version of the EUPL.
+# Please see LICENSE file for your rights under this license.
+
+###############################################################################
+#      FILE AUTOMATICALLY POPULATED BY PI-HOLE INSTALL/UPDATE PROCEDURE.      #
+# ANY CHANGES MADE TO THIS FILE AFTER INSTALL WILL BE LOST ON THE NEXT UPDATE #
+#                                                                             #
+#             CHANGES SHOULD BE MADE IN A SEPARATE CONFIG FILE                #
+#                    WITHIN /etc/dnsmasq.d/yourname.conf                      #
+###############################################################################
+
+# RFC 6761: Caching DNS servers SHOULD recognize
+#     test, localhost, invalid
+# names as special and SHOULD NOT attempt to look up NS records for them, or
+# otherwise query authoritative DNS servers in an attempt to resolve these
+# names.
+server=/test/
+server=/localhost/
+server=/invalid/
+
+# The same RFC requests something similar for
+#     16.172.in-addr.arpa.  22.172.in-addr.arpa.  27.172.in-addr.arpa.
+#     17.172.in-addr.arpa.  30.172.in-addr.arpa.  28.172.in-addr.arpa.
+#     18.172.in-addr.arpa.  23.172.in-addr.arpa.  29.172.in-addr.arpa.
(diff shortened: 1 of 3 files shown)


automated install/basic-install.sh
```diff
@@ -1284,7 +1286,7 @@ version_check_dnsmasq() {
     fi
     # Copy the new Pi-hole DNS config file into the dnsmasq.d directory
     install -D -m 644 -T "${dnsmasq_pihole_01_snippet}" "${dnsmasq_pihole_01_location}"
-    printf "%b  %b Copying 01-pihole.conf to /etc/dnsmasq.d/01-pihole.conf\\n" "${OVER}"  "${TICK}"
+    printf "%b  %b Copied 01-pihole.conf to /etc/dnsmasq.d/01-pihole.conf\\n" "${OVER}"  "${TICK}"
```
PromoFaux, line 1289: I think only on the slowest of all the slowest machines in the world will one ever see the word `Copying` in this case. But sure makes more sense gramatically.

advanced/06-rfc6761.conf
```diff
@@ -0,0 +1,41 @@
...
+#     19.172.in-addr.arpa.  24.172.in-addr.arpa.  31.172.in-addr.arpa.
+#     20.172.in-addr.arpa.  25.172.in-addr.arpa.  168.192.in-addr.arpa.
+# Pi-hole implements this via the dnsmasq option "bogus-priv" (see
+# 01-pihole.conf) because this also covers IPv6.
+
+# OpenWRT furthermore blocks    bind, local, onion    domains, this seems meaningful
+# see https://git.openwrt.org/?p=openwrt/openwrt.git;a=blob_plain;f=package/network/services/dnsmasq/files/rfc6761.conf;hb=HEAD
+# and https://www.iana.org/assignments/special-use-domain-names/special-use-domain-names.xhtml
+server=/bind/
+server=/local/
```
PromoFaux, line 40: I think we changed the default Pi-hole DHCP domain name from `local` to `lan` some time ago (as one should never use `.local` for their local domain (apparently!)) however...

As it is a custom field on the web interface, folk may have decided use `.local` for their local domain, or it just simply has been that way since they first installed Pi-hole (as that is what we used to ship).

This will break those people's networks, right? My understanding of `server=/local/` means it is telling `dnsmasq` to return `NXDOMAIN` to any lookups for `.local`

DL6ER: > My understanding of `server=/local/` means it is telling `dnsmasq` to return `NXDOMAIN` to any lookups for `.local`

No, it means: Never **forward** this query. When the domain is known locally (DHCP server or `/etc/hosts` or local DNS records), it'll be answered.

DL6ER: What could break is, however, the following situation: The router is the DHCP server with domain `.local` and the router is configured as the only upstream.

Then this would stop local queries, e.g. `smarttv.local` from working as expected - because it wouldn't be allowed to be forwarded. Reverse-lookups would still work because the query would be of `.arpa` type.

I guess OpenWRT simply doesn't care about this case because they will almost always be the DHCP server of the network.

I'll remove `local`.

automated install/basic-install.sh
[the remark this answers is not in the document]
DL6ER: Obviously there is no reason. I changed the output to use the variable instead.

advanced/06-rfc6761.conf
dschaper: .local is what mDNS/Bonjour uses.

PromoFaux (approved):

HVR88:
Great idea, now look at the 100's of opening posts and 1000's of replies trying to figure out why .local addresses don't work with PiHole.

HVR88:
Yes. Have you read any of the threads? Have you tried these TLDs in the local DNS records in pihole?

PR #4287, opened 2021-08-20
Conditional forwarding: Also forward unqualified host names

DL6ER:

---

Ensure conditional forwarding will forward unqualified host names if we have no local answer for them.

Add `server=//a.b.c.d` config option.

`man dnsmasq`:
> An empty domain specification, `//` has the special meaning of "unqualified names only" ie names without any dots in them.

Conditional forwarding can unleash its full power only when these two check boxes are unticked, but this should be fairly clear from the hint below them (highlighted in the screenshot):

Screenshot from 2021-08-20 15-14-23

None

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -54,7 +54,7 @@ add_setting() {
 }
 
 delete_setting() {
-    sed -i "/${1}/d" "${setupVars}"
+    sed -i "/^${1}/d" "${setupVars}"
 }
 
 change_setting() {
@@ -67,7 +67,7 @@ addFTLsetting() {
 }
 
 deleteFTLsetting() {
-    sed -i "/${1}/d" "${FTLconf}"
+    sed -i "/^${1}/d" "${FTLconf}"
 }
 
 changeFTLsetting() {
@@ -84,7 +84,7 @@ add_dnsmasq_setting() {
 }
 
 delete_dnsmasq_setting() {
-    sed -i "/${1}/d" "${dnsmasqconfig}"
+    sed -i "/^${1}/d" "${dnsmasqconfig}"
 }
 
 SetTemperatureUnit() {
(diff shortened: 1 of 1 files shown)


Bucking-Horn (commented):
Looking good as far as adding that option goes.
How is removal handled if *Conditional Forwardingh* get's unticked?

DL6ER:
Conditional forwarding adds three lines: One `rev-server` and two `server` lines. *Before* conditional forwarding lines are written to the config file, the two `server` lines are removed:

https://github.com/pi-hole/pi-hole/blob/4736e03108763cc2d5659f48d8a1e8a64d9b2608/advanced/Scripts/webpage.sh#L149

and then later added when conditional forwarding is enabled. If it is not enabled, they will not be added anew.

Removing the `rev-server` is currenty missing, I'll add this. This results in that it is not possible to disable conditional forwarding right now in Pi-hole.

DL6ER:
My last statement was actually incorrect. **Everything worked as expected (and intended)**.

This is because `delete_dnsmasq_setting "server" ` removed without *anchoring* so "delete `server`" also matched and removed `rev-server`.
I still added `delete_dnsmasq_setting "rev-server" ` now because I tweaked the `delete_dnsmasq_setting` to only delete exactly what it is called for. If you want, deleting the `rev-server` was just a side-effect of deleting `server`. However, we should try to minimize side-effects as good as we can.

dschaper:
Since this is a behavior change I think it would be better to have this selectable. 

Can we have the Conditional Forwarding section add a checkbox to let users allow or deny to forward non-FQDNs to the rev server?

Most OSes will automatically add the local domain to the query if the hostname query is not successful. Windows does this and Linux will add the search domains from `/etc/resolv.conf`.

Bucking-Horn:
I think that ***Never forward non-FQDNs*** already allows you to control that behaviour.
Setting that option nullifies `server=//ip`.

I'd also argue that most users would expect local names to be passed on for resolution if *Conditional Forwarding* is enabled (just as the Discourse OP triggering this PR did).

If you add another, similar option just for *Conditional Forwarding*, you'd always have to tick two boxes to make that work.

DL6ER:
I agree with @Bucking-Horn that this option is already there. This PR rather fixes that they are still not forwarded via conditional forwarding, even when the box is unticked.

(We should probably rename "conditional forwarding" to something better but this is a separate thing)

dschaper:
What happens when there are multiple `rev-server` lines? In the cases where people have network ranges that are not handled by `dnsmasq` (/22's)? 

That may be a separate concern to address but if this is to fix that non-fqdn's were not being forwarded when the box was unticked then this is a fix and not an enhancement?

DL6ER:
Yeah, definitely a fix.

On the other issue: Multiple `rev-server` lines will not work reliably if they give multiple upstreams (because how would we know whom to ask?).

dschaper (approved):

PR #4263, opened 2021-08-05
Update debug for new back end.

dschaper:
**By submitting this pull request, I confirm the following:** 
*please fill any appropriate checkboxes, e.g: [X]*

- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md), as well as this entire template.
- [ ] I have made only one major change in my proposed changes.
- [x] I have commented my proposed changes within the code.
- [x] I have tested my proposed changes, and have included unit tests where possible.
- [x] I am willing to help maintain this change if there are issues with it later.
- [x] I give this submission freely and claim no ownership.
- [x] It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
- [x] I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

---
**What does this PR aim to accomplish?:**

Debug can now use port 443 for uploads to back end. No longer need fallback to `nc` option if the TLS certs did not rotate correctly. No longer need protection from null or zero byte terminated token return strings. (And you need curl to run Gravity so we know it's there.)

Also clarify that logs are removed after 48 hours. 

**How does this PR accomplish the above?:**

Screenshot 2021-08-04 210048

**What documentation changes (if any) are needed to support this PR?:**
None

PR #4304, opened 2021-08-30
Fix pipe to pihole tricorder

DL6ER:

---

Fix pipe to `pihole tricorder`

Change the upload mechanism to using `curl` as done in https://github.com/pi-hole/pi-hole/pull/4263 for the debug script.

None


dschaper:
I think we should still have the warnings/failure output there just in case.

DL6ER:
I added a test similar to what we have in the debug script.

Instead of `--silent`, I added `--no-progress-bar` to get error output if something fails, e.g. when there is an issue with the connection:
Screenshot from 2021-08-30 20-21-54
or when the DNS server is down:
Screenshot from 2021-08-30 20-23-42

With `--silent` we'd always only see the last line, without knowing the reason for the failure:
Screenshot from 2021-08-30 20-21-54

DL6ER:
And, finally, an example for a successful upload:
Screenshot from 2021-08-30 20-26-51

dschaper:
Are those changes something to port to the debugger script?

dschaper (approved):

PR #4305, opened 2021-08-30
Use --no-progress-meter instead of --silent so error reasons are printed

DL6ER:

---

Keep and print possible errors when `curl` fails to upload to Tricorder.

Use `--no-progress-meter` instead of `--silent` in the `curl` call

None

### DIFF
--- a/advanced/Scripts/piholeDebug.sh
+++ b/advanced/Scripts/piholeDebug.sh
@@ -1334,7 +1334,7 @@ curl_to_tricorder() {
     # Users can submit their debug logs using curl (encrypted)
     log_write "    * Using ${COL_GREEN}curl${COL_NC} for transmission."
     # transmit he log via TLS and store the token returned in a variable
-    tricorder_token=$(curl --silent --upload-file ${PIHOLE_DEBUG_LOG} https://tricorder.pi-hole.net)
+    tricorder_token="$(curl --no-progress-meter --upload-file ${PIHOLE_DEBUG_LOG} https://tricorder.pi-hole.net)"
     if [ -z "${tricorder_token}" ]; then
         log_write "    * ${COL_GREEN}curl${COL_NC} failed, contact Pi-hole support for assistance."
     fi
--- a/advanced/Scripts/piholeDebug.sh
+++ b/advanced/Scripts/piholeDebug.sh
@@ -1333,7 +1333,7 @@ analyze_pihole_log() {
 curl_to_tricorder() {
     # Users can submit their debug logs using curl (encrypted)
     log_write "    * Using ${COL_GREEN}curl${COL_NC} for transmission."
-    # transmit he log via TLS and store the token returned in a variable
+    # transmit the log via TLS and store the token returned in a variable
     tricorder_token=$(curl --silent --fail --show-error --upload-file ${PIHOLE_DEBUG_LOG} https://tricorder.pi-hole.net 2>&1)
     if [[ "${tricorder_token}" != "https://tricorder.pi-hole.net/"* ]]; then
         log_write "    * ${COL_GREEN}curl${COL_NC} failed, contact Pi-hole support for assistance."
--- a/advanced/Scripts/piholeDebug.sh
+++ b/advanced/Scripts/piholeDebug.sh
@@ -1338,7 +1338,7 @@ curl_to_tricorder() {
     if [[ "${tricorder_token}" != "https://tricorder.pi-hole.net/"* ]]; then
         log_write "    * ${COL_GREEN}curl${COL_NC} failed, contact Pi-hole support for assistance."
         # Log curl error (if available)
-        if [ ! -z "${tricorder_token}" ]; then
+        if [ -n "${tricorder_token}" ]; then
             log_write "    * Error message: ${COL_RED}${tricorder_token}${COL_NC}\\n"
             tricorder_token=""
         fi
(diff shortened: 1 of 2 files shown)


advanced/Scripts/piholeDebug.sh
```diff
@@ -1334,7 +1334,7 @@ curl_to_tricorder() {
     # Users can submit their debug logs using curl (encrypted)
     log_write "    * Using ${COL_GREEN}curl${COL_NC} for transmission."
     # transmit he log via TLS and store the token returned in a variable
-    tricorder_token=$(curl --silent --upload-file ${PIHOLE_DEBUG_LOG} https://tricorder.pi-hole.net)
+    tricorder_token="$(curl --no-progress-meter --upload-file ${PIHOLE_DEBUG_LOG} https://tricorder.pi-hole.net)"
```
dschaper, line 1337: I don't think we need the quotes since it's not a variable that is being used.

dschaper (changes requested):

advanced/Scripts/piholeDebug.sh
MichaIng: In bash and dash at least (not sure if this is true for all bourne shells) in such cases, when the variable call or command substitution is assigned directly to a variable, quotes are never required. No globbing and no word splitting happens, also multiple spaces are preserved correctly.
Same in:
```sh
var=$value
var=$(command)
```
___________
If only errors are wanted, no warnings or other possible informational output, '-sS' used in the Pi-hole install command would be the way to go.

> -s, --silent
> Silent or quiet mode. Don't show progress meter or error messages. Makes Curl mute. It will still output the data you ask for, potentially even to the terminal/stdout unless you redirect it.
> Use -S, --show-error in addition to this option to disable progress meter but still show error messages.
> 
> -S, --show-error
> When used with -s, --silent, it makes curl show an error message if it fails.
> See also --no-progress-meter.
> 
> --no-progress-meter
> Option to switch off the progress meter output without muting or otherwise affecting warning and informational messages like -s, --silent does.

And I'm personally fans of an additional `-f`:

> -f, --fail
> (HTTP) Fail silently (no output at all) on server errors. This is mostly done to better enable scripts etc to better deal with failed attempts. In normal cases when an HTTP server fails to deliver a document, it returns an HTML document stating so (which often also describes why and more). This flag will prevent curl from outputting that and return error 22.
> This method is not fail-safe and there are occasions where non-successful response codes will slip through, especially when authentication is involved (response codes 401 and 407).

Probably doesn't matter here where you have full control of the webservers 4xx handling, but in general this prevents full HTML error pages from being passed to STDOUT and processed, and in combination `-sSf` instead a one-line error message with the HTTP response code is passed to STDERR.

DL6ER: Okay, I'm fine with both
```
dominik@ubuntu-server:~$ curl -sSf http://www.notexistant.pi-hole.com
curl: (6) Could not resolve host: www.notexistant.pi-hole.com

dominik@ubuntu-server:~$ curl --no-progress-meter http://www.notexistant.pi-hole.com
curl: (6) Could not resolve host: www.notexistant.pi-hole.com
```

dschaper: I prefer the second with the full command written out so you don't have to go to the man page to see what the shorthand means.

MichaIng: Just note that this means that there is possibly other stuff stored to `tricorder_token`, so the below checks `[ -z "${tricorder_token}" ]` and `[[ -n "${tricorder_token}" ]]` to detect whether the upload was successful or not may not be accurate anymore. But I have no example or way to actually trigger such an unexpected STDOUT with this particular upload, just know it from other cases:
```
# curl --no-progress-bar https://api.github.com/repos/none/none/releases/latest 2> /dev/null
{
  "message": "Not Found",
  "documentation_url": "https://docs.github.com/rest/reference/repos#get-the-latest-release"
}
# curl -s https://api.github.com/repos/none/none/releases/latest 2> /dev/null
{
  "message": "Not Found",
  "documentation_url": "https://docs.github.com/rest/reference/repos#get-the-latest-release"
}
# curl -sS https://api.github.com/repos/none/none/releases/latest 2> /dev/null
{
  "message": "Not Found",
  "documentation_url": "https://docs.github.com/rest/reference/repos#get-the-latest-release"
}
# curl -sSf https://api.github.com/repos/none/none/releases/latest 2> /dev/null
# curl -sSf https://api.github.com/repos/none/none/releases/latest
curl: (22) The requested URL returned error: 404
```
As can be seen, in the first four cases, `curl` returns the GitHub APIs 404 JSON to STDOUT, and does not even return an error exit code, as long as it successfully received a response from the webserver. Only the last combination with `--fail` makes it returning and error exit code, as well as an error message to STDERR. `--silent` and `--show-error` then assures that nothing else is printed to STDOUT besides the actual non-4xx document content.

DL6ER: My solution is to store everything in the variable (`2>&1`) and if the returned string does not start in `https://tricorder.pi-hole.net/`, then this is not a valid response -> error message. See more details below.

DL6ER:
Successful upload to tricorder:

Screenshot from 2021-09-07 11-20-16

Failed upload to tricorder (forced by messing up the URL in the script, real errors would show identically):

Screenshot from 2021-09-07 11-11-22


---


And the same changes applied to pipe to `pihole tricorder`:

Screenshot from 2021-09-07 11-16-14

and (failed)

Screenshot from 2021-09-07 11-15-43

dschaper (approved):

PR #4328, opened 2021-09-13
Set file permission for querie database in pihole-FTL.service

yubiuser:

---
Set file permission for querie database in pihole-FTL.service to 664 so that the pihole group has wirte permissions as well. This is necessary to give the web interface write permission to the `message` table.

Goes together with https://github.com/pi-hole/AdminLTE/pull/1886 and https://github.com/pi-hole/FTL/pull/1162


MichaIng:
1. Is `/etc/pihole/macvendor.db` actually assured to be present? Else we should move it from old `chmod` to new `chmod -f` so that startup does not fail when it's still missing.
2. Does the web interface also need to edit `/etc/pihole/gravity.db` in cases?

yubiuser:
> Is /etc/pihole/macvendor.db actually assured to be present?

It should. But I can easily add the `-f` to make it safe.

https://github.com/pi-hole/pi-hole/blob/2673c2c0720fb3871a4c3cae6a3f38f9ccede24e/automated%20install/basic-install.sh#L2228
____

> Does the web interface also need to edit /etc/pihole/gravity.db in cases?

Yes. But it is allowed to do so already - all the group management happens in `gravity.db`

MichaIng:
It make sense to have this separated:
- Use `chmod` without `-f` for files/dirs which must exist or are pre-created before, so that the FTL service fails as early as possible when something is not as expected.
- Use `chmod -f` only in cases where it is okay when files are missing as they are created by FTL.

But ah, actually if we need to change the mode to non-default (`644` => `664`), then we need to pre-create the file as well. Now, when the database is missing, FTL will create it, but with `644` as long as umask is not changed. Changing umask again implies unwanted write permissions for `www-data` to all other files/dirs FTL creates. Or does FTL "manually" change/set the mode when creating it? If not, I suggest to add `/etc/pihole/pihole-FTL.db` to the files which are touched, so that the mode change is assured to be applied. Then `-f` can be skipped.

In case of `gravity.db` it's strange, since I do not see where `664` mode is applied to it, despite the installer having a comment that `www-data` is added to `pihole` group to be able to write to that database directly 🤔.

In case of `macvendor.db` the question is whether FTL fails if it is not present, and whether there is any benefit of having this empty compared to not having it at all. If I'm not mistaken, it is downloaded on install/upgrade only, and not written to by FTL or web interface. If so, there is no point to pre-create and change mode of it, when it's missing and we may simply skip handling it at all?

yubiuser:
> n case of macvendor.db the question is whether FTL fails if it is not present

No it doesn't. I manually moved the file, restarted FTL and flushed the network table. It gets populated with the devices, only missing the vendor information. Not even a log not in `/var/log/pihole-FTL.log`

MichaIng:
Then I think it's fine like it is so the FTL service is as graceful as possible.

advanced/Templates/pihole-FTL.service
```diff
@@ -24,9 +24,13 @@ start() {
     touch /run/pihole-FTL.pid /run/pihole-FTL.port /var/log/pihole-FTL.log /var/log/pihole.log /etc/pihole/dhcp.leases
     # Ensure that permissions are set so that pihole-FTL can edit all necessary files
     chown pihole:pihole /run/pihole-FTL.pid /run/pihole-FTL.port /var/log/pihole-FTL.log /var/log/pihole.log /etc/pihole/dhcp.leases /run/pihole /etc/pihole
-    chmod 0644 /run/pihole-FTL.pid /run/pihole-FTL.port /var/log/pihole-FTL.log /var/log/pihole.log /etc/pihole/dhcp.leases /etc/pihole/macvendor.db
+    chmod 0644 /run/pihole-FTL.pid /run/pihole-FTL.port /var/log/pihole-FTL.log /var/log/pihole.log /etc/pihole/dhcp.leases
+    # Ensure that permissions are set so that pihole-FTL can edit the files. We ignore errors as the file may not (yet) exist
+    chmod -f 0644/etc/pihole/macvendor.db
```
MichaIng, line 29: Space missing:
```sh
chmod -f 0644 /etc/pihole/macvendor.db
```

yubiuser: Thanks. Fixed.

DL6ER (approved):

advanced/Templates/pihole-FTL.service
```diff
@@ -24,9 +24,13 @@ start() {
     touch /run/pihole-FTL.pid /run/pihole-FTL.port /var/log/pihole-FTL.log /var/log/pihole.log /etc/pihole/dhcp.leases
     # Ensure that permissions are set so that pihole-FTL can edit all necessary files
     chown pihole:pihole /run/pihole-FTL.pid /run/pihole-FTL.port /var/log/pihole-FTL.log /var/log/pihole.log /etc/pihole/dhcp.leases /run/pihole /etc/pihole
-    chmod 0644 /run/pihole-FTL.pid /run/pihole-FTL.port /var/log/pihole-FTL.log /var/log/pihole.log /etc/pihole/dhcp.leases /etc/pihole/macvendor.db
+    chmod 0644 /run/pihole-FTL.pid /run/pihole-FTL.port /var/log/pihole-FTL.log /var/log/pihole.log /etc/pihole/dhcp.leases
+    # Ensure that permissions are set so that pihole-FTL can edit the files. We ignore errors as the file may not (yet) exist
+    chmod -f 0644 /etc/pihole/macvendor.db
     # Chown database files to the user FTL runs as. We ignore errors as the files may not (yet) exist
     chown -f pihole:pihole /etc/pihole/pihole-FTL.db /etc/pihole/gravity.db /etc/pihole/macvendor.db
+    # Chown database file persmissions so that the pihole group (web interface) can edit the file. We ignore errors as the files may not (yet) exist
```
PromoFaux, line 32:
```suggestion
    # Chown database file permissions so that the pihole group (web interface) can edit the file. We ignore errors as the files may not (yet) exist
```

Minor :)

PromoFaux (changes requested):

DL6ER (approved):

DL6ER:
@PromoFaux You need to approve as well?

Screenshot from 2021-09-29 20-32-14

PromoFaux (approved):

PR #4327, opened 2021-09-13
Export DNS_FQDN_REQUIRED and DNS_BOGUS_PRIV to setupVars.conf during installation

yubiuser:
- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md), as well as this entire template.
- [x] I have made only one major change in my proposed changes.
- [x] I have commented my proposed changes within the code.
- [x] I have tested my proposed changes, and have included unit tests where possible.
- [x] I am willing to help maintain this change if there are issues with it later.
- [x] I give this submission freely and claim no ownership.
- [x] It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
- [x] I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

Please make sure you [Sign Off](https://github.com/pi-hole/pi-hole/wiki/How-to-signoff-your-commits.) all commits. Pi-hole enforces the [DCO](https://github.com/pi-hole/pi-hole/wiki/Contributing-to-the-project).

---
**What does this PR aim to accomplish?:**
On a fresh installation there is a mismatch between the options "Never formward non-FQDN" and "Never forward reverse lookups for private IP ranges" which are apparently enabled on the web interface but the resp. options (domain-needed and bogus-priv) are missing in the `01-pihole.conf` dnsmasq config file.

Although the options are set in the template of `01-pihole.conf` located in the local repo clone, they are going to be removed at the end of the installation script because it calls 
https://github.com/pi-hole/pi-hole/blob/2673c2c0720fb3871a4c3cae6a3f38f9ccede24e/advanced/Scripts/webpage.sh#L147

from webpage.sh. The options would have been re-added by the script but only if `DNS_FQDN_REQUIRED` and `DNS_BOGUS_PRIV` are true. But they are not, because they have not been exported to `setupVars.conf` in the install scripts

https://github.com/pi-hole/pi-hole/blob/2673c2c0720fb3871a4c3cae6a3f38f9ccede24e/automated%20install/basic-install.sh#L1806-L1817

___
For more information see here: https://github.com/pi-hole/AdminLTE/pull/1873#issuecomment-917707168
___

I see two solutions to this:
1) Add DNS_FQDN_REQUIRED and DNS_BOGUS_PRIV to `setupVars.conf`
2) Don't tick the boxes in the web interface by default and remove the options from `01-pihole.conf` template

[...]

PR #4354, opened 2021-10-02
Do not export `DNS_FQDN_REQUIRED` and `DNS_BOGUS_PRIV` unconditionally

yubiuser:

---
https://github.com/pi-hole/pi-hole/pull/4327 added `DNS_FQDN_REQUIRED` and `DNS_BOGUS_PRIV` to the final export variables. 
However, this export was static and unconditionally. On every update, repair or core branch switch this would have added `variable=true` to `setupVars.conf`. 
This PR does two things:
1) if the variables have never been set (first install), it sets them to `true` (same as the other PR)
2) if they have been set before, preserve them during export.

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1774,15 +1774,15 @@ finalExports() {
     echo "CACHE_SIZE=${CACHE_SIZE}"
     }>> "${setupVars}"
 
-    # if this runs the first time set the varibale to true otherwiese keep the old value
-    if ! grep -q '^DNS_FQDN_REQUIRED=' ${setupVars}; then
+    # if this runs the first time (variable not set yet) set the varibale to true otherwiese keep the old value
+    if [[ ! -v DNS_FQDN_REQUIRED ]]; then
       echo "DNS_FQDN_REQUIRED=true" >> "${setupVars}"
     else
       echo "DNS_FQDN_REQUIRED=${DNS_FQDN_REQUIRED}" >> "${setupVars}"
     fi
 
-    # if this runs the first time set the varibale to true otherwiese keep the old value
-    if ! grep -q '^DNS_BOGUS_PRIV=' ${setupVars}; then
+    # if this runs the first time (variable not set yet) set the varibale to true otherwiese keep the old value
+    if [[ ! -v DNS_BOGUS_PRIV ]]; then
       echo "DNS_BOGUS_PRIV=true" >> "${setupVars}"
     else
       echo "DNS_BOGUS_PRIV=${DNS_BOGUS_PRIV}" >> "${setupVars}"
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1772,14 +1772,9 @@ finalExports() {
     echo "INSTALL_WEB_INTERFACE=${INSTALL_WEB_INTERFACE}"
     echo "LIGHTTPD_ENABLED=${LIGHTTPD_ENABLED}"
     echo "CACHE_SIZE=${CACHE_SIZE}"
+    echo "DNS_FQDN_REQUIRED=${DNS_FQDN_REQUIRED:-true}"
+    echo "DNS_BOGUS_PRIV=${DNS_BOGUS_PRIV:-true}"
     }>> "${setupVars}"
-
-    # if this runs the first time (variable not set yet) set the varibale to true otherwiese keep the old value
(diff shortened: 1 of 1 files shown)


yubiuser:
Of course I can't grep for a line I have removed 3 steps above... fixed.

automated install/basic-install.sh
```diff
@@ -1772,9 +1772,22 @@ finalExports() {
     echo "INSTALL_WEB_INTERFACE=${INSTALL_WEB_INTERFACE}"
     echo "LIGHTTPD_ENABLED=${LIGHTTPD_ENABLED}"
     echo "CACHE_SIZE=${CACHE_SIZE}"
-    echo "DNS_FQDN_REQUIRED=true"
-    echo "DNS_BOGUS_PRIV=true"
     }>> "${setupVars}"
+
+    # if this runs the first time (variable not set yet) set the varibale to true otherwiese keep the old value
+    if [[ ! -v DNS_FQDN_REQUIRED ]]; then
```
DL6ER, line 1778: `-v` has been added in `bash 4.2`. Are we sure this is not another round of the gravity error we've just seen with the last release? (this was a bash v4.4 feature, though)

automated install/basic-install.sh
```diff
@@ -1772,9 +1772,22 @@ finalExports() {
...
+      echo "DNS_FQDN_REQUIRED=true" >> "${setupVars}"
+    else
+      echo "DNS_FQDN_REQUIRED=${DNS_FQDN_REQUIRED}" >> "${setupVars}"
+    fi
```
DL6ER, line 1782:
```suggestion
    echo "DNS_FQDN_REQUIRED=${DNS_FQDN_REQUIRED:-true}" >> "${setupVars}"
```

automated install/basic-install.sh
```diff
@@ -1772,9 +1772,22 @@ finalExports() {
...
+
+    # if this runs the first time (variable not set yet) set the varibale to true otherwiese keep the old value
+    if [[ ! -v DNS_BOGUS_PRIV ]]; then
+      echo "DNS_BOGUS_PRIV=true" >> "${setupVars}"
+    else
+      echo "DNS_BOGUS_PRIV=${DNS_BOGUS_PRIV}" >> "${setupVars}"
+    fi
+
```
DL6ER, line 1790:
```suggestion
    echo "DNS_BOGUS_PRIV=${DNS_BOGUS_PRIV:-true}" >> "${setupVars}"
```

automated install/basic-install.sh
DL6ER: From the `bash` documentation:
``` bash
${parameter:-word}
#    If parameter is unset or null, the expansion of word is substituted. 
#    Otherwise, the value of parameter is substituted.
```

Example (`a` is not set initially):
``` bash
$ echo "$a"

$ echo "${a:-default}"
default

$ a="abc"
$ echo "${a:-default}"
abc
```

automated install/basic-install.sh
dschaper: Bash 4.2 is safe to use, every supported distro is 4.3 or above. But the suggested bashisms below are good to use instead. I'm not a huge fan of bashisms but I guess we might as well use them while we depend on bash as the shell.

PromoFaux: Is `! -v` functionally equivalent to `-n` ? Might be moot, as a I see you've made a different suggestion all together below.

yubiuser: @PromoFaux 
` -v` checks for unset variable, `-n` for non-empty string.

automated install/basic-install.sh
```diff
@@ -1772,9 +1772,14 @@ finalExports() {
     echo "INSTALL_WEB_INTERFACE=${INSTALL_WEB_INTERFACE}"
     echo "LIGHTTPD_ENABLED=${LIGHTTPD_ENABLED}"
     echo "CACHE_SIZE=${CACHE_SIZE}"
-    echo "DNS_FQDN_REQUIRED=true"
-    echo "DNS_BOGUS_PRIV=true"
     }>> "${setupVars}"
```
PromoFaux, line 1775:
```suggestion
    echo "DNS_FQDN_REQUIRED=${DNS_FQDN_REQUIRED:-true}"
    echo "DNS_BOGUS_PRIV=${DNS_BOGUS_PRIV:-true}"
    }>> "${setupVars}"
```

These can just go back into here now, right?

PromoFaux (approved):

PR #4372, opened 2021-10-05
Download all adlists that are enabled during gravity run

yubiuser:

---
So far, gravity will only download adlists that are in the `vw_adlist` which means
1) enabled
**AND**
2) in the default group or in at least one active group

https://github.com/pi-hole/pi-hole/blob/dad6247cb0c90fe8cc1ad541e0ccf8f02e1dd2be/gravity.sh#L383-L385

https://github.com/pi-hole/pi-hole/blob/dad6247cb0c90fe8cc1ad541e0ccf8f02e1dd2be/advanced/Templates/gravity.db.sql#L146-L151

This can confuse users as the web interface lists them as "active" suggesting they will be downloaded and stored in the database, even if the are only in a group that is (at the moment) not active.

This has been discussed here:
https://discourse.pi-hole.net/t/group-blocking-later-fails-if-blocklists-are-updated-when-the-group-is-disabled/50189

Instead of using `vw_adlist`, gravity downloads all adlists that are enabled, regardless of their group assignment.

None.

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -379,8 +379,8 @@ gravity_DownloadBlocklists() {
 
   # Retrieve source URLs from gravity database
   # We source only enabled adlists, sqlite3 stores boolean values as 0 (false) or 1 (true)
-  mapfile -t sources <<< "$(sqlite3 "${gravityDBfile}" "SELECT address FROM vw_adlist;" 2> /dev/null)"
-  mapfile -t sourceIDs <<< "$(sqlite3 "${gravityDBfile}" "SELECT id FROM vw_adlist;" 2> /dev/null)"
+  mapfile -t sources <<< "$(sqlite3 "${gravityDBfile}" "SELECT address FROM adlist WHERE enabled = 1;" 2> /dev/null)"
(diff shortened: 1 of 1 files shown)


dschaper:
This is how the database was accessed when the scheme was first adopted. 

https://github.com/pi-hole/pi-hole/commit/465a39427b1a0f34b83ede17a9dbfb000a19a3bb#diff-1aa8de157651defa7dd9bb1eec010df250d3a336800e2bad8a4a3e297587747cR236

I'm not sure where this was changed to the setup we have now, I've gone through the git blame but I don't see the commit that modified it.

yubiuser:
It's been used in one other place which I just replaced. I also removed `vw_adlist` from the gravity template and created a migration step.
It's also used in one FTL test suit. PR XXX will remove it.

yubiuser:
It's not used in the web repo (@PromoFaux I checked twice this time)

yubiuser:
I like it a lot. Before starting to dissect the commits I started fresh here:
https://github.com/pi-hole/pi-hole/pull/4379

PR #4373, opened 2021-10-06
Remove .idea/

friederbluemle:
Remove and ignore the `.idea/` folder (configuration data for JetBrains IDEs).

None of these files are necessary in order to open them in IntelliJ-based IDEs. In fact, removing them even fixes a problem when opening it in the latest version, where files would be shown as not belonging to the project.


PromoFaux:
Presumably Idea respects the `.editorconfig` file?

I must admit, it's been a long time since I used Idea - I've largely switched to vscode these days - I'm not sure if anyone else on the team still uses it - we can find out: @pi-hole/core-approvers ?

PromoFaux (approved):
Good enough for me! I'll await a second approval on this though just in case someone suddenly wonders where there .idea folder has gone :D

friederbluemle:
> Just for clarity:
> 
> > IntelliJ supports the .editorconfig file as well
> 
> Do you want to say that mean it could read it or that it actually does so?

Yes, it reads it:

screenshot-2021-10-06_181729_721579412

PR #4421, opened 2021-11-03
Recommend apt instead of apt-get if updating the package cache failed

yubiuser:

---
Recommend `apt` instead of `apt-get` for updating the package cache on deb based distribution as it gives more user-friendly output.

Inspired by https://cubiclenate.com/2021/10/31/pi-hole-update-troubles/

Check for the presence of `apt`, if available use this.

Additional
We don't switch to `apt` for installing the packages as it does not provide all the necessary options.

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -266,13 +266,8 @@ if is_command apt-get ; then
     # Set some global variables here
     # We don't set them earlier since the installed package manager might be rpm, so these values would be different
     PKG_MANAGER="apt-get"
-    # A variable to store the command used to update the package cache. If apt is available use this as we have seen it
-    # gives more user-friendly (interactive) advice, else fall-back to apt-get
-    if is_command apt ; then
-        UPDATE_PKG_CACHE="apt update"
-    else
-        UPDATE_PKG_CACHE="${PKG_MANAGER} update"
-    fi
+    # A variable to store the command used to update the package cache
+    UPDATE_PKG_CACHE="${PKG_MANAGER} update"
     # The command we will use to actually install packages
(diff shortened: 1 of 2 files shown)


DL6ER (dismissed):

DL6ER:
You will need to mock `apt` in the Debian tests (see `test_update_package_cache_failure_no_errors`)

yubiuser:
Need to fix the test before merging.

zebulon501:
Is this change reasonable?
From [`man apt`](https://manpages.debian.org/buster/apt/apt.8#SCRIPT_USAGE_AND_DIFFERENCES_FROM_OTHER_APT_TOOLS), `apt` is for the user usage and `apt-get` for scripting, which is the case here.

yubiuser:
We only change it in one place and if it is available: updating the package cache. After the last Pi-hole update users sent numerous support requests because they had also upgrades their OS  (Buster -> Bullseye) and the installer would fail because users needed to acknowledged the repo suite change from stable to oldstable. Using apt they would simple be presented with a yes/no question. All the other places will still use `apt-get`.

dschaper:
I'd leave things as is for what we use in the scripts. (Leave it all as `apt-get`). The `apt` front end is not meant for scripting, it used to have a warning to only use in interactively.

yubiuser:
I changed the PR to keep `apt-get` but recommend `apt` in case it is available and `apt-get update` failed.

yubiuser:
```
rockpi@rockpi-4b:~$ pihole -up
  [✗] Update local cache of available packages
  Error: Unable to update package cache. Please try "sudo apt update"
  
  ```

DL6ER (approved):

PR #2347, opened 2018-08-11
Don't let one blocklist block an other blocklist

Gusher123:
If a blocklist is blocked by an other and 'pihole -g' is run, update the blocked blocklist by useing an alternate dns server (in this case 1.1.1.1) to acquire the IP adres and download the blocklist using the 'curl --resolve' (https://curl.haxx.se/docs/manpage.html#--resolve)

This is my first pull request. My apologies for any noob mistakes 

**By submitting this pull request, I confirm the following:** 
*please fill any appropriate checkboxes, e.g: [X]*

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md), as well as this entire template.
- [X] I have made only one major change in my proposed changes.
- [X] I have commented my proposed changes within the code.
- [X] I have tested my proposed changes, and have included unit tests where possible.
- [X] I am willing to help maintain this change if there are issues with it later.
- [X] I give this submission freely and claim no ownership.
- [X] It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
- [x] I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

Please make sure you [Sign Off](https://github.com/pi-hole/pi-hole/wiki/How-to-signoff-your-commits.) all commits. Pi-hole enforces the [DCO](https://github.com/pi-hole/pi-hole/wiki/Contributing-to-the-project).

---
**What does this PR aim to accomplish?:**
*A detailed description, screenshots (if necessary), as well as links to any relevant GitHub issues*
Prevent updates using 'pihole -g' from being blocked by Pi-hole

**How does this PR accomplish the above?:**
*A detailed description (such as a changelog) and screenshots (if necessary) of the implemented fix*
By using an other dns server to lookup the host and passing the ip, port and domain directly to curl using the --resolve option.

See https://discourse.pi-hole.net/t/pi-hole-v4-0-released-with-ftldns-improved-blocking-modes-regex-docker-and-more/11462/21

**What documentation changes (if any) are needed to support this PR?:**
*A detailed list of any necessary changes*
Don't know.

---
* You must follow the template instructions. Failure to do so will result in your pull request being closed.
* Please respect that Pi-hole is developed by volunteers, who can only reply in their spare time.

PR #4450, opened 2021-12-03
Unblock adlist domain during gravity run in NODATA mode

yubiuser:

---
https://github.com/pi-hole/pi-hole/pull/2347 added a way to circumvent blocking during a gravity download in case the adlist domain is on another adlist. However, blocking mode `NODATA` was added later (https://github.com/pi-hole/FTL/pull/433) and the circumvent mechanism was never extended for `NODATA`

Check for `NODATA` as `BLOCKINGMODE`, check if `dig domain` results in `NOERROR` but empty response. If so, use the available upstream DNS server

**Example**: block `raw.githubusercontent.com` and having `NODATA` as blocking mode

Before
Bildschirmfoto zu 2021-12-03 09-18-23

After
Bildschirmfoto zu 2021-12-03 09-20-56

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -597,7 +597,7 @@ gravity_DownloadBlocklistFromUrl() {
         blocked=true
       fi;;
     "NODATA")
-      if [[ $(dig "${domain}" | grep "NOERROR" -c) -ge 1 ]] && [[ -z $(dig +noall +answer "${domain}" |awk '{print $5}') ]]; then
+      if [[ $(dig "${domain}" | grep "NOERROR" -c) -ge 1 ]] && [[ -z $(dig +short "${domain}") ]]; then
          blocked=true
       fi;;
     "NULL"|*)
--- a/gravity.sh
+++ b/gravity.sh
@@ -596,6 +596,10 @@ gravity_DownloadBlocklistFromUrl() {
       if [[ $(dig "${domain}" | grep "NXDOMAIN" -c) -ge 1 ]]; then
         blocked=true
       fi;;
+    "NODATA")
+      if [[ $(dig "${domain}" | grep "NOERROR" -c) -ge 1 ]] && [[ -z $(dig +noall +answer "${domain}" |awk '{print $5}') ]]; then
+         blocked=true
+      fi;;
     "NULL"|*)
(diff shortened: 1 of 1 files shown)


gravity.sh
```diff
@@ -596,6 +596,10 @@ gravity_DownloadBlocklistFromUrl() {
       if [[ $(dig "${domain}" | grep "NXDOMAIN" -c) -ge 1 ]]; then
         blocked=true
       fi;;
+    "NODATA")
+      if [[ $(dig "${domain}" | grep "NOERROR" -c) -ge 1 ]] && [[ -z $(dig +noall +answer "${domain}" |awk '{print $5}') ]]; then
```
DL6ER, line 600: The second test checks for the absence of an `A` record by checking if there is an answer record and if it is empty. It is very unlikely that an answer record without a resource target shows up, do you have an example for this?

Otherwise, how about this more simple way without `awk` ?

```suggestion
      if [[ $(dig "${domain}" | grep "NOERROR" -c) -ge 1 ]] && [[ -z $(dig +short "${domain}") ]]; then
```

It should do the same thing.

DL6ER:
We should revise these checks in a future PR. They test for return codes and decide, e.g., that `NXDOMAIN` is a bad thing even when it may not even come from FTL (e.g. blocking mode is still `NULL`). However, it may actually be a typo in the address and the domain really doesn't exist. In this case, the users gets a hint that the domain is blocked (while it isn't) making the error analysis unnecessarily harder.

gravity.sh
yubiuser: You're right. My thoughts went around a few corners and made it to complicated.

DL6ER (approved):

yubiuser:
Our test is a bit crude: we do not actually test if the domain is blocked, but rather if the return code and answer fits to what we would expect. We would need some combination of `pihole -q` and `pihole-FTL regex-test` to pass the domain along. (`pihole-FTL regex-test` is not even sufficient enough to test all the regex FTL is able to understand - think of blocking based on query type, inverse regex,...). 
Another (not elegant) way would be to grep `pihole.log` for the domain and keywords (like `gravity blocked`, `regex blacklisted`)

PR #4475, opened 2021-12-19
Check for updates on master based on tags not commits

yubiuser:

---
Check for updates in `pihole -up` based on tags (not commits) when branch is `master`.

So far, we compared commit hashes to decide if an update is available, but later reset the local repo back to the latest tag if on master.
https://github.com/pi-hole/pi-hole/blob/a3cc5df317ffeec2b6bf78d37e075f33aeb0f79c/automated%20install/basic-install.sh#L519-L522

This led to an update loop if non-taged commits have been pushed to master.

If the local repo is on `master` we check for the latest tag to decide if an update is available.

P.S. `updatecheck.sh` already checks based on tags https://github.com/pi-hole/pi-hole/blob/a3cc5df317ffeec2b6bf78d37e075f33aeb0f79c/advanced/Scripts/updatecheck.sh#L52-L57

### DIFF
--- a/advanced/Scripts/update.sh
+++ b/advanced/Scripts/update.sh
@@ -49,7 +49,7 @@ GitCheckUpdateAvail() {
         # get the latest local tag
         LOCAL=$(git describe --abbrev=0 --tags)
         # get remote repo_name and URL
-        repo_name="$(basename -s .git $(git remote get-url origin))"
+        repo_name="$(basename -s .git "$(git remote get-url origin)")"
         repo_url="https://api.github.com/repos/pi-hole/${repo_name}/releases/latest"
         # get the latest tag from remote
         REMOTE=$(curl -s "${repo_url}"  2> /dev/null | grep tag_name | awk 'BEGIN { FS = "\"" } ; { print $4}')
--- a/advanced/Scripts/update.sh
+++ b/advanced/Scripts/update.sh
@@ -35,6 +35,7 @@ source "/opt/pihole/COL_TABLE"
 
 GitCheckUpdateAvail() {
     local directory
+    local repo_name repo_url
     directory="${1}"
     curdir=$PWD
     cd "${directory}" || return
@@ -42,18 +43,33 @@ GitCheckUpdateAvail() {
     # Fetch latest changes in this repo
     git fetch --quiet origin
 
-    # @ alone is a shortcut for HEAD. Older versions of git
-    # need @{0}
-    LOCAL="$(git rev-parse "@{0}")"
+    # Check current branch. If it is master, then check for the latest available tag instead of latest commit.
+    curBranch=$(git rev-parse --abbrev-ref HEAD)
+    if [[ "${curBranch}" == "master" ]]; then
+        # get the latest local tag
+        LOCAL=$(git describe --abbrev=0 --tags)
+        # get remote repo_name and URL
+        repo_name="$(git config --get remote.origin.url |  awk -F '[/.]' '{ print $6}')"
+        repo_url="https://api.github.com/repos/pi-hole/${repo_name}/releases/latest"
+        # get the latest tag from remote
+        REMOTE="$(curl -s ${repo_url}  2> /dev/null |grep tag_name | awk 'BEGIN { FS = "\"" } ; { print $4}')"
 
-    # The suffix @{upstream} to a branchname
-    # (short form <branchname>@{u}) refers
-    # to the branch that the branch specified
-    # by branchname is set to build on top of#
-    # (configured with branch.<name>.remote and
-    # branch.<name>.merge). A missing branchname
-    # defaults to the current one.
-    REMOTE="$(git rev-parse "@{upstream}")"
+
+    else
+        # @ alone is a shortcut for HEAD. Older versions of git
+        # need @{0}
+        LOCAL="$(git rev-parse "@{0}")"
+
+        # The suffix @{upstream} to a branchname
+        # (short form <branchname>@{u}) refers
+        # to the branch that the branch specified
+        # by branchname is set to build on top of#
+        # (configured with branch.<name>.remote and
+        # branch.<name>.merge). A missing branchname
+        # defaults to the current one.
+        REMOTE="$(git rev-parse "@{upstream}")"
+    fi
+
 
     if [[ "${#LOCAL}" == 0 ]]; then
         echo -e "\\n  ${COL_LIGHT_RED}Error: Local revision could not be obtained, please contact Pi-hole Support"
--- a/advanced/Scripts/update.sh
+++ b/advanced/Scripts/update.sh
@@ -49,11 +49,10 @@ GitCheckUpdateAvail() {
         # get the latest local tag
         LOCAL=$(git describe --abbrev=0 --tags)
         # get remote repo_name and URL
-        repo_name="$(git config --get remote.origin.url |  awk -F '[/.]' '{ print $6}')"
+        repo_name="$(basename -s .git $(git remote get-url origin))"
         repo_url="https://api.github.com/repos/pi-hole/${repo_name}/releases/latest"
         # get the latest tag from remote
-        REMOTE="$(curl -s "${repo_url}"  2> /dev/null |grep tag_name | awk 'BEGIN { FS = "\"" } ; { print $4}')"
-
+        REMOTE=$(curl -s "${repo_url}"  2> /dev/null | grep tag_name | awk 'BEGIN { FS = "\"" } ; { print $4}')
 
     else
         # @ alone is a shortcut for HEAD. Older versions of git
--- a/advanced/Scripts/update.sh
+++ b/advanced/Scripts/update.sh
@@ -52,7 +52,7 @@ GitCheckUpdateAvail() {
         repo_name="$(git config --get remote.origin.url |  awk -F '[/.]' '{ print $6}')"
         repo_url="https://api.github.com/repos/pi-hole/${repo_name}/releases/latest"
         # get the latest tag from remote
-        REMOTE="$(curl -s ${repo_url}  2> /dev/null |grep tag_name | awk 'BEGIN { FS = "\"" } ; { print $4}')"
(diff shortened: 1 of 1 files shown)


yubiuser:
Limiting this to `master` to allow to update individual feature branches by commits via `pihole -up`

advanced/Scripts/update.sh
```diff
@@ -35,25 +35,41 @@ source "/opt/pihole/COL_TABLE"
...
-    # branch.<name>.merge). A missing branchname
-    # defaults to the current one.
-    REMOTE="$(git rev-parse "@{upstream}")"
+    # Check current branch. If it is master, then check for the latest available tag instead of latest commit.
+    curBranch=$(git rev-parse --abbrev-ref HEAD)
+    if [[ "${curBranch}" == "master" ]]; then
+        # get the latest local tag
+        LOCAL=$(git describe --abbrev=0 --tags)
+        # get remote repo_name and URL
+        repo_name="$(git config --get remote.origin.url |  awk -F '[/.]' '{ print $6}')"
```
DL6ER, line 52: This does not work on my own Pi-hole because I manually switched from `https` to `git` protocol:
```
$ git config --get remote.origin.url
git@github.com:pi-hole/pi-hole.git
```

```suggestion
        repo_name="$(basename "$(git rev-parse --show-toplevel)")"
```

another way could be (closer to yours) where `.git` is stripped from the end of the URL (if present)
```suggestion
        repo_name="$(basename -s .git $(git remote get-url origin))"
```

advanced/Scripts/update.sh
```diff
@@ -35,25 +35,41 @@ source "/opt/pihole/COL_TABLE"
...
+        repo_url="https://api.github.com/repos/pi-hole/${repo_name}/releases/latest"
+        # get the latest tag from remote
+        REMOTE="$(curl -s "${repo_url}"  2> /dev/null |grep tag_name | awk 'BEGIN { FS = "\"" } ; { print $4}')"
+
+
```
DL6ER, line 57:
```suggestion
```

advanced/Scripts/update.sh
```diff
@@ -35,25 +35,41 @@ source "/opt/pihole/COL_TABLE"
...
+    # Check current branch. If it is master, then check for the latest available tag instead of latest commit.
+    curBranch=$(git rev-parse --abbrev-ref HEAD)
+    if [[ "${curBranch}" == "master" ]]; then
+        # get the latest local tag
+        LOCAL=$(git describe --abbrev=0 --tags)
+        # get remote repo_name and URL
+        repo_name="$(git config --get remote.origin.url |  awk -F '[/.]' '{ print $6}')"
+        repo_url="https://api.github.com/repos/pi-hole/${repo_name}/releases/latest"
+        # get the latest tag from remote
+        REMOTE="$(curl -s "${repo_url}"  2> /dev/null |grep tag_name | awk 'BEGIN { FS = "\"" } ; { print $4}')"
```
DL6ER, line 55:
```suggestion
        REMOTE="$(curl -s "${repo_url}"  2> /dev/null | grep tag_name | awk 'BEGIN { FS = "\"" } ; { print $4}')"
```

advanced/Scripts/update.sh
DL6ER: And, just because you have `var="$()"` here and without quotes above: It is generally not needed as word splitting is not performed right-hand side of an assignment (see https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Shell-Parameters) (Section 3.5.4).

DL6ER (changes requested):

DL6ER (dismissed):

DL6ER:
Nonetheless, I wonder if this can be resolved in an easier manner without fetching and parsing API output from GitHub. In the end, the API output could be something different (like rate-limiting happening to quite a few users every now and then), we'd not get a tag and this could generate an incorrect signal of an available update when there is none.

Looking at the various `man git` pages, it seems like
``` bash
git fetch --tags
upstream_tag=$(git describe --tags $(git rev-list --tags -1))
```
could get the latest upstream tag for us as well:

1. `git fetch --tags` will pull the tags from upstream
2. `git rev-list --tags` will give us the hashes of all annotated tags while `-1` will ensure we will actually get only one has here: the most recent one
3. `git describe --tags ...` will translate the hash into its pretty name (the tag name itself)

I'm not aware of a way to combine 2+3 into one `git` command but you may know how to do this.

What do you think?

DL6ER:
Another idea, even simpler:
``` bash
git fetch --tags
local_tag=$(git describe --abbrev=0 --tags master)
upstream_tag=$(git describe --abbrev=0 --tags origin/master)
```

dschaper:
> ```shell
> git fetch --tags
> local_tag=$(git describe --abbrev=0 --tags master)
> upstream_tag=$(git describe --abbrev=0 --tags origin/master)
> ```

That looks like it works for me.

yubiuser:
I did not include the `git fetch --tags` as we already fetch everything right above the master branch if clause.

https://github.com/pi-hole/pi-hole/blob/300c63127d092c1038ff600fac38cf3f659274d0/advanced/Scripts/update.sh#L44-L48

advanced/Scripts/update.sh
```diff
@@ -35,25 +35,37 @@ source "/opt/pihole/COL_TABLE"
 
 GitCheckUpdateAvail() {
     local directory
+    local curBranch
     directory="${1}"
     curdir=$PWD
     cd "${directory}" || return
 
     # Fetch latest changes in this repo
     git fetch --quiet origin
```
DL6ER, line 44:
```suggestion
    git fetch --tags --quiet origin
```
This one is lacking `--tags`

`man git fetch` says:
> **-t, --tags**
> **Fetch all tags from the remote** (i.e., fetch remote tags refs/tags/* into local tags with the same name), **in addition to whatever else would otherwise be fetched**. Using this option alone does not subject tags to pruning, even if --prune is used (though tags may be pruned anyway if they are also the destination of an explicit refspec; see --prune).

(emphasis added by me). It sounds like a good idea to have it in addition.

yubiuser: I did not include it because from https://git-scm.com/docs/git-fetch

> By default, any tag that points into the histories being fetched is also fetched; the effect is to fetch tags that point at branches that you are interested in. This default behavior can be changed by using the --tags or --no-tags options or by configuring remote.<name>.tagOpt. By using a refspec that fetches tags explicitly, you can fetch tags that do not point into branches you are interested in as well.

I concluded that any tags pointed to `master` are fetched. But I can include it just to be safe.

DL6ER (approved):

PromoFaux (approved):

ashishkotnala29:
This does not work on CentOS 7 which uses git 1.8.3.1. Always shows pi-hole core and web already up to date although FTL worked.

I verified by reverting this change and update worked again.

PromoFaux:
I've just played around with an install on a CentOS 7 droplet (confirmed git version 1.8.3.1, and i'm not able to reproduce this PR being the cause of update issues on CentOS. 

I also ran all of the commands individually and they all gave an output.

yubiuser:
@ashishkotnala29 

Please upload a debug log and post **just the  token URL** that is generated after the log is uploaded by running the following command from the Pi-hole host terminal:
```
pihole -d
```
or do it through the Web interface:

***Tools > Generate Debug Log***

ashishkotnala29:
@yubiuser https://tricorder.pi-hole.net/KhGLo5uu/

thanks!

yubiuser:
Thanks for letting us know. I'm not sure why it failed for you. Currently, we see the strangest things with this update on CentOS 7 (including the addition of those backticks in the upstream repo - which was not the issue in your case). Hope it is resolved now and we don't run into issues with the next update again.

ashishkotnala29:
@yubiuser Time will tell. Although I can consistently reproduce this issue if I have pihole 5.7 installed and try to update it. It doesn't update unless I revert the changes of this PR in the update.sh file.

ashishkotnala29:
@yubiuser An update from my side. Someone helped me by compiling Git 2.34 RPMs for armv7hf.
This issue is resolved after upgrading git to v2.34 (probably not the minimum needed).

image

image

image

There are definitely some git command changes in this PR that don't work with git 1.8.x

yubiuser:
OK, please post the output of each of the used commands from your stock CentOS 7 system. Perform these commands in `/etc/.pihole/`

```
git fetch --tags origin  
git rev-parse --abbrev-ref HEAD
git describe --abbrev=0 --tags master
git describe --abbrev=0 --tags origin/master
```

ashishkotnala29:
@yubiuser Here it is, this is done on v5.7 deliberately so that I don't get a false-positive scenario with an already up-to-date instance.

image

I'll upgrade git now and see if the output changes without updating pihole.

yubiuser:
Thanks. It seems like this is the place which fails

`git describe --abbrev=0 --tags origin/master`

ashishkotnala29:
@yubiuser Yes, exactly.

yubiuser:
Please try `git describe --tags $(git rev-list --tags -1)`

ashishkotnala29:
image
Seems to work @yubiuser

yubiuser:
I'll write a patch for this.

ashishkotnala29:
Thanks, much appreciated.

ashishkotnala29:
@yubiuser This is interesting...
image

`git fetch --tags origin` did something extra in git v2.34 and now `git describe --abbrev=0 --tags origin/master` works. So the culprit is not the suspected command but the fetch command with the `--tags` parameter.

yubiuser:
Thanks for investigating. Please repeat, omitting the `--tags`

```
git fetch origin  
git rev-parse --abbrev-ref HEAD
git describe --abbrev=0 --tags master
git describe --abbrev=0 --tags origin/master
```

ashishkotnala29:
Let me downgrade and try this on v1.8 again. I'm sure this will work as fetch will just fetch everything.

ashishkotnala29:
Thought to restore my 5.7 image snapshot entirely. Tried the commands again on that and here's the output.

image

Works as expected!

PR #4485, opened 2021-12-26
Return the port FTL is listening on in pihole status function

yubiuser:

---

Includes the port which FTL is running on in `pihole status` which is used in the web interface. 
It does this by using the FTL's telnet API endpoint `dns-port` introduced by https://github.com/pi-hole/FTL/pull/1264

As the API is used via netcat, this is (re-) added to the dependencies.

Corresponding `web` PR: https://github.com/pi-hole/AdminLTE/pull/2031

### DIFF
--- a/pihole
+++ b/pihole
@@ -285,7 +285,7 @@ Options:
 analyze_ports() {
   # FTL is listening at least on at least one port when this
   # function is getting called
-  echo -e "  ${TICK} DNS service is listening"
+  echo -e "  ${TICK} DNS service is listening on port 53"
   # Check individual address family/protocol combinations
   # For a healthy Pi-hole, they should all be up (nothing printed)
   if grep -q "IPv4.*UDP" <<< "${1}"; then
--- a/pihole
+++ b/pihole
@@ -285,7 +285,6 @@ Options:
 analyze_ports() {
   # FTL is listening at least on at least one port when this
   # function is getting called
-  echo -e "  ${TICK} DNS service is listening on port 53"
   # Check individual address family/protocol combinations
   # For a healthy Pi-hole, they should all be up (nothing printed)
   if grep -q "IPv4.*UDP" <<< "${1}"; then
--- a/pihole
+++ b/pihole
@@ -325,7 +325,7 @@ statusFunc() {
   else
     #get the port pihole-FTL is listening on by using FTL's telnet API
     port="$(echo ">dns-port >quit" | nc 127.0.0.1 4711)"
-    listening="$(lsof -Pni:53)"
+    listening="$(lsof -Pni:${port})"
     if [[ "${port}" == "0" ]]; then
       case "${1}" in
         "web") echo "-1";;
@@ -333,7 +333,8 @@ statusFunc() {
       esac
       return 0
     else
-      if [[ "${1}" != "web" ]] && [[ "$port" -eq 53 ]]; then
+      if [[ "${1}" != "web" ]]; then
+        echo -e "  ${TICK} FTL is listening on port ${port}"
         analyze_ports "${listening}"
       fi
     fi
@@ -350,8 +351,7 @@ statusFunc() {
     # Configs are set
     case "${1}" in
       "web") echo "$port";;
-      *) echo -e "  ${TICK} Pi-hole blocking is enabled. FTL is listening on port ${port}"
-      ;;
+      *) echo -e "  ${TICK} Pi-hole blocking is enabled";;
     esac
   else
     # No configs were found
--- a/pihole
+++ b/pihole
@@ -337,7 +337,7 @@ statusFunc() {
       esac
       return 0
     fi
-
+  fi
   # Determine if Pi-hole's blocking is enabled
   if grep -q "BLOCKING_ENABLED=false" /etc/pihole/setupVars.conf; then
     # A config is commented out
--- a/pihole
+++ b/pihole
@@ -323,43 +323,43 @@ statusFunc() {
     esac
     return 0
   else
-      #get the port pihole-FTL is listening on
-      port="$(lsof -Pni UDP -p ${pid} -a | grep -m1 : | awk -F ":" '{print $2}')"
-      listening="$(lsof -Pni:53)"
-      if [[ ! -z "$port" ]]; then
-        if [[ "${1}" != "web" ]] && [[ "$port" -eq 53 ]]; then
-          analyze_ports "${listening}"
-        fi
-      else
-        case "${1}" in
-          "web") echo "-1";;
-          *) echo -e "  ${CROSS} DNS service is NOT listening";;
-        esac
-        return 0
+    #get the port pihole-FTL is listening on
+    port="$(lsof -Pni UDP -p ${pid} -a | grep -m1 : | awk -F ":" '{print $2}')"
+    listening="$(lsof -Pni:53)"
+    if [[ ! -z "$port" ]]; then
+      if [[ "${1}" != "web" ]] && [[ "$port" -eq 53 ]]; then
+        analyze_ports "${listening}"
       fi
+    else
+      case "${1}" in
+        "web") echo "-1";;
+        *) echo -e "  ${CROSS} DNS service is NOT listening";;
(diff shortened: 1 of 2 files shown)


yubiuser:
No idea why it is failing the tests

```
[gw0] [ 50%] FAILED test_automated_install.py::test_installPihole_fresh_install_readableBlockpage[True] 

....

        echo "INSTALL_WEB_SERVER=${INSTALL_WEB_SERVER}"
        ''')
>       assert 0 == installWeb.rc
E       assert 0 == 1
E         +0
E         -1
```

https://github.com/pi-hole/pi-hole/blob/1b809e4e8ea442cd2a3cc54206429fabd58cce05/test/test_automated_install.py#L391

dschaper:
```plain
Check for existing repository in /var/www/html/admin\n  
[i] Clone https://github.com/pi-hole/AdminLTE.git into /var/www/html/admin...\n  \x1b[1;31mError: Could not update local repository. Contact support.\x1b[0m\n", 


stderr=b'debconf: delaying package configuration, since apt-utils is not installed\ndebconf: delaying package configuration, since apt-utils is not installed\n'
```

pihole
```diff
@@ -312,42 +312,55 @@ analyze_ports() {
...
+      return 0
   else
-    case "${1}" in
-      "web") echo "-1";;
-      *) echo -e "  ${CROSS} DNS service is NOT listening";;
-    esac
-    return 0
-  fi
+      #get the port pihole-FTL is listening on
+      port="$(lsof -Pni UDP -p ${pid} -a | grep -m1 : | awk -F ":" '{print $2}')"
```
DL6ER, line 327: This is not going to work reliably as you should not rely on port 53 being the first one reported by `lsof`.

Mind that FTL may bind to other port as well, such as 67 for DHCP or 547 for DHCPv6.
A quick check indeed suggests that DHCP ports will show up earlier (the order depends on the order `dnsmasq`/FTL bound to these ports) as both

https://github.com/pi-hole/FTL/blob/3b7bd2c5224c3ed13b983abcab26dd676cdc7cd0/src/dnsmasq/dnsmasq.c#L329

https://github.com/pi-hole/FTL/blob/3b7bd2c5224c3ed13b983abcab26dd676cdc7cd0/src/dnsmasq/dnsmasq.c#L348

are before

https://github.com/pi-hole/FTL/blob/3b7bd2c5224c3ed13b983abcab26dd676cdc7cd0/src/dnsmasq/dnsmasq.c#L422

I see three possibilities right now, alternatives are welcomed:

1. Simply ask FTL for its port over the Telnet API. This only needs a simple API extension in FTL.
2. Have FTL write the DNS port into some file which is then read by PHP. This is pretty indirect and the file could be outdated.
3. Check which port FTL bound to both on UDP and TCP. This *should* only happen for the DNS port. Even when this would work without an FTL change, it clearly looks like the most hacky solution among all three.

pihole
```diff
@@ -312,42 +312,55 @@ analyze_ports() {
...
+  # Determine if there is pihole-FTL service is listening on any UDP port
+  local listening pid port
+
+  pid="$(getFTLPID)"
+  if [[ "$pid" -eq "-1" ]]; then
+      case "${1}" in
+        "web") echo "-1";;
+        *) echo -e "  ${CROSS} DNS service is NOT running";;
+      esac
+      return 0
```
DL6ER, line 324:
```suggestion
  if [[ "$pid" -eq "-1" ]]; then
    case "${1}" in
      "web") echo "-1";;
      *) echo -e "  ${CROSS} DNS service is NOT running";;
    esac
    return 0
```

The indentation level is off.

pihole
```diff
@@ -312,42 +312,55 @@ analyze_ports() {
...
+        esac
+      else
+        # No configs were found
+        case "${1}" in
+          "web") echo -2;;
+          *) echo -e "  ${INFO} Pi-hole blocking will be enabled";;
+        esac
+        # Enable blocking
+        "${PI_HOLE_BIN_DIR}"/pihole enable
+      fi
```
DL6ER, line 362:
```suggestion
    #get the port pihole-FTL is listening on
    port="$(lsof -Pni UDP -p ${pid} -a | grep -m1 : | awk -F ":" '{print $2}')"
    listening="$(lsof -Pni:53)"
    if [[ ! -z "$port" ]]; then
      if [[ "${1}" != "web" ]] && [[ "$port" -eq 53 ]]; then
        analyze_ports "${listening}"
      fi
    else
      case "${1}" in
        "web") echo "-1";;
        *) echo -e "  ${CROSS} DNS service is NOT listening";;
      esac
      return 0
    fi

  # Determine if Pi-hole's blocking is enabled
  if grep -q "BLOCKING_ENABLED=false" /etc/pihole/setupVars.conf; then
    # A config is commented out
    case "${1}" in
      "web") echo 0;;
      *) echo -e "  ${CROSS} Pi-hole blocking is disabled";;
    esac
  elif grep -q "BLOCKING_ENABLED=true" /etc/pihole/setupVars.conf;  then
    # Configs are set
    case "${1}" in
      "web") echo "$port";;
      *) echo -e "  ${TICK} Pi-hole blocking is enabled";;
    esac
  else
    # No configs were found
    case "${1}" in
      "web") echo -2;;
      *) echo -e "  ${INFO} Pi-hole blocking will be enabled";;
    esac
    # Enable blocking
    "${PI_HOLE_BIN_DIR}"/pihole enable
  fi
```

pihole
```diff
@@ -312,42 +312,55 @@ analyze_ports() {
...
+          analyze_ports "${listening}"
+        fi
+      else
+        case "${1}" in
+          "web") echo "-1";;
+          *) echo -e "  ${CROSS} DNS service is NOT listening";;
+        esac
+        return 0
+      fi
+
```
DL6ER, line 340: Isn't there a `fi` missing here?

DL6ER (changes requested):

pihole
DL6ER: Concerning my suggestion no.1: https://github.com/pi-hole/FTL/pull/1264

yubiuser: Thanks for pointing out possible pitfalls. I think no. 1 is the most elegant one. I'll try to change the code accordingly.

pihole
yubiuser: Yes. Thanks.

DL6ER (approved):
Using `netcat` seems to be the easiest solution. Package `netcat` is 9.5 KB download and 18 KB installed (incl. `man` page). That's not something we should be concerned about, really.

DL6ER:
Merging this PR makes the core `development` branch depend on latest FTL `development`

moovy2:
> * [x]  I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md), as well as this entire template.
> * [x]  I have made only one major change in my proposed changes.
> * [x]  I have commented my proposed changes within the code.
> * [x]  I have tested my proposed changes, and have included unit tests where possible.
> * [x]  I am willing to help maintain this change if there are issues with it later.
> * [x]  I give this submission freely and claim no ownership.
> * [x]  It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
> * [x]  I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
> 
> Please make sure you [Sign Off](https://docs.pi-hole.net/guides/github/how-to-signoff/) all commits. Pi-hole enforces the [DCO](https://docs.pi-hole.net/guides/github/contributing/).
> 
> **What does this PR aim to accomplish?:**
> 
> Includes the port which FTL is running on in `pihole status` which is used in the web interface. It does this by using the FTL's telnet API endpoint `dns-port` introduced by [pi-hole/FTL#1264](https://github.com/pi-hole/FTL/pull/1264)
> 
> As the API is used via netcat, this is (re-) added to the dependencies.
> 
> Corresponding `web` PR: [pi-hole/AdminLTE#2031](https://github.com/pi-hole/AdminLTE/pull/2031)

Issue #3429, opened 2020-05-22
4 DNS Upstreams + router in dashboard  ... but only 2 defined in pihole ? (Conditional Forwarding)


roe1974:
Hi

i just activated "Conditional Forwarding" in pihole DNS settings with IP from my router + domain name (x.home.arpa). Now i see in the dashboard 4 upstream DNS (1.1.1.1, 1.0.0.1, 1.1.1.2, 1.0.02,my router) ... ist this normal ?
<img width="381" alt="1" src="https://user-images.githubusercontent.com/30420568/82667912-0b9b0880-9c39-11ea-906c-50710d5bed69.png">
<img width="757" alt="2" src="https://user-images.githubusercontent.com/30420568/82667921-0f2e8f80-9c39-11ea-855e-fa2d8fc479c0.png">

br, richard

roe1974:
home.arpa is reserven for local domains at home see RFC8375
https://tools.ietf.org/html/rfc8375
richard

dschaper:
Interesting, I don't do much with HNCP nets.

Closing as answered.

PR #4500, opened 2022-01-01
Update default DHCP domain name to home.arpa

thiagowfx:
from lan. RFC 8375[1] suggests it as a sensible default.

[1]: https://datatracker.ietf.org/doc/html/rfc8375

Other supporting sources in order of relevance:

- pfSense also followed suit c.f. https://redmine.pfsense.org/issues/10533
  a year ago.
- https://superuser.com/q/1566825/165044 ("Best choice for
private/home network TLD? Overlap with mDNS?")
- https://github.com/pi-hole/pi-hole/issues/3429#issuecomment-632984174

Related but not worthy of mentioning in the commit message:

- https://datatracker.ietf.org/doc/html/draft-chapin-rfc2606bis-00 seems to suggest .lan as problematic.
- https://www.ctrl.blog/entry/homenet-domain-name.html: This guy does a great job at explaining the rationale.
- https://en.wikipedia.org/wiki/.arpa

- [N/A] I have commented my proposed changes within the code.

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -387,7 +387,7 @@ ProcessDHCPSettings() {
         fi
 
         if [[ "${PIHOLE_DOMAIN}" == "" ]]; then
-            PIHOLE_DOMAIN="lan"
+            PIHOLE_DOMAIN="home.arpa"
             change_setting "PIHOLE_DOMAIN" "${PIHOLE_DOMAIN}"
         fi
 


thiagowfx:
Oh I am sorry I accidentally cut some characters in my first reference. It's meant to be https://redmine.pfsense.org/issues/10533, not https://redmine.pfsense.org/issues/10. Will reupload.

dschaper:
Thank you for the submission.

I doubt that `.lan` will ever become a TLD since it's used in so many consumer routers.

I read the draft RFC and I don't agree that it suggests .lan as problematic. I think it spells out .lan as being a candidate for IETF reserved status so that it can't be used as a valid TLD.

I think it will lead to confusion by users since they will see the .arpa top level and conflate that with in-addr apra reverse queries.

DL6ER:
@thiagowfx  Thank you for your contribution and your thoughts. I do appreciate it, however, I also fully agree on what @dschaper said. Pi-hole is often the first networking project (let alone Linux) users ever set up in the life. We should make it as easy as possible for them. I do see the possible implications of suddenly having hostnames as `mylaptop.home.arpa` instead of the more self-explanatory `mylaptop.lan`. Mind that there are some largely distributed operating systems that will always try to append the DHCP domain to their hostnames...

So I propose we close this PR but not without seeing what you say to this.

I'm surely looking forward to other contributions from you!

thiagowfx:
@dschaper and @DL6ER: Thank you for your comments and rationale, I understand.

I don't have too much to counter on this, there's nothing wrong with `.lan` per se, `.home.arpa` is meant to be an improvement due to that RFC, but since it's not enforced, it's just a recommendation, keeping `.lan` is perfectly fine as well.

I will just add one piece of anecdote: Whenever I am setting up a local network, I keep forgetting what the default domain suffix is. I know it's one of `local` or `localdomain` or `lan`, but I never remember which one is right, the fact they all start with an `l` doesn't help much. In this regard, `home.arpa` is (IMHO) way more memorable, after you see it for the first time it's very hard to forget it, and it's unambiguous.

This was one of the reasons why I wanted to change the default.

That said, I totally understand if you prefer to keep `.lan` in order to make it easier for beginners to remember the prefix. Hopefully people that never saw `.local` and `.localdomain` do not get confused as much as I do :)

thiagowfx:
Closing this for bookkeeping. If you ever change your mind, feel free to reopen and merge. :)

PR #4543, opened 2022-01-09
Include ip addr show and ip route show in debug log

DL6ER:

---

Include `ip addr show` and `ip route show` in the network section of the debug log to help with local-service issues (where hops-away is measured). It does not make sense to not have this information there as the entire purpose of the debug log is to get all relevant information in one place (whether or not users decide to upload).

Add `ip addr show` and `ip route show` without any further processing.

None

### DIFF
--- a/advanced/Scripts/piholeDebug.sh
+++ b/advanced/Scripts/piholeDebug.sh
@@ -779,6 +779,21 @@ check_required_ports() {
     done
 }
 
+ip_command() {
+    # Obtain and log information from "ip XYZ show" commands
+    echo_current_diagnostic "${2}"
+    local entries=()
+    mapfile -t entries < <(ip "${1}" show)
+    for line in "${entries[@]}"; do
+        log_write "   ${line}"
+    done
+}
+
+check_ip_command() {
+    ip_command "addr" "Network interfaces and addresses"
+    ip_command "route" "Network routing table"
+}
+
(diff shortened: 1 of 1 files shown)


DL6ER:
Screenshot from 2022-01-09 11-32-13

advanced/Scripts/piholeDebug.sh
```diff
@@ -779,6 +779,26 @@ check_required_ports() {
     done
 }
 
+ip_command() {
+    # Obtain and log information from "ip XYZ show" commands
+    echo_current_diagnostic "${2}"
+    OLD_IFS="$IFS"
+    IFS=$'\n'
```
dschaper, line 786: You don't really need to save IFS and restore it. Changing it inside a function just affects the time inside the function.

dschaper (changes requested):

advanced/Scripts/piholeDebug.sh
DL6ER: Thanks, I just copy-pasted from another function (DHCP discovery), I will just make `IFS` local :+1:

DL6ER: Actually, `IFS` is already `\n` by default. I don't think we really have to set it explicitly.

dschaper (approved):

PR #4547, opened 2022-01-10
Pi-hole Core v5.9

PromoFaux:


pihole
```diff
@@ -223,8 +223,7 @@ Time:
       fi
 
       local str="Pi-hole Disabled"
-      sed -i "/BLOCKING_ENABLED=/d" "${setupVars}"
-      echo "BLOCKING_ENABLED=false" >> "${setupVars}"
+      sed -i "s/^BLOCKING_ENABLED=true/BLOCKING_ENABLED=false/" "${setupVars}"
```
MichaIng, line 226: I'd revert this to keep handling a missing `BLOCKING_ENABLED` line gracefully. Shouldn't be the case, but there are/will be always such cases.

PromoFaux: Ah, Yes. (Sorry didn't see this comment before)

Fresh install from dev branches is currently broken because `BLOCKING_ENABLED` is missing from `setupVars.conf`... interestingly, switching to master and then back to dev fixes it, but I have not yet worked out why!!

image

MichaIng: `BLOCKING_ENABLED` is not added directly by the installer on any branch, FTL itself however treats it as enabled when unset: https://github.com/pi-hole/FTL/blob/7459d82/src/setupVars.c#L236

Not sure how AdminLTE checks the status, but if is used FTL, it should be actually able to derive the effective status correctly. Only disabling via `pihole` script would not work until the setting is present.

Not sure whether it is added indirectly at some other point on `master` branch via installer/updater?

PromoFaux: >Not sure whether it is added indirectly at some other point on master branch via installer/updater?

I guess before it was added on the first run of `pihole enable` or even `pihole disable`, because it was just doing a brute force "delete if exists and then append new value". See https://github.com/pi-hole/pi-hole/pull/4585 for a solution that I vaguely understand

MichaIng: Ah lol, I mean if it works reliably, good. The previous solution how did as well and at least everyone should understand and additionally assured that always a valid value is present. When preserving any existing value, in theory it could be an invalid one, possible also due to filesystem corruption and such, so had some benefits.

DL6ER (approved):

yubiuser (approved):

PR #4575, opened 2022-01-22
Fix updating based on tags on older git versions by doing a full fetch

yubiuser:

---
PR #4475 introduced updating based on tags instead of commits for master branch. This works well, except for older versions of git which are shipped by default on CentOS 7 (currently supported OS). It turned out, that fetching using `--tags` was the issue here (https://github.com/pi-hole/pi-hole/pull/4475#issuecomment-1019357036)

This PR does the fetching without `--tags` (which is what we used before #4475). It has been confirmed working on CentOS 7 (https://github.com/pi-hole/pi-hole/pull/4475#issuecomment-1019358878).

### DIFF
--- a/advanced/Scripts/update.sh
+++ b/advanced/Scripts/update.sh
@@ -41,7 +41,7 @@ GitCheckUpdateAvail() {
     cd "${directory}" || return
 
     # Fetch latest changes in this repo
-    git fetch --tags --quiet origin
+    git fetch --quiet origin
 
     # Check current branch. If it is master, then check for the latest available tag instead of latest commit.
     curBranch=$(git rev-parse --abbrev-ref HEAD)


DL6ER:
We added `--tags` because of what the `man git fetch` page said for this option (emphasis added by me):
> **-t, --tags**
>     Fetch all tags from the remote (i.e., fetch remote tags `refs/tags/*` into local tags with the same name), **in addition to whatever else would otherwise be fetched.**

This makes me wondering if fetching *less* can actually be the solution.

ashishkotnala29:
@DL6ER git fetch --tags pulls all the tags from the history of the repository meanwhile a normal fetch (without the parameter) fetches only new tags from the point the local repo's HEAD.

Reading the description from the man page and from what I know it does make sense to use `--tags` BUT the same does not work in exactly the same manner with git v1.8.3.1 which is bundled with CentOS 7 compared to modern versions (git 2.x).

More specifically git 1.8.x does not update the remote commit history when `fetch` is used with `--tags` as observed here: https://github.com/pi-hole/pi-hole/pull/4475#issuecomment-1019358878 
Hence running `git describe --abbrev=0 --tags origin/master` would still output an old version as observed here: https://github.com/pi-hole/pi-hole/pull/4475#issuecomment-1019357036

This could have been a bug in git 1.8.x or an improvement done in git 2.x. Either ways to use `git describe --abbrev=0 --tags origin/master` we need latest commit history in the local git repo's metadata (not really in the working copy). This activity is done by git 2.x but doesn't seem like it's done by git 1.8.x.

This is my understanding but I may very well be slightly incorrect.

rdwebdesign:
I'm not sure if this helps, but this link has information comparing `git fetch --tags` in different Git versions:

https://stackoverflow.com/questions/1204190/does-git-fetch-tags-include-git-fetch#20608181

ashishkotnala29:
@rdwebdesign right, thanks for pointing out these. So as said in the second one for git v1.8...
`git fetch --tags` fetches all tags, all commits necessary for them. **It will not update branch heads**, even if they are reachable from the tags which were fetched.
And the branch head should be updated for `git describe --abbrev=0 --tags origin/master` to give expected output for the update to be done.
So yeah doing a normal `git fetch` seems right.

DL6ER (approved):

PR #4582, opened 2022-01-28
add parameter to set filename for teleporter

lschloetterer:

---

Make it possible to write pihole -a -t myname.tar.gz to configure the filename however you want.
Using only pihole -a -t still gives you the same auto-generated filename as before.

Check if parameter is given and use it for the redirect.

Add second option for the teleporter command with a filename parameter

---

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -37,15 +37,16 @@ Example: pihole -a -p password
 Set options for the Admin Console
 
 Options:
-  -p, password        Set Admin Console password
-  -c, celsius         Set Celsius as preferred temperature unit
-  -f, fahrenheit      Set Fahrenheit as preferred temperature unit
-  -k, kelvin          Set Kelvin as preferred temperature unit
-  -e, email           Set an administrative contact address for the Block Page
-  -h, --help          Show this help dialog
-  -i, interface       Specify dnsmasq's interface listening behavior
-  -l, privacylevel    Set privacy level (0 = lowest, 3 = highest)
-  -t, teleporter      Backup configuration as an archive. Optionally specify a custom filename"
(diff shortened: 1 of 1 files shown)


DL6ER:
Please add a comment about the possibility to specify a custom filename to the `helpFunc` in the same file.

lschloetterer:
Like this? You could probably use something like `teleporter [filename]` too, but it's not like this for the other commands.

PromoFaux:
> Like this?

This is fine for me, maybe we could go as far as adding something like `e.g pihole teleporter myfilename.tar.gz`

Another thought, maybe we should make the `tar.gz` just happen by default, so the usage would be `pihole teleporter myfile`

rdwebdesign:
Every test will fail to CentOS 8 (EOL).

lschloetterer:
> maybe we could go as far as adding something like e.g pihole teleporter myfilename.tar.gz

Done.

> maybe we should make the tar.gz just happen by default

I'm not too much a fan of this idea, as it would break for example `pihole -a -t $(mktemp)`. Also when using tar it doesn't add it automatically either.

> Every test will fail for CentOS 8 (EOL).

I'm not sure why this is, the only thing I changed after the first commit which passed was text

yubiuser:
> I'm not sure why this is, the only thing I changed after the first commit which passed was text

This is kind of an issue on our side. CentOS 8 is end of file and the test container we used was unable to install software from an now old centos repo. This was fixed now as we switched to CentOS Steam (https://github.com/pi-hole/pi-hole/pull/4596)

lschloetterer:
> This was fixed now as we switched to CentOS Steam (https://github.com/pi-hole/pi-hole/pull/4596)

Ah I see, the checks are now passing again 👍

PromoFaux (approved):

Issue #4580, opened 2022-01-27
[CentOS7] [WEB-UI] Health status of Adlists is not updating


ashishkotnala29:
### Versions

- Pi-hole:  5.8.1
- AdminLTE:  5.10.1
- FTL:  5.13

### Platform

- OS and version:  CentOS 7 (7.9.2009)

- Platform:  Linux Deploy container (on Android) and VM (x86_64)

### Expected behavior

The status of the Adlists should be updated after running `pihole -g` for the first time or subsequent times. A green checkmark should be shown if the list was updated a different icon if the list was not updated but downloaded.

### Actual behavior / bug

A question mark icon is shown and status is "**Health status of this list:	List was not downloaded so far**"
This does not affect functionality however. Just seems to be a visual bug.

![image](https://user-images.githubusercontent.com/18037179/151407753-96c699c9-4f7d-4929-9db4-599d583b156f.png)

### Steps to reproduce

1) Have a CentOS 7 VM/docker container and pi-hole on it.
2) `pihole -g`
3) Web UI -> Group Management -> Adlists

You would see no status/checkmarks beside the list.

## Debug Token

- URL: 
(Linux deploy container) https://tricorder.pi-hole.net/I5623RBr/
(VM x86_64) https://tricorder.pi-hole.net/IQvaadx4/

DesktopECHO:
FYI, not exactly a fix but a pretty good indicator of what the problem is...  
Updating sqlite3 and reinstalling Pi-hole seems to have fixed the issue.   My RPM's are [here](https://github.com/DesktopECHO/CentOS7-ARMv7)

![image](https://user-images.githubusercontent.com/33142753/151433720-0eb346c3-561f-4188-b296-8825d8370129.png)

ashishkotnala29:
Confirmed, fixed by updating sqlite3. @DesktopECHO Many thanks again!

ashishkotnala29:
On a second thought seems like this affects Ubuntu 16.04 as well https://www.reddit.com/r/pihole/comments/sdxq2k/incorrect_health_status_of_adlist
And since CentOS7 and Ubuntu 16.04 both are supported OS it's better we make changes to support older sqlite3 versions. Reopening the issue.

DL6ER:
> better we make changes to support older sqlite3 versions. Reopening the issue.

Do you have any SQLite3 error messages for us so we can see what would need to be changed and if that's possible at all. Which version was installed before your forced upgrade?

[...]

PR #4584, opened 2022-01-30
Replace calls to sqlite3 by calls to pihole-FTL sqlite3

DL6ER:

---

Solve issues with ancient system-provided SQLite3 version as this is limiting us and is even causing bugs from time to time (see #4580).

Replace calls to `sqlite3` by calls to `pihole-FTL sqlite3`.

We also remove `sqlite3` as explicit dependency from the installer. Users can always use `pihole-FTL sqlite3` (`pihole-FTL sql` for short) to get the most recent SQLite3 engine embedded into FTL. In current development, this has all the features user expect, including auto-completion and history. We also add convenience features such as `pihole-FTL sql -h ...` for more human-friendly table printing that is not present in the official `sqlite3` package.

I had to rewrite some parts in gravity as the logic was incorrect. Using our embedded SQLite3 engine, this was revealed as it was complaining about a lot of foreign key violations because our embedded version is sufficiently new (compared to the system-provided ones) that foreign key enforcement is a thing.

The issue is that we are creating the database and inject `gravity` domains as the first thing. However, each gravity domain refers to an `adlist` entry and, since this table was still empty, each insertion triggered a violation. This is now fixed by ensuring we make the entire database ready before inserting the gravity domains in a final step.

Three new `DELETE FROM ... WHERE NOT IN (...)` statements in the gravity copying script removes orphaned entries which could be in the database when users interacted with the database manually. We remove these orphans as their presence would also trigger foreign key violations when the domains/adlists/clients they refer to are gone.

None

### DIFF
--- a/advanced/Scripts/database_migration/gravity-db.sh
+++ b/advanced/Scripts/database_migration/gravity-db.sh
@@ -19,13 +19,13 @@ upgrade_gravityDB(){
 	auditFile="${piholeDir}/auditlog.list"
 
 	# Get database version
-	version="$(sqlite3 "${database}" "SELECT \"value\" FROM \"info\" WHERE \"property\" = 'version';")"
+	version="$(pihole-FTL sqlite3 "${database}" "SELECT \"value\" FROM \"info\" WHERE \"property\" = 'version';")"
 
 	if [[ "$version" == "1" ]]; then
 		# This migration script upgrades the gravity.db file by
 		# adding the domain_audit table
 		echo -e "  ${INFO} Upgrading gravity database from version 1 to 2"
-		sqlite3 "${database}" < "${scriptPath}/1_to_2.sql"
+		pihole-FTL sqlite3 "${database}" < "${scriptPath}/1_to_2.sql"
 		version=2
 
 		# Store audit domains in database table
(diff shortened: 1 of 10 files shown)


DL6ER:
I should have caught all of them. The only remainders of something not starting in `pihole-FTL sqlite3` seems to be comments and PHP:
Screenshot from 2022-01-30 11-19-24

DL6ER:
Confirmed working by OP in https://github.com/pi-hole/pi-hole/issues/4580#issuecomment-1025108547

PromoFaux (dismissed):
Also confirmed working by me, I just got distracted by another bug!

yubiuser (changes requested):
https://github.com/pi-hole/pi-hole/blob/e80a7731c98d3ee1c03e0fd62621b49a4c621b43/automated%20install/basic-install.sh#L335

Remove it from RPM distros as well

yubiuser (approved):

yubiuser:
Build on CentOS 8 failed due to EOL.
https://stackoverflow.com/questions/70926799/centos-through-vm-no-urls-in-mirrorlist

Adding 

```
RUN sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-Linux-* &&\
    sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-Linux-*
```

Could fix the issue (see answer 2 from above). Or we switch to CentOS Steam

DL6ER:
@dschaper How do we want to proceed here? First remove or fix CentOS 8 and merge the modified test instructions into this PR or merge despite the failed tests?

dschaper:
Merge this with admin override. Then remove the CentOS 8 tests?

PR #4586, opened 2022-01-30
POC for internal discussion

PromoFaux:
This PR is to give an idea of what it might look like to have a file that contains self contained functions, that could not only be easily sourced by other scripts, but also easy to add tests to these functions. To further demonstrate this example, I have added a test.


PromoFaux:
(ignore stickler for the time being)

yubiuser:
I like the idea to where this is going. A lot can be re-used in https://github.com/pi-hole/pi-hole/blob/master/advanced/Scripts/webpage.sh as well I think

PromoFaux:
>https://github.com/pi-hole/pi-hole/blob/master/advanced/Scripts/webpage.sh as well I think

Agree that

PromoFaux:
Commits added to #4585

PR #4597, opened 2022-02-04
Validating domains including uppercase

rdwebdesign:
**By submitting this pull request, I confirm the following:** 

---
**What does this PR aim to accomplish?:**
 Fix last PR (4578)

**How does this PR accomplish the above?:**

**What documentation changes (if any) are needed to support this PR?:**
none

PR #4598, opened 2022-02-04
Use case insensitive deletion when removing custom CNAME/DNS records

PromoFaux:
Presented as an alternative to #4597


yubiuser (approved):

PromoFaux:
Fixes https://github.com/pi-hole/AdminLTE/issues/2094

PR #4653, opened 2022-03-16
Tweaks to functions in utils.sh and refactored usages in webpage.sh

PromoFaux:

---

Move some functionality out of webpage.sh and into utils.sh - keep the wrapper functions in webpage.sh for now as docker is currently using them - but these can go once we update docker container

### DIFF
--- a/advanced/Scripts/utils.sh
+++ b/advanced/Scripts/utils.sh
@@ -56,11 +56,11 @@ addOrEditKeyValPair() {
 # Deletes a key from target file
 #
 # Example usage:
-# removeKey "PIHOLE_DNS_1" "/etc/pihole/setupVars.conf"
+# removeKey "/etc/pihole/setupVars.conf" "PIHOLE_DNS_1"
 #######################
 removeKey() {
-  local key="${1}"
-  local file="${2}"
+  local file="${1}"
+  local key="${2}"
   sed -i "/^${key}/d" "${file}"
 }
 
--- a/advanced/Scripts/utils.sh
+++ b/advanced/Scripts/utils.sh
@@ -15,7 +15,10 @@
 #  - New functions must have a test added for them in test/test_any_utils.py
 
 #######################
-# Takes three arguments key, value, and file.
+# Takes either
+#   - Three arguments: key, value, and file.
+#   - Two arguments: key, and file
+#
 # Checks the target file for the existence of the key
 #   - If it exists, it changes the value
 #   - If it does not exist, it adds the value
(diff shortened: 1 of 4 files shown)


advanced/Scripts/utils.sh
```diff
@@ -25,15 +28,48 @@
...
   local key="${1}"
-  local value="${2}"
-  local file="${3}"
-  if grep -q "^${key}=" "${file}"; then
-    sed -i "/^${key}=/c\\${key}=${value}" "${file}"
+  local value
+  local file
+  
+  # If two arguments have been passed, then the second one is the file - there is no value
+  if [ $# -lt 3 ]; then
```
dschaper, line 35:
```suggestion
  if [[ $# -lt 3 ]]; then    
```

Either double here or single on line 42, just so they match.

advanced/Scripts/utils.sh
```diff
@@ -15,7 +15,10 @@
...
 
 #######################
-# Takes three arguments key, value, and file.
+# Takes either
+#   - Three arguments: key, value, and file.
+#   - Two arguments: key, and file
+#
 # Checks the target file for the existence of the key
 #   - If it exists, it changes the value
 #   - If it does not exist, it adds the value
```
dschaper, line 24: Thank you for function comments!!

dschaper (changes requested):
Can you make sure that you don't use any bashisms? I'm working on POSIX compliance so we can run in non-bash native distributions. (i.e. Alpine).

advanced/Scripts/utils.sh
rdwebdesign: I may be talking nonsense, but wouldn't it be easier to remember, more readable (and simpler to test) if we used a different parameters order? 

Since the filename and key must always be present, wouldn't it be better if it were this way?

```sh
addOrEditKeyValuePair {filename} {key} {value (optional)}

addOrEditKeyValuePair "/etc/pihole/setupVars.conf" "BLOCKING_ENABLED" "true"
addOrEditKeyValuePair "/etc/pihole/setupVars.conf" "KEY_WITH_NO_VALUE"
```

yubiuser: I agree with @rdwebdesign, changing the order makes a lot of sense to me.

test/test_any_utils.py
```diff
@@ -2,15 +2,33 @@ def test_key_val_replacement_works(host):
...
     output = host.run('''
     cat ./testoutput
     ''')
-    expected_stdout = 'KEY_ONE=value3\nKEY_TWO=value2\nKEY_FOUR=value4\n'
+    expected_stdout = 'KEY_ONE=value3\nKEY_TWO=value2\nKEY_FOUR=value4\nKEY_FIVE_NO_VALUE\n'
+    assert expected_stdout == output.stdout
+
+
+def test_key_val_removal_works(host):
+    ''' Confirms addOrEditKeyValPair provides the expected output '''
```
yubiuser, line 20:
```suggestion
    ''' Confirms removeKey provides the expected output '''
```
```suggestion
    ''' Confirms addOrEditKeyValPair provides the expected output '''
```

yubiuser, line 20:
```suggestion
    ''' Confirms removeKey provides the expected output '''
```
```suggestion
    ''' Confirms addOrEditKeyValPair provides the expected output '''
```

advanced/Scripts/utils.sh
```diff
@@ -15,46 +17,74 @@
...
 }
 
+#######################
+# Takes two arguments key, and file.
+# Deletes a key from target file
+#
+# Example usage:
+# removeKey "PIHOLE_DNS_1" "/etc/pihole/setupVars.conf"
+#######################
+removeKey() {
```
yubiuser, line 61: Is it cheeky to ask to change the order of arguments here as well? Just to have the same as in `addOrEditKeyValPair`: file first, key second?

yubiuser, line 61: Is it cheeky to ask to change the order of arguments here as well? Just to have the same as in `addOrEditKeyValPair`: file first, key second?

yubiuser (commented):
L263 and 273 in `pihole` can be expressed with the new functions form `utils.sh` as well

advanced/Scripts/utils.sh
```diff
@@ -15,46 +17,74 @@
...
+    # value has no value, so it is just a key. Add it if it does not already exist
+    if ! grep -q "^${key}" "${file}"; then
+      # Key does not exist, add it.
+      echo "${key}" >> "${file}"
+    fi
   fi
 }
 
+#######################
+# Takes two arguments key, and file.
```
yubiuser, line 55:
```suggestion
# Takes two arguments file, and key.
```

advanced/Scripts/utils.sh
```diff
@@ -15,46 +17,74 @@
...
-  local value="${2}"
-  local file="${3}"
-  if grep -q "^${key}=" "${file}"; then
-    sed -i "/^${key}=/c\\${key}=${value}" "${file}"
+  local file="${1}"
+  local key="${2}"
+  local value="${3}"
+
+  if [ "${value}" != "" ]; then
+    # value has a value, so it is a key pair
```
yubiuser, line 37:
```suggestion
    # value has a value, so it is a key-value pair
```

yubiuser (approved):

Issue #3977, opened 2021-01-10
endles loop for test_dpkg_lock() although has counter..


MatthewCKelly:
### Versions

- Pi-hole:   v5.2.2.
- AdminLTE:  v5.2.2.
- FTL:  v5.3.4.

### Platform

- OS and version:  Linux XXX 4.4.190.x86_64.1 #1 SMP Thu Oct 22 04:41:26 UTC 2020 x86_64 GNU/Linux

- Platform:  Netgear ReadyNAS 314

### Expected behavior

for some reason my system /var/lib/dpkg/lock is always in use.. 
In the function  test_dpkg_lock() there looks like there is an '`# An iterator used for counting loop iterations`'  However, there is no additional code to break out of that function when that value reaches a given value..

### Actual behavior / bug
endless loop, will continue until user breaks execution..

### Steps to reproduce

```
wget -O basic-install.sh https://install.pi-hole.net
PIHOLE_SKIP_OS_CHECK=true   bash basic-install.sh
update
```

![image](https://user-images.githubusercontent.com/16784124/104111357-c0f40b00-5345-11eb-872a-b1a02f2dce26.png)

Inserted line after `i` was incremented :             `echo "in function test_dpkg_lock() i=$i"`

## Debug Token

- URL:  https://tricorder.pi-hole.net/culvx9ablr

### Screenshots

_If applicable, add screenshots to help explain your problem._

### Additional context

any change to this function will need to be updated back at where it was called from..

dschaper:
Looks like it was there originally and was changed before merge. This was back in version 3, so approx 4 years or so ago?

@DL6ER Check https://github.com/pi-hole/pi-hole/pull/1286/commits/a535ca9db4504dc75c57097fec57b123d6b2ce90 and see if that makes sense to you.

As for your system, I don't think you'll be able to install if the apt package lock is always there. Even if you skip the lock check you'll fail when `apt install` fails for the lock file existing.

MatthewCKelly:
@dschaper strangely bypassing the lock check allows all the install to complete, and I have it is working well on my system..

DL6ER:
@dschaper Doesn't seem to make sense. As the `i` counter is still there, I can only assume the `while` was removed accidentally.

sjt5:
Lovely unintentional infinite while loop.
Could add a simple check and bail to prevent infinite loop:
```
[...]

PR #1286, opened 2017-03-03
Test for dpkg lock before using apt-get

DL6ER:
**By submitting this pull request, I confirm the following (please check boxes, eg [X]) _Failure to fill the template will close your PR_:**

***Please submit all pull requests against the `development` branch. Failure to do so will delay or deny your request***

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] I have checked that [another pull request](https://github.com/pi-hole/pi-hole/pulls) for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**How familiar are you with the codebase?:** 10

---

This will solve a number of issues for people seeing the package installer crashing. Basically, this change will ensure to check that the package manager can actually be run before trying to update e.g. the package lists and/or install packages.

This kind of problems can e.g. be seen on fresh installed virtual machines. They commonly run `unattended-upgrade` at the first boot to get all security updates. Depending on the resources, this can take several minutes within which the Pi-hole installer crashes with no obvious reason. This PR solves this problem as the installer will always wait until it is safe to use the package manager
```
root@vultr:~# cd /etc/.pihole
root@vultr:/etc/.pihole# git checkout tweak/dpkg_lock 
Branch tweak/dpkg_lock set up to track remote branch tweak/dpkg_lock from origin.
Switched to a new branch 'tweak/dpkg_lock'
root@vultr:/etc/.pihole# ./automated\ install/basic-install.sh
:::
::: You are root.
::: Verifying free disk space...
:::
::: Updating local cache of available packages...
```
and continues only afterwards.

Note that this implementation is Debian specific, an there is no implementation for Fedora based systems so far (I don't even know if `dnf` uses a lock as well).

_This template was created based on the work of [`udemy-dl`](https://github.com/nishad/udemy-dl/blob/master/LICENSE)._

PR #4682, opened 2022-04-04
Exit installer if dpkg lock is held for more then 30 seconds

yubiuser:

---
Fixes https://github.com/pi-hole/pi-hole/issues/3977.

When installing via `apt` we use a function to check if `dpkg` is holding a lock on `/va/lib/dpkg/lock`. If so we wait...until the lock is lifted. If there is no lifting we wait until forever.
This PR adds a safety check to stop the whole installer if the lock is not lifted within 30 seconds.The original PR https://github.com/pi-hole/pi-hole/pull/1286 already introduced a counter but never used it. We do use it now.

P.S. The PR additionally moves the function `test_dpkg_lock` out of `package_manager_detect` because it is not used in there. (Only later in `install_dependent_packages`)

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -262,7 +262,7 @@ os_check() {
 # This function waits for dpkg to unlock, which signals that the previous apt-get command has finished.
 test_dpkg_lock() {
     i=0
-    printf "  %b   Waiting for package manager to finish\\n" "${INFO}"
+    printf "  %b Waiting for package manager to finish\\n" "${INFO}"
     # fuser is a program to show which processes use the named files, sockets, or filesystems
     # So while the lock is held,
     while fuser /var/lib/dpkg/lock >/dev/null 2>&1
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -262,7 +262,7 @@ os_check() {
 # This function waits for dpkg to unlock, which signals that the previous apt-get command has finished.
 test_dpkg_lock() {
     i=0
-    printf "  %b Waiting for package manager to finish\\n" "${INFO}"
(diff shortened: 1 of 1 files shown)


PromoFaux:
No issue with this in theory, nobody wants an infinite loop with no user feedback, is 30 seconds a sensible default?

yubiuser:
We can easily make it 60 seconds or more.... but at some point users will cancle the installer hard because they think it got stuck.

automated install/basic-install.sh
```diff
@@ -259,6 +259,29 @@ os_check() {
     fi
 }
 
+# This function waits for dpkg to unlock, which signals that the previous apt-get command has finished.
+test_dpkg_lock() {
+    i=0
+    printf "  %b Waiting for package manager to finish\\n" "${INFO}"
```
PromoFaux, line 265:
```suggestion
    printf "  %b Waiting for package manager to finish (up to 30 seconds)\\n" "${INFO}"
```

Or something like this?

yubiuser: Happy to add this after we agreed on a time

PromoFaux: I think 30 seconds is probably fine - I just wonder if there are any legitimate processes we could end up killing/cutting short if we exit after 30 seconds - but I may be thinking about it the wrong way around

yubiuser: We do not kill any other process. If 30 seconds have passed, we exit our installer. The other process is allowed to keep the lock.

PromoFaux: Right, in that case I totally got the wrong end of the stick here - and I think 30 seconds is fine!

PR #4698, opened 2022-04-14
[utils.sh] Split `addOrEditKeyValuePair` to do `addKey` in a separate function

PromoFaux:

---

Having a function with a variable number of arguments in this case was too complicated for something that is only really used once

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -155,7 +155,7 @@ ProcessDNSSettings() {
         if [ -z "${!var}" ]; then
             break;
         fi
-        addOrEditKeyValPair "${dnsmasqconfig}" "server" "${!var}"
+        addKey "${dnsmasqconfig}" "server=${!var}"
         (( COUNTER++ ))
     done
 
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -88,7 +88,7 @@ delete_dnsmasq_setting() {
 }
 
 SetTemperatureUnit() {
-    change_setting "TEMPERATUREUNIT" "${unit}"
+    addOrEditKeyValPair "${setupVars}" "TEMPERATUREUNIT" "${unit}"
     echo -e "  ${TICK} Set temperature unit to ${unit}"
 }
 
@@ -123,7 +123,7 @@ SetWebPassword() {
         echo ""
 
         if [ "${PASSWORD}" == "" ]; then
-            change_setting "WEBPASSWORD" ""
+            addOrEditKeyValPair "${setupVars}" "WEBPASSWORD" ""
             echo -e "  ${TICK} Password Removed"
             exit 0
         fi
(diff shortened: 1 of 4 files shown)


yubiuser (changes requested):
If splitting to add `addkey` the wrappers in `webpage.sh` need adjustment.

advanced/Scripts/webpage.sh
```diff
@@ -147,67 +147,67 @@ SetWebPassword() {
...
+    removeKey "${dnsmasqconfig}" "server"
 
     COUNTER=1
     while true ; do
         var=PIHOLE_DNS_${COUNTER}
         if [ -z "${!var}" ]; then
             break;
         fi
-        add_dnsmasq_setting "server" "${!var}"
+        addOrEditKeyValPair "${dnsmasqconfig}" "server" "${!var}"
```
yubiuser, line 158: I think this is not going to work. If I understand the whole part right, it is reading `PIHOLE_DNS_X` from `setupVars` and adding it to `01-pi-hole.conf`. 
`addOrEditKeyValPair` is preventing adding multiple `server` lines, as it changes them always. You need

```suggestion
        addKey "${dnsmasqconfig}" "server=${!var}"
```

PromoFaux: Ah yes. That is a nice point. Currently mobile, so difficult to look - but do existing values get cleared out first, before we loop through PIHOLE_DNS_ values?

yubiuser: Yes.

https://github.com/pi-hole/pi-hole/blob/23e6fa1ec56e7e24054d359ef5da0114e2f9b77f/advanced/Scripts/webpage.sh#L147-L154

test/test_any_utils.py
```diff
@@ -6,8 +6,8 @@ def test_key_val_replacement_works(host):
     addOrEditKeyValPair "./testoutput" "KEY_TWO" "value2"
     addOrEditKeyValPair "./testoutput" "KEY_ONE" "value3"
     addOrEditKeyValPair "./testoutput" "KEY_FOUR" "value4"
-    addOrEditKeyValPair "./testoutput" "KEY_FIVE_NO_VALUE"
-    addOrEditKeyValPair "./testoutput" "KEY_FIVE_NO_VALUE"
+    addKey "./testoutput" "KEY_FIVE_NO_VALUE"
+    addKey "./testoutput" "KEY_FIVE_NO_VALUE"
```
yubiuser, line 10: I'm not sure what you want to do here. Why do you add it twice?

It this is what you really want to check for, adjust the check in L15 to

```
expected_stdout = 'KEY_ONE=value3\nKEY_TWO=value2\nKEY_FOUR=value4\nKEY_FIVE_NO_VALUE\nKEY_FIVE_NO_VALUE\n'
``

PromoFaux: Ah, no. Thinking on it... AddKey is missing some logic to check for dupes before adding

yubiuser: Right. Good Point!

yubiuser: I was wrong... we do have a check on L54
````
if ! grep -q "^${key}" "${file}"; then
`````
But still I think it makes not much sense  to add it twice.

PromoFaux: Ah. Then - we are ok.

The original logic of the test was to prove that adding something that already existed would not duplicate. Maybe we can create a separate test to make this clearer?

yubiuser: This is a good idea. 
We have separted the functions and as you wrote in `utils.sh` ;-)
> New functions must have a test added for them in test/test_any_utils.py

yubiuser (approved):

Issue #4697, opened 2022-04-14
CentOS8/EL8/AlmaLinux8 aarch64 container - pihole-FTL architecture detection doesnt work


spmfox:
Trying to install Pi-hole in a CentOS (AlmaLinux 8) podman container - host is a Raspberry Pi 4 running the 64-bit Raspberry Pi OS.

I've discovered that the architecture check logic for pihole-FTL does not work due to "ldd /bin/ls" not working.

After more digging I found that the CentOS 8, Red Hat UBI 8, and AlmaLinux 8 (aarch64 specifically) containers are not compatible with this check method because "ls" is not a binary - its a text file that contains one line (#!/usr/bin/coreutils --coreutils-prog-shebang=ls).

CentOS 7 aarch64 container has ls as a normal binary as expected, but all the 8 variants use this script to coreutils instead.

Since "CentOS" is an officially supported OS, is there a supported way to fix this? Or even override the architecture detection and just specify via environment variable?

spmfox:
I created a pull request with a proposed change. I notice there are several system checks, I wonder do these checks run in containers? I don't know enough how these tests work, however it might be valuable to start checking some of these distros using the aarch64 versions - however I can see how that may increase exponentially. Testing every distro on every architecture. However, still, since this project is heavily marketed to Raspberry Pi, it may be good to test aarch64 container images.

spmfox:
Just tested, everything works.

Thanks again for the help everyone.

PR #4700, opened 2022-04-16
Changed ldd binary check for better compatibility

spmfox:

---
https://github.com/pi-hole/pi-hole/issues/4697

The basic-install.sh script checks which architecture it is running on so it downloads the correct pihole-FTL binary file. It uses a simple command, `ldd /bin/ls` to determine which libraries the underlying system is using. The problem comes in when using the aarch64 container version of CentOS 8 (also RHEL 8, AlmaLinux 8, and all RHEL clones). These container images do not have 'ls' or some of the other core utilities. They use '/usr/bin/coreutils' with a argument to invoke 'ls'.

This change is simple, instead of using '/bin/ls' as the binary check - we use '/bin/bash instead'. This should not break any compatibility and after all the script is being invoked by bash.

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -2216,7 +2216,7 @@ get_binary_name() {
         local rev
         rev=$(uname -m | sed "s/[^0-9]//g;")
         local lib
-        lib=$(ldd $(which sh) | grep -E '^\s*/lib' | awk '{ print $1 }')
+        lib=$(ldd "$(which sh)" | grep -E '^\s*/lib' | awk '{ print $1 }')
         if [[ "${lib}" == "/lib/ld-linux-aarch64.so.1" ]]; then
             printf "%b  %b Detected AArch64 (64 Bit ARM) processor\\n" "${OVER}" "${TICK}"
             # set the binary to be used
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -2216,7 +2216,7 @@ get_binary_name() {
         local rev
         rev=$(uname -m | sed "s/[^0-9]//g;")
         local lib
-        lib=$(ldd /bin/bash | grep -E '^\s*/lib' | awk '{ print $1 }')
+        lib=$(ldd $(which sh) | grep -E '^\s*/lib' | awk '{ print $1 }')
         if [[ "${lib}" == "/lib/ld-linux-aarch64.so.1" ]]; then
             printf "%b  %b Detected AArch64 (64 Bit ARM) processor\\n" "${OVER}" "${TICK}"
             # set the binary to be used
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -2216,7 +2216,7 @@ get_binary_name() {
         local rev
         rev=$(uname -m | sed "s/[^0-9]//g;")
         local lib
-        lib=$(ldd /bin/ls | grep -E '^\s*/lib' | awk '{ print $1 }')
+        lib=$(ldd /bin/bash | grep -E '^\s*/lib' | awk '{ print $1 }')
         if [[ "${lib}" == "/lib/ld-linux-aarch64.so.1" ]]; then
             printf "%b  %b Detected AArch64 (64 Bit ARM) processor\\n" "${OVER}" "${TICK}"
             # set the binary to be used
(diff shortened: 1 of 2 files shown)


PromoFaux (dismissed):

yubiuser (commented):
Maybe use `/usr/bin/sh`? We might be more Posix compatible one day...

spmfox:
Sure, that's what I wanted to do but I wasn't sure if /bin was preferred. I didn't want to change too much.

I'll update the pull request in a bit.

yubiuser:
`bin` is fine

dschaper:
Use `$(which sh)` and you don't have to worry about path.

automated install/basic-install.sh
```diff
@@ -2216,7 +2216,7 @@ get_binary_name() {
         local rev
         rev=$(uname -m | sed "s/[^0-9]//g;")
         local lib
-        lib=$(ldd /bin/ls | grep -E '^\s*/lib' | awk '{ print $1 }')
+        lib=$(ldd /bin/bash | grep -E '^\s*/lib' | awk '{ print $1 }')
```
dschaper, line 2219:
```suggestion
        lib=$(ldd $(which sh) | grep -E '^\s*/lib' | awk '{ print $1 }')
```

dschaper (changes requested):

spmfox:
@dschaper I'm not sure how to update the test/test_any_automated_install.py for this change though.

dschaper:
Okay, I think the test should be okay, but if they are not then we can mock the `ldd` call and basically echo back what ever we say the proper response should be. The test isn't about `ldd` returning a proper value but about the function knowing how to handle responses we know about and responses that we are not expecting.

So I'd test two scenarios, one that `ldd` returns the intended response (pick an architecture, doesn't matter which one really) and the other when `ldd` returns nothing or something we don't expect...

test/test_any_automated_install.py
```diff
@@ -679,7 +679,7 @@ def test_FTL_detect_aarch64_no_errors(host):
     mock_command(
         'ldd',
         {
-            '/bin/ls': (
+            '/bin/bash': (
```
dschaper, line 682:
```suggestion
            *: (
```
I don't remember offhand what the proper way to say "Return this no matter what is called"

test/test_any_automated_install.py
```diff
@@ -807,7 +807,7 @@ def test_FTL_detect_armv8a_no_errors(host):
     # mock uname to return armv8a platform
     mock_command('uname', {'-m': ('armv8a', '0')}, host)
     # mock ldd to respond with ld-linux-armhf shared library
-    mock_command('ldd', {'/bin/ls': ('/lib/ld-linux-armhf.so.3', '0')}, host)
+    mock_command('ldd', {'/bin/bash': ('/lib/ld-linux-armhf.so.3', '0')}, host)
     detectPlatform = host.run('''
     source /opt/pihole/basic-install.sh
     create_pihole_user
```
dschaper, line 813: Add a last test here that returns `No such file or directory` as the last test.

dschaper (changes requested):

spmfox:
Yeah it looks like the current test is failing, which it passed before. I don't know much python nor do I understand fully how the test even works. In my original change, I just changed 'ls' to 'bash' so it was pretty straightforward.

dschaper:
I'm not sure if we need to mock the `which` command as well so that `which sh` returns something like `/bin/sh` and then the existing tests should behave like we expect.

test/test_any_automated_install.py
dschaper: Let's leave this as it is and mock `which` to return `/bin/sh` for `which sh`.

automated install/basic-install.sh
```diff
@@ -2216,7 +2216,7 @@ get_binary_name() {
         local rev
         rev=$(uname -m | sed "s/[^0-9]//g;")
         local lib
-        lib=$(ldd /bin/ls | grep -E '^\s*/lib' | awk '{ print $1 }')
+        lib=$(ldd $(which sh) | grep -E '^\s*/lib' | awk '{ print $1 }')
```
dschaper, line 2219:
```suggestion
        lib=$(ldd "$(which sh)" | grep -E '^\s*/lib' | awk '{ print $1 }')
```

PromoFaux (approved):

Issue #3361, opened 2020-05-13
Codespell report for "Pi-hole" (on fossies.org)


jschleus:
Since you encourage to report also typos here a list of some few spelling errors and typos that the server "[fossies.org](https://fossies.org/features.html#codespell)" have found with the help of the tool "[codespell](https://github.com/codespell-project/codespell)":

 https://fossies.org/linux/misc/dns/pi-hole-5.0.tar.gz/codespell.html

If it would make sense I could also implement a "continuous" check of the current master version.
But that would be done within a restricted special test folder that isn't fully integrated into the standard Fossies services.

PromoFaux:
I wonder if there is something we can implement using Github Actions, that checks Pull Requests and the `development` branch

jschleus:
Not the optimal solution and probably not what you want but fossies.org can offer a report at

 https://fossies.org/linux/test/pi-hole-development-snap.tar.gz/codespell.html

That testwise realized URL should always redirect to a codespell report for the current `development` branch (currently with a delay of max. 30 mins). As an additional information that page contains also a spelling error history report.

dschaper:
https://github.com/codespell-project/codespell/issues/1499

jschleus:
Oh, nice information. That are very interesting and promising approaches.

peternewman:
Hi @dschaper / @jschleus ,

Glad you found that, given it seems to work already, feel free to just go with it if you want.

You just need a workflow file like this:
https://github.com/peternewman/actions-codespell/blob/master/.github/workflows/testing.yml

But change `uses: ./` to `uses: peternewman/actions-codespell` and it should just work.

I'm going to add more options soon (or they're as simple as this https://github.com/TrueBrain/actions-flake8/pull/9/files ), and I expect move it to the codespell-project organisation/publish on Marketplace etc.

Are there particular codespell arguments you need to make it work for you?

dschaper:
Thanks for the note, @PromoFaux if you want to investigate?

peternewman:
Just to let you know, the repo is now transferred, so please use https://github.com/codespell-project/actions-codespell going forward.

yubiuser:
Pinging @PromoFaux

PromoFaux:
@peternewman Apologies for this taking me sooooo long to look into. Not really sure where time went there.

[...]

PR #4713, opened 2022-04-22
Add code-spell action / Move editorconfig-checker action to smoke-tests

PromoFaux:

---

Addresses https://github.com/pi-hole/pi-hole/issues/3361


.codespellignore
```diff
@@ -0,0 +1,3 @@
+doubleclick
+wan
+nwe
```
yubiuser, line 3: Does it work with
```suggestion
\nWe
```

peternewman: The ignore file is about effectively removing entries from the codespell dictionary, so this change would stop codespell checking for that typo, rather than ignoring that phrase.

The solution would be to use the --ignore-regex option, which will essentially vanish the pattern `\nwe` from the source text that codespell then spellchecks, so it will never find those instances of nwe:
https://github.com/codespell-project/codespell/blob/c807a6cdd1cc88e1a7c75970c855b0b53d4521ac/codespell_lib/_codespell.py#L292-L299

Although I don't think this is in actions-codespell yet, so you'd need to have a codespell config file and use that.
https://github.com/codespell-project/codespell#using-a-config-file
Although it's not yet merged in for reasons I can't immediately remember:
https://github.com/codespell-project/actions-codespell/pull/19

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,18 @@
+name: Codespell
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+jobs:
+  spell-check:
+    if: github.event.pull_request.draft == false
```
peternewman, line 8: ❤️ 
That's cool, although does mean someone couldn't find typos in their PR until it's open for review...

.codespellignore
yubiuser: Thanks for your input. So, the way to go would be: you merge https://github.com/codespell-project/actions-codespell/pull/19, then we could create a config file which would allow to enable `-ignore-regex`?

.github/workflows/codespell.yml
PromoFaux: I just copy pasted the full template from one of our other actions and changed the parts below the `if:`, wasn't really a considered choice! 🙂

I've also considered running it along side the [`smoke-tests`](https://github.com/pi-hole/pi-hole/blob/master/.github/workflows/test.yml#L19) here, and preventing `distro-test` from running until code-spell is satistfied.

yubiuser:
Found one more:
https://github.com/pi-hole/pi-hole/blob/853f6b7da2b5130e33d76840f463ab374eb77596/automated%20install/basic-install.sh#L198

yubiuser:
@PromoFaux what is the state of this PR?

PromoFaux:
Is it not waiting on another PR (per [this comment](https://github.com/pi-hole/pi-hole/pull/4713#discussion_r857150469))

Else we can just add it the same as we have with the other repos

.codespellignore
yubiuser: @peternewman 
Do you have any roadmap for https://github.com/codespell-project/actions-codespell/pull/19?

PromoFaux: @yubiuser - I'm happy with this PR in it's current state, and then add enhancements later?

yubiuser (approved):

Issue #4751, opened 2022-06-06
E: Unable to locate package php0.0-common


bradydean:
### Versions
```
  Pi-hole version is v5.10 (Latest: v5.10)
  AdminLTE version is v5.12 (Latest: v5.12)
  FTL version is v5.12.1 (Latest: v5.15)
```

### Platform
Debian 11 arm, Raspberry Pi

### Expected behavior

Installer script updates pi-hole to latest version.

### Actual behavior / bug

```
  [i] Checking for / installing Required dependencies for Pi-hole software...
  [✓] Checking for cron
  [✓] Checking for curl
  [✓] Checking for iputils-ping
  [✓] Checking for psmisc
  [✓] Checking for sudo
  [✓] Checking for unzip
  [✓] Checking for idn2
  [✓] Checking for libcap2-bin
  [✓] Checking for dns-root-data
  [✓] Checking for libcap2
  [✓] Checking for netcat-openbsd
  [✓] Checking for procps
  [✓] Checking for lighttpd
  [i] Checking for php0.0-common (will be installed)
  [i] Checking for php0.0-cgi (will be installed)
  [i] Checking for php0.0-sqlite3 (will be installed)
  [i] Checking for php0.0-xml (will be installed)
  [i] Checking for php0.0-intl (will be installed)
  [i] Checking for php0.0-json (will be installed)
  [i] Waiting for package manager to finish (up to 30 seconds)
  [i] Processing apt-get install(s) for: php0.0-common php0.0-cgi php0.0-sqlite3 php0.0-xml php0.0-intl php0.0-json, please wait...

E: Unable to locate package php0.0-common
E: Couldn't find any package by glob 'php0.0-common'
E: Couldn't find any package by regex 'php0.0-common'
E: Unable to locate package php0.0-cgi
E: Couldn't find any package by glob 'php0.0-cgi'
E: Couldn't find any package by regex 'php0.0-cgi'
E: Unable to locate package php0.0-sqlite3
E: Couldn't find any package by glob 'php0.0-sqlite3'
E: Couldn't find any package by regex 'php0.0-sqlite3'
E: Unable to locate package php0.0-xml
E: Couldn't find any package by glob 'php0.0-xml'
E: Couldn't find any package by regex 'php0.0-xml'
E: Unable to locate package php0.0-intl
E: Couldn't find any package by glob 'php0.0-intl'
E: Couldn't find any package by regex 'php0.0-intl'
E: Unable to locate package php0.0-json
[...]

PR #4757, opened 2022-06-13
Exit installer if no valid PHP version is found

yubiuser:

Exit the installer if the detected PHP version is not a number. This was reported here: https://github.com/pi-hole/pi-hole/issues/4751

Check if `$phpInsMajor` and `$phpInsMinor` are `[:digit:]`

None.

---

---
[x] I have read the above and my PR is ready for review. _Check this box to confirm_

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -301,8 +301,8 @@ package_manager_detect() {
         local phpVer="php"
         if is_command php ; then
             printf "  %b Existing PHP installation detected : PHP version %s\\n" "${INFO}" "$(php <<< "<?php echo PHP_VERSION ?>")"
-            printf -v phpInsMajor "$(php <<< "<?php echo PHP_MAJOR_VERSION ?>")"
-            printf -v phpInsMinor "$(php <<< "<?php echo PHP_MINOR_VERSION ?>")"
+            printf -v phpInsMajor "%s" "$(php <<< "<?php echo PHP_MAJOR_VERSION ?>")"
+            printf -v phpInsMinor "%s" "$(php <<< "<?php echo PHP_MINOR_VERSION ?>")"
             if [[ "$phpInsMajor" =~ [^[:digit:]] || "$phpInsMinor" =~ [^[:digit:]]  ]]; then
                  printf "  %b No valid PHP version detected\\n" "${CROSS}"
                 # so exit the installer
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -298,7 +298,7 @@ package_manager_detect() {
         # Update package cache
         update_package_cache || exit 1
         # Check for and determine version number (major and minor) of current php install
-        local phpVer
+        local phpVer="php"
         if is_command php ; then
             phpVer="$(php -v 2> /dev/null | head -n1 | cut -d '-' -f1 | cut -d ' ' -f2)"
             # Check if the first character of the string is numeric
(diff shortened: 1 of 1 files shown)


automated install/basic-install.sh
```diff
@@ -301,8 +301,13 @@ package_manager_detect() {
...
             printf "  %b Existing PHP installation detected : PHP version %s\\n" "${INFO}" "$(php <<< "<?php echo PHP_VERSION ?>")"
-            printf -v phpInsMajor "%d" "$(php <<< "<?php echo PHP_MAJOR_VERSION ?>")"
-            printf -v phpInsMinor "%d" "$(php <<< "<?php echo PHP_MINOR_VERSION ?>")"
+            printf -v phpInsMajor "%s" "$(php <<< "<?php echo PHP_MAJOR_VERSION ?>")"
+            printf -v phpInsMinor "%s" "$(php <<< "<?php echo PHP_MINOR_VERSION ?>")"
+            if [[ "$phpInsMajor" =~ [^[:digit:]] || "$phpInsMinor" =~ [^[:digit:]]  ]]; then
+                 printf "  %b No valid PHP version detected\\n" "${CROSS}"
+                # so exit the installer
+                exit
+            fi
```
dschaper, line 310: `PHP_VERSION` versions should always be in the format of `1.1x.1x`. 

I think a good way to structure it is:

```pseudo
if is_command php then
  PHP_VERSION = $(php <<< "<?php echo PHP_VERSION ?>")"
  if first char of PHP_VERSION is not numeric then
   print an error message that PHP is not functioning correctly
   print what we see is the PHP_VERSION (most likely a null or empty string but might have the error message returned by `php`.
   print an explanation that this is a fatal error and we are going to quit.
   exit
 else:
  do all the good stuff(TM)
 endif
```

dschaper (changes requested):

automated install/basic-install.sh
yubiuser: > `if first char of PHP_VERSION is not numeric then`

I'll do it the other way round, because it's easier to define what's valid

dschaper: SGTM.

I don't know why I always test negations instead of the condition.

yubiuser:
Not sure why the tests are failing. It fails to install the `php` packages, because `phpVer` is empty which produced

`[i] Checking for -intl (will be installed)\n  [i] Checking for -json`

Looks like it's not entering `if [[ ${phpVer:0:1} =~ [1-9] ]]; then`....

automated install/basic-install.sh
```diff
@@ -300,10 +300,19 @@ package_manager_detect() {
         # Check for and determine version number (major and minor) of current php install
         local phpVer="php"
         if is_command php ; then
-            printf "  %b Existing PHP installation detected : PHP version %s\\n" "${INFO}" "$(php <<< "<?php echo PHP_VERSION ?>")"
-            printf -v phpInsMajor "%d" "$(php <<< "<?php echo PHP_MAJOR_VERSION ?>")"
-            printf -v phpInsMinor "%d" "$(php <<< "<?php echo PHP_MINOR_VERSION ?>")"
-            phpVer="php$phpInsMajor.$phpInsMinor"
+            phpVer="$(php -v 2> /dev/null | head -n1 | cut -d '-' -f1 | cut -d ' ' -f2)"
+            # Check if the first character of the string is numeric
```
dschaper, line 304: I don't know what a corrupt `php` setup will return for `php -v`. I do know that the call to `php` echoing `PHP_VERSION` will give us output that we already know about. Plus that call doesn't need any pipes or cuts.

dschaper (approved):

PR #4790, opened 2022-07-07
[Meta] Reformat yml files

PromoFaux:
Also addresses stickler issues in the release PR : https://github.com/pi-hole/pi-hole/pull/4789/checks?check_run_id=7238526403

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -13,14 +13,14 @@ jobs:
       issues: write
 
     steps:
-    - uses: actions/stale@v5
-      with:
-        repo-token: ${{ secrets.GITHUB_TOKEN }}
-        days-before-stale: 30
-        days-before-close: 5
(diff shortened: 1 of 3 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -23,18 +23,14 @@ jobs:
...
+        uses: actions/checkout@v3
+      # Initializes the CodeQL tools for scanning.
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v2
+        with:
+          languages: "python"
+      - name: Autobuild
+        uses: github/codeql-action/autobuild@v2
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@v2
```
dschaper, line 36: I kind of like the old format, with the space to break up the steps. We can set some of the rules with:

https://github.com/adrienverge/yamllint

Add the config to stickler: https://stickler-ci.com/docs#yaml

PR #4833, opened 2022-07-24
Explicit set minor and patch version on used actions

yubiuser:

Allow updates for GH `actions` by dependabot also for minor and patch versions. So far, we only specify major versions in our workflows. Therefore, dependabot will only check if new major versions exist and won't update the workflow on new minor or patch versions.

By explicitly setting minor and patch version we allow dependabot to update those as well.

E.g. `action/stale@v5` is currently missing the `close-issue-reason`. This was added in `v5.1.0`. However, dependabot fails to recognize the update atm.

P.S. I expect some dependabot PRs after this has been merged. I did not update the versions manually to get the dependabot changelog.

---

---

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -29,12 +29,12 @@ jobs:
     # Initializes the CodeQL tools for scanning.
     -
       name: Initialize CodeQL
-      uses: github/codeql-action/init@v2.0.0
+      uses: github/codeql-action/init@v2
       with:
         languages: 'python'
     -
       name: Autobuild
-      uses: github/codeql-action/autobuild@v2.0.0
+      uses: github/codeql-action/autobuild@v2
     -
       name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@v2.0.0
+      uses: github/codeql-action/analyze@v2
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -25,16 +25,16 @@ jobs:
     steps:
     -
       name: Checkout repository
-      uses: actions/checkout@v3
+      uses: actions/checkout@v3.0.0
     # Initializes the CodeQL tools for scanning.
     -
       name: Initialize CodeQL
-      uses: github/codeql-action/init@v2
+      uses: github/codeql-action/init@v2.0.0
       with:
         languages: 'python'
     -
       name: Autobuild
-      uses: github/codeql-action/autobuild@v2
+      uses: github/codeql-action/autobuild@v2.0.0
     -
       name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@v2
+      uses: github/codeql-action/analyze@v2.0.0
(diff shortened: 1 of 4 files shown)


yubiuser:
Smoke tests fails due to an upstream issue
https://github.com/editorconfig-checker/editorconfig-checker/issues/213

yubiuser:
CodeQl does not seem to use `.minor.patch` versioning.

PromoFaux:
Are the minors and patches not rolled into the v3 action? I've never seen an example workflow use the full version number...

DL6ER:
@PromoFaux I don't think so, when you go to any action page and click on "Use latest version", you get a yaml code suggestion like

Screenshot_2022-07-24-15-19-30-48_3aea4af51f236e4932235fdada7d1643~2

PromoFaux:
👀 - and that's not even the latest version!

DL6ER:
Screenshot_2022-07-24-15-28-47-05_3aea4af51f236e4932235fdada7d1643~2

yubiuser:
Currently latest versions:

`checkout v3.0.2`
`stale v5.1.0`
`action-add-labels v1.1.0`
`action-editorconfig-checker v1.0.0`
`setup-python v4.1.0`

PromoFaux:
Just throwing [this](https://docs.github.com/en/actions/creating-actions/about-custom-actions#versioning-your-action:~:text=Instead%2C%20you%20can%20recommend%20that%20your%20users%20specify%20a%20major%20version%20when%20using%20your%20action%2C%20and%20only%20direct%20them%20to%20a%20more%20specific%20version%20if%20they%20encounter%20issues.) in here. 

Like Is said, I've never seen any docs point to using an exact version before - upstream actions owners should be moving the `v2`/`v3` etc tags to point to the latest minor/patch version

It just feels like Dependabot will be opening unnecessary PRs at with this change. Taking the actions/checkout repo as an example:

image


`v3` and `v3.0.2` point to the same commit -Dependabot does not need to do anything 🤷

PromoFaux:
It appears only `stale` has not done this. 

image

image

image

yubiuser:
I see this is an issue with 'stale' not moving their release to the latest commit.
However, I still think this a valid PR: we know exactly which action version we use, we see what's changed between versions and can easily revert if something breaks.
I'd like to know which software we are using. We always tell users to read our release notes, but for the actions we use we go mostly blindly so far.

yubiuser:
https://github.com/actions/stale/issues/783

yubiuser:
To support my statement from above: currently, we have two issues with actions, `stale` and `editorconfig-checker`. If we would use explicit versions, one would not have occurred and the other one could mitigated by simply revert to the previous action version.

PromoFaux:
You might just have convinced me. I notice you have set everything to `x.0.0`, is that to test the dependabot action? Might be safer to set it off on a test repo first of all and then just pin these to whatever version we want to (rather than deal with a million different dependabot PRs after/if this is merged)

PromoFaux (dismissed):

yubiuser:
Needs re-approval after force-push to amend pgp signature.

PromoFaux (approved):

PR #4852, opened 2022-08-01
Only run the full test suite if required files are touched.

PromoFaux:

Only run full test suite if absolutely necessary

---

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,6 +8,30 @@ permissions:
   contents: read
 
 jobs:
+  # JOB to run change detection
+  changes:
+    runs-on: ubuntu-latest
+    # Set job outputs to values from filter step
+    outputs:
+      requires-distro-test: ${{ steps.filter.outputs.requires-distro-test }}
+    steps:
+    -
+        name: Checkout Repo
+        uses: actions/checkout@v3
+    -
+        name: Check which files have been touched
+        uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            requires-distro-test:
+              - 'advanced/**'
+              - 'automated install/**'
+              - 'manpages/**'
+              - 'test/**'
+              - 'gravity.sh'
+              - 'pihole'
+
   smoke-tests:
     if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
@@ -17,6 +41,7 @@ jobs:
       uses: actions/checkout@v3.0.2
     -
       name: Check scripts in repository are executable
+      if: ${{ needs.changes.outputs.requires-distro-test == 'true' }}
       run: |
         IFS=$'\n';
         for f in $(find . -name '*.sh'); do if [[ ! -x $f ]]; then echo "$f is not executable" && FAIL=1; fi ;done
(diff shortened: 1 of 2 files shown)


rdwebdesign (approved):

rdwebdesign (approved):

.github/workflows/test.yml
```diff
@@ -35,11 +61,10 @@ jobs:
       name: Run editorconfig-checker
       run: editorconfig-checker
 
-
   distro-test:
-    if: github.event.pull_request.draft == false
+    needs: [ smoke-tests, changes ]
+    if: ${{ github.event.pull_request.draft == false }}
```
yubiuser, line 66: Why do you not add the check of `needs.changes.outputs.requires-distro-test == 'true'` here instead of adding it to each single step?

PromoFaux: This is what I did originally - but then the whole job is skipped and therefore do not count towards the `Required` checks for PR to be merged (without admin crimes) Though I agree it is not 100% ideal in it's current form. I'll stick it to draft and give it some more thought.

yubiuser: Strange, it should have worked, as

>If a job in a workflow is skipped due to a conditional, it will report its status as "Success".

You would only skip the job and not the whole workflow.

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks

PromoFaux: Well, when I had it like that before - RD approved the PR and it still was expecting required checks. Which didn't seem right. I might try it on a fork later - I'm doing `Real Work™` right now

yubiuser: How about:

```
  distro-test:
    needs: [ smoke-tests, changes ]
    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
      matrix:
        distro: [debian_10, debian_11, ubuntu_18, ubuntu_20, ubuntu_22, centos_8, fedora_34]
    env:
      DISTRO: ${{matrix.distro}}
    steps:
    -
      name: Passed with nothing to do
      if: ${{ needs.changes.outputs.requires-distro-test == 'false' }}
      run: |
       echo "Nothing to test!"
       exit 0
    -
      name: Checkout repository
      uses: actions/checkout@v3.0.2
    -
      name: Set up Python 3.8
      uses: actions/setup-python@v4.1.0
      with:
        python-version: 3.8
    -
      name: Install dependencies
      run: pip install -r test/requirements.txt
    -
      name: Test with tox
      run: tox -c test/tox.${DISTRO}.ini

```

PromoFaux: That will just exit from that step, not the job  :) I'm yet to find a way to exit a job early with success without adding an `if:` section to the following steps

yubiuser: Dammit! I see this is a more general issue: https://github.com/actions/runner/issues/662

yubiuser: You could put back `needs.changes.outputs.requires-distro-test == 'true` to the whole job, create a subsequent job that does nothing else than checking if `distro-test` did not fail (hence, skipped or successful) and make this new `distro-test-checker` a requirement.

PromoFaux:
Closing for the time being

PR #4862, opened 2022-08-06
Add security-related headers to Pi-hole web interface

DL6ER:

**Please make sure you**

---

Add security-related headers to Pi-hole web interface. Justification for the additions:
- [`Content-Security-Policy`](https://content-security-policy.com/) tells the browser where resources are allowed to be
  loaded and if it’s allowed to parse/run inline styles or Javascript. This is
  important because it prevents content injection attacks, such as Cross Site
  Scripting (XSS).
  - We set the `default-src` directive (defines the default policy for fetching
    resources such as JavaScript, Images, CSS, Fonts, AJAX requests, Frames,
    HTML5 Media) to `self` which allows loading resources from the same origin
    (same scheme, host and port) only.
  - We add `'unsafe-inline'` inline source elements such as style attribute,
    onclick, or script tag bodies.
- `X-XSS-Protection` sets the configuration for the cross-site scripting filters
  built into most browsers. This is important because it tells the browser to
  block the response if a malicious script has been inserted from a user input.
- `X-Content-Type-Options` stops a browser from trying to MIME-sniff the content
  type and forces it to stick with the declared content-type. This is important
  because the browser will only load external resources if their content-type
  matches what is expected, and not malicious hidden code.
- `X-Permitted-Cross-Domain-Policies` is an XML document that grants a web client,
  such as Adobe Flash Player or Adobe Acrobat (though not necessarily limited to
  these), permission to handle data across domains.
- `Referrer-Policy` allows control/restriction of the amount of information present
  in the referral header for links away from your page - the URL path or even if the
  header is sent at all.

Modify the installed `lighttpd` configuration files.

None

---

---
- [X] I have read the above and my PR is ready for review. _Check this box to confirm_

### DIFF
--- a/advanced/lighttpd.conf.debian
+++ b/advanced/lighttpd.conf.debian
@@ -78,10 +78,21 @@ include_shell "find /etc/lighttpd/conf-enabled -name '*.conf' -a ! -name 'letsen
 
 # If the URL starts with /admin, it is the Web interface
 $HTTP["url"] =~ "^/admin/" {
-    # Create a response header for debugging using curl -I
+    # X-Pi-hole is a response header for debugging using curl -I
+    # X-Frame-Options prevents clickjacking attacks and helps ensure your content is not embedded into other sites via < frame >, < iframe > or < object >.
+    # X-XSS-Protection sets the configuration for the cross-site scripting filters built into most browsers. This is important because it tells the browser to block the response if a malicious script has been inserted from a user input.
+    # X-Content-Type-Options stops a browser from trying to MIME-sniff the content type and forces it to stick with the declared content-type. This is important because the browser will only load external resources if their content-type matches what is expected, and not malicious hidden code.
+    # Content-Security-Policy tells the browser where resources are allowed to be loaded and if it’s allowed to parse/run inline styles or Javascript. This is important because it prevents content injection attacks, such as Cross Site Scripting (XSS).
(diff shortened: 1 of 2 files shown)


rdwebdesign:
Do we still need this?

https://github.com/pi-hole/AdminLTE/blob/1714b082c3a233ea3a0c344b50352c59818676fc/scripts/pi-hole/php/header.php#L171

DL6ER:
@rdwebdesign Thanks for the pointer, I'll remove it in a follow-up PR.

> The meta support is handy when you can't set a HTTP response header, but in most cases using a HTTP response header is a stronger approach.
>
> [source](https://content-security-policy.com/examples/meta/)

yubiuser (approved):

bstnbuck:
Hello together!
Thanks for setting the security headers by default.
However, the `X-XSS-Protection` header is deprecated and can have the contrary effect and leak XSS attacks, among other things. Since most browsers have the XSS filter disabled and it only works on old versions, it is recommended to set it to "0" or remove it completely.
My suggestion is to set it off. -> `"X-XSS-Protection" => "0"`
In all new types of browsers, `CSP` is functional by default and also serves as protection against XSS

See https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Headers_Cheat_Sheet.html and https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection for more information

>As an example, Bing and Cloudflare have not set the headers at all, Google has turned this off.

PR #4863, opened 2022-08-06
Fix git indention

yubiuser:

Fixes the indention shown when switching branches and the branch is "Already up to date"

Bildschirmfoto zu 2022-08-06 19-01-06

The check for the output string was checking for the wrong string: it checked for "up-to-date", while `git` outputs `Already up to date` (without hyphens) since https://github.com/git/git/commit/7560f547e614244fe1d4648598d4facf7ed33a56

This PR fixes this by removing the check completely and always indent the output.

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -2126,7 +2126,7 @@ checkout_pull_branch() {
 
     git_pull=$(git pull --no-rebase || return 1)
 
-    if [[ "$git_pull" == *"up-to-date"* ]]; then
+    if [[ "$git_pull" == *"up to date"* ]]; then
         printf "  %b %s\\n" "${INFO}" "${git_pull}"
     else
         printf "%s\\n" "$git_pull"
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -2126,11 +2126,7 @@ checkout_pull_branch() {
 
     git_pull=$(git pull --no-rebase || return 1)
 
-    if [[ "$git_pull" == *"up to date"* ]]; then
-        printf "  %b %s\\n" "${INFO}" "${git_pull}"
-    else
-        printf "%s\\n" "$git_pull"
-    fi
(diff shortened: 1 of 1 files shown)


DL6ER (changes requested):
I was wondering if we should simply check for both variants. However, following the commit you linked, I've seen this string was changed as of `git` tag [`v2.15.0-rc0`](https://github.com/git/git/releases/tag/v2.15.0-rc0).
- `Debian Stretch` shipped with `2.11.0` (`up-to-date`) but is gone now :heavy_check_mark: 
- `Debian Buster` ships `2.20.1-2` (`up to date`) :heavy_check_mark: 
- `Ubuntu 18.04` has `2.17` (at least when users installed all regular updates) :heavy_check_mark: 
- RHEL/CentOS 7 is starting to feel a somewhat dated, but it reaches the end of its support lifetime only end of June 2024. It still installs `1.8.3.1` (not even a `v2.x` here). You can only get a recent `git` version by adding external repositories. At first glance, [it doesn't seem like EPEL would add in more recent `git` versions](https://serverfault.com/q/709433) (see the question's text) :heavy_multiplication_x:

automated install/basic-install.sh
```diff
@@ -2126,7 +2126,7 @@ checkout_pull_branch() {
 
     git_pull=$(git pull --no-rebase || return 1)
 
-    if [[ "$git_pull" == *"up-to-date"* ]]; then
+    if [[ "$git_pull" == *"up to date"* ]]; then
```
DL6ER, line 2129:
```suggestion
    if [[ "$git_pull" == *"up to date"* || "$git_pull" == *"up-to-date"* ]]; then
```

yubiuser: An other solution would be to remove the up to date check completly and always indent.

`printf "  %b %s\\n" "${INFO}" "${git_pull}"`

DL6ER (approved):

PR #4875, opened 2022-08-22
Store versions as key/value pairs rather than space delimeted values

PromoFaux:

Store Pi-hole component versions as key/value pairs in a single file. Storing as space delimited may save (a small amount of) space, but it can get complicated to ensure you are using the correct index when reading it out to an array.

This makes things a lot clearer.

AdminLTE will need to be updated to reflect this change too (@rdwebdesign said he would look at that)

This additionally adds the value of `GITHUB_DOCKER_TAG` into the mix so that updates to the docker container are visible, too.

---

---

### DIFF
--- a/advanced/Scripts/updatecheck.sh
+++ b/advanced/Scripts/updatecheck.sh
@@ -41,33 +41,19 @@ function get_local_version() {
 # shellcheck disable=SC1091
 . /etc/pihole/setupVars.conf
 
-utilsfile="/opt/pihole/utils.sh"
-source "${utilsfile}"
+# Source the utils file
+# shellcheck disable=SC1091
+. /opt/pihole/utils.sh
 
 # Remove the below three legacy files if they exist
-GITHUB_VERSION_FILE="/etc/pihole/GitHubVersions"
-LOCAL_BRANCH_FILE="/etc/pihole/localbranches"
-LOCAL_VERSION_FILE="/etc/pihole/localversions"
-
-if [ -f "${GITHUB_VERSION_FILE}" ]; then
-    rm "${GITHUB_VERSION_FILE}"
-fi
-
-if [ -f "${LOCAL_BRANCH_FILE}" ]; then
-    rm "${LOCAL_BRANCH_FILE}"
-fi
-
-if [ -f "${LOCAL_VERSION_FILE}" ]; then
-    rm "${LOCAL_VERSION_FILE}"
-fi
+rm -f "/etc/pihole/GitHubVersions"
+rm -f "/etc/pihole/localbranches"
+rm -f "/etc/pihole/localversions"
 
 # Create new versions file if it does not exist
 VERSION_FILE="/etc/pihole/versions"
-
-if [ ! -f "${VERSION_FILE}" ]; then
-    touch "${VERSION_FILE}"
-    chmod 644 "${VERSION_FILE}"
-fi
+touch "${VERSION_FILE}"
+chmod 644 "${VERSION_FILE}"
 
 if [[ "$2" == "remote" ]]; then
 
--- a/advanced/Scripts/updatecheck.sh
+++ b/advanced/Scripts/updatecheck.sh
@@ -41,54 +41,78 @@ function get_local_version() {
 # shellcheck disable=SC1091
 . /etc/pihole/setupVars.conf
 
+utilsfile="/opt/pihole/utils.sh"
+source "${utilsfile}"
+
+# Remove the below three legacy files if they exist
+GITHUB_VERSION_FILE="/etc/pihole/GitHubVersions"
+LOCAL_BRANCH_FILE="/etc/pihole/localbranches"
+LOCAL_VERSION_FILE="/etc/pihole/localversions"
+
+if [ -f "${GITHUB_VERSION_FILE}" ]; then
+    rm "${GITHUB_VERSION_FILE}"
+fi
+
+if [ -f "${LOCAL_BRANCH_FILE}" ]; then
+    rm "${LOCAL_BRANCH_FILE}"
+fi
+
+if [ -f "${LOCAL_VERSION_FILE}" ]; then
+    rm "${LOCAL_VERSION_FILE}"
+fi
+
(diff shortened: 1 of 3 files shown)


advanced/Scripts/updatecheck.sh
```diff
@@ -41,54 +41,78 @@ function get_local_version() {
 # shellcheck disable=SC1091
 . /etc/pihole/setupVars.conf
 
+utilsfile="/opt/pihole/utils.sh"
+source "${utilsfile}"
```
rdwebdesign, line 45: stickerCI

dschaper, line 45:
```suggestion
source "${utilsfile:-/opt/pihole/utils.sh}"
```

dschaper: I guess unless `utilsfile` is used elsewhere in the script.

PromoFaux: Heh, I actually just took this from another place where we source it (webpage.sh) but it certainly can stand to be changed

rdwebdesign, line 45: ShellCheck is still complaining.
Suggestion:

```suggestion
utilsfile="/opt/pihole/utils.sh"
# shellcheck disable=SC1091
. "${utilsfile}"
```

advanced/Scripts/version.sh
```diff
@@ -90,16 +90,17 @@ getRemoteVersion(){
...
     local cachedVersions
     local arrCache
-    cachedVersions="/etc/pihole/GitHubVersions"
+    cachedVersions="/etc/pihole/versions"
 
     #If the above file exists, then we can read from that. Prevents overuse of GitHub API
     if [[ -f "$cachedVersions" ]]; then
-        IFS=' ' read -r -a arrCache < "$cachedVersions"
+
+        source "$cachedVersions"
```
rdwebdesign, line 98:
```suggestion
        # shellcheck disable=SC1091
        . "$cachedVersions"
```

advanced/Scripts/updatecheck.sh
```diff
@@ -41,54 +41,78 @@ function get_local_version() {
...
+    rm "${GITHUB_VERSION_FILE}"
+fi
+
+if [ -f "${LOCAL_BRANCH_FILE}" ]; then
+    rm "${LOCAL_BRANCH_FILE}"
+fi
+
+if [ -f "${LOCAL_VERSION_FILE}" ]; then
+    rm "${LOCAL_VERSION_FILE}"
+fi
```
dschaper, line 62: You can use `rm -f <filename>` instead and the file will be deleted if found and`rm` will not error out if the file doesn't exist.

I'd also just directly `rm` the file instead of the variable since the file locations are in a known path and couldn't be located anywhere else.

advanced/Scripts/updatecheck.sh
```diff
@@ -41,54 +41,78 @@ function get_local_version() {
...
+
+# Create new versions file if it does not exist
+VERSION_FILE="/etc/pihole/versions"
+
+if [ ! -f "${VERSION_FILE}" ]; then
+    touch "${VERSION_FILE}"
+    chmod 644 "${VERSION_FILE}"
```
dschaper, line 69: It should be safe to `touch` and `chmod` the file always. `touch` will just modify the timestamps if the file already exists (it won't change the content). And `chmod` is safe to do anytime.

advanced/Scripts/updatecheck.sh
```diff
@@ -41,54 +41,78 @@ function get_local_version() {
...
+        addOrEditKeyValPair "${VERSION_FILE}" "GITHUB_WEB_VERSION" "${GITHUB_WEB_VERSION}"
     fi
 
     GITHUB_FTL_VERSION="$(json_extract tag_name "$(curl -s 'https://api.github.com/repos/pi-hole/FTL/releases/latest' 2> /dev/null)")"
-    echo -n " ${GITHUB_FTL_VERSION}" >> "${GITHUB_VERSION_FILE}"
+    addOrEditKeyValPair "${VERSION_FILE}" "GITHUB_FTL_VERSION" "${GITHUB_FTL_VERSION}"
 
-else
+    if [[ "${PIHOLE_DOCKER_TAG}" ]]; then
+        GITHUB_DOCKER_VERSION="$(json_extract tag_name "$(curl -s 'https://api.github.com/repos/pi-hole/docker-pi-hole/releases/latest' 2> /dev/null)")"
```
dschaper, line 90: Should we add `jq` as a dependency? I think we could utilize it in a few places.

yubiuser: Good idea. I can make a PR for that. `jq` will be a requirement for `PADD v6` anyway.

yubiuser: I'll PR this after this PR has been merged. I can then remove the `json_extract` in a few places.

yubiuser (commented):
Should we add "LOCAL_" to the local versions/branches as we do for the remote once with "GITHUB_"?

PromoFaux:
I did wonder that, but I just used the variable names that were already in use. I'm not against the idea, though

rdwebdesign:
Personally, I don't think we need to add a prefix to the "local" versions/branches variables.

advanced/Scripts/updatecheck.sh
PromoFaux: Huh, I don't know how this came to pass, I definitely added the comment to disable 1091! Nevermind... it will be there in next push

advanced/Scripts/updatecheck.sh
PromoFaux: Yeah, For sure a good idea, though I will keep this PR as is

dschaper (approved):
My requested changes all look good.

rdwebdesign (approved):

rdwebdesign:
@PromoFaux 

Maybe the update function should be changed to always run "updateCheck" and create the new file for current installs.

PR #4857, opened 2022-08-05
Exit init script with error result code if FTL fails to start

dschaper:
- **What does this PR aim to accomplish?:**

Exit init script with error result code if FTL fails to start

- **How does this PR accomplish the above?:**

Exits init script with non-zero result code if FTL fails to start.

- **What documentation changes (if any) are needed to support this PR?:**

Nothing.

---
**By submitting this pull request, I confirm the following:** 

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against. 
2. I have commented my proposed changes within the code and I have tested my changes.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

---
- [x] I have read the above and my PR is ready for review. _Check this box to confirm_

PR #4887, opened 2022-09-02
Cleanup if startup failed

yubiuser:

Cleanup files if `pihole-FTL` did not start properly after getting invoked by `pihole-FTL.service`.

With https://github.com/pi-hole/pi-hole/pull/4857 we added the exit code, however do not use ist so far. We do now.

---

---

### DIFF
--- a/advanced/Templates/pihole-FTL.service
+++ b/advanced/Templates/pihole-FTL.service
@@ -9,11 +9,10 @@
 # Description:       Enable service provided by pihole-FTL daemon
 ### END INIT INFO
 
-#source utils.sh for getFTLPIDFile(), getFTLPID ()
+#source utils.sh for getFTLPIDFile(), getFTLPID()
 PI_HOLE_SCRIPT_DIR="/opt/pihole"
 utilsfile="${PI_HOLE_SCRIPT_DIR}/utils.sh"
 . "${utilsfile}"
-
 
 is_running() {
   if [ -d "/proc/${FTL_PID}" ]; then
@@ -23,7 +22,7 @@ is_running() {
 }
 
 cleanup() {
-  rm -f /run/pihole/FTL.sock /dev/shm/FTL-* "${FTL_PID_FILE}" "${FTL_PORT_FILE}"
+    rm -f /run/pihole/FTL.sock /dev/shm/FTL-* "${FTL_PID_FILE}" "${FTL_PORT_FILE}"
 }
 
 # Start the service
@@ -114,6 +113,9 @@ status() {
 
 
 ### main logic ###
+
+# catch sudden termination
+trap 'cleanup; exit 1' INT HUP TERM ABRT
 
 # Get file paths
 FTL_PID_FILE="$(getFTLPIDFile)"
--- a/advanced/Templates/pihole-FTL.service
+++ b/advanced/Templates/pihole-FTL.service
@@ -22,7 +22,7 @@ is_running() {
 }
 
 cleanup() {
-    rm -f /run/pihole/FTL.sock /dev/shm/FTL-* "${FTL_PID_FILE}" "${FTL_PORT_FILE}"
+    rm -f /run/pihole/FTL.sock /dev/shm/FTL-* "${FTL_PID_FILE}"
 }
 
 # Start the service
@@ -65,9 +65,9 @@ start() {
       /usr/bin/pihole-FTL || ec=$?
     fi
     # Cleanup if startup failed
-    if [ "${ec}" != 0 ]; then
+    if [ -n "${ec}" ] && [ "${ec}" != 0 ]; then
         cleanup
-        exit "${ec}"
+        exit $ec
     fi
     echo
   fi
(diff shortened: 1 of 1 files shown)


advanced/Templates/pihole-FTL.service
```diff
@@ -22,6 +22,9 @@ is_running() {
   return 1
 }
 
+cleanup() {
```
dschaper, line 25: Looks good. Do you think this would be a good candidate for a `trap` approach? You have the cleanup function written so all that would need to change is 67-72. 

See https://www.shellscript.sh/trap.html

yubiuser: To you want to trap more then just `EXIT`?

dschaper: Trapping for cleanup would be helpful in any condition that isn't `return 0`. I trust your judgement on what you think should be trapped and when the cleanup function should be called.

yubiuser: I've implemented a `trap` but after writing the code I doubt if this is a good idea/implementation. We don't want to execute `cleanup` on every exit. E.g.

https://github.com/pi-hole/pi-hole/blob/d80d248ca61d6dde3688050304e5568e86db8bcc/advanced/Templates/pihole-FTL.service#L142-L143

Would delete the files if the service was called with an unknown argument. 
___

Maybe it's simpler and less error-prone if we just execute `cleanup` when we need it and don't trap `EXIT`, but only e.g. `INT`

rdwebdesign: Can you test the previous return code? If it's zero you don't execute the cleanup?

yubiuser: But we want the cleanup sometimes with zero exit code (`stop()`).

I think it's getting overly complicated...

yubiuser: I changed it to trap only `INT HUP TERM ABRT` (not EXIT) to avoid cleaning files we still need if users only want to get the help text or request the status

advanced/Templates/pihole-FTL.service
```diff
@@ -57,10 +59,15 @@ start() {
...
+      su -s /bin/sh -c "/usr/bin/pihole-FTL" pihole || ec=$?
     else
       echo "Warning: Starting pihole-FTL as root because setting capabilities is not supported on this system"
-      /usr/bin/pihole-FTL || exit $?
+      /usr/bin/pihole-FTL || ec=$?
+    fi
+    # Cleanup if startup failed
+    if [ "${ec}" != 0 ]; then
+        cleanup
+        exit "${ec}"
```
dschaper, line 70: Since you're now trapping for non-zero exit you don't need any of this code. 
`/usr/bin/pihole-FTL` is all you need.

advanced/Templates/pihole-FTL.service
```diff
@@ -57,10 +59,15 @@ start() {
     fi
 
     if setcap CAP_NET_BIND_SERVICE,CAP_NET_RAW,CAP_NET_ADMIN,CAP_SYS_NICE,CAP_IPC_LOCK,CAP_CHOWN+eip "/usr/bin/pihole-FTL"; then
-      su -s /bin/sh -c "/usr/bin/pihole-FTL" pihole || exit $?
+      su -s /bin/sh -c "/usr/bin/pihole-FTL" pihole || ec=$?
```
dschaper, line 62: You don't need the `|| ec=$?`, the non-zero exit will trigger the trap.

advanced/Templates/pihole-FTL.service
```diff
@@ -108,6 +114,9 @@ status() {
 
 ### main logic ###
 
+# catch sudden termination
+trap 'cleanup; exit 1' INT HUP TERM ABRT
```
dschaper, line 118: You can move the `exit 1` to the `cleanup` function.

advanced/Templates/pihole-FTL.service
```diff
@@ -9,19 +9,21 @@
...
 is_running() {
   if [ -d "/proc/${FTL_PID}" ]; then
     return 0
   fi
   return 1
 }
 
+cleanup() {
+    rm -f /run/pihole/FTL.sock /dev/shm/FTL-* "${FTL_PID_FILE}" "${FTL_PORT_FILE}"
+}
```
dschaper, line 26:
```suggestion
cleanup() {
    rm -f /run/pihole/FTL.sock /dev/shm/FTL-* "${FTL_PID_FILE}" "${FTL_PORT_FILE}"
    exit 1
}
```

dschaper (changes requested):

advanced/Templates/pihole-FTL.service
yubiuser: We can't, because we also use the cleanup function when we just stop the  service. See L99.

yubiuser:
Despite all test passed, it wasn't working on real systems. Reason was: it's not enough to check if the exit code is different then `zero`, it also must be set at all...

yubiuser:
Superseded by https://github.com/pi-hole/pi-hole/pull/5052 after resolving merge conflicts introduced by https://github.com/pi-hole/pi-hole/pull/4924

PR #4926, opened 2022-09-19
Update python tests and add black code formatter action

yubiuser:

1) Replace `whitelist_externals` as it is deprecated (https://tox.wiki/en/latest/config.html#conf-allowlist_externals)
2) Remove `pytest-cov` as we did not use it.
3) Add github action to tests `black` python formatter

---

---


.github/workflows/test.yml
```diff
@@ -34,7 +34,11 @@ jobs:
     -
       name: Run editorconfig-checker
       run: editorconfig-checker
-
+    -
+      name: Run black python code formatter
```
PromoFaux, line 38: This action is just doing checks, right? It will flag any issues and present them to the PRing user? (note to self - install black locally) Can we change this text to something like `Check python code formatting with black` ?

https://black.readthedocs.io/en/stable/integrations/github_actions.html Suggests default options passed are `--check --diff`

PromoFaux (changes requested):

.github/workflows/test.yml
rdwebdesign: Yes. The current config is using `--check --diff`:
image

Note: _if you want to change the options, I suggest using `--check --diff --color`._

PromoFaux: Ha. I didn't even think to look at the last check run 🙈

```suggestion
      name: Check python code formatting with black
```

PromoFaux (approved):
black changes are expected (though odd in a couple of places, I can see why they've ended up like that)

happy with the yml formatting changes - they are a good compromise between what we had, and what a linter/formatter does to it

PR #4941, opened 2022-09-24
Remove refrences to the now removed pihole-FTL.port file

PromoFaux:
Will update this description later - I think it needs more eyes on it, so creating as draft


PromoFaux:
I now see the `no_port` branch...

PR #4945, opened 2022-09-26
Remove pihole-FTL.port file

yubiuser:

Removes code related to the `pihole-FTL.port` file. Accompanies https://github.com/pi-hole/FTL/pull/1445

---

---
- [ ] I have read the above and my PR is ready for review. _Check this box to confirm_


advanced/Scripts/utils.sh
```diff
@@ -70,47 +70,29 @@ removeKey() {
...
+        expr "${ftl_api_port}" : "[^[:digit:]]" > /dev/null && unset ftl_api_port
+    else
+        # if there is no pihole-FTL.conf, use the default port
+        ftl_api_port="${DEFAULT_FTL_PORT}"
     fi
 
-    # echo the port found in the portfile or default to the default port
-    echo "${ftl_api_port:=$DEFAULT_FTL_PORT}"
+    # If the ftl_api_port contained malicious stuff, substitute with -1
+    ftl_api_port=${ftl_api_port:=-1}
```
rdwebdesign, line 94: Just a question:
Why `-1`? Why not `${DEFAULT_FTL_PORT}`?

yubiuser: To signal that something went wrong. If we echo `${DEFAULT_FTL_PORT}`, the calling function will not know that something went wrong.

yubiuser:
Merge conflicts resolved and rebased on development

advanced/Scripts/utils.sh
rdwebdesign: Why don't we show an error message, but use the default value to avoid an error?

yubiuser: `ftl_api_port=${ftl_api_port:=-1}` this syntax means: if `ftl_api_port` is set, use it's value, otherwise set it to `-1`

> ${parameter:=[word]}
Assign Default Values. If parameter is unset or null, the expansion of word (or an empty string if word is omitted) shall be assigned to parameter.

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html
___

I would not error here, but we could add a check for each calling function for `-1`

rdwebdesign: I know what `:=` means. My point is: `-1` will fail in other places.

> I would not error here, but we could add a check for each calling function for -1

Why not here? I think this function should return the port.

Other tests are returning the default value, in case of failure:
- If no `.conf` file => `ftl_api_port="${DEFAULT_FTL_PORT}"` **(OK)**;
- If there's a file, but no `FTLPORT` set => `ftl_api_port="${DEFAULT_FTL_PORT}"` **(OK)**;

Why not use the same logic if the value is invalid:
- If there's a file and `FTLPORT` set as an invalid value => `ftl_api_port=-1` (NOT OK and **pihole won't be able to access the API**).

rdwebdesign: I think a better question is:
**What FTL does when an invalid port is set?**

This code should use a similar logic.

yubiuser: It defaults to `4711`....
https://github.com/pi-hole/FTL/blob/4c8abe63a1c4605c58171b18d92b20c1499ff8a7/src/config.c#L235-L244

yubiuser: Resolved by https://github.com/pi-hole/pi-hole/pull/4945/commits/276c480f5001465d994dacf6e30d1e1c2d0a3b0b

PromoFaux (approved):

PR #4956, opened 2022-10-01
Fix default DHCPv4 leasetime value

PromoFaux:

Reported here: https://github.com/pi-hole/pi-hole/issues/4957

Originally introduced by myself back in 2017 😬https://github.com/pi-hole/pi-hole/pull/1646

Discussed indirectly here: https://github.com/pi-hole/pi-hole/issues/4142#issuecomment-827066762

This fixes it by ensuring that the `leasetime` local variable is set to a valid value before it is applied to the DHCP config. It is still stored in `setupVars.conf` as an integer value

---

---

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -395,11 +395,6 @@ ProcessDHCPSettings() {
         elif [[ "${DHCP_LEASETIME}" == "" ]]; then
             leasetime="24h"
             addOrEditKeyValPair "${setupVars}" "DHCP_LEASETIME" "24"
-        elif [[ "${DHCP_LEASETIME}" == "24h" ]]; then
-            #Installation is affected by known bug, introduced in a previous version.
-            #This will automatically clean up setupVars.conf and remove the unnecessary "h"
-            leasetime="24h"
-            addOrEditKeyValPair "${setupVars}" "DHCP_LEASETIME" "24"
         else
             leasetime="${DHCP_LEASETIME}h"
         fi
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -393,13 +393,13 @@ ProcessDHCPSettings() {
         if [[ "${DHCP_LEASETIME}" == "0" ]]; then
             leasetime="infinite"
         elif [[ "${DHCP_LEASETIME}" == "" ]]; then
-            leasetime="24"
-            addOrEditKeyValPair "${setupVars}" "DHCP_LEASETIME" "${leasetime}"
+            leasetime="24h"
+            addOrEditKeyValPair "${setupVars}" "DHCP_LEASETIME" "24"
         elif [[ "${DHCP_LEASETIME}" == "24h" ]]; then
             #Installation is affected by known bug, introduced in a previous version.
             #This will automatically clean up setupVars.conf and remove the unnecessary "h"
-            leasetime="24"
-            addOrEditKeyValPair "${setupVars}" "DHCP_LEASETIME" "${leasetime}"
+            leasetime="24h"
+            addOrEditKeyValPair "${setupVars}" "DHCP_LEASETIME" "24"
         else
             leasetime="${DHCP_LEASETIME}h"
         fi


yubiuser (commented):
I think the whole part 

```
        elif [[ "${DHCP_LEASETIME}" == "24h" ]]; then
            #Installation is affected by known bug, introduced in a previous version.
            #This will automatically clean up setupVars.conf and remove the unnecessary "h"
            leasetime="24h"
            addOrEditKeyValPair "${setupVars}" "DHCP_LEASETIME" "24"
```

can be removed - this was a special case handling introduced **5 years** ago (https://github.com/pi-hole/pi-hole/pull/1646).
____
Why is there no `addOrEditKeyValPair` in case `"${DHCP_LEASETIME}" == "0"`?

advanced/Scripts/webpage.sh
```diff
@@ -393,13 +393,13 @@ ProcessDHCPSettings() {
...
-            addOrEditKeyValPair "${setupVars}" "DHCP_LEASETIME" "${leasetime}"
+            leasetime="24h"
+            addOrEditKeyValPair "${setupVars}" "DHCP_LEASETIME" "24"
         elif [[ "${DHCP_LEASETIME}" == "24h" ]]; then
             #Installation is affected by known bug, introduced in a previous version.
             #This will automatically clean up setupVars.conf and remove the unnecessary "h"
-            leasetime="24"
-            addOrEditKeyValPair "${setupVars}" "DHCP_LEASETIME" "${leasetime}"
+            leasetime="24h"
+            addOrEditKeyValPair "${setupVars}" "DHCP_LEASETIME" "24"
```
dschaper, line 402: This entire section can be removed. There was one specific case in version 3 that had the possibility of `setupVars.conf` containing `DCHP_LEASETIME=24h` because we used to store that config with the time character. We originally stored it as seconds but people complained that math was hard and who knew what `86400` meant. 

We automatically added `h` to everything so in that specific case the `dnsmasq` configuration file ended up with `24hh` as the `leasetime`. I really don't think we need this shim anymore, no one should have `DCHP_LEASETIME=24h` in `setupVars.conf` now.

Keep it simple. If the var is `0` then it's infinite. If it's empty then `24h`, else it's the value of the var with `h` appended. You can not enter anything other than an integer via the web interface. 

The only other case I can see that should be accounted for is when the var contains anything other than an integer. In that case I think we use the default `24h`.

dschaper (changes requested):

advanced/Scripts/webpage.sh
PromoFaux: Funilly enough, I was going to remove it - and then I backed out at the last minute. Should have gone with my gut and saved you the typing ;)

PromoFaux:
> Why is there no `addOrEditKeyValPair` in case `"${DHCP_LEASETIME}" == "0"`?

Because we don't want to add anything to `setupVars.conf` here... it already contains `0` in this case, and so doesn't need to be changed :)

advanced/Scripts/webpage.sh
```diff
@@ -393,13 +393,8 @@ ProcessDHCPSettings() {
...
-            addOrEditKeyValPair "${setupVars}" "DHCP_LEASETIME" "${leasetime}"
-        elif [[ "${DHCP_LEASETIME}" == "24h" ]]; then
-            #Installation is affected by known bug, introduced in a previous version.
-            #This will automatically clean up setupVars.conf and remove the unnecessary "h"
-            leasetime="24"
-            addOrEditKeyValPair "${setupVars}" "DHCP_LEASETIME" "${leasetime}"
+            leasetime="24h"
+            addOrEditKeyValPair "${setupVars}" "DHCP_LEASETIME" "24"
         else
             leasetime="${DHCP_LEASETIME}h"
```
rdwebdesign, line 399: > The only other case I can see that should be accounted for is when the var contains anything other than an integer. In that case I think we use the default `24h`.

To filter non-integer values, we could use here:


```suggestion
            leasetime="${DHCP_LEASETIME//[^0-9]/}h"
```

rdwebdesign (dismissed):

advanced/Scripts/webpage.sh
dschaper: Is that a POSIX solution?

rdwebdesign: To be honest, I have no idea.

rdwebdesign: I think this is just a variable expansion pattern, but I'm not sure if this is a POSIX solution.

yubiuser: It's bash only: https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion

Not listed in https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html

PromoFaux:
Rebased on development to satisfy new fed version checks

advanced/Scripts/webpage.sh
rdwebdesign: OK.
We can use `sed`, but I'm sure there is a better choice than this:

```suggestion
            leasetime=$(echo "$DHCP_LEASETIME" | sed 's/[^0-9]*//g; s/$/h/')
```

dschaper: I think this PR covers what it says on the tin so we can look at changes in a separate PR.

dschaper (approved):

advanced/Scripts/webpage.sh
PromoFaux: TBF, the rest of the function is not POSIX compatible - but that's for another time.

I think that there is probably little point in sanitising the value at this point. The web interface will only put integers into this setting - if someone has entered a bad value, then it should fall over 😉

yubiuser: `leasetime=$(echo "$DHCP_LEASETIME" |  tr -cd [:digit:])`

rdwebdesign (approved):

Issue #4759, opened 2022-06-18
Version numbers not displayed correctly anymore in Chronometer


Jiab77:
### Versions

- Pi-hole:  5.10
- AdminLTE:  5.12
- FTL:  5.15

### Platform

- OS and version:  Ubuntu Focal 20.04

- Platform:  Raspeberry Pi

### Expected behavior

When running the `chronometer` feature at boot, I'm expecting to see at the top right:

```
Core: v5.10
Web: v512
FTL: v5.15
```

### Actual behavior / bug

But what I'm seeing currently is:

```
Core: ERROR
Web: ERROR
FTL: v5.15
```

### Steps to reproduce

1. Check line [#L230](https://github.com/pi-hole/pi-hole/blob/master/advanced/Scripts/chronometer.sh#L230) in [chronometer.sh](https://github.com/pi-hole/pi-hole/blob/master/advanced/Scripts/chronometer.sh)
2. Run the used command: `pihole -v -c 2> /dev/null`
3. See error

But if you run the same command with `sudo`, it will work as expected and return version numbers correctly.

### Screenshots

* `pi-hole` version command tests:

  ![image](https://user-images.githubusercontent.com/9881407/174452474-1fbde6f2-d741-4603-8cde-c8ed00c42351.png)

### Debug and solution

I then decided to dig a little bit further in the code and found that the version returned by `pihole -v -c` was in fact managed by [version.sh](https://github.com/pi-hole/pi-hole/blob/master/advanced/Scripts/version.sh) where the most interesting line was [#L32](https://github.com/pi-hole/pi-hole/blob/master/advanced/Scripts/version.sh#L32). It executes the command `git describe --tags --always` to get the version numbers.

But when I've tried to run it from the `/etc/.pihole` folder (variable `COREGITDIR`), I got the following error:

```
$ cd /etc/.pihole/
$ git describe --tags --always
fatal: unsafe repository ('/etc/.pihole' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /etc/.pihole
```

Which is true as the `/etc/.pihole` folder is owned by __root__.

I then did the same test for the LTE admin folder (variable `WEBGITDIR`) and got the same error:

```
$ cd /var/www/html/admin/
$ git describe --tags --always
fatal: unsafe repository ('/var/www/html/admin' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /var/www/html/admin
```

[...]

PR #4939, opened 2022-09-24
Remove cronjob that checks local versions every 10 minutes - only check when required

jpgpi250:
Signed-off-by: jpgpi250 <jpgpi250@gmail.com>

## Thank you for your contribution to the Pi-hole Community! 

Please read the comments below to help us consider your Pull Request.

We are all volunteers and completing the process outlined will help us review your commits quicker.

**Please make sure you**
 1. Base your code and PRs against the repositories developmental branch.
 2. [Sign Off](https://docs.pi-hole.net/guides/github/how-to-signoff/) all commits as we enforce the [DCO](https://docs.pi-hole.net/guides/github/dco/) for all contributions

---
- **What does this PR aim to accomplish?:**

see https://discourse.pi-hole.net/t/pi-hole-stoppen-zwecks-backup/57989/10

- eliminate updatechecker local from the pihole cronjob
- add updatecheck to piholeDebug.sh
- check if api.github.com returns a valid IP address

- **How does this PR accomplish the above?:**
changed:
- pihole.cron
- piholeCheckout.sh
- piholeDebug.sh
- updatecheck.sh

- **What documentation changes (if any) are needed to support this PR?:**

none

**By submitting this pull request, I confirm the following:** 

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against. 
2. I have commented my proposed changes within the code and I have tested my changes.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

---
- [x] I have read the above and my PR is ready for review. _Check this box to confirm_

PR #4663, opened 2022-03-30
Only use sudo power if needed

yubiuser:
- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md), as well as this entire template.
- [x] I have made only one major change in my proposed changes.
- [x] I have commented my proposed changes within the code.
- [x] I have tested my proposed changes, and have included unit tests where possible.
- [x] I am willing to help maintain this change if there are issues with it later.
- [x] I give this submission freely and claim no ownership.
- [x] It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
- [x] I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

Please make sure you [Sign Off](https://docs.pi-hole.net/guides/github/how-to-signoff/) all commits. Pi-hole enforces the [DCO](https://docs.pi-hole.net/guides/github/contributing/).

---
**What does this PR aim to accomplish?:**
So far, the `pihole` command needed `sudo` power for all kind of function except of `help`. This PR allows to use functions, that do not need root privileges without the `sudo` check.

These functions are
```
  "-h" | "help" | "--help"     
  "-v" | "version"             
  "-c" | "chronometer"          
  "-q" | "query"                
  "status"                     
  "-t" | "tail"                
  "tricorder"                  
```

Fixes https://github.com/pi-hole/pi-hole/issues/2927

**How does this PR accomplish the above?:**
Extend the `case"${1}"` for the above mentioned functions, then check for sudo power and later check for all other functions.

PR #4959, opened 2022-10-06
Simplify versions.sh

yubiuser:

Simplifies `version.sh`. Rationale: PR https://github.com/pi-hole/pi-hole/pull/4663 removed the need for `sudo` powers if `pihole -v` is called. However, this leads to issues with newer `git` versions (seen in https://github.com/pi-hole/pi-hole/issues/4759) as `version.sh` uses some `git` commands to get the local version and branches. Without `sudo`, `git` complains now about `fatal: unsafe repository ('/var/www/html/admin' is owned by someone else)`. 
Since PR https://github.com/pi-hole/pi-hole/pull/4875 we store local and remote version (and branch) info in `/etc/pihole/versions`. This PR moves the remaining functions from `version.sh` to `updatechecker.sh`. The latter is always called with `sudo` powers and therefore does not run into the described `git` issue. The remaining `version.sh` does nothing else than to print the requested output of `/etc/pihole/versions`.

Fixes: https://github.com/pi-hole/pi-hole/issues/4759

1. Add a new function `get_remote_version()` to `updatechecker.sh`
2. Removes the distinction between updating the remote and local info from `updatechecker`. That way, we can be sure, that the local branch is updated before we get the remote hash of the the current branch. If you look at the current code, there was only one occurrence where `local` and `remote` were not called consecutively: the daily cron job. We just agreed to remove the local update from the cron job (https://github.com/pi-hole/pi-hole/pull/4939), but this was about updating every 10 minutes. I think updating once every 24h should be fine. We already ensure that `updatechecker` is called at every relevant place: after a fresh install, after each `pihole -up` and after each `checkout`.
3. Move the `hash` functions to `updatechecker.sh`
4. Let `pihole -v` gather everything from `/etc/pihole/versions`

---

5. I have commented my proposed changes within the code and I have tested my changes.
6. I am willing to help maintain this change if there are issues with it later.
7. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
8. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

---

### DIFF
--- a/advanced/Scripts/updatecheck.sh
+++ b/advanced/Scripts/updatecheck.sh
@@ -20,6 +20,10 @@ function get_local_version() {
     git describe --long --dirty --tags 2> /dev/null || return 1
 }
 
+function get_remote_version() {
+    curl -s "https://api.github.com/repos/pi-hole/${1}/releases/latest" 2> /dev/null | jq --raw-output .tag_name || return 1
+}
+
 # Source the setupvars config file
 # shellcheck disable=SC1091
 . /etc/pihole/setupVars.conf
@@ -52,19 +56,19 @@ if [[ "$2" == "remote" ]]; then
         sleep 30
     fi
 
-    GITHUB_CORE_VERSION="$(curl -s 'https://api.github.com/repos/pi-hole/pi-hole/releases/latest' 2> /dev/null | jq --raw-output .tag_name)"
+    GITHUB_CORE_VERSION="$(get_remote_version pi-hole)"
     addOrEditKeyValPair "${VERSION_FILE}" "GITHUB_CORE_VERSION" "${GITHUB_CORE_VERSION}"
 
     if [[ "${INSTALL_WEB_INTERFACE}" == true ]]; then
-        GITHUB_WEB_VERSION="$(curl -s 'https://api.github.com/repos/pi-hole/AdminLTE/releases/latest' 2> /dev/null | jq --raw-output .tag_name)"
+        GITHUB_WEB_VERSION="$(get_remote_version AdminLTE)"
         addOrEditKeyValPair "${VERSION_FILE}" "GITHUB_WEB_VERSION" "${GITHUB_WEB_VERSION}"
     fi
 
-    GITHUB_FTL_VERSION="$(curl -s 'https://api.github.com/repos/pi-hole/FTL/releases/latest' 2> /dev/null | jq --raw-output .tag_name)"
+    GITHUB_FTL_VERSION="$(get_remote_version FTL)"
     addOrEditKeyValPair "${VERSION_FILE}" "GITHUB_FTL_VERSION" "${GITHUB_FTL_VERSION}"
 
     if [[ "${DOCKER_TAG}" ]]; then
-        GITHUB_DOCKER_VERSION="$(curl -s 'https://api.github.com/repos/pi-hole/docker-pi-hole/releases/latest' 2> /dev/null | jq --raw-output .tag_name)"
+        GITHUB_DOCKER_VERSION="$(get_remote_version docker-pi-hole)"
         addOrEditKeyValPair "${VERSION_FILE}" "GITHUB_DOCKER_VERSION" "${GITHUB_DOCKER_VERSION}"
     fi
 
--- a/advanced/Scripts/updatecheck.sh
+++ b/advanced/Scripts/updatecheck.sh
@@ -49,10 +49,8 @@ rm -f "/etc/pihole/localversions"
 
 # Create new versions file if it does not exist
 VERSION_FILE="/etc/pihole/versions"
-if [ ! -f ${VERSION_FILE} ]; then
-    touch "${VERSION_FILE}"
-    chmod 644 "${VERSION_FILE}"
-fi
+touch "${VERSION_FILE}"
+chmod 644 "${VERSION_FILE}"
 
 # if /pihole.docker.tag file exists, we will use it's value later in this script
 DOCKER_TAG=$(cat /pihole.docker.tag 2>/dev/null)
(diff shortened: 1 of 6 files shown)


advanced/Scripts/version.sh
```diff
@@ -8,179 +8,93 @@
...
-source /etc/pihole/setupVars.conf
+. /etc/pihole/setupVars.conf
 
-getLocalVersion() {
-    # FTL requires a different method
-    if [[ "$1" == "FTL" ]]; then
-        pihole-FTL version
-        return 0
-    fi
+# Sourece the versions file poupulated by updatechecker.sh
```
rdwebdesign, line 15:
```suggestion
# Source the versions file poupulated by updatechecker.sh
```

yubiuser: Why do we have codespell at all?

dschaper: @PromoFaux

PromoFaux: I guess souerce is not in its misspellings dictionary.

We can remove if if doesn't work

yubiuser: Maybe. I got the feeling it's only detecting issues when it likes to do some work, otherwise it just says LGTM. We had some misspellings in the past that weren't detected for a long time and suddenly it's complaining.

However, we should keep it - better to correct us from time to time than never.

rdwebdesign: > I guess souerce is not in its misspellings dictionary.

I know, but this is not the expected behavior.
That's weird. We expect it to find all errors.

It found the error here (`versons` => `versions`):
https://github.com/pi-hole/pi-hole/blob/79c0b446e194c4d7f2be249a9e250be987e7601b/advanced/Scripts/version.sh#L22

But it didn't find `Sourece`.

yubiuser: I looked through all of the `codespell` dictionaries at https://github.com/codespell-project/codespell/tree/master/codespell_lib/data but none contains `Sourece`.

__

I know, but this is not the expected behavior.
That's weird. We expect it to find all errors.

Actually it is. `Codespell` only finds errors that are listed in their dictionaries. If an misspelling is not part of it, it won't pop up.

rdwebdesign: I know how it works and I know its not easy to get every spell error, but I think it works differently than the general "expected behavior".

The "expected" behavior from a spell checker action is to find all spell errors, not to find just a small set of errors.

This action is helpful, but there are some "holes" to be fixed.

yubiuser: Sure. I'm not sure why they did choose this path. Maybe because users use all kind of strange function names and variables. If the would catch everything that is not in a usual dictionary they would get way too many false-positives. Their approach produces almost non, but misses real spelling mistakes in exchange.

advanced/Scripts/version.sh
```diff
@@ -8,179 +8,93 @@
...
+cachedVersions="/etc/pihole/versions"
 
-    # Get the tagged version of the local repository
-    local directory="${1}"
-    local version
+if [ -f ${cachedVersions} ]; then
+    # shellcheck disable=SC1090
+    . "$cachedVersions"
+else
+    echo "Could not find /etc/pihole/versions. Exiting."
```
PromoFaux, line 22: I think we should just have this run `/opt/pihole/updatechecker.sh` in the case of the versions file not existing. If that requires that the version script go back behind a root check - then so be it

advanced/Scripts/updatecheck.sh
```diff
@@ -35,8 +49,10 @@ rm -f "/etc/pihole/localversions"
 
 # Create new versions file if it does not exist
 VERSION_FILE="/etc/pihole/versions"
-touch "${VERSION_FILE}"
-chmod 644 "${VERSION_FILE}"
+if [ ! -f ${VERSION_FILE} ]; then
```
PromoFaux, line 52: No need to guard this, see https://github.com/pi-hole/pi-hole/pull/4875#discussion_r951935177

advanced/Scripts/updatecheck.sh
```diff
@@ -46,54 +62,73 @@ if [[ ! "${DOCKER_TAG}" =~ $regex ]]; then
...
+addOrEditKeyValPair "${VERSION_FILE}" "GITHUB_CORE_VERSION" "${GITHUB_CORE_VERSION}"
 
-    if [[ "${INSTALL_WEB_INTERFACE}" == true ]]; then
-        WEB_VERSION="$(get_local_version /var/www/html/admin)"
-        addOrEditKeyValPair "${VERSION_FILE}" "WEB_VERSION" "${WEB_VERSION}"
-    fi
+GITHUB_CORE_HASH="$(get_remote_hash pi-hole "${CORE_BRANCH}")"
+addOrEditKeyValPair "${VERSION_FILE}" "GITHUB_CORE_HASH" "${GITHUB_CORE_HASH}"
+
+if [[ "${INSTALL_WEB_INTERFACE}" == true ]]; then
```
PromoFaux, line 117: I think to save multiple uses of the `if web interface installed` block, we could maybe reorganise this section and group by component, rather than local/remote etc

pseudo:

```
core_version
core_branch
core_hash
github_core_version
github_core_hash

if [ web stuff ]
  web_version
  etc
fi

ftl...

docker...
```

advanced/Scripts/updatecheck.sh
```diff
@@ -15,16 +15,30 @@ function get_local_branch() {
...
+}
+
+function get_remote_version() {
+    curl -s "https://api.github.com/repos/pi-hole/${1}/releases/latest" 2> /dev/null | jq --raw-output .tag_name || return 1
+}
+
+
+function get_remote_hash(){
+    git ls-remote "https://github.com/pi-hole/${1}" --tags "${2}" | awk '{print substr($0, 0,9);}' || return 1
+}
```
rdwebdesign, line 35: I think some of these functions could be used by the debug log to check versions.

advanced/Scripts/version.sh
yubiuser: I run the `updatechecker` via `pihole updatechecker` which will check for root powers.

```
rockpi@rockpi-4b:~$ ls -lha /etc/pihole/versions
ls: cannot access '/etc/pihole/versions': No such file or directory
rockpi@rockpi-4b:~$ pihole -v
Could not find /etc/pihole/versions. Running update now.
[sudo] password for rockpi: 
  Pi-hole version is simplify_versions.sh v5.12.2-53-g52283478 (Latest: v5.13)
  AdminLTE version is 99 v5.16-5-g8abfde2e (Latest: v5.16)
  FTL version is fix/socket_warning vDev-bcdf653 (Latest: v5.18.2)


```

PromoFaux (approved):

PR #4970, opened 2022-10-12
master -> dev

PromoFaux:
Need to look at why the action is not firing at some point in the future...


rdwebdesign:
> Need to look at why the action is not firing at some point in the future...

**Note:**

Other times this action failed:
- #4897
- #4912 
- #4925

PR #4988, opened 2022-10-20
Prevent CNAME loop

yubiuser:

Check when adding CNAME records by `pihole -a addcustomcanme` that domain and target are different. Prevents `FTL` failing with `[2022-10-20 06:41:46.574 22783M] FATAL ERROR in dnsmasq core: CNAME loop involving jesse.chahal`.

Fixes the backend side of https://github.com/pi-hole/AdminLTE/issues/2406

---

---

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -729,7 +729,12 @@ AddCustomCNAMERecord() {
     if [[ -n "${validDomain}" ]]; then
         validTarget="$(checkDomain "${target}")"
         if [[ -n "${validTarget}" ]]; then
-            echo "cname=${validDomain},${validTarget}" >> "${dnscustomcnamefile}"
+            if [ "${validDomain}" = "${validTarget}" ]; then
+                echo "  ${CROSS} Domain and target are the same!"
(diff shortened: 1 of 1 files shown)


advanced/Scripts/webpage.sh
```diff
@@ -729,7 +729,12 @@ AddCustomCNAMERecord() {
     if [[ -n "${validDomain}" ]]; then
         validTarget="$(checkDomain "${target}")"
         if [[ -n "${validTarget}" ]]; then
-            echo "cname=${validDomain},${validTarget}" >> "${dnscustomcnamefile}"
+            if [ "${validDomain}" = "${validTarget}" ]; then
+                echo "  ${CROSS} Domain and target are the same!"
```
DL6ER, line 733: Shouldn't we print a small info text about why this is a bad thing?

Like
```suggestion
                echo "  ${CROSS} Domain and target are the same. Not adding to prevent looping."
```
(but maybe better)

yubiuser: Maybe not better, but different?

DL6ER (approved):

Issue #4989, opened 2022-10-20
local CNAME records delete wrong entry


ethanliu:
### Versions

-  Pi-hole version is v5.13 (Latest: v5.13)
-  AdminLTE version is v5.16 (Latest: v5.16)
-  FTL version is v5.18.2 (Latest: v5.18.2)

### Platform

- Raspbian GNU/Linux 10 (buster)

- Platform:  Raspberry Pi

### Expected behavior

int the local CNAME records list, say we have the following cname records

aaa.section.test.local
aaa-section.test.local
bbb.section.test.local
bbb-section.test.local

delete "aaa.section.test.local" should only delete this record only.

### Actual behavior / bug

delete "aaa.section.test.local" or "bbb.section.test.local" actually delete all four records.

delete "aaa-section.test.local" only delete "aaa-section.test.local" as expect.

yubiuser:
The issue is here:

https://github.com/pi-hole/pi-hole/blob/e773e3302ca66a6d918a40c8a8c6282f223d4906/advanced/Scripts/webpage.sh#L758

```
rockpi@rockpi-4b:~$ cat /etc/dnsmasq.d/05-pihole-custom-cname.conf
cname=aaa.section,q
cname=aaa-section,q
cname=bbb.section,q
cname=bbb-section,q
rockpi@rockpi-4b:~$ sudo sed -i "/cname=bbb.section,q$/Id" /etc/dnsmasq.d/05-pihole-custom-cname.conf 
rockpi@rockpi-4b:~$ cat /etc/dnsmasq.d/05-pihole-custom-cname.conf
cname=aaa.section,q
cname=aaa-section,q

```

yubiuser:
We need to escape the `.`

```
rockpi@rockpi-4b:~$ sudo sed -i "/^cname=bbb\.section,q$/Id" /etc/dnsmasq.d/05-pihole-custom-cname.conf 
rockpi@rockpi-4b:~$ cat /etc/dnsmasq.d/05-pihole-custom-cname.conf
cname=bbb-section,q
```

yubiuser:
@ethanliu 

Thanks for the report. You discovered a real but. PR https://github.com/pi-hole/pi-hole/pull/4990 should fix this.
Can you please try it with `pihole checkout core escape_dot`.

You can go back on track anytime and before the next Pi-hole update with `pihole checkout core master`

ethanliu:
[...]

PR #4990, opened 2022-10-21
Escape dots in local DNS records/CNAMES before removing them

yubiuser:

Fixes a bug in our code when we remove local DNS records/CNAMEs. We use `sed` to remove the entry from the relevant file, but don't escape dots, leading to a bug described in https://github.com/pi-hole/pi-hole/issues/4989

Escapes dots before removing them by `sed`

---

---

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -629,6 +629,8 @@ checkDomain()
 
 escapeDots()
 {
+    # shellcheck disable=SC2001
+    # SH suggest bashism ${variable//search/replace}
     escaped=$(sed 's/\./\\./g' <<< "$1" )
     echo "${escaped}"
 }
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -629,9 +629,9 @@ checkDomain()
 
 escapeDots()
 {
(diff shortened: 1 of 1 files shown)


yubiuser:
Confirmed working in the linked issue

dschaper (dismissed):

advanced/Scripts/webpage.sh
```diff
@@ -629,6 +629,7 @@ checkDomain()
 
 escapeDots()
 {
+    # shellcheck disable=SC2001
```
dschaper, line 632: This is to disable a false positive? If so, could you add a comment that says why the check was disabled?

yubiuser: See the title of the second commit. SH suggests `${variable//search/replace}` which is bash-only.

yubiuser: I forced push with a more detailed explanation.

rdwebdesign: I don't think this is a false positive. `<<<` is bash only.

I think the POSIX alternative is:

```sh
    escaped=$(echo "$1" | sed 's/\./\\./g')
```

rdwebdesign (changes requested):

rdwebdesign (approved):

PR #4995, opened 2022-10-29
Fix static IP not set if IP is already present in any way inside dhcpcd.conf

StephanPillhofer:
Signed-off-by: Stephan Pillhofer <43667664+StephanPillhofer@users.noreply.github.com>

This PR solves a bug which prevents the `dhcpcd.conf` file from being changed when the user wants to assign a static IP address to the RaspberryPi. The bug occurs because the target static IP address may already be present in any form in the (default) `dhcpcd.conf` file. (e.g. the IP address may be mentioned in another setting than "static ip_address" or it may be present in a commented example setting)

The bug is resolved by utilizing a regular expression for matching a valid non-commented static IP addess entry (e.g. `static ip_address=192.168.10.2/24`)

none

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -828,8 +828,11 @@ It is also possible to use a DHCP reservation, but if you are going to do that, 
 
 # Configure networking via dhcpcd
 setDHCPCD() {
-    # Check if the IP is already in the file
-    if grep -q "${IPV4_ADDRESS}" /etc/dhcpcd.conf; then
+    # Regex for matching a non-commented static ip address setting
+    local regex="^[ \t]*static ip_address[ \t]*=[ \t]*${IPV4_ADDRESS}[ \t]*$"
+
+    # Check if static IP is already set in file
+    if grep -xq "${regex}" /etc/dhcpcd.conf; then
         printf "  %b Static IP already configured\\n" "${INFO}"
     # If it's not,
     else


yubiuser (approved):

automated install/basic-install.sh
```diff
@@ -828,8 +828,11 @@ It is also possible to use a DHCP reservation, but if you are going to do that,
 
 # Configure networking via dhcpcd
 setDHCPCD() {
-    # Check if the IP is already in the file
-    if grep -q "${IPV4_ADDRESS}" /etc/dhcpcd.conf; then
+    # Regex for matching a non-commented static ip address setting
+    local regex="[ ]*static ip_address[ ]*=[ ]*${IPV4_ADDRESS}"
```
dschaper, line 832: I think the regex needs to be tweaked a bit.

`[ ]*` would only match a space character. The manpage for [`dhcpcd.conf`](https://manpages.ubuntu.com/manpages/focal/man5/dhcpcd.conf.5.html) says that:
>The first word on the line is the option and the rest of the line is the value.
     **Leading and trailing whitespace for the option and value are trimmed.**

dschaper (changes requested):

automated install/basic-install.sh
dschaper: And I don't think spaces are allowed around the `=` but I don't have a setup to check that right now.

StephanPillhofer: I also looked at the same document and couldn't find any other information anywhere about allowed spaces. I added the multiple `[ ]*` in case it may work, just to be on the safe side, but I myself only tested it without the spaces.

I can test it with spaces in a few days or remove the  `[ ]*`  from the regex if you want. In this case we can be sure that a valid static IP setting is configured.

dschaper: `[ ]*` only matches a literal space character. I understand 'whitespace' to include tabs. 

The first check _should_ start with a newline anchor and then zero or more whitespace characters. I don't know if the checks around `=` should include tabs or not. I don't _think_ it's a problem to include a space check around `=`, I'm just mentioning it in case others have insight about the case.

dschaper: And I guess it depends on how **option** and **value** is interpreted. I think `static` is the **option** and `ip_address=ip.ad.dd.r` is the **value**. 

I guess `static ip_address` could be the **option** and `ip.ad.dd.r` is the **value** and `=` separates them?

StephanPillhofer: Personally I think that `static ip_address` being the option and `ip.ad.dd.r` being the value makes more sense, since `=` is often used to seperate key value pairs. Also you are definitely right with the whitespaces not only being spaces and new line anchoring (I had it in the beginning but somehow it got lost during testing).

Based on the above assumtion I would rewrite the regex like this: 
`regex="^[ \t\r\n\f]*static ip_address[ \t\r\n\f]*=[ \t\r\n\f]*${ip}[ \t\r\n\f]*$"`

StephanPillhofer: Correction:

Since grep -xq only looks at a text line by line it's enought to check for white spaces and tabs. 

`regex="^[ \t]*static ip_address[ \t]*=[ \t]*{IPV4_ADDRESS}[ \t]*$"`

StephanPillhofer: Actually, checking for ony tabs and spaces should be enough, since a new option/value pair must start on a new line:
`regex="^[ \t]*static ip_address[ \t]*=[ \t]*${IPV4_ADDRESS}[ \t]*$"`

rdwebdesign: I'm not sure if this is relevant, but looking at 3 different manpages makes it more confusing:
- https://manpages.ubuntu.com/manpages/trusty/man5/dhcpcd.conf.5.html (already linked above)
- https://manpages.debian.org/testing/dhcpcd5/dhcpcd.conf.5.en.html;
- https://linux.die.net/man/5/dhcpd-options.

The last one shows spaces around the equal sign. The other 2 don't.

Also, there are a lot of places talking about using spaces (or commas) as separators.

dschaper: The only place that really matters is Raspberry Pi OS/Raspbian (or what ever it's currently called.) That is the only case that triggers the `dhcpcd.conf` detection. We stopped doing any kind of IP addressing with every other OS and it was my request that we keep holding hands with Raspbian. 

I don't think including spaces/tabs around the `=` is a problem, it allows for more cases than may be needed but it doesn't exclude any valid configurations. 

I would like to keep the `^` anchor for the newline, if only for making the code self-documenting.

StephanPillhofer: I agree with including spaces/tabs around the  `=`.

Keeping the `^ ... $` for self documentation is a good point. It doesn't change anything.

Regarding changing `[ \t]*` to `\s*`: I intentionally avoided it since in some sources I read that `\s` apparently shouldn't be supported with the flavor of regex used by grep. (However the regex still worked with `\s` on my local Linux machine.)

http://www.greenend.org.uk/rjk/tech/regexp.html 
https://stackoverflow.com/questions/1074941/what-flavour-of-regular-expression-is-grep

dschaper: >I agree with including spaces/tabs around the =.

Sounds good.

> Keeping the ^ ... $ for self documentation is a good point. It doesn't change anything.

Yeah, consider it stylistic.

> Regarding changing [ \t]* to \s*: I intentionally avoided it

Okay, that makes sense to me. `\s` would add newlines but your earlier comments show that newlines do not need to be considered.

StephanPillhofer: I updated the regex/commit accordingly.

dschaper (approved):

yubiuser:
With the anchors, the `-x` option should not be necessary anymore.
__

The current code with the ending anchor will fail with something like this

`static ip_address = 192.168.0.5 #comment`

I think the trailing anchor should be removed.

StephanPillhofer:
You are right, didn't think about the possibility of adding a comment at the end. Also the `-x` can be removed. Should I create a new PR/Commit to fix it? 

New Regex will be: `regex="^[ \t]*static ip_address[ \t]*=[ \t]*${IPV4_ADDRESS}"` (tested)

yubiuser:
I also tested with different regex/grep combinations and came to the same conclusion. Please open a new PR if you want :)

dschaper:
Is an inline trailing comment a valid configuration?

yubiuser:
The manual says

> Comments may be placed anywhere within the file (except within quotes). Comments begin with the # character and end at the end of the line.

dschaper:
Can you link to that please? 

The only thing I have found is 

> The first word on the line is the option and the rest of the line is the value. Leading and trailing whitespace for the option and value are trimmed. You can escape characters in the value using the \ character. Comments can be prefixed with the # character. String values should be quoted with the " character.

From https://manpages.debian.org/testing/dhcpcd5/dhcpcd.conf.5.en.html

StephanPillhofer:
So in the worst case the comment may be part of the "value" part which would not be good.

yubiuser:
The relevant code section should be 

https://github.com/NetworkConfiguration/dhcpcd/blob/6a9994bbf00712ace77a835cb7e45558383be673/src/common.c#L181-L195

StephanPillhofer:
Apparently it seems to be valid:

```
noipv6rs                 # disable routing solicitation
denyinterfaces eth2      # Don't touch eth2 at all
interface eth0
  ipv6rs                 # enable routing solicitation for eth0
  ia_na 1                # request an IPv6 address
  ia_pd 2 eth1/0         # request a PD and assign it to eth1
  ia_pd 3 eth2/1 eth3/2  # req a PD and assign it to eth2 and eth3
```

From: https://www.daemon-systems.org/man/dhcpcd.conf.5.html (scroll a bit down)

StephanPillhofer:
> The relevant code section should be
> 
> https://github.com/NetworkConfiguration/dhcpcd/blob/6a9994bbf00712ace77a835cb7e45558383be673/src/common.c#L181-L195

My c is not the best but I would also see the code as an indication that it is valid to do inline comments.

dschaper:
Yeah, the `c` code that Yubi linked to does truncate the lines at the first `#`, so trailing line comments are perfectly valid.

dschaper:
My `c` is not great either but my Stack Overflow is Ninja level.

StephanPillhofer:
New PR created: #4998

PR #4872, opened 2022-08-18
Use utils.sh in install script

yubiuser:
- **What does this PR aim to accomplish?:**

Inspired by https://github.com/pi-hole/pi-hole/pull/3590, with this PR the install script now uses the `addKey`, `addOrEditKeyValPair` and `removeKey` from `utils.sh`. After `installScripts()` finishes, this file should be available on all installations.
To modify `01-pihole.conf` the template file needed adjustment.

Additionally, `FTL_CONFIG_FILE="${PI_HOLE_CONFIG_DIR}/pihole-FTL.conf"` was introduced.

---
**By submitting this pull request, I confirm the following:** 

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against. 
2. I have commented my proposed changes within the code and I have tested my changes.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

---
- [x] I have read the above and my PR is ready for review. _Check this box to confirm_

PR #4999, opened 2022-11-03
Touch files before addKey or addOrEditKeyValPair

yubiuser:

Fixes a bug introduced by https://github.com/pi-hole/pi-hole/pull/4872.
Using `addOrEditKeyValPair` in `finalExports()` will result in an `grep` error that `setupVars.conf` could not be found if `setupVars.conf` does not exist, when this is the first run of `finalExports()`. Reason is, that `addOrEditKeyValPair` does not check if the file exists before `greping` for the "key".
https://github.com/pi-hole/pi-hole/blob/e773e3302ca66a6d918a40c8a8c6282f223d4906/advanced/Scripts/utils.sh#L29-L40

Before https://github.com/pi-hole/pi-hole/pull/4872 we simply `echo`ed the (new) content of `setupVars.conf`, which created the file if it did not exist.

~~Create `setupVars.conf` before running `addOrEditKeyValPair` on it by `touching` it.~~
Touch files before `addKey` or `addOrEditKeyValPair`

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1785,6 +1785,9 @@ create_pihole_user() {
 finalExports() {
     # set or update the variables in the file
 
+    # create the file if it does not exist
+    touch "${setupVars}"
+
     addOrEditKeyValPair "${setupVars}" "PIHOLE_INTERFACE" "${PIHOLE_INTERFACE}"
     addOrEditKeyValPair "${setupVars}" "PIHOLE_DNS_1" "${PIHOLE_DNS_1}"
     addOrEditKeyValPair "${setupVars}" "PIHOLE_DNS_2" "${PIHOLE_DNS_2}"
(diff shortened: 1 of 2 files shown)


automated install/basic-install.sh
```diff
@@ -1785,6 +1785,9 @@ create_pihole_user() {
 finalExports() {
     # set or update the variables in the file
 
+    # create the file if it does not exist
+    touch "${setupVars}"
```
PromoFaux, line 1789: I wonder if it might be better to introduce a `touch` into `addOrEditKeyValPair` instead? That would reduce the need to type out a `touch` command before calling `addOrEditKeyValPair` on the off chance we create a new file at a later date

yubiuser: I thought about this too. I would then also add this to `addKey`

PromoFaux (approved):

Issue #5008, opened 2022-11-08
Regex not supported in adlists but not considered as invalid


jankytay:
**Versions**

- Pi-hole: v5.13
- AdminLTE: v5.16
- FTL: v5.18.2

Gravity update only complains about some RegEx domains in adlists as invalid.
For example:
`*.example.com` will show up in "Sample of invalid domains"
But `(^|\.)example\.com` won't. It'll just be discarded silently.

Also I'd suggest making it clear in [RegEx blocking documentation](https://docs.pi-hole.net/regex/overview/) that RegEx syntax is not supported in Adlists.

dschaper:
Can you provide any information like a list that shows you these errors?

yubiuser:
I can change the code to something like this

```
  [i] Target: https://raw.githubusercontent.com/jankytay/pihole-troubleshooting/master/issue5008.txt
  [✓] Status: Retrieval successful
  [i] Analyzed 8 domains, 5 domains invalid!
      Sample of invalid domains:
      - *.example3.com
      - (^|\.)example5\.com
      - (\.|^)example6\.com$
      - ^example77?
      - (^|\.)example8\.com;reply=none
```

1332mm:
SXP

1332mm:
M👁Ⓜ️

1332mm:
- [ ] - 😂🤣😆😁😄🙃🤣😂🇹🇷🇺🇳🙏🤲🇺🇸🇬🇧🇦🇪🍼🌶🫑🪆👩‍🦳👨‍🦳👩‍🦲🅱️👨‍👩‍👧👨‍👩‍👧‍👧🧑‍🌾🏅☺️🌄🎉☢️⭕️🟤🌀🥹👁Ⓜ️🇹🇷🇺🇳🇹🇨🧠💁🏾⏱🤑🧨👺💵😈💰👹👹😂🥹💀🍷🍷🍷🤬🎃😾para yoksa işaret****_![image](https://user-images.githubusercontent.com/110432476/201821924-2fa5e8f3-d404-4976-93d1-79bf324b2178.jpeg)_****** 🙅

yubiuser:
Fix released with https://github.com/pi-hole/pi-hole/releases/tag/v5.15

PR #5009, opened 2022-11-09
Improve detection of incorrect lines during gravity

yubiuser:

When `gravity` downloads a file we use a two-step "cleaning" mechanism (`gravity_ParseFileIntoDomains` and `parseList`).
https://github.com/pi-hole/pi-hole/blob/e773e3302ca66a6d918a40c8a8c6282f223d4906/gravity.sh#L686-L688

In `gravity_ParseFileIntoDomains` we already remove lines which don't match domain names 
https://github.com/pi-hole/pi-hole/blob/e773e3302ca66a6d918a40c8a8c6282f223d4906/gravity.sh#L742

Later, in `parseList` we use a different regex to filter for invalid domains
https://github.com/pi-hole/pi-hole/blob/e773e3302ca66a6d918a40c8a8c6282f223d4906/gravity.sh#L533

However, we use the output of `gravity_ParseFileIntoDomains` to see if there are "incorrect_lines" in the file
https://github.com/pi-hole/pi-hole/blob/e773e3302ca66a6d918a40c8a8c6282f223d4906/gravity.sh#L535

Due to the cleaning already in `gravity_ParseFileIntoDomains` the number of `incorrect_lines` can be underestimated.
This is reported here: https://github.com/pi-hole/pi-hole/issues/5008

The regex from `gravity_ParseFileIntoDomains` is now part of `parseList`, meaning the "input" is less filtered (preventing underestimation). Within `parseList` both regex are used to filter incorrect lines separately (because they find different - although partly overlapping - incorrect lines), and sorted for unique entries afterwards.

---

---

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -524,15 +524,20 @@ num_target_lines=0
 num_source_lines=0
 num_invalid=0
 parseList() {
-  local adlistID="${1}" src="${2}" target="${3}" incorrect_lines
+  local adlistID="${1}" src="${2}" target="${3}" incorrect_lines sample_incorrect_lines
   # This sed does the following things:
+  # 0. Remove all lines containing no domains
   # 1. Remove all domains containing invalid characters. Valid are: a-z, A-Z, 0-9, dot (.), minus (-), underscore (_)
   # 2. Append ,adlistID to every line
   # 3. Remove trailing period (see https://github.com/pi-hole/pi-hole/issues/4701)
   # 4. Ensures there is a newline on the last line
-  sed -e "/[^a-zA-Z0-9.\_-]/d;s/\.$//;s/$/,${adlistID}/;/.$/a\\" "${src}" >> "${target}"
-  # Find (up to) five domains containing invalid characters (see above)
-  incorrect_lines="$(sed -e "/[^a-zA-Z0-9.\_-]/!d" "${src}" | head -n 5)"
+  sed -r  "/([^\.]+\.)+[^\.]{2,}/!d;/[^a-zA-Z0-9.\_-]/d;s/\.$//;s/$/,${adlistID}/;/.$/a\\" "${src}" >> "${target}"
+
+  # Find lines containing no domains or with invalid characters (see above)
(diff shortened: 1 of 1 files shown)


PromoFaux:
Does the removed sed/regex work in the first place? Seems to just delete everything when I try on sed.js.org

image


The one we are left with appears to behave correctly except for happily accepting an empty line as a domain:

image

yubiuser:
Don't omit the `-r` (regex-extended). For `sed.js.org` use the "option" menu.

PromoFaux:
Cool - I missed that! 

One thing to note, then is that the regex that this PR intends to remove will mean that single words will get matched as domains (where they were removed before)

image


image

yubiuser:
Ok, I see the problem. I'll think about a solution.

yubiuser:
I improved the code: 

- `gravity_ParseFileIntoDomains` will now remove empty lines
- the regex from `gravity_ParseFileIntoDomains` is now part of `parseList`
- within `parseList` both regex are used to filter **incorrect** lines separately (because the find different - although partly overlapping - incorrect lines)
  - they are added to an array which is sorted for unique entries and limited to 5 examples

Because we enter `parseList` with a less filtered source the number of invalid lines reported by gravity will increase.

gravity.sh
```diff
@@ -524,15 +524,20 @@ num_target_lines=0
 num_source_lines=0
 num_invalid=0
 parseList() {
-  local adlistID="${1}" src="${2}" target="${3}" incorrect_lines
+  local adlistID="${1}" src="${2}" target="${3}" incorrect_lines sample_incorrect_lines
   # This sed does the following things:
+  # 0. Remove all lines containing no domains
```
PromoFaux, line 529: Can we make this a 1-based list rather than a 0-based one, please?

yubiuser: Sure.

PromoFaux (approved):
Tested working in theory with with regex tester. Happy with your sample output too

PR #5048, opened 2022-12-04
Install wheel before other python packages

yubiuser:

We use `pip` to install the necessary packages for the distro tests.

https://github.com/pi-hole/pi-hole/blob/0034538794a3754ab532a53480da39d3fd13d7ed/.github/workflows/test.yml#L72-L73

Some of the required packages have dependencies themselves. Not for all of those pre-packed binary exists ("wheel"), so they need to be compiled during installation. However, they sometimes lack a `pyproject.toml` to define the builtd environment. Hence, we get a deprecation warning which will be a hard error once `pip` reaches v23.1.

```
  DEPRECATION: docopt is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559
```

This PR fixes the issue as suggested by the warning: we install `wheel` before all other requirements. It would be easy to add `wheel` to our `requirements.txt`, however, this does not work (`wheel` is not used when installed at the same time). See https://stackoverflow.com/questions/59104686/wheel-installation-from-within-requirements-txt

Therefore, we need a two-step approach.

Install `wheel` by `pip` before all other dependencies.

---

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -70,7 +70,7 @@ jobs:
           python-version: "3.10"
 
       - name: Install dependencies
-        run: pip install -r test/requirements.txt
+        run:  pip install wheel && pip install -r test/requirements.txt
 
       - name: Test with tox
         run: tox -c test/tox.${DISTRO}.ini
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,8 +69,11 @@ jobs:
         with:
           python-version: "3.10"
 
+      - name: Install wheel
+        run:  pip install wheel
+
       - name: Install dependencies
-        run:  pip install wheel && pip install -r test/requirements.txt
+        run: pip install -r test/requirements.txt
 
       - name: Test with tox
         run: tox -c test/tox.${DISTRO}.ini


.github/workflows/test.yml
```diff
@@ -70,7 +70,7 @@ jobs:
           python-version: "3.10"
 
       - name: Install dependencies
-        run: pip install -r test/requirements.txt
+        run:  pip install wheel && pip install -r test/requirements.txt
```
dschaper, line 73: Should this be an `&&` that could fail, thus skipping the `install -r test/requirements.txt` portion of the step?

Can it be broken down to two `run` steps?

yubiuser: I can break it down into two steps. `install -r test/requirements.txt` should only be executed if `wheel` is installed successfully. At the moment it won't matter, as `wheel` is not a hard requirement. But it will be once `pip` is v23.1

rdwebdesign: If the first one fails, it will be possible to continue?

dschaper (approved):

PR #5065, opened 2022-12-15
Create new lighttpd pihole-admin.conf for admin interface

gstrauss:
**What does this PR aim to accomplish?:**

lighttpd pihole-admin.conf for admin interface

First steps to simplifying the convoluted historical pi-hole configuration of lighttpd.

**How does this PR accomplish the above?:**

PR creates pihole-admin.conf and update install and uninstall scripts.

---
**By submitting this pull request, I confirm the following:**

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against.
2. I have commented my proposed changes within the code and I have tested my changes.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
6. I have checked that another pull request for this purpose does not exist.
7. I have considered, and confirmed that this submission will be valuable to others.
8. I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
9. I give this submission freely, and claim no ownership to its content.

---
- [x] I have read the above and my PR is ready for review. *Check this box to confirm*

PR #5066, opened 2022-12-15
lighttpd: remove obsolete pihole block page error handler

gstrauss:

* lighttpd: remove obsolete pihole block page error handler
    Remove the advanced functionality of the 404 page (Blockpage)
    https://github.com/pi-hole/pi-hole/pull/3910

* lighttpd: remove obsolete exclusion of letsencrypt.conf
   (Anyone impacted by this exclusion probably worked around it by including letsencrypt.conf in external.conf)

PR inspired by discussion with @PromoFaux in discourse

This PR also replaces `/usr/share/lighttpd/include-conf-enabled.pl`, which is deprecated since Debian 10.

---

---


PromoFaux:
Code changes look sound in theory - I will spin up a droplet and do some testing over the weekend.

For clarification - am I right in thinking that if these changes are implemented - then a user wishing to make use of the optional IP blocking functionality will need to create an additional config in the conf.d folder to add their own 404 handler?

Not a problem in theory, I just think we should probably document this somewhere.

The removed code was there to ease things for the end users, but I'm not averse to stepping away from the mentality that we need to hold _everyone_'s hand through things.

gstrauss:
> For clarification - am I right in thinking that if these changes are implemented - then a user wishing to make use of the optional IP blocking functionality will need to create an additional config in the conf.d folder to add their own 404 handler?
>
> Not a problem in theory, I just think we should probably document this somewhere.

Yes, they can configure their own 404 handler, if desired.  As you pointed out in #3910, this is not the typical config and is no longer supported out of the box by pi-hole.

> The removed code was there to ease things for the end users, but I'm not averse to stepping away from the mentality that we need to hold everyone's hand through things.

The aim of this PR is to separate the pi-hole admin page config into an include file separate from lighttpd.conf so that we can discuss approaches to making future new installations *avoid* overwriting lighttpd.conf.   ...Hopefully existing users can opt-in to a future less-invasive pihole lighttpd config simply by uninstalling, the reinstalling pihole.

gstrauss:
I re-pushed (force-pushed) with my commits signed.

gstrauss:
> Not a problem in theory, I just think we should probably document this somewhere.

@PromoFaux do you have a suggestion where?  If this mode is not directly supported out-of-the-box anymore, where is the doc to let people know how to optionally set that up in pi-hole DNS config, and then we can perhaps add instructions for an extra included .conf file for lighttpd.

yubiuser:
We could add a section/page at https://docs.pi-hole.net/

Code is at https://github.com/pi-hole/docs

gstrauss:
@PromoFaux 
> > Not a problem in theory, I just think we should probably document this somewhere.
> 
> @PromoFaux do you have a suggestion where? If this mode is not directly supported out-of-the-box anymore, where is the doc to let people know how to optionally set that up in pi-hole DNS config, and then we can perhaps add instructions for an extra included .conf file for lighttpd.

@yubiuser 
> We could add a section/page at https://docs.pi-hole.net/
> 
> Code is at https://github.com/pi-hole/docs

I posted PR https://github.com/pi-hole/docs/pull/824 for discussion.

PromoFaux:
Please rebase on `development`, which should fix the test issues

gstrauss:
> Please rebase on development, which should fix the test issues

rebased on development and tests now pass

PromoFaux:
One potentially undesirable consequence of this PR is that there is no longer an automatic redirect from `pi.hole` to `pi.hole/admin` 

Is it possible to maintain this behaviour? So effectively add something that says "if host is pi.hole (or configured custom host), then go to /admin"

rdwebdesign:
I'm not sure if this is the correct config, but I think we need something like this:

```ini
$HTTP["host"] == "pi.hole" {
  server.document-root = "/var/www/html/admin/" 
}
```

gstrauss:
I added the following to #5065 
```
$HTTP["host"] == "pi.hole" {
    $HTTP["url"] == "/" {
        url.redirect = ("" => "/admin/")
    }
}
```

gstrauss:
I added the following to #5065 
```
$HTTP["host"] == "pi.hole" {
    $HTTP["url"] == "/" {
        url.redirect = ("" => "/admin/")
    }
}
```

gstrauss:
I rebased this PR on #5065 now that #5065 has been merged.  Trivial merge conflicts have been resolved.

gstrauss:
@PromoFaux with #5075 mocked up, it would appear possible to preserve `server.error-handler-404 = "/pihole/index.php"` for existing usage -- though omitted for new installs -- if I do a bit more work.  Is that desirable?

PromoFaux:
I'm in two minds here:
 1 - Remove the config per this PR, also delete `pihole/index.php` if it is detected, as it is now useless. Any changes users may or may not have made to that file would be overwritten on update anyway..
 2 - Be a little bit kind, and per your latest suggestion, if we detect that `pihole/index.php` exists on the system, keep the 404 handler in place.

2 would cause the least friction maybe, but honestly I think we need to make it about value, and I can't see that anyone is getting any value from `index.php` at the moment. It doesn't do a lot other than redirect to the `/admin` url.

Custom 404's are handled by:

https://github.com/pi-hole/pi-hole/blob/8d2e023ec0a2c52c309fd7971d76eeaef124ad65/automated%20install/basic-install.sh#L1403-L1406

(actually, whichever way - that bit will need changing)

gstrauss:
> 1 - Remove the config per this PR, also delete pihole/index.php if it is detected, as it is now useless. Any changes users may or may not have made to that file would be overwritten on update anyway..

pihole/index.php is still removed on uninstall.  Do you want me to add code to also remove it during install?

> 2 - Be a little bit kind, and per your latest suggestion, if we detect that pihole/index.php exists on the system, keep the 404 handler in place.

Yes, kinder to existing users.  OTOH, it does leave around technical debt baggage.

Regarding the use of disrecommended `BLOCK_IPV4` and `BLOCK_IPV6`
>> I posted PR https://github.com/pi-hole/docs/pull/824 for discussion.

LMK what you'd like and I'll make the appropriate modifications.

PromoFaux:
> Do you want me to add code to also remove it during install?

I think we should kill off `index.php` entirely, yes. It serves no purpose if we leave it there - as custom block pages and the like will have been made in `custom.php`.  I would like that we remove the logic that adds the 404 back in for `custom.php`, too.

We can include a link to the docs additions in the release notes (nobody ever reads the release notes) and/or point users at it as and when they notice the 404 page functionality is missing entirely. Better the few people that are likely still using it make a one time change to their configuration that we don't overwrite, than to always make assumptions in the installer.

gstrauss:
> I think we should kill off index.php entirely, yes. It serves no purpose if we leave it there - as custom block pages and the like will have been made in custom.php. I would like that we remove the logic that adds the 404 back in for custom.php, too.

Already removed in this PR.

I rearranged the order of commits in #5075 to stop overwriting lighttpd.conf and then (this PR) to remove obsolete pihole block page error handler.  If (if if if) we were to change our minds and choose not to kill off `index.php`, then we would remove those commits near the tip of #5075 and would have to replace with a smaller commit which disabled tests for `index.php` on new installs (for which `server.error-handler-404` would not be enabled by default).

I am going to close this PR.  Follow-up in #5075

Issue #2471, opened 2018-10-20
Password change failed: webpage.sh uses Bash variable expansion where it may not


TheCapsLock:
**In raising this issue, I confirm the following:** `{please fill the checkboxes, e.g: [X]}`

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the the source code relevant to this issue?:**

`7`

---
**Expected behaviour:**

When using the command

```bash
$ pihole -a -p
```

I expect the password to be entered allows me to access admin web UI regardless the complexity of my password.

**Actual behaviour:**

If I use some special charracters like "\" I cannot login anymore.

**Steps to reproduce:**

Use this password as an example:

```
@6AZzxR,(![c\h{v_$u{urVM/!Fy3]6a@PCp!,%%*.uyn7U/x9;u"Xvvg+65HPj'>y4Lj%tjfpvki5mK^L'3#,Ng(3
```

(Here, bash variable expension will replace "\h" by "h".)

Try to login using the password above: it fails. if you use the one below, it works:

```
@6AZzxR,(![ch{v_$u{urVM/!Fy3]6a@PCp!,%%*.uyn7U/x9;u"Xvvg+65HPj'>y4Lj%tjfpvki5mK^L'3#,Ng(3
```

**Debug token provided by [uploading `pihole -d` log](https://discourse.pi-hole.net/t/the-pihole-command-with-examples/738#debug):**

(upload failed, here is the basic information of this installation):

```
*** [ DIAGNOSING ]: Core version                                                                       
[i] Core: v4.0 (https://discourse.pi-hole.net/t/how-do-i-update-pi-hole/249) 
[i] Branch: master                                                                 
[i] Commit: v4.0-0-gddbdb51                                                                                                                                                                                                                                                    
                                                                                                                                
*** [ DIAGNOSING ]: Web version                                                                                                                                                                                                                                                
[...]

PR #5096, opened 2022-12-28
Added specific domain forwarding to other name server

ImDaBigBoss:

This PR is made so you can create a list of domains that would be forwarded to a different name server.
This is useful if you have run Pi-hole but don't maintain the internal domains. For example, you are on a network that relies on services that use domain names, domain names that are issued internally by a DNS server that you don't control. If you still want access to these local domains, what they point to and their subdomains, you can do so without relying on that DNS server for all requests.
It also applies to VLANs and their hostname resolution (https://discourse.pi-hole.net/t/multiple-subnets-vlans-and-hostname-resolution/31746)

Much like the DNS records (A/AAAA records) or the CNAME records, this adds a section under "Local DNS" where you can register domain forwarding.
This creates a new file in the dnsmasq directory and adds `server=/my.domain/1.2.3.4` into it.

None

---

---

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -19,6 +19,7 @@ readonly FTLconf="/etc/pihole/pihole-FTL.conf"
 readonly dhcpstaticconfig="/etc/dnsmasq.d/04-pihole-static-dhcp.conf"
 readonly dnscustomfile="/etc/pihole/custom.list"
 readonly dnscustomcnamefile="/etc/dnsmasq.d/05-pihole-custom-cname.conf"
+readonly dnscustomnsfile="/etc/dnsmasq.d/07-pihole-custom-ns.conf"
 
 readonly gravityDBfile="/etc/pihole/gravity.db"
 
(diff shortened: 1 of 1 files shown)


ImDaBigBoss:
This depends on [#2471](https://github.com/pi-hole/AdminLTE/pull/2471)

PromoFaux:
Echo comments: https://github.com/pi-hole/AdminLTE/pull/2471#issuecomment-1368432808

jpgpi250:
If I understand this correctly, I've been doing this for a very long time, with an additional /etc/dnsmasq.d file, entries like:
```
server=/b2discourse.b-cdn.net/127.10.10.5#5555
server=/b2discourse.pi-hole.net/127.10.10.5#5555
```
27.10.10.5#5555 being knot-resolver, also running on my pi, the default upstream resolver is unbound (27.10.10.2#5552)

however....

In the files changed section of this PR, I noticed this code uses **RestartDNS** to apply the changes, If I'm correct RestartDNS actually does **service pihole-FTL restart**. This is totally unnecessary.
the dnsmasq man page states:
```
--servers-file=<file>
A special case of --conf-file which differs in two respects. Firstly, only --server and --rev-server are allowed in the configuration file included. Secondly, the file is re-read and the configuration therein is updated when dnsmasq receives SIGHUP.
```
this implies adding a configuration line (I added this line in a separate config file - could be added to 01-pihole.conf):
```
servers-file=/etc/dnsmasq.d/07-pihole-custom-ns.conf
```
doesn't require **RestartDNS** but would activate the changes, using  **RestartDNS reload** which actually does:
```
kill -HUP ${pid}
```

PR #5103, opened 2023-01-03
Adjusting `fastcgi.server` config

### LINKED ISSUE #5102: Generate debug log DoS lighttpd
### Versions
```
  Pi-hole version is v5.14.2 (Latest: v5.14.2)
  AdminLTE version is v5.18 (Latest: v5.18)
  FTL version is v5.20 (Latest: v5.20)
```

### Platform
- OS and version:  Ubuntu 22.04
- Platform:  Bare-metal x64 machine 

### Expected behavior

A debug log should be generated

### Actual behavior / bug

The log freezes at "Dashboard headers" every time.
Then `lighttpd` freezes/locks up and the entire web UI is inaccessible (even a local `curl http://127.0.0.1` just hangs) until I restart `lighttpd`.

### Steps to reproduce

1. Tools > Generate debug log
2. Leave all options unchecked
3. Click "Generate Debug Log"

## Debug Token

- URL:  https://tricorder.pi-hole.net/k8zCIey3/

### Screenshots

image

### Additional context

This is a brand new install.

`pihole -d -w` runs fine in the terminal.

I have uninstalled pihole, lighttpd, and php, wiped `/etc/www`, `/etc/httpd`, and the error still happens exactly the same.

CPU and RAM do not spike at all. System looks as "normal" as it ever does.

rdwebdesign:

Fix #5102 

Recent changes created a new `fastcgi.server` config on Debian and derivative OS, but some configs related to PHP were forgotten.
This PR fixes the config.

none

---

---

### DIFF
--- a/advanced/pihole-admin.conf
+++ b/advanced/pihole-admin.conf
@@ -12,9 +12,13 @@
 # ANY CHANGES MADE TO THIS FILE AFTER INSTALL WILL BE LOST ON THE NEXT UPDATE #
 ###############################################################################
 
+server.errorlog := "/var/log/lighttpd/error-pihole.log"
+
 $HTTP["url"] =~ "^/admin/" {
     server.document-root = "/var/www/html"
     server.stream-response-body = 1
+    accesslog.filename = "/var/log/lighttpd/access-pihole.log"
+    accesslog.format = "%{%s}t|%h|%V|%r|%s|%b"
 
     fastcgi.server = (
         ".php" => (
--- a/advanced/pihole-admin.conf
+++ b/advanced/pihole-admin.conf
@@ -15,6 +15,7 @@
 $HTTP["url"] =~ "^/admin/" {
     server.document-root = "/var/www/html"
     server.stream-response-body = 1
+
     fastcgi.server = (
         ".php" => (
             "localhost" => (
@@ -22,6 +23,14 @@ $HTTP["url"] =~ "^/admin/" {
                 "bin-path" => "/usr/bin/php-cgi",
                 "min-procs" => 0,
                 "max-procs" => 1,
+                "bin-environment" => (
+                    "PHP_FCGI_CHILDREN" => "4",
+                    "PHP_FCGI_MAX_REQUESTS" => "10000",
+                ),
+                "bin-copy-environment" => (
+                    "PATH", "SHELL", "USER"
+                ),
+                "broken-scriptfilename" => "enable",
             )
         )
     )


yubiuser (dismissed):

rdwebdesign:
I edited the PR to reflect the decision we made to change the config.

Tests are passing.

yubiuser:
What's about the log file location?

rdwebdesign:
I will add one more commit to change the log path. I just need to run a few tests before.

advanced/pihole-admin.conf
```diff
@@ -15,13 +15,22 @@
 $HTTP["url"] =~ "^/admin/" {
     server.document-root = "/var/www/html"
     server.stream-response-body = 1
+
     fastcgi.server = (
         ".php" => (
             "localhost" => (
-                "socket" => "/tmp/pihole-php-fastcgi.socket",
+                "socket" => "/run/lighttpd/php.socket",
```
PromoFaux, line 22: I think we can safely leave this as `/tmp/pi-hole-php-fastcgi.socket` - less chance of collision, then.

```suggestion
                "socket" => "/tmp/pihole-php-fastcgi.socket",
```

dschaper: Depends on the user running the daemons and the access for the directories.

rdwebdesign:
Ohhh... I forgot about that. I do it today.

advanced/pihole-admin.conf
PromoFaux: How do you mean? As to whether or not it would cause conflicts? Unless you can think of a reason why not, I feel like the socket being in /tmp for the pi-hole specific fastcgi.server config makes sense?

dschaper: A couple of reasons offhand.

* Do we know that the daemon can write to `/tmp`? I honestly don't know if it can or can not.
* Do we know that `/tmp` is safe to write to, security wise? Can anyone see or access that socket and what are the implications of that.
* Most pressing, do we know that `/tmp` won't get purged by the system since it's really just meant for temporary storage. What happens if `/tmp/pihole-php-fastcgi.socket` is deleted while `lighttpd` is still running?

dschaper: rd mentioned this in the PR message as well.

>Also, the new socket path was causing a permission error when the debug log was executed from the web interface.
This PR fixes the config.

PromoFaux: Just looking over the `lighttpd` docs, all the examples are in `/tmp`, is all.

https://redmine.lighttpd.net/projects/lighttpd/wiki/Mod_fastcgi#:~:text=Starting%20a%20FastCGI%2DPHP

rdwebdesign: I'm really undecided here.

I agree with the idea of using a specific path. I searched the `lighttpd` docs and found only 2 or 3 examples using `/run`... the rest always uses `/tmp`.

On the other hand, I'm not sure about permissions on this path on each distro.
Also, I downloaded the packages for Debian 10 and 11. The original value used on `15-fastcgi-php.conf` file is always using `/run/lighttpd/` _(the one we always used)_.
On Fedora, the default path is initially defined as `/var/lib/lighttpd/sockets/`.

rdwebdesign: After considering the whole discussion here, a lot of testing and reviewing the previous settings, I decided to use the default `15-fastcgi-php.conf` (installed using `lighty-enable-mod fastcgi-php`) and remove the `fastcgi.server` option from the new file.

I will change the PR files.

PromoFaux (approved):

Issue #4783, opened 2022-07-05
[Tracking] Installer assumes users always enter two custom upstream DNS server even if they enter only one


yubiuser:
The whole "Select custom DNS server" part in the installer needs re-working. It assumes, users enter two DNS servers. If I enter only one, both PIHOLE_DNS_1 and PIHOLE_DNS_2 are set to the same address.

mattpace1:
Also, when a user enters more than two DNS servers, only the first and last values are shown (as _PIHOLE_DNS_1_ and _PIHOLE_DNS_2_ respectively)

Issue #4902, opened 2022-09-08
[Tracking] Raspberry Pi OS switch from dhcpd to NM


yubiuser:
Our installer should be compatible with the upcoming change to Raspberry Pi OS 

https://www.raspberrypi.com/news/the-latest-update-to-raspberry-pi-os/

dschaper:
I think our consensus is that we will not be changing to support a new IP stack manager.

yubiuser:
Yes. That's why I created the linked pull request. Once Raspbian makes the switch we can merge the PR (surely will need adaption) and close the issue.

PR #5109, opened 2023-01-05
Feature/network manager dhcp support

libklein:

This PR implements support for setting a static IP via NetworkManager. See #4783.

The codes adds a check if network-manager is installed and if it manages the selected interface. If so, nmcli is used to configure the interface.

Note that this is a preliminary implementation that will need to get tested once Raspberry OS makes the change. Until then it could be used to support more operating systems.

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -716,6 +716,24 @@ find_IPv6_information() {
     fi
 }
 
+# Check if network manager manages the specified device
+doesNMManageIface() {
+    # Does nmcli exist?
+    if ! command -v "nmcli" &> /dev/null; then
+        return 1
+    fi
+
+    iface="$1"
+
+    # Get managed state
+    nm_state=$(nmcli -f GENERAL.NM-MANAGED dev show "$iface")
+    # Output of command
+    if [[ "$nm_state" =~ no$ ]]; then
+        return 1
+    fi
+    return 0
+}
+
 # A function to collect IPv4 and IPv6 information of the device
 collect_v4andv6_information() {
     find_IPv4_information
@@ -725,6 +743,11 @@ collect_v4andv6_information() {
     if [[ -f "/etc/dhcpcd.conf" ]]; then
         # configure networking via dhcpcd
         getStaticIPv4Settings
+        setDHCPCD
+    # Check if network manager is enabled and manages the pihole interface
+    elif doesNMManageIface "${PIHOLE_INTERFACE}" ; then
+        getStaticIPv4Settings
+        setDHCPNM
     fi
     find_IPv6_information
     printf "  %b IPv6 address: %s\\n" "${INFO}" "${IPV6_ADDRESS}"
@@ -822,7 +845,6 @@ It is also possible to use a DHCP reservation, but if you are going to do that, 
             done
             ;;
        esac
-       setDHCPCD
 }
 
 # Configure networking via dhcpcd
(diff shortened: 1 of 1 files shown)


yubiuser:
Thanks for your PR.

However, this will need some discussion. Some time ago we decided to drop `dhcpcd` as a dependency (https://github.com/pi-hole/pi-hole/pull/4260) to not interfere with user's selected network configuration software. We decided to only offer setting a static IP addresses when `dhcpcd` is already installed to lower entry barriers for new users (mainly targeting Raspbian). With the announced switch from `dhcpcd` to NM (https://github.com/pi-hole/pi-hole/issues/4902) our decision needs to be revised.

Do we want to continue to offer setting static address during installation? Do we want to include NM already now? Or only after Raspbian made the change?

__

In any case, we would greatly appreciate a contribution to our documentation (https://docs.pi-hole.net/main/prerequisites/) on how to set a static address for different network managers. We had a draft PR https://github.com/pi-hole/docs/pull/547 but it was abandoned. Maybe you could build on that and write a guide for NM?

PromoFaux:
My personal take on this is that we drop the static IP hand-holding once Raspberry Pi OS goes to NM, though it's not a hill I will die on.

Since we dropped the static address setting on all but systems with `dhcpd5` pre-installed - is there a rough idea of how many people have had to ask us how they set a static IP address on their machine? A quick search and skim on Discourse does not turn up much in the way of requests.

`NetworkManager` support is [already included in Raspberry Pi OS](https://www.raspberrypi.com/news/the-latest-update-to-raspberry-pi-os/#:~:text=closes%20the%20menu.-,NetworkManager%20compatibility,-In%20previous%20releases), though it has to be enabled manually and any configuration re-done. It's probably worth nothing that so far there is [no other mention of `NetworkManager` on their blog](https://www.raspberrypi.com/news/archive/?q=NetworkManager&from_month=&from_year=&to_month=&to_year=), so I don't think we should be making any steps towards this just yet, unless someone can point to some documentation/blog/announcment beyonf the one already linked (which states "but in some future release, NetworkManager will become the default.")

Reads to me rather like it's coming [soon™](https://wowpedia.fandom.com/wiki/Soon) 

---

I'm also of the opinion that static IP addressing should be handled at the router via dhcp reservation, but other peoples opinions on that are certainly valid, too 🙃

dschaper:
I'm happy to drop configuring server networking completely. 

When we started 6 years ago it was easier to be a one-shot install that tried to do as much work for people as possible. We didn't have the resources to walk people through everything and explain the basics of IP addressing along the way. Things certainly have changed since then, there are tons of guides and resources and help with getting people that are new to Linux or new to networking set up. A great driver of that has been the Raspberry Pi and lowering the barrier to entry for people to have access. 

I think we are at the point that we can let this go. Originally the idea was to drop static networking completely but I argued for keeping `dhcpcd5` support there to help people with Raspberry Pi directly. It looks like they are going away from that IP manager and I assume that they are prepared with support and documentation for a major change like `dhcpcd5` to NetworkManager. Their OS is rather opinionated and I don't think we should be in the middle of that, let the OS handle the IP stack using whatever tools they have or will have. 

When you get down to it, Pi-hole is a DNS server and needs to have a stable IP address for clients to use. How you get that stability is up to the user now, we may get an increase of support requests for people that lose Pi-hole from not having a stable IP address but the debugger is pretty accurate with `dhcp` leases and payloads so we can often tell when the issue is a bad or changed IP address.

libklein:
This is my first PR to this project so weight my opinion accordingly. 

I do not see enough value in further offering (automatic) static IP configuration by the installer. In fact, I assume that it creates more harm than good: 
Users who have a valid use case for a configuration like this will know how to configure it manually and probably prefer that. For new users, it's just another option that adds to the confusion.

Moreover, it's something that potentially limits the number of supported distros and causes a maintenance overhead.

Im happy to close this PR in favour if #5111.

dschaper:
How does #4783 relate to the IP address of the Pi-hole node?

libklein:
> How does #4783 relate to the IP address of the Pi-hole node?

Sorry, i guess that was a typo on my side. I was referring to the original issue (#4902). To the best of my knowledge, but correct me if I'm wrong here, there is no further action on the side of Pi-Hole required to support NM.

PR #5128, opened 2023-01-17
Reduce number of false positives when gravity unable to import domains - change the wording of the output

PromoFaux:

Currently when we analyse an adlist, we print out a warning if any invalid domains are detected. However, it is not always strictly true that there is something wrong with the entry - it is more that they are incompatible with Pi-hole

https://discourse.pi-hole.net/t/fresh-install-steven-black-list-invalid-domains/60523

This PR introduces a small list of common localhost domains (as included at the top of StevenBlack's list to discount these as being counted as "invalid", and also changes the wording of invalid domains to unusable entries.

Example output before these changes:
```console
  [i] Analyzed 168834 domains, 15 domains invalid!
      Sample of invalid domains:
      - 0.0.0.0
      - broadcasthost
      - fe
      - ff
  [i] List stayed unchanged

  [i] Analyzed 168834 domains, 15 domains invalid!
      Sample of invalid domains:
      - 0.0.0.0
      - broadcasthost
      - fe
      - ff
  [i] List stayed unchanged
```

Example output after these changes:
```console
  [i] Imported 168819 domains, ignoring 2 non-domains
      Sample of non-domain entries:
      - fe
      - ff

  [i] Target: https://raw.githubusercontent.com/jankytay/pihole-troubleshooting/master/issue5008.txt
  [i] Imported 3 domains, ignoring 5 non-domains
      Sample of non-domain entries:
      - (\.|^)example6\.com$
      - (^|\.)example5\.com
      - (^|\.)example8\.com;reply=none
      - *.example3.com
      - ^example77?
```

The copy could probably be further clarified to mention that this is not an error - just for information, and that nothing is wrong, but it is late and I'm struggling to find the words

---

---

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -244,7 +244,7 @@ database_adlist_number() {
     return;
   fi
 
-  output=$( { printf ".timeout 30000\\nUPDATE adlist SET number = %i, invalid_domains = %i WHERE id = %i;\\n" "${num_source_lines}" "${num_invalid}" "${1}" | pihole-FTL sqlite3 "${gravityDBfile}"; } 2>&1 )
+  output=$( { printf ".timeout 30000\\nUPDATE adlist SET number = %i, invalid_domains = %i WHERE id = %i;\\n" "${num_source_lines}" "${num_unusable}" "${1}" | pihole-FTL sqlite3 "${gravityDBfile}"; } 2>&1 )
   status="$?"
 
   if [[ "${status}" -ne 0 ]]; then
@@ -522,9 +522,9 @@ gravity_DownloadBlocklists() {
 # num_target_lines does increase for every correctly added domain in pareseList()
 num_target_lines=0
 num_source_lines=0
-num_invalid=0
+num_unusable=0
 parseList() {
-  local adlistID="${1}" src="${2}" target="${3}" incorrect_lines sample_incorrect_lines
+  local adlistID="${1}" src="${2}" target="${3}" unusable_lines sample_unusable_lines tmp_unusuable_lines_str false_positive
   # This sed does the following things:
   # 1. Remove all lines containing no domains
   # 2. Remove all domains containing invalid characters. Valid are: a-z, A-Z, 0-9, dot (.), minus (-), underscore (_)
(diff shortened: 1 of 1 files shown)


PromoFaux:
@jfb-pihole can  I have your thoughts on this version, too, please?

jfb-pihole:
"Sample of unusable entries:"

I would change this to "sample of entries not imported into Pi-hole"

rdwebdesign:
The expression "unusable entries" tries to make it clear Pi-hole can't use those entries... 

I know the difference is very small (if any), but I think the expression "not imported into Pi-hole" will raise a lot of questions like "Why Pi-hole didn't import everything? What is wrong?".

We can explain better using something like "X lines without domains not imported by Pi-hole", but this looks too long.

jfb-pihole:
I'm OK with that change.  We just need to indicate that the domains weren't imported.

PromoFaux:
```
  [i] Imported 168819 domains, ignoring 2 non-domains
      Sample of non-domain entries:
      - fe
      - ff

  [i] Target: https://raw.githubusercontent.com/jankytay/pihole-troubleshooting/master/issue5008.txt
  [i] Imported 3 domains, ignoring 5 non-domains
      Sample of non-domain entries:
      - (\.|^)example6\.com$
      - (^|\.)example5\.com
      - (^|\.)example8\.com;reply=none
      - *.example3.com
      - ^example77?
```

gravity.sh
```diff
@@ -522,9 +522,9 @@ gravity_DownloadBlocklists() {
 # num_target_lines does increase for every correctly added domain in pareseList()
 num_target_lines=0
 num_source_lines=0
-num_invalid=0
+num_unusable=0
 parseList() {
-  local adlistID="${1}" src="${2}" target="${3}" incorrect_lines sample_incorrect_lines
+  local adlistID="${1}" src="${2}" target="${3}" unusable_lines sample_unusable_lines
```
yubiuser, line 527:
```suggestion
  local adlistID="${1}" src="${2}" target="${3}" unusable_lines sample_unusable_lines tmp_unusuable_lines_str false_positive
```

gravity.sh
```diff
@@ -534,10 +534,41 @@ parseList() {
...
+    "localhost"
+    "ip6-localhost"
+    "ip6-loopback"
+    "lo0 localhost"
+    "ip6-localnet"
+    "ip6-mcastprefix"
+    "ip6-allnodes"
+    "ip6-allrouters"
+    "ip6-allhosts"
+    "0.0.0.0"
```
yubiuser, line 558:
```suggestion

```

I'm not sure if we should remove `0.0.0.0` here. It's not a domain nor a hostname.

rdwebdesign: I agree. We don't use IP addresses. They are not useful.

rdwebdesign (dismissed):

yubiuser (dismissed):

gravity.sh
```diff
@@ -534,36 +534,65 @@ parseList() {
...
+
+  local tmp_new_imported_total
+  # Get the new number of lines in destination file
+  tmp_new_imported_total="$(grep -c "^" "${target}")"
+  # Number of usable lines is the difference between the new total and the old total. (Or, the number of lines we just added.)
+  num_usable_lines="$(( tmp_new_imported_total-num_total_imported_lines ))"
+  # Replace the total with the new total.
+  num_total_imported_lines="$tmp_new_imported_total"
+  # Get the number of unusable lines
+  num_unusable_lines="${#unusable_lines[@]}"
```
yubiuser, line 581: Maybe this is a clash of semantics vs syntax, but it's not clear what we want to report here: number of unusable lines or unusable domains. How accurate should we be? This number is determined after we removed false-positives and after we removed duplicates.

Image a list has 100 "valid" lines (e.g. no comments) but only 80 domains and 20 non-domains of which 5 are duplicates one one is a false-positive. Which number do you want to show?

rdwebdesign: I'm not sure if we need to be precise about what is not used.
I think the number of imported domains is important and should reflect the final number of imported domains.

To tell truth, the most important feature didn't change: 
Pi-hole is able to filter unusable lines and import real domains found on the list.

jfb-pihole: Report unusable domains.

PromoFaux: Look at removing the false postives and duplicates as being treated the same as removing comments. We don't include the number of comments in the unsuable total, why would we include the false postives and duplicates?

At the end of the process, either one or two numbers are generated per list. 
 - How many domains we could import (the important one)
 - How many non-domain items were left in the file after stripping out all the comments, duplicates, and false positives.

PromoFaux: made the variable names even more concise

yubiuser: Ok, I see your point. But then we should modify

> How many domains we could import (the important one)

because at the moment this number contains all the duplicated imported domains. We should not de-dup the list before sending it to `${target}` because `FTL` will handel duplicates well. Only before counting.

PromoFaux: We are really getting into splitting hairs here...

There is no sense in importing duplicate domains from a single list, but a domain existing on several lists is valid - so it makes sense that the would ultimately contain duplicates. (This is reflected in the final total count output by gravity)

How often are domains duplicated In a single list?

yubiuser (approved):

gravity.sh
yubiuser: > How often are domains duplicated In a single list?

I guess it depends on the quality of the list. SB does not contain any duplicates.

I think you convinced me. And with the new wording it's clearer.

thomasmerz:
Great PR! I **love** it. Because now it's not "something wrong" anymore, but only "an info in a footnote" 😆 

grafik

dschaper:
These changes will lead to a high system load when fed a list that contains nothing of value or use. Any list that is in ABP format will try to filter out the common localhosts from a long string of invalid information (read ABP regex) and Pi-hole will try its hardest to salvage the list. 

Looks like we need to have an intermediate step that looks for any hosts entry in the supplied list and if there are none then stop. I thought we had ABP filtering before but that may have been to try to convert ABP in to hosts format. If I can find that detection regex then we can reuse it to just remove the invalid entries and then count the length of the array of hosts entries. If there are none then we don't need to proceed. Sample list that doesn't work for Pi-hole anymore is https://big.oisd.nl. The older https://dbl.oisd.nl has been removed and we safely note that the list is gone. But people will try to replace dbl with big and we shouldn't crash just because they are feeding garbage to Pi-hole now.

yubiuser:
> Looks like we need to have an intermediate step that looks for any hosts entry in the supplied list and if there are none then stop.

If you want to go for an all-or-nothing approach, just check the number of lines of `${target} and if it's `0` skip the removal of false positive.

dschaper:
Okay, ideally we should only remove false positives from potential positives but I haven't looked too deeply at the code to see how it all is being processed now. (`sed` and `regex` is such a nightmare in `bash`...)

yubiuser:
The `false_positive` is misleading here. They are `false_positive` of being "unusable". We first get all non-domains and remove the `false_positive` from this list leaving the real "unusable". Therefore we first need all the garbage to sort out a few items.

PR #5132, opened 2023-01-18
No need to enable accesslog

yubiuser:
**What does this PR aim to accomplish?:**

Don't enable `accesslog` module during installation. Undo a part of https://github.com/pi-hole/pi-hole/pull/5121.
Currently, consecutive updated of `development` will break, because `lighttpd` won't start. It's complaining about

```
sudo lighttpd -tt -f /etc/lighttpd/lighttpd.conf
Duplicate config variable in conditional 0 global: accesslog.filename
2023-01-18 13:54:31: configfile.c.1970) source: /etc/lighttpd/conf-enabled/10-accesslog.conf line: 4 pos: 1 parser failed somehow near here: (EOL)
2023-01-18 13:54:31: configfile.c.1970) source: /etc/lighttpd/lighttpd.conf line: 75 pos: 1 parser failed somehow near here: (EOL)
```

Reason is, with the enabled `accesslog` module, it will enable `10-accesslog.conf` containing two lines

```
server.modules += ( "mod_accesslog" )

accesslog.filename = "/var/log/lighttpd/access.log"

```

However, we already set a `accesslog.filename` in `15-pihole-admin.conf` and PR https://github.com/pi-hole/pi-hole/pull/5121 also added `server.modules +=  "mod_accesslog"` to that config. So I think enabling `accesslog` is not necessary.

---
**By submitting this pull request, I confirm the following:**

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against.
2. I have commented my proposed changes within the code and I have tested my changes.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
6. I have checked that another pull request for this purpose does not exist.
7. I have considered, and confirmed that this submission will be valuable to others.
8. I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
9. I give this submission freely, and claim no ownership to its content.

---
- [x] I have read the above and my PR is ready for review. *Check this box to confirm*

PR #5133, opened 2023-01-18
Only remove the `server.modules+=` from new config if lighttpd version is < 1.4.56

PromoFaux:

Offered as an alternative to https://github.com/pi-hole/pi-hole/pull/5132, which aims to fix an issue caused by https://github.com/pi-hole/pi-hole/pull/5121. 

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1432,14 +1432,23 @@ installConfigs() {
         elif [[ -d "/etc/lighttpd/conf-available" ]]; then
             conf=/etc/lighttpd/conf-available/15-pihole-admin.conf
             install -D -m 644 -T ${PI_HOLE_LOCAL_REPO}/advanced/pihole-admin.conf $conf
-            # disable server.modules += ( ... ) in $conf to avoid module dups
-            # (needed until Debian 10 no longer supported by pi-hole)
-            # (server.modules duplication is ignored in lighttpd 1.4.56+)
-            if awk '!/^server\.modules/{print}' $conf > $conf.$$ && mv $conf.$$ $conf; then
+
+            # Get the version number of lighttpd
+            version=$(lighttpd -v | grep -o -E '[0-9]+\.[0-9]+\.[0-9]+')
+            # Test if that version is greater than or euqal to 1.4.56
+            if dpkg --compare-versions "$version" "ge" "1.4.56"; then
+                # If it is, then we don't need to disable the modules
                 :
             else
-                rm $conf.$$
+                # disable server.modules += ( ... ) in $conf to avoid module dups
+                # (server.modules duplication is ignored in lighttpd 1.4.56+)
+                if awk '!/^server\.modules/{print}' $conf > $conf.$$ && mv $conf.$$ $conf; then
+                :
+                else
+                    rm $conf.$$
+                fi
             fi
+
             chmod 644 $conf
             if is_command lighty-enable-mod ; then
                 lighty-enable-mod pihole-admin access accesslog redirect fastcgi setenv > /dev/null || true
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1434,14 +1434,14 @@ installConfigs() {
             install -D -m 644 -T ${PI_HOLE_LOCAL_REPO}/advanced/pihole-admin.conf $conf
 
             # Get the version number of lighttpd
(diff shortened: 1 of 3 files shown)


automated install/basic-install.sh
```diff
@@ -1432,14 +1432,23 @@ installConfigs() {
...
+            # Get the version number of lighttpd
+            version=$(dpkg-query -f='${Version}\n' --show lighttpd)
+            # Test if that version is greater than or euqal to 1.4.56
+            if dpkg --compare-versions "$version" "ge" "1.4.56"; then
+                # If it is, then we don't need to disable the modules
                 :
             else
-                rm $conf.$$
+                # disable server.modules += ( ... ) in $conf to avoid module dups
+                # (server.modules duplication is ignored in lighttpd 1.4.56+)
```
yubiuser, line 1444:
```suggestion
                # (server.modules duplication is ignored in lighttpd <=1.4.56)
```

yubiuser (changes requested):
```
rockpi@rockpi-4b:/var/www/html/admin$ sudo lighttpd -tt -f /etc/lighttpd/lighttpd.conf
Duplicate config variable in conditional 0 global: accesslog.filename
2023-01-20 10:24:57: configfile.c.1970) source: /etc/lighttpd/conf-enabled/10-accesslog.conf line: 4 pos: 1 parser failed somehow near here: (EOL)
2023-01-20 10:24:57: configfile.c.1970) source: /etc/lighttpd/lighttpd.conf line: 74 pos: 1 parser failed somehow near here: (EOL)
```

This PR will still have `accesslog.filename` doubled (and conflicting) on upgrades with Debian 11. I think it's then part of `10-access-log.conf` and `15-pihole-admin.conf`

automated install/basic-install.sh
PromoFaux: I think you mean `>=`

PromoFaux: Ah yes I think we should be using `:=` on the following line

https://github.com/pi-hole/pi-hole/blob/9048429bbbc002c851365d3ccef1c5923ef8845c/advanced/pihole-admin.conf#L20

Do we even need to be overriding the access log file name? I don't think any of our scripts depend on it.  (Actually the same could be said for the error log - I don't recall why we prepended `pihole-` to their filenames in the first place... )

yubiuser: Reason must be somewhere here: https://github.com/pi-hole/pi-hole/pull/3732

yubiuser: > I think you mean >=

As we check `"$version" "ge" "1.4.56` and do nothing in that case, the removal happens when the version is smaller than 1.4.56

PromoFaux: Ah, I see what you mean. Looking back at it, this line would be better positioned in the do nothing case.

PromoFaux: @yubiuser - I've just FPd another change - can you try on an existing install again please?

PromoFaux: Though I've just tested this on a friends pi (shh, don't tell them) and it seems OK:

```
pi@raspberrypi:~ $ lighttpd -tt -f /etc/lighttpd/lighttpd.conf
2023-01-21 14:43:57: configfile.c.2274) server.upload-dirs doesn't exist: /var/cache/lighttpd/uploads
```

I think the `/var/cache/` notice is irrelevant in this case?

yubiuser: I'll try later but think it won't work. The path in ` *.debian.conf` and `15-pihole-admin.conf` don't conflict, they are the same. But they conflict with `10-accesslog.conf` because the path is different.

PromoFaux: The paths that were conflicting were the one in `*.debian.conf` and `10-accesslog.conf`. The one in `15-pihole-admin.conf` is employed only when the request falls into the `$HTTP["url"] =~ "^/admin/" {` section of the config.

See https://redmine.lighttpd.net/projects/lighttpd/wiki/Mod_accesslog for an example where multiple acceslog paths are defined.

Like I said - I tested this and found it to be working OK - my only question is the `server.upload-dirs` part, but that is actually a false flag and nothing to do with this PR/issue

yubiuser (approved):

automated install/basic-install.sh
yubiuser: Happy to be proven wrong.

PR #5190, opened 2023-02-28
Change FTLcheckUpdate logic

### LINKED ISSUE #5189: Pi-hole update treats github unavailability as FTL update available
### Versions
$ pihole -v
  Pi-hole version is v5.15.5 (Latest: v5.15.5)
  AdminLTE version is v5.18.4 (Latest: v5.18.4)
  FTL version is v5.21 (Latest: v5.21)

### Platform
- OS and version:  Ubuntu 22.04.2 LTS
- Platform:  VM on VMware Fusion 12

### Expected behavior

When github is unavailable due to being blocked or network down, running a Pi-hole update should report current versions and fail to check for new versions gracefully.

  [i] Checking for updates...
  [i] Pi-hole Core:	up to date
  [i] Web Interface:	up to date
  [i] FTL:		up to date

### Actual behavior / bug

When github is unavailable due to being blocked or network down, running a Pi-hole update causes it to report that an update is available for the FTL component and launch into the update routine, which then fails due to said unavailability.

 [i] Checking for updates...
fatal: unable to access 'https://github.com/pi-hole/pi-hole.git/': Couldn't connect to server
  [i] Pi-hole Core:	up to date
fatal: unable to access 'https://github.com/pi-hole/AdminLTE.git/': Couldn't connect to server
  [i] Web Interface:	up to date
  [i] FTL:		update available
  [i] FTL out of date, it will be updated by the installer.
  [✓] Root user check
  [...]
  Unable to complete update, please contact Pi-hole Support

### Steps to reproduce

1. Render github.com unavailable by either blocking in Pi-hole (if Pi-hole is using itself for DNS) or taking the network connection offline
2. Run `pihole -up`
3. Observe the error above
4. Compare with expected behaviour when github.com is available

## Debug Token

- URL:  https://tricorder.pi-hole.net/GWecwgka/

### Screenshots

n/a

### Additional context

Test Ubuntu VM created for discussion in https://discourse.pi-hole.net/t/error-when-updating-pihole/61177/42

github.com being blocked was the cause of the user's problem, but in testing I observed this behaviour which appears to be a bug.

rdwebdesign:

Fix #5189

Changing the logic of the function:
- execute `curl` in a separate step
- use `api.github.com` and `jq` to retrieve _tag_name_
- show the error message if the variable is empty

none

---

---


automated install/basic-install.sh
```diff
@@ -2442,15 +2442,15 @@ FTLcheckUpdate() {
         if [[ ${ftlLoc} ]]; then
             local FTLversion
             FTLversion=$(/usr/bin/pihole-FTL tag)
+
             local FTLlatesttag
+            FTLlatesttag=$(curl -s https://api.github.com/repos/pi-hole/FTL/releases/latest | jq -r .tag_name)
 
-            if ! FTLlatesttag=$(curl -sI https://github.com/pi-hole/FTL/releases/latest | grep --color=never -i Location: | awk -F / '{print $NF}' | tr -d '[:cntrl:]'); then
+            if [ -z "${FTLlatesttag}" ]; then
```
dschaper, line 2449: This will fail if there is a bad response from the `curl` command.
```sh
Viking:~$ tag=$(curl -s https://api.github.com/repos/pi-hole/FTL/releases/lates | jq -r .tag_name)
Viking:~$ echo $tag
null
```

dschaper (changes requested):
I do think this section of code needs to be redone but I think there's a lot of missed opportunities here. We are using `jq` now so we should actually use it to it's abilities and not just as a dumb parser. 

`jq` has a very rich conditionals and comparisons ability that we can use for checking the output from `curl` and to process that output for validity. See https://manpages.ubuntu.com/manpages/xenial/man1/jq.1.html#conditionals%20and%20comparisons

automated install/basic-install.sh
rdwebdesign: I tested and never got `null`, but I will replace the command with:

```
curl -s https://api.github.com/repos/pi-hole/FTL/releases/lates | jq -r '.tag_name | values'
```

rdwebdesign: Do you want to use `jq` for the test, like this?

```
curl -s https://api.github.com/repos/pi-hole/FTL/releases/lat | jq '.tag_name == "${FTLversion}"'
```

automated install/basic-install.sh
```diff
@@ -2442,15 +2442,15 @@ FTLcheckUpdate() {
...
                 # There was an issue while retrieving the latest version
                 printf "  %b Failed to retrieve latest FTL release metadata" "${CROSS}"
                 return 3
-            fi
-
-            if [[ "${FTLversion}" != "${FTLlatesttag}" ]]; then
+            elif [[ "${FTLversion}" != "${FTLlatesttag}" ]]; then
```
yubiuser, line 2453: You could re-use the `version_converter` we have in PADD to really check if the new version is bigger

https://github.com/pi-hole/PADD/blob/a9681264c83529a2bd826bc86dd933d85b9b15bf/padd.sh#L1078-L1082

automated install/basic-install.sh
rdwebdesign: The current code will work for any `curl` response:
- if `curl` returns a valid JSON, `jq` will try to find and return `tag_name`. 
- if no `tag_name` is found in JSON response, an empty string will be returned.
- if `curl` responds with anything that is not a valid JSON, `jq` command will return an empty string.

Command details:
```console
jq -sRr 'fromjson? | .tag_name | values'
```

- `-s` will read the entire input file into a large array before apply the filters;
- `-R` will not parse the input as JSON. When combined with `-s`, it reads the whole response into a single long string;
- The `?` is the error suppression operator and works like a simple `try/catch`.
- `fromjson?` will **try** to parse the text into JSON format. If not possible will return `null`. 
- `.tag_name` will return the value of "tag_name" (or `null` if not found);
- `values` will filter the result and return only "non null values" (returning an empty string in case of `null`).
- `-r` will output the result as plain text (without the quotes).

automated install/basic-install.sh
rdwebdesign: Included on the new code.

yubiuser (approved):

yubiuser:
```
chrko@ThinkPad-X230:~$ FTLlatesttag=$(curl -s https://api.github.com/repos/pi-hole/FTL/releases/not_available | jq -sRr 'fromjson? | .tag_name | values')
chrko@ThinkPad-X230:~$ echo $FTLlatesttag
```

dschaper:
What are you showing there Yubi? Is that output expected or unexpected.

yubiuser:
It's the expected output. When address is not reachable the output of the whole function is an empty string. As opposed to 
https://github.com/pi-hole/pi-hole/pull/5190#discussion_r1120631680

yubiuser:
Re-requesting review from @dschaper

dschaper:
Is this still needed?

dschaper:
Last call for alcohol, you don't have to go home but you can't stay here... (Close this PR if it's no longer needed please.)

rdwebdesign:
Before this PR was created, an [issue](https://github.com/pi-hole/pi-hole/issues/5189) was opened reporting `curl` command was failing when it didn't receive a response from Github server (network error, server error, Pi-hole blocking github.com, etc).

At that time, you suggested to use `jq` to read the tag name. 

This PR was created to use `curl` + `jq` and answer correctly, no matter what was the `curl` output.

---

The current v6 code uses `Location:` header (like the original code) to identify if tag name is present and return its value.

I don't think we need to use this PR, unless the original issue is still happening.

rdwebdesign:
Reopening because [the original issue is still happening](https://github.com/pi-hole/pi-hole/issues/5189#issuecomment-2870052412) and this PR fixes it.

rdwebdesign:
I will simplify the PR removing some code not currently used in v6.

rdwebdesign:
I opened a [new PR](https://github.com/pi-hole/pi-hole/pull/6229) using the current `development` branch as base and removed all unnecessary code.

Issue #5199, opened 2023-03-06
pihole -q is case sensitive, pihole-FTL regex-test isn't


jpgpi250:
Pi-hole version is v5.15.5 (Latest: v5.15.5)
  AdminLTE version is v5.18.4 (Latest: v5.18.4)
  FTL version is new/adb_style_blocking vDev-da118e8 (Latest: v5.21)

### Platform
raspberrypi 3B
Linux raspberrypi 5.15.84-v7+ #1613 SMP Thu Jan 5 11:59:48 GMT 2023 armv7l GNU/Linux

### Expected behavior

pihole -q reports the same result as pihole-FTL for regexes

### Actual behavior / bug

pihole -q is case sensitive, it shouldn't be!

### Steps to reproduce

my regex:
![image](https://user-images.githubusercontent.com/24669386/223077962-5c488566-5fcc-46a7-a811-38cf430d3ff6.png)

pihole-FTL regex-test **C**ookiebot cookiebot
```
[i] Compiling regex filter...
    Compiled regex filter in 0.117 msec

Checking domain...
    cookiebot matches
   Time: 0.076 msec
```
 pihole -q **C**ookiebot -> **NO regex match reported**
```
 Match found in file:///home/pi/blocklistproject/blocklistproject:
   consent.cookiebot.com
   cookiebot.com
 Match found in http://localhost/mylist.txt:
   consent.cookiebot.com
 Match found in https://v.firebog.net/hosts/neohostsbasic.txt:
   consent.cookiebot.com
 Match found in https://v.firebog.net/hosts/RPiList-Malware.txt:
   resources.cookiebot.com
 Match found in https://www.github.developerdan.com/hosts/lists/ads-and-tracking-extended.txt:
   consent.cookiebot.com
   afdverify.consent.cookiebot.com
```
pihole -q cookiebot
```
 Match found in regex blacklist
   cookiebot
...
```

edit
possible fix (query.sh), **using bash 4.0**: add **,,** to "${options}" e.g. **"${options,,}"**
```
options=$(sed -E 's/ ?-(all|exact) ?//g' <<< "${options,,}")
```

possible fix (query.sh), **using tr**: replace **"${options}"** with **```$```(echo "${options}") | tr '[:upper:]' '[:lower:]')**
```
options=$(sed -E 's/ ?-(all|exact) ?//g' <<< $(echo "${options}") | tr '[:upper:]' '[:lower:]')
```
/edit

dschaper:
[...]

PR #5200, opened 2023-03-06
Convert domain to lowercase in pihole -q

yubiuser:

Converts the domain passed to `pihole -q` to lowercase before running the actual search.

Fixes https://github.com/pi-hole/pi-hole/issues/5199

` tr '[:upper:]' '[:lower:]'`

---

---

### DIFF
--- a/advanced/Scripts/query.sh
+++ b/advanced/Scripts/query.sh
@@ -57,12 +57,12 @@ options=$(sed -E 's/ ?-(all|exact) ?//g' <<< "${options}")
 case "${options}" in
     ""             ) str="No domain specified";;
     *" "*          ) str="Unknown query option specified";;
-    *[![:ascii:]]* ) domainQuery=$(idn2 "${options}");;
-    *              ) domainQuery="${options}";;
+    *[![:ascii:]]* ) rawDomainQuery=$(idn2 "${options}");;
+    *              ) rawDomainQuery="${options}";;
 esac
 
 # convert the domain to lowercase
-domainQuery=$(echo "${domainQuery}" | tr '[:upper:]' '[:lower:]')
+domainQuery=$(echo "${rawDomainQuery}" | tr '[:upper:]' '[:lower:]')
 
 if [[ -n "${str:-}" ]]; then
     echo -e "${str}${COL_NC}\\nTry 'pihole -q --help' for more information."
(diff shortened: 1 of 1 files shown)


advanced/Scripts/query.sh
```diff
@@ -57,6 +57,9 @@ scanList(){
     esac
 }
 
+# convert all options (including the domain to lowercase)
+options=$(echo "${options}" | tr '[:upper:]' '[:lower:]')
```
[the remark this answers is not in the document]
dschaper: Doing that will convert all options to lowercase. Linux is case sensitive and only the items that we know should be case insensitive should be converted. Just convert the domains to lower please.

advanced/Scripts/query.sh
```diff
@@ -88,11 +61,41 @@ case "${options}" in
     *              ) domainQuery="${options}";;
 esac
 
+# convert the domain to lowercase
+domainQuery=$(echo "${domainQuery}" | tr '[:upper:]' '[:lower:]')
```
orazioedoardo, line 65:
```suggestion
domainQuery="${domainQuery,,}"
```

PromoFaux: Although it looks nicer - going forward we are trying to avoid bashisms (we're not strictly enforcing this at the moment), with the ultimate goal being posix compatibility and increased portability.

PromoFaux: Absolutely we are, which is why I said going forward.

I think the unwritten/unspoken rule at the moment is "If we can do it without a bashism, then that is the preference". At some point we will sit down and convert entire scripts, but as you can imagine that is a daunting task and there is only so much free time!

utils.sh and version.sh are already posix compliant (as is PADD)

advanced/Scripts/query.sh
```diff
@@ -88,11 +61,41 @@ case "${options}" in
     *              ) domainQuery="${options}";;
```
dschaper, line 61: Suggest you use two different variable names for this process. I don't like modifying the variable and copying it back to itself when shell is the language. 

`*   ) rawDomainQuery="${options}";;`

`domainQuery=$(stuff with $rawDomainQuery)`

dschaper (changes requested):

advanced/Scripts/query.sh
yubiuser: Pushed

dschaper (approved):

PR #5211, opened 2023-03-17
Fix addKey to handle substrings of existing keys

wblew:

This PR intends to address the case where a second upstream DNS server's address is a substring of the first upstream DNS
server's address. For example: using the Pi-hole docker image, with `PIHOLE_DNS_="192.168.1.178;192.168.1.1".`

Fix `addKey` in `utils.sh` to handle the case where a key is being added, and that key is the leading substring of an already existing key within that file. The new test `test_key_addition_substr' is added, to verify this behaviour is correct.

For example: add `"server=192.168.1.1"`, when `"server=192.168.1.178"` already exists within the `etc/dnsmasq.d/01-pihole.conf` file.

Check pihole docker with `PIHOLE_DNS_="192.168.1.178;192.168.1.1"`. Its `/etc/dnsmasq/01-pihole.conf` will be missing its second `server=` entry.

None necessary, this is a bug fix. There are no intended changes in observable behaviour.

---

---
- [XXX ] I have read the above and my PR is ready for review. *Check this box to confirm*


wblew:
I added a pretty heft comment to the addKey function, to describe the situation. Depending on project expectations, that comment might need trimming, or even complete removal.

Such a decision is better in the hands of more experienced project contributors.

advanced/Scripts/utils.sh
```diff
@@ -57,7 +57,12 @@ addKey(){
   # touch file to prevent grep error if file does not exist yet
   touch "${file}"
 
-  if ! grep -q "^${key}" "${file}"; then
+  # match $key against entire line. Using -x assumes that
+  # the key within file always lacks bounding whitespace.
+  # The entire line test is needed to correctly determine
+  # key is absent when its a leading substring of another
+  # existing key within file.
+  if ! grep -x -q "${key}" "${file}"; then
```
yubiuser, line 65:
```suggestion
  if ! grep -q "^${key}$" "${file}"; then
```

yubiuser (commented):
Thanks for your PR. This clearly an issue which should be fixed.
Maybe it's easier just to add the anchor at the end of the line as well?

wblew:
Thank you @yubiuser for the quick review. Using anchors is certainly clearer, and that makes it better, right?

Updated to reflect your review comment.

wblew:
Thank you @yubiuser for the quick review. Using anchors is certainly clearer, and that makes it better, right?

Updated to reflect your review comment.

wblew:
> Why did you close the PR? __
> 
> I think the long comment can go now and please squash all your commits down. I think afterwards it's ready to go...

Because I am an experienced idiot? How do I get it open again?

yubiuser (dismissed):

dschaper:
Sorry, I'd like the comment to be there please. 

@PromoFaux This potentially affects environment variables for docker. Please check this over.

wblew:
> Sorry, I'd like the comment to be there please.
> 
> @PromoFaux This potentially affects environment variables for docker. Please check this over.

I failed to add my signoff, in the first attempt. Here is another try to restore the comment, with my signoff.

PromoFaux:
I started off a little confused as to what was being fixed/changed here..

The issue you describe with docker does not exist as you describe it - at least, I cannot reproduce it:

image

Which is because they are added as `PIHOLE_DNS_1=`, and `PIHOLE_DNS_2=` respectively...

But then I re-read the issue and realised that the issue is actually nothing to do with docker at all, rather it is  when `setupVars.conf` is transposed into `/etc/dnsmasq.d/01-pihole.conf` in webpage.sh, here:

https://github.com/pi-hole/pi-hole/blob/d86b325dfe4d39b7bc523672bb22f1bdbd12265f/advanced/Scripts/webpage.sh#L157

---

That was all a very long way of saying "I've tested this and confirm it does what it says on the tin"

PromoFaux (approved):

wblew:
Yes, my problem statement is woefully incomplete. Thus causing confusion. Why is hindsight always so clear?

wblew:
Thank you all. For your time, and patience, and advice.

Issue #5246, opened 2023-04-05
Adblock Plus style adlist domain entries with CSS element selector should be ignored


RoHapp:
### Versions
- Pi-hole:  v5.16.2
- AdminLTE:  v5.19
- FTL:  v5.22

### Expected behavior

Entries with a CSS element selector should not be blocked.

### Actual behavior / bug

Example line from easylist:
```
coin360.com##.b5BiRm
```
The site `coin360.com` gets blocked because it has entries in the `easylist.txt` following a CSS element selector.  
The CSS selector `##.b5BiRm` is an [element selector](https://adblockplus.org/filter-cheatsheet#elementselection).
Lines containing this selector should be ignored.
The [Extended CSS selectors (Adblock Plus specific)](https://adblockplus.org/filter-cheatsheet#elementhideemulation) should probably also be ignored.

In the lower part of the easylist are a lot of domains affected by this bug.

yubiuser:
Please post the output of `pihole -q coin360.com`

RoHapp:
Output of `pihole -q coin360.com`
```
$ pihole -q coin360.com
 Match found in exact whitelist
   coin360.com
 Match found in https://easylist.to/easylist/easylist.txt:
   coin360.com
   coin360.com
   coin360.com
   coin360.com
   coin360.com
   coin360.com
   coin360.com
```

yubiuser:
Thanks.

This is a tricky task. The reason we see this behaviour is because we remove trailing comments from lists. We do this, because it allows users to annotate (comment) on list entries. This was not an issue so far, because we did not supported any ABP style at all.

https://github.com/pi-hole/pi-hole/blob/d885e92674e8d8d9a673b35ae706b2c49ea05840/gravity.sh#L769

How do you want do discriminate between a user comment (could also have two ## in a row) and a ABP CSS selector?

RoHapp:
Oh, yeah this is tricky. Do you have a example list with such a comment?. I would suspect a comment to have a space or tab before the double pound sign (##).  
Example:
```
example.com ##Great Website
```
Is it a good idea to discriminate with the space or tab before the comment?

yubiuser:
After some internal discussion we're likely going to remove all lines from the input file which contain a patter like `#[?@$](0,1)#.*$`. That should strip out all lines with extended CSS selectors. We're going to do this before removing trailing comment so users can keep existing comments. Only limitation is that comments can't include `##, `#$#`,... anymore.

[...]

PR #5247, opened 2023-04-07
Filter ABP CSS selectors

yubiuser:

Some ABP style lists contain lines with [(extended) CSS selectors](https://adblockplus.org/filter-cheatsheet#elementhideemulation): `##`, `#$#`, `#?#` and `#@#`

Currently our code strips all comments from the input file, converting e.g. `good.com##bad-element` to `good.com` leading to overblocking.
This PR deletes lines with CSS selectors **before** stripping comments.

Fixes https://github.com/pi-hole/pi-hole/issues/5246

---

---

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -770,7 +770,7 @@ gravity_ParseFileIntoDomains() {
   # 2) Remove carriage returns
   # 3) Remove lines starting with ! (ABP Comments)
   # 4) Remove lines starting with [ (ABP Header)
-  # 5) Remove lines containing ABP extended CSS selectors ("##", "#!#", "#@#", "#?#")
+  # 5) Remove lines containing ABP extended CSS selectors ("##", "#!#", "#@#", "#?#") preceded by a letter
   # 6) Remove comments (text starting with "#", include possible spaces before the hash sign)
   # 7) Remove leading tabs, spaces, etc. (Also removes leading IP addresses)
   # 8) Remove empty lines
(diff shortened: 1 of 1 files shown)


yubiuser:
I also removed the  `sed` that would remove lines containing `/` prior to paring. Such lines (e.g. whole URLs) should be marked as invalid which the are not when they were removed before parsing.

gravity.sh
```diff
@@ -768,25 +768,21 @@ gravity_ParseFileIntoDomains() {
...
+  # 5) Remove lines containing ABP extended CSS selectors ("##", "#!#", "#@#", "#?#")
+  # 6) Remove comments (text starting with "#", include possible spaces before the hash sign)
+  # 7) Remove leading tabs, spaces, etc. (Also removes leading IP addresses)
+  # 8) Remove empty lines
+
+    sed -i -r \
+    -e 's/\r$//' \
+    -e 's/\s*!.*//g' \
+    -e 's/\s*\[.*//g' \
+    -e '/\#[!|?|@]{0,1}\#/d' \
```
rdwebdesign, line 782: 1. You don't need the `|` here. This `[!?@]` already means "Match a single character present in the list".
2. `#$#` is also a possible rule. You forgot to include the `$` to the regex.
3. **Note:** I'm not sure if there is a `#!#` rule.

rdwebdesign (dismissed):

gravity.sh
yubiuser: You're 100% correct.

PromoFaux (approved):

PR #5252, opened 2023-04-15
Trigger stale workflow on issue comments to remove stale label immediately

yubiuser:

We use the `stale` action to mark issues stale after 30 days of inactivity. The actions runs only once a day to mark stale issues. If updates to issues occur (e.g. comments) the action removes the stale label. However, this happens so far only once a day - in the worst case leaving a "gap" of 23:59 between comments and label removal.

This PR adds a trigger to the stale workflow to run on `issue_comment`so that new comments should remove the label immediately.

---

---

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -4,6 +4,7 @@ on:
   schedule:
     - cron: '0 8 * * *'
   workflow_dispatch:
+  issue_comment:
 
 jobs:
   stale:
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -6,6 +6,9 @@ on:
   workflow_dispatch:
   issue_comment:
 
+env:
+  stale_label: stale
+
 jobs:
   stale_action:
     if: github.event_name != 'issue_comment'
@@ -20,7 +23,7 @@ jobs:
           days-before-stale: 30
           days-before-close: 5
           stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Please comment or update this issue or it will be closed in 5 days.'
-          stale-issue-label: 'stale'
+          stale-issue-label: $stale_label
           exempt-issue-labels: 'Internal, Fixed in next release, Bug: Confirmed, Documentation Needed'
           exempt-all-issue-assignees: true
           operations-per-run: 300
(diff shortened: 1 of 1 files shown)


dschaper:
Will this both mark and unmark on every comment or just unmark on every comment?

yubiuser:
As this triggers the whole workflow, it will also trigger marking issues as stale if they are due.

PromoFaux:
I think running it on every `issue_comment` is overkill, and may even lead us into hitting rate limits on the Github API (1000 requests per hour per repo).. maybe not solely this, but in combination with other actions.

> However, this happens so far only once a day - in the worst case leaving a "gap" of 23:59 between comments and label removal.

Does this cause an issue? We know that once a comment is added to an issue then next time the action runs the stale label will be removed.

How about instead just changing the schedule to be every `n` hours - with `n` to be decided. (But maybe 12?)

yubiuser:
> Does this cause an issue? We know that once a comment is added to an issue then next time the action runs the stale label will be removed.

It can be irritating for users if the label is not removed shortly after a new comment was added. I agree with
https://github.com/pi-hole/AdminLTE/issues/2534#issuecomment-1484106470

__

We're not the getting that many comments... How about changing the workflow to not trigger the whole `stale action` but only removing the label if there is any?

.github/workflows/stale.yml
```diff
@@ -24,3 +25,18 @@ jobs:
...
+    if: github.event_name == 'issue_comment'
+    permissions:
+      contents: read #  for actions/checkout
+      issues: write #  to edit issues label
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3.4.0
+      - name: Remove 'stale' label
+        run: gh issue edit ${{ github.event.issue.number }} --remove-label 'stale'
```
dschaper, line 39: Can you pull out the label in to a workflow variable and use that here on this line and in the `stale-issue-label` on line 23?

This will prevent changing either this line or line 23 in the future and neglecting to change both of them. 

https://docs.github.com/en/actions/learn-github-actions/variables#defining-environment-variables-for-a-single-workflow

dschaper (changes requested):

dschaper (approved):

PR #5259, opened 2023-04-21
Allow to checkout tags via `pihole checkout`

yubiuser:

Allow to checkout specific tags via `pihole checkout` for `core` and `web`. So far only branches were allowed.

P.S. Checking out tags for `FTL`was possible already (as we don't download via git repo but a binary directly from our own server).

1. Fetch with `--tags` added when getting upstream references.
2. When performing a "checkout" we check the upstream repo if the requested `ref` is a branch or a `tag` and depending on that we create a new branch with or without `--tracking` set (which is set by default when using `checkout` without `-b/B`)
3. "Guard" all `git pull` executions by first checking if the current branch is a branch or tag upstreams and skipping the pull in the latter case (does not make any sense to pull in case we checked out a tag)
4. Moves all relevant functions from `basic-install.sh` to `piholeCheckout.sh` as they have been used only there and nowhere else.

---

6. I have commented my proposed changes within the code and I have tested my changes.
7. I am willing to help maintain this change if there are issues with it later.
8. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
9. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
10. I have checked that another pull request for this purpose does not exist.
11. I have considered, and confirmed that this submission will be valuable to others.
12. I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
13. I give this submission freely, and claim no ownership to its content.

---


rdwebdesign:
I tried the branch _(I know it's still a draft)_.

This is the current output:
```console
root@pihole:/# pihole checkout web v5.15          
  Please note that changing branches or tags severely alters your Pi-hole subsystems

  [✓] Fetching branches/tags from https://github.com/pi-hole/AdminLTE.git
  [i] 127 branches/tags available for Web Admin

  [✓] Switching to branch/tag: 'v5.15' from 'refs/heads/devel'
You are not currently on a branch.
Please specify which branch you want to merge with.
See git-pull(1) for details.

    git pull <remote> <branch>
```

The tag was correctly used:
```console
root@pihole:/var/www/html/admin# git status
HEAD detached at v5.15
nothing to commit, working tree clean
```

But `pihole -v` still shows the branch `devel` and the original commit:
```console
root@pihole:/var/www/html/admin# pihole -v
  Pi-hole version is checkout/tags v5.15.5-81-g6d2ceca6 (Latest: v5.16.2)
  AdminLTE version is devel v5.19-33-gefddf9c7 (Latest: v5.19)
  FTL version is development vDev-b0bf9c0 (Latest: v5.22)
```

advanced/Scripts/piholeCheckout.sh
```diff
@@ -111,59 +193,59 @@ checkout() {
...
     elif [[ "${1}" == "core" ]] ; then
-        str="Fetching branches from ${piholeGitUrl}"
+        str="Fetching branches/tags from ${piholeGitUrl}"
         echo -ne "  ${INFO} $str"
         if ! fully_fetch_repo "${PI_HOLE_FILES_DIR}" ; then
             echo -e "${OVER}  ${CROSS} $str"
             exit 1
         fi
-        corebranches=($(get_available_branches "${PI_HOLE_FILES_DIR}"))
+        corerefs=($(get_available_refs "${PI_HOLE_FILES_DIR}"))
```
rdwebdesign, line 202: https://www.shellcheck.net/wiki/SC2207

```suggestion
        corerefs=( "$(get_available_refs "${PI_HOLE_FILES_DIR}")" )
```

rdwebdesign (dismissed):

yubiuser:
This PR needed more changes than I thought. I squashed and forced push, now there are two commits: the first includes all necessary changes (for easier review) and the second moves the relevant functions from `basic-install.sh` to `piholeCheckout.sh`.
I'll adjust the PR message to explain all changes. The PR should be ready now for testing.

advanced/Scripts/update.sh
```diff
@@ -56,14 +58,26 @@ GitCheckUpdateAvail() {
...
 
-        # The suffix @{upstream} to a branchname
-        # (short form <branchname>@{u}) refers
-        # to the branch that the branch specified
-        # by branchname is set to build on top of#
-        # (configured with branch.<name>.remote and
-        # branch.<name>.merge). A missing branchname
-        # defaults to the current one.
-        REMOTE="$(git rev-parse "@{upstream}")"
+        # check if the local branch ef is a branch or a tag on remote repo
```
DL6ER, line 61:
```suggestion
        # check if the local branch ref is a branch or a tag on remote repo
```

automated install/basic-install.sh
```diff
@@ -472,11 +473,26 @@ update_repo() {
...
     git clean --quiet --force -d || true # Okay for already clean directory
-    # Pull the latest commits
-    git pull --no-rebase --quiet &> /dev/null || return $?
-    # Check current branch. If it is master, then reset to the latest available tag.
+
+    # Check current branch.
+    # Remove any leading "heads/" as $curBranch is later used to check if local branch is remote branch or tag
+    curBranch=$(git rev-parse --abbrev-ref HEAD| sed "s/^heads\///g")
+
+    # check if the local branch ef is a branch or a tag on remote repo
```
DL6ER, line 481:
```suggestion
    # check if the local branch ref is a branch or a tag on remote repo
```

DL6ER (commented):
Really very minor criticism at this point (only two typos). I will have to test this in on a new install with a shallow cloned repository to be meaningful

advanced/Scripts/update.sh
```diff
@@ -56,14 +58,26 @@ GitCheckUpdateAvail() {
...
-        # (configured with branch.<name>.remote and
-        # branch.<name>.merge). A missing branchname
-        # defaults to the current one.
-        REMOTE="$(git rev-parse "@{upstream}")"
+        # check if the local branch ref is a branch or a tag on remote repo
+        if git show-ref -q --verify "refs/remotes/origin/$curBranch" 2>/dev/null; then
+            ref_is_tag=false
+        elif git show-ref -q --verify "refs/tags/$curBranch" 2>/dev/null; then
+            ref_is_tag=true
+        fi
```
DL6ER, line 66: I see no `else` here, what happens if neither of the two are true? `ref_is_tag` seems to stay random in this case (empty, likely). Can this happen? (I'm thinking about users doing local changes and committing them so the head is not pointing to anything on `origin` and also isn't a tag)

automated install/basic-install.sh
```diff
@@ -472,11 +473,26 @@ update_repo() {
...
+    # Check current branch.
+    # Remove any leading "heads/" as $curBranch is later used to check if local branch is remote branch or tag
+    curBranch=$(git rev-parse --abbrev-ref HEAD| sed "s/^heads\///g")
+
+    # check if the local branch ref is a branch or a tag on remote repo
+    if git show-ref -q --verify "refs/remotes/origin/$curBranch" 2>/dev/null; then
+        ref_is_branch=true
+    elif git show-ref -q --verify "refs/tags/$curBranch" 2>/dev/null; then
+        ref_is_branch=false
+    fi
```
DL6ER, line 486: See my comment in `update.sh`, this is the same code.

DL6ER (commented):
What I did during review was creating a fresh VM and installing latest Pi-hole on it. Then I replaced the files in ´/opt/pihole` with the versions proposed in this PR.

Then, I tried `pihole checkout web v5.18`:
Screenshot from 2023-05-12 10-18-28

Seems to work fine.

Then I repeated this for `core` (where `v5.18` does not exist), and this seems to work, too:
Screenshot from 2023-05-12 10-19-15
(cropped screenshot)

This also revealed some interesting tag that does not show up on Github:
Screenshot from 2023-05-12 10-21-56

All in all this seems to be doing what it supposed to do. I don't think we'll be able to anticipate any possible user system (mis)configuration but for a plain simple-and-easy standard installation, this seems to work fine.

As changing the `update.sh` script has the potential to break things in a bad way, I'd liek to ask @dschaper for his opinion. The change in this script is small.

yubiuser:
The failure is expected as you only replaced the files in `/opt/pihole` but not the local repo in `/etc/.pihole`. We call the installer from the repo after checkout of `core`

https://github.com/pi-hole/pi-hole/blob/ff91db7b9af422f1ce6c6aaeadb2d5f4b5e2260a/advanced/Scripts/piholeCheckout.sh#L282-L283

The old file is missing the handling of `tags`

https://github.com/pi-hole/pi-hole/blob/ff91db7b9af422f1ce6c6aaeadb2d5f4b5e2260a/automated%20install/basic-install.sh#L488-L492

DL6ER:
I also replaced them in the repo (and committed that to `master` to not be in a dirty state), but this was obviously overwritten with
```
  [✓] Switching to branch/tag: 'v5.16.2' from 'refs/heads/master'
  [i] Running installer to upgrade your installation
```

So is this a general issue here? Whenever users will want to checkout something *before* (the coming) v5.17 possibly including this, they will *always* run the installer that is not aware of tags or am I mistaken here?

yubiuser:
I fear you are right. Checking out any tag before this is merged and tagged will fail on `core` because the handing of tags obviously was not implemented back then.
How to proceed? We could prevent checking out tags before 5.17?

DL6ER:
> We could prevent checking out tags before 5.17?

I'd say it depends on how much code would be required to do this. We could also simply show an additional warning when we detect that users try to checkout core + tag telling them that below v5.17 won't work and giving the option to cancel without actually doing anything `[y/N]` at this point

yubiuser:
It would be quite some code.... I tried yesterday: we would separate getting `branches` and `tags`, convert the tags to numbers (could re-use some code from PADD) and only include tags above a certain number.

yubiuser:
```
pi@s740:~$ pihole checkout core v5.9.1
  Please note that changing branches or tags severely alters your Pi-hole subsystems

  [✓] Fetching branches/tags from https://github.com/pi-hole/pi-hole.git
  [i] 122 branches/tags available for Pi-hole Core

  [i] Switching to branch/tag: 'v5.9.1' from 'refs/heads/checkout/tags'

  [i] Your are trying to checkout the tag v5.9.1 on the core repo
  Please note that checking out core tags <v5.17  will break your local repo 
  Proceed anyway? [y/N] 
```

PR #5266, opened 2023-05-02
Remove unused code from query.sh

yubiuser:

Removes unused code from query.sh.

Within the `scanList()` function was a `case` selection for 3 different cases - however `scanList()`was only called once within `query.sh` - with `regex` passed as argument. This PR removed all but the `regex` case

---

---

### DIFF
--- a/advanced/Scripts/query.sh
+++ b/advanced/Scripts/query.sh
@@ -69,31 +69,16 @@ if [[ -n "${str:-}" ]]; then
     exit 1
 fi
 
-# Scan an array of files for matching strings
-scanList(){
-    # Escape full stops
-    local domain="${1}" esc_domain="${1//./\\.}" lists="${2}" list_type="${3:-}"
+# Scan a domain again a list of RegEX
+scanRegExList(){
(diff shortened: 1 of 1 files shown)


advanced/Scripts/query.sh
```diff
@@ -69,31 +69,16 @@ if [[ -n "${str:-}" ]]; then
...
-                fi
-            done;;
-        *       ) grep -i "${esc_domain}" "${lists}" /dev/null 2>/dev/null;;
-    esac
+# Scan a domain again a list of RegEX
+scanRegExList(){
+    local domain="${1}" list="${2}"
+
+    for entry in ${list}; do
+        if [[ "${domain}" =~ ${entry} ]]; then
```
DL6ER, line 77: This code here is not really true but this is unrelated to this particular PR. Does this even work at all with regular expressions that use Pi-hole's extensions? I very much doubt so. I'd propose to use `pihole-FTL regex-test` here, however, I feel we are sufficiently advanced on our way to v6.0 and can simply do it over there (ask the API).

DL6ER (approved):

PR #5270, opened 2023-05-07
Don't source the install script in webpage.sh

yubiuser:

While working on the installation script locally, I noticed a sourcing loop which prevents testing certain changes to the install script.
During installation, we source `webpage.sh` 

https://github.com/pi-hole/pi-hole/blob/d885e92674e8d8d9a673b35ae706b2c49ea05840/automated%20install/basic-install.sh#L1857

And one of the first things we do in `webpage.sh` is to source the install script **from the repo**
https://github.com/pi-hole/pi-hole/blob/d885e92674e8d8d9a673b35ae706b2c49ea05840/advanced/Scripts/webpage.sh#L27C8-L29

This leads to overwriting any function in the install script that initially called `webpage.sh` by the functions defined in the repo (which is stock master branch). Hence, downstream modifications to the install script will be reset.

___

We source the installer script in`webpage.sh` for a few variables and `valid_ip()` and `valid_ip6()`.I tried to move the functions to `utils.sh` but this does not work because the installer requires the function early and before we download the `utils.sh` script.

Duplicate  `valid_ip()` and `valid_ip6()` into `webpage.sh` and remove sourcing of the installer script.

---

---

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -22,12 +22,14 @@ readonly dnscustomcnamefile="/etc/dnsmasq.d/05-pihole-custom-cname.conf"
 
 readonly gravityDBfile="/etc/pihole/gravity.db"
 
-# Source install script for ${setupVars}, ${PI_HOLE_BIN_DIR} and valid_ip()
-readonly PI_HOLE_FILES_DIR="/etc/.pihole"
-# shellcheck disable=SC2034  # used in basic-install to source the script without running it
-SKIP_INSTALL="true"
-source "${PI_HOLE_FILES_DIR}/automated install/basic-install.sh"
 
+readonly setupVars="/etc/pihole/setupVars.conf"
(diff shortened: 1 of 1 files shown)


PromoFaux:
Can I suggest that if we're going to be duplicating them, we duplicate them into `utils.sh` anyway, and source that in `webpage.sh`? `webpage.sh` will be disappearing in v6 I think (It's been a while since I worked on the `development-v6` branch in anger, so I may not be as up to speed as I remember!)

yubiuser:
> webpage.sh will be disappearing in v6 I think

This is the reason I would not put in `utils.sh`- it will be gone in `v6` "automatically". If it would be in `utils.sh` we would need to remember to remove it.

DL6ER (approved):
This file will be largely removed in v6.0, I'm not sure what we will keep else than a convenience CLI wrapper for setting the API password.

PR #5275, opened 2023-05-14
Use list parsing functionality of FTL

DL6ER:

Use https://github.com/pi-hole/FTL/pull/1559 for list parsing during gravity for both a speedup and reduction in memory footprint. While the current gravity implementation needs 3x the memory of the lists (downloaded files, temporary file and database), the new implementation uses only the minimum (downloaded files are directly stored in the database). Needs the FTL branch to be checked out to work.

---

## Checklist:

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -129,7 +129,7 @@ gravity_swap_databases() {
   echo -e "${OVER}  ${TICK} ${str}"
 
   if $oldAvail; then
-    echo -e "  ${TICK} The old database remains available."
+    echo -e "  ${TICK} The old database remains available"
   fi
 }
 
@@ -140,18 +140,6 @@ update_gravity_timestamp() {
 
   if [[ "${status}" -ne 0 ]]; then
     echo -e "\\n  ${CROSS} Unable to update gravity timestamp in database ${gravityDBfile}\\n  ${output}"
-    return 1
-  fi
-  return 0
-}
-
-# Update timestamp when the gravity table was last updated successfully
-set_abp_info() {
-  pihole-FTL sqlite3 "${gravityDBfile}" "INSERT OR REPLACE INTO info (property,value) VALUES ('abp_domains',${abp_domains});"
-  status="$?"
-
-  if [[ "${status}" -ne 0 ]]; then
-    echo -e "\\n  ${CROSS} Unable to update ABP domain status in database ${gravityDBfile}\\n  ${output}"
     return 1
   fi
   return 0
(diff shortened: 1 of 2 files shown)


rdwebdesign (dismissed):

gravity.sh
```diff
@@ -239,17 +227,6 @@ database_table_from_file() {
     echo -e "  ${CROSS} Unable to remove ${tmpFile}"
 }
 
-# Update timestamp of last update of this list. We store this in the "old" database as all values in the new database will later be overwritten
-database_adlist_updated() {
```
yubiuser, line 243: This function was removed but I don't see a corresponding function in https://github.com/pi-hole/FTL/pull/1559

gravity.sh
```diff
@@ -145,18 +145,6 @@ update_gravity_timestamp() {
   return 0
 }
 
-# Update timestamp when the gravity table was last updated successfully
-set_abp_info() {
```
yubiuser, line 149: We can't remove this function. It collects info about all lists and updates the `abp_domains` info field depending on the ABP status of all lists. It was replaced by the following logic in FTL

https://github.com/pi-hole/FTL/blob/c2bdc24772487fd4870e71f1ec2f1a2c2a729f05/src/gravity-tools.c#L299-L310

But this will fail: it only considers a single list and sets the value always to `1`. It does not take into account the situation where users had ABP style domains and removed the relevant adlists and now there are no ABP style domains left. It won't change the field back to `0`

yubiuser (dismissed):

gravity.sh
rdwebdesign: We need to add an `else` to mark as `0`.

PromoFaux: Perhaps the code on the other side should be more like:

```c
if(abp_domains > 0)
	{
		sql = "INSERT OR REPLACE INTO info (property,value) VALUES ('abp_domains',1);";
else
        {
                sql = "INSERT OR REPLACE INTO info (property,value) VALUES ('abp_domains',0);";
}

// Execute SQL


```

yubiuser: This is not enough. This can't be handled by FTL as long it does work on every list on it's own. `abp_domains` is a summary field which can only be set after all lists have been parsed.

rdwebdesign: You are correct... the current FTL code is seeing lists individually. The value should be set per "gravity.sh" run, as a whole.

PromoFaux: Ah right, OK. Fair point.

Silly question, then. Any argument against a bit/boolean in the adlist table? If we want to know if there are any abp domains it's simply `(Select count(1) from adlist where abp=1) > 0`

rdwebdesign: > Any argument against a bit/boolean in the adlist table?

> It's used to signal FTL if there is any ABP style domains in gravity to activate the ABP style checking


We can do that, but we will need to find a way to keep this value in sync with gravity table.

Currently we store the global `abp_domains` flag in `info` table.
This value only changes when we run gravity. FTL reads the value when needed. 
When we delete adlists (via web interface, for example) nothing changes until we run gravity again. 
Both `gravity` table and the `abp_domains` flag change at the same time, only when we run gravity.

On the other hand, if we store a flag per adlist, in `adlists` table, FTL would need to read this value from each adlist (`SELECT count(1) FROM adlist WHERE abp=1`).
I'm not sure how serious this could be (probably not very serious), but a problem could happen if we delete adlists without updating gravity. 
`gravity` table will only change if we run gravity. The `SELECT count` will change immediately and the values won't be in sync.

**Example:**
We start with 2 lists and one of them uses ABP style. Then we delete this list from the table without running gravity.
That `SELECT count` result would be "0", but gravity table will still have ABP entries.

DL6ER: The logic behind this is the following: We always start with the empty database where `abp_domains` is unset. So whenever FTL sees an ABP list, it adds the `abp_domains = 1`. If there is at least one ABP list during gravity, the property is set. It is never reset but always unset for new gravity databases created when running `pihole -g`. When there are no ABP domains during a gravity run, the property remains completely unset so FTL knows there are no ABP lists. - I think this is correct

gravity.sh
DL6ER: https://github.com/pi-hole/FTL/pull/1565

gravity.sh
DL6ER: There was still a logic quirk as the previous value for `abp_domains` has been copied from the old database. https://github.com/pi-hole/pi-hole/pull/5275/commits/c92826c1525c80cddca3d6fde7e2d8ce8a96bb46 fixes this.

yubiuser: With the mentioned commit it works as expected. `abp_domains` is unset if no ABP domains are present and set to 1 if present.

PromoFaux (approved):
image

image


👍

PR #5320, opened 2023-06-18
Etag

yoshimo:
Signed-off-by: Yoshimo [yoshimo@users.noreply.github.com](mailto:yoshimo@users.noreply.github.com)

What does this PR aim to accomplish?:
  
  It tries to save bandwith by avoiding unneccessary downloads of files that haven't changed on the remote location since we last downloaded them.

How does this PR accomplish the above?:

  If the remote server supplies an etag header we store it for every list locally.
      When we refresh a list we check if there is an etag stored for this entry and if we have an etag we add a "if-none-match: " header to the request.
      The remote server compares our etag with its local database and if the conclusion is, that there hasn't been any change on the file, it will send us back a 304 not-modified status code.
      We then use our local copy, avoid an unneccessary re-download and move on to the next entry on our list.

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -563,9 +563,9 @@ gravity_DownloadBlocklistFromUrl() {
     cmd_ext="--resolve $domain:$port:$ip"
   fi
 
-  if [ -f "${piholeDir}/list.${id}.${domain}.etag" ]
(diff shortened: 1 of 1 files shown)


JapCLoN:
+1

PR #5341, opened 2023-07-17
If ${USER} variable is blank, then populate it with whoami

PromoFaux:

Running `pihole -g` inside the container never appears to fall into the trap that runs `pihole -g` as user `pihole` because the `${USER}` variable is always blank. (Tested by adding echos into gravity)

This is fine when `root` runs `pihole -g` on the command line, but when `pihole-FTL` (which is running as `pihole`) attempts to run it like this, then we get an error because `pihole` cannot run `pihole` as `pihole`....

The other side to this fix is to add `pihole` into the `sudoers` file in the container so that it can run `/usr/local/bin/pihole` (PR to follow)

---

---

### DIFF
--- a/pihole
+++ b/pihole
@@ -544,6 +544,13 @@ if [[ ! $EUID -eq 0 && need_root -eq 1 ]];then
     exit 1
   fi
 fi
+
+# In the case of alpine running in a container, the USER variable appears to be blank
+# which prevents the next trap from working correctly. Set it by running whoami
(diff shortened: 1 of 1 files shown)


DL6ER (approved):
The failing tests seem unrelated to your change

PromoFaux:
> failing tests seem unrelated

Indeed, fixed in #5344  (and indeed #5343)

PR #5300, opened 2023-06-01
Do not remove -all|exact when not surrounded by space in query.sh

yubiuser:
**What does this PR aim to accomplish?:**

Fix issue https://github.com/pi-hole/pi-hole/issues/5299.

Within `query.sh` we remove the arguments `-all` and `-exact` from the string to query the adlists. However, we used `?` as a multiplier after a space seperator which translates to 'zero or more'. But we want to have **one or more** (spaces) which is a `+` multiplier.

---
**By submitting this pull request, I confirm the following:**

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against.
2. I have commented my proposed changes within the code and I have tested my changes.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
6. I have checked that another pull request for this purpose does not exist.
7. I have considered, and confirmed that this submission will be valuable to others.
8. I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
9. I give this submission freely, and claim no ownership to its content.

---
- [x] I have read the above and my PR is ready for review. *Check this box to confirm*

PR #5349, opened 2023-07-20
Fix the regex to strip options even when they are before the domain

rdwebdesign:

After #5300, when using `pihole -q` with options (`-all` or `-exact`), if the option is used before the domain the regex fails to remove the option and the domain becomes invalid (example: domain will be `-all domain`).

This happens because, in this case, the value of the variable `$options` doesn't start with an space (`-all domain`). 
If the option is used in the end (`domain -all`), the regex is able to remove the `-all` (there is a space before the match). 

Changing the regex to remove the option on the start of the line or after a space.
 
none.

---

---

### DIFF
--- a/advanced/Scripts/query.sh
+++ b/advanced/Scripts/query.sh
@@ -50,7 +50,7 @@ fi
 
 # Strip valid options, leaving only the domain and invalid options
 # This allows users to place the options before or after the domain
-options=$(sed -E 's/ +-(all|exact) ?//g' <<< "${options}")
+options=$(sed -E 's/(^|\s)-(all|exact) ?//g' <<< "${options}")
 
 # Handle remaining options
 # If $options contain non ASCII characters, convert to punycode
--- a/advanced/Scripts/query.sh
+++ b/advanced/Scripts/query.sh
@@ -50,7 +50,7 @@ fi
 
 # Strip valid options, leaving only the domain and invalid options
 # This allows users to place the options before or after the domain
-options=$(sed -E 's/(^|\s)-(all|exact) ?//g' <<< "${options}")
+options=$(sed -E 's/(^|\s)-(all|exact)($|\s)//g' <<< "${options}")
 
 # Handle remaining options
 # If $options contain non ASCII characters, convert to punycode


PromoFaux:
Just to throw a spanner into the works here...

> This allows users to place the options before or after the domain

I know this comment has always been here, and you're just fixing broken behaviour, however...

Maybe we should rethink this function (or perhaps even the whole script?). Why do we allow the options to come before _or_ after the domain? 

Does it need to be 256 lines of less-portable `bash` (Maybe we could make it `posix` compliant), or maybe could we utilise some internal function of `pihole-FTL` (`pihole-FTL --query somedomain`?) Would be interested in @DL6ER's thoughts on that.

rdwebdesign:
> Maybe we should rethink this function (or perhaps even the whole script?). Why do we allow the options to come before or after the domain?

Usually I see options BEFORE the other arguments, but this was created a long time ago and I was just fixing it.
I have no idea why this option can be used after the domain. 

> Maybe we could make it posix compliant
 
I think we would need to change how we use long (multiple letters) arguments to use `--` instead of `-` (`pihole -up`, as an example should be changed to `pihole --up`). https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html

yubiuser:
> Does it need to be 256 lines of less-portable bash (Maybe we could make it posix compliant), or maybe could we utilise some internal function of pihole-FTL (pihole-FTL --query somedomain?)

I currently try to implement that `list` function by using the API via bash. It's basically a port of what I use in `PADD`. I have the API part ready and push a branch likely tmw. The handling of the returned data needs some `jq` magic but I think I'm already quite advanced.

```
nanopi@nanopi:~$ ./test.sh 
Authentication failed. Please enter your Pi-hole password
****
Authentication successful.
[
  {
    "address": "http://sysctl.org/cameleon/hosts",
    "domains": [
      "data.flurry.com",
      "dev.flurry.com",
      "flurry.com"
    ]
  },
  {
    "address": "https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts",
    "domains": [
      "flurry.com",
      "adlog.flurry.com",
      "ads.flurry.com",
      "cdn.flurry.com",
      "cfg.flurry.com",
      "data.flurry.com",
      "gw.flurry.com",
      "proton.flurry.com",
      "dev.flurry.com"
    ]
  },
  {
    "address": "https://raw.githubusercontent.com/r-a-y/mobile-hosts/master/AdguardDNS.txt",
    "domains": [
      "flurry.com"
    ]
  },
  {
    "address": "https://small.oisd.nl/",
    "domains": [
      "||flurry.com^"
    ]
  }
]
Session successfully deleted.
nanopi@nanopi:~$ 

```
___

> pihole-FTL --query somedomain

I'm not sure if we should go down this road. We'll end up with exposing almost all endpoints via `pihole-FTL command`. Instead, I think we should interact with the provided API via bash.

rdwebdesign:
@yubiuser 

This PR is not about the returned value.

The problem is: the current command has 2 possible formats:
1. `pihole -q -all www.domain.com` (every option is added **before** the domain) or
2. `pihole -q www.domain.com -all` (some options are allowed **after** the domain).

Both cases are valid on `v5`, but the current regex broke the command when `-all` or `-exact` appears before the domain.
I just tried to fix the first format.

> ... Why do we allow the options to come before or after the domain?

I agree  with ***promofaux***. We don't need both formats.
I think the first one is the most used and we should only accept one format.

The use of `pihole-FTL` is a different topic and we should discuss it before change the code.

yubiuser:
> This PR is not about the returned value.

I know. But I will change the way we deal with supplied options as well during the overhaul.

```
# Process all options (if present)
while [ "$#" -gt 0 ]; do
  case "$1" in
    "-h" | "--help"     ) help;;
    "--exact"           ) partial="false";;
    "--all"             ) results=10000;;
    *                   ) domain=$1;;
  esac
  shift
done

main "${domain}"
```

rdwebdesign:
I think there are 2 different subjects here:
1. current `v6` code is broken when you use `pihole -q -all www.domain.com`;
2. we want to change the way this command accepts options (every option should be added before the domain). 

> I know. But I will change the way we deal with supplied options as well during the overhaul.

At the moment, the change made by this PR won't change that. It only fixes the first item.
I can add a new commit to also change how the options for this command are handled.

rdwebdesign:
This is not needed any more. Options are correctly handled now.

PR #5354, opened 2023-07-29
Use port 8080 only if 80 is already taken

yubiuser:

User expect to open the web interface via `pi.hole/admin/` and not `pi.hole:8080/admin/`. 

This PR changes the port of the web server only to 8080 if port 80 is already taken. This will be the case on updates, where `lighttpd` is still installed or anther webserver is running, but not on fresh installs.

Adds a check for used ports during install/updates after FTL has been stopped and before it's restarted.

In conjunction with: https://github.com/pi-hole/FTL/pull/1608

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -2243,6 +2243,14 @@ main() {
         mv /var/log/pihole*.* /var/log/pihole/ 2>/dev/null
     fi
 
+    # v5 -> v6 upgrade path
+    # if port 80 is already taken, e.g. on update where lighttpd is installed, change the default port to 8080
+    # this should be no issue on upgrades/repairs if FTL is the only webserver as we stop it right above and it should free port 80
+
+    if ss --listening --numeric --tcp --udp | grep -q ":80" ; then
+        setFTLConfigValue webserver.port 8080
+    fi
+
     restart_service pihole-FTL
 
     # Download and compile the aggregated block list


DL6ER:
Are we planning on doing something about the still running web server on upgrades? Like checking if everything is still default config and the content of the directory is still only pi-hole stuff? Or is this considered too much effort?

yubiuser:
I think we can't check everything, there are so many possibilities. I think we should leave things untouched and only cover this case here. Users need to adjust their individual setup afterwards. We can tell them to remove `lighttpd` if they don't need it and change the webport manually after successful installation.

DL6ER:
Two thoughts:
1. During the upgrade, this seems to be at a place where the old FTL has just been stopped and the new FTL has not been started. Hence, can we be sure about the permissions for user `pihole` in `/etc/pihole`? FTL will try to create `/etc/pihole/config_backup` and move the old files there. Also, if we start it like this in the installer, it means FTL is started as `root` (?) possibly causing more ownership issues down the road.
2. I'm now wondering if we should do the same for port `443` which is taken by FTL for the first time for TLS/SSL. A fallback could be `https-alt` (`8443`). I see that we are not yet enabling TLS by default (which we definitely should do) - something for a separate PR.

Summary: I wonder if we should move the check of whether port 80 (and possibly 443) into FTL during initialization (could even be done every time we start FTL, to be discussed). What do you think?

PromoFaux:
> 1. Also, if we start it like this in the installer, it means FTL is started as `root` (?) possibly causing more ownership issues down the road.

Did we (you) not put something into FTL fairly recently to ensure that the FTL conf file gets chowned to pihole? We came across some similar issues in docker

yubiuser:
> What do you think?

I'm fine with FTL handling things - means less shell code and smaller installer (will also prevent us from needing to remove this at some distant future). I don't think we should check on every restart - this can lead to side effects when users experiment with their setup and restart FTL and don't expect the port to changes on it's own. It should only change/set when there is no `*.toml` - hence on upgrades.

DL6ER:
https://github.com/pi-hole/FTL/pull/1609

yubiuser:
Superseded by https://github.com/pi-hole/FTL/pull/1609

PR #5361, opened 2023-08-13
Update query.sh to use FTL's API instead of directly interacting with the database

yubiuser:

This is a complete rewrite of `query.sh`

So far, we have been using direct access to the gravity database to implement the functions provided by `pihole -q` (query the adlist/denylist for a given domain). With FTL v6, FTL provides an API endpoint `/search/DOMAIN` which will do all the low-level work for us and outputs a json containing all the relevant data (and some more). 
This PR changes the implementation of `pihole -q` to use the provided API. To be able to do that, I added a helper file (`api.sh`) which provides the necessary functions to authenticate with the API and request data from it and finally to delete the session if everything is done. It's the same function that I use to access FTL'API int the v6 branch of PADD. As we need to rewrite some more function to use FTL's API they can be reused in other `pihole` cli calls. Both, the `query.sh` as well as `api` are POSIX style. The output colors are similar to that the v6 web interface is using. As we source `COL_TABLE` this was made POSIX compatible as well.

Screenshot at 2023-08-13 22-07-25

Most of the code in `query.sh` is now the generation of the output (from the json returned by FTL). While working on it, I also changed the way we handle provided option (`--exact`, `--all`). They can be in an arbitrary position and handled gracefully. This
PR supersedes https://github.com/pi-hole/pi-hole/pull/5349

---

---

### DIFF
--- a/advanced/Scripts/COL_TABLE
+++ b/advanced/Scripts/COL_TABLE
@@ -1,5 +1,5 @@
 # Determine if terminal is capable of showing colors
-if ([[ -t 1 ]] && [[ $(tput colors) -ge 8 ]]) || [[ "${WEBCALL}" ]]; then
+if ([ -t 1 ] && [ $(tput colors) -ge 8 ]) || [ "${WEBCALL}" ]; then
   # Bold and underline may not show up on all clients
   # If something MUST be emphasized, use both
   COL_BOLD='[1m'
--- a/advanced/Scripts/api.sh
+++ b/advanced/Scripts/api.sh
@@ -1,0 +1,140 @@
+#!/usr/bin/env sh
+# shellcheck disable=SC3043 #https://github.com/koalaman/shellcheck/wiki/SC3043#exceptions
+
+# Pi-hole: A black hole for Internet advertisements
+# (c) 2017 Pi-hole, LLC (https://pi-hole.net)
+# Network-wide ad blocking via your own hardware.
+#
+# Script to hold api functions for use in other scripts
+#
+# This file is copyright under the latest version of the EUPL.
+# Please see LICENSE file for your rights under this license.
+
+
+# The basic usage steps are
+# 1) Test Availability of the API
+# 2) Try to authenticate (read password if needed)
+# 3) Get the data from the API endpoint
+# 4) Delete the session
+
+
+TestAPIAvailability() {
+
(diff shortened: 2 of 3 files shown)


yubiuser:
P.S. the function from `api.sh` are not in `utils.sh` as they are not self-contained - which is one requirement of `utils.sh`

PromoFaux:
I like this as an overall idea - the only "don't like" from me at the moment is that it requires entering the api password to run `pihole -q`, whereas this was never needed before.

Is this an issue? I don't know - perhaps one that we should discuss before moving ahead however.

DL6ER:
It can be circumvented by excluding local API users from authentication (FTL has a dedicated option for this since the very early days). I agree this may require further discussion. We might want to add an extra option (defaulting to true?) to exclude only *some* local requests from auth - permitting only a few harmless things without password such as reading the lists

DL6ER:
@PromoFaux @yubiuser Any comments on my last idea (opt-out: exclude local API users from authentication)?

yubiuser:
I like the idea of having a additional option - like "excluding light" for local users.

DL6ER:
Dedicated option incoming via https://github.com/pi-hole/FTL/pull/1662

yubiuser:
With the last commit, no authentication is required when one of `webserver.api.loaclAPIauth` or `webserver.api.searchAPIauth` is false.

DL6ER:
Thanks, this code seems to work as expected (I also verified the `searchAPIauth = true` fallback to entering a password). 

However, given that ABP searching does not go along with partial matching, I wonder if exact matching would actually be the better default? On the web interface <kbd>Enter</kbd> also executes exact searching.

```
pihole -q 0.000dn.com
```
Screenshot from 2023-10-14 18-48-09

```
pihole -q 0.000dn.com --exact
```
Screenshot from 2023-10-14 18-48-17

yubiuser:
I can change the default behavior. I just kept it as it was in v5 -> partial was the default.

yubiuser:
Yes I want to improve anti-gravity support and also print a warning if 'N' is to small (FTL PR needs to be merged). I was busy reviewing your PRs 😉

yubiuser:
I decided to merge already now - tweaking can happen later and others could use what we have in `api.sh` already now.

PR #5367, opened 2023-08-21
Remove fake user agent when downloading adlist

yubiuser:
Same as https://github.com/pi-hole/pi-hole/pull/5366 but for v6

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -382,7 +382,7 @@ gravity_DownloadBlocklists() {
     unset sources
   fi
 
-  local url domain agent str target compression
+  local url domain str target compression
   echo ""
 
   # Prepare new gravity database
@@ -438,9 +438,6 @@ gravity_DownloadBlocklists() {
     saveLocation="${piholeDir}/list.${id}.${domain}.${domainsExtension}"
     activeDomains[$i]="${saveLocation}"
 
-    # Default user-agent (for Cloudflare's Browser Integrity Check: https://support.cloudflare.com/hc/en-us/articles/200170086-What-does-the-Browser-Integrity-Check-do-)
-    agent="Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/70.0.3538.102 Safari/537.36"
-
     echo -e "  ${INFO} Target: ${url}"
     local regex check_url
     # Check for characters NOT allowed in URLs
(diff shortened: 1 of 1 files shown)


gravity.sh
```diff
@@ -438,8 +438,15 @@ gravity_DownloadBlocklists() {
...
-    # Default user-agent (for Cloudflare's Browser Integrity Check: https://support.cloudflare.com/hc/en-us/articles/200170086-What-does-the-Browser-Integrity-Check-do-)
-    agent="Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/70.0.3538.102 Safari/537.36"
+    # Array of user agents to randomly select from
+    agent_array=("Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:101.0) Gecko/20100101 Firefox/101.0")
+    agent_array+=("Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:99.0) Gecko/20100101 Firefox/99.0")
+    agent_array+=("Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/104.0.0.0 Safari/537.36")
+    agent_array+=("Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/99.0.4844.84 Safari/537.36")
+    agent_array+=("Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_3) AppleWebKit/537.75.14 (KHTML, like Gecko) Version/7.0.3 Safari/7046A194A")
+
+    random=$((RANDOM % 5))
```
prichert, line 448: Instead hardcoded array length, set length based on current array lengh with:
`RANDOM % ${#agent_array[@]}`
The advantage of this, if someone adds a new user-agent to the array this line would work without future modifications.

gravity.sh
```diff
@@ -438,8 +438,15 @@ gravity_DownloadBlocklists() {
     saveLocation="${piholeDir}/list.${id}.${domain}.${domainsExtension}"
     activeDomains[$i]="${saveLocation}"
 
-    # Default user-agent (for Cloudflare's Browser Integrity Check: https://support.cloudflare.com/hc/en-us/articles/200170086-What-does-the-Browser-Integrity-Check-do-)
-    agent="Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/70.0.3538.102 Safari/537.36"
+    # Array of user agents to randomly select from
+    agent_array=("Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:101.0) Gecko/20100101 Firefox/101.0")
```
prichert, line 442: You can find way more user agents with some generators:
[](https://user-agents.net/random)

prichert (commented):
Some improvements

gravity.sh
yubiuser: Thanks for your review. I planed to do this but then forget it when actually writing the code.

yubiuser:
@prichert Thanks for your review. After some internal discussion, we decided to remove the fake user agent and to go with the default curl provided. I'm going to force push away the latter two commits so your remarks don't apply anymore.

yubiuser:
Ready for review

PromoFaux (approved):

PR #5444, opened 2023-10-18
Logrotate config file needs to be owned by root

yubiuser:

Currently, logrotate of the log files is broken. Reason is, that the logrotate config file needs to be owned by `root` and must not be writable by group or others. Otherwise errors occure.

```
nanopi@nanopi:~$ pihole flush once
  [i] Flushing /var/log/pihole/pihole.log ...warning: Potentially dangerous mode on /etc/pihole/logrotate: 0660
error: Ignoring /etc/pihole/logrotate because it is writable by group or others.

nanopi@nanopi:~$ sudo chmod g-w /etc/pihole/logrotate 
nanopi@nanopi:~$ pihole flush once
  [i] Flushing /var/log/pihole/pihole.log ...error: Ignoring /etc/pihole/logrotate because the file owner is wrong (should be root or user with uid 0).


nanopi@nanopi:~$ sudo chown root /etc/pihole/logrotate 
nanopi@nanopi:~$ sudo pihole flush once
  [✓] Flushed /var/log/pihole/pihole.log


```

---

---

### DIFF
--- a/advanced/Templates/pihole-FTL-prestart.sh
+++ b/advanced/Templates/pihole-FTL-prestart.sh
@@ -16,6 +16,10 @@ chown -R pihole:pihole /etc/pihole /var/log/pihole
 chmod -R 0640 /var/log/pihole
 chmod -R 0660 /etc/pihole
 
+# Logrotate config file need to be owned by root and must not be writable by group and others
+chown root:root /etc/pihole/logrotate
+chmod 0644 /etc/pihole/logrotate
+
 # allow all users to enter the directories
 chmod 0755 /etc/pihole /var/log/pihole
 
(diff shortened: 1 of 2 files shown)


DL6ER:
Should we really do this on every FTL restart? Having said that I do see what the intention is: fix it whenever it is broken, have `service pihole-FTL restart` fix things in magic ways.

This question does not want to imply I'm against this, this question just came to my mind when seeing this.

---

On other matters: You did not check the checkbox ;-)

yubiuser:
> On other matters: You did not check the checkbox ;-)

The confirmation checkbox is to far down.

___

I think this is a nice way to handle stuff in a save way - we have such a complex bash design with all the paths triggering updates, checkouts, repairs... they often call the install script and source and execute various parts of it.

DL6ER (approved):

PR #5445, opened 2023-10-19
*BREAKING* Drop support for ancient ARMv4 and ARMv5

DL6ER:

Drop support for ancient ARMv4 and ARMv5, the Debian binaries turn out to cause unsolvable SIGFPE on ARMv6 and lower. Fortunately, we can still use the Alpine binaries on all Raspberry Pi models (even the oldest ones). Use them on ARMv6 devices to restore compatibility of Pi-hole with the very early boards like Raspberry Pi 1B and Zero (W).

Both feature a `BCM2835` chip which is a member of the ARM11 family (type `ARM1176JZF-S `) which turns out to be the `armv6KZ` architecture so exactly what `alpine/arm/v6` is designed for.

**Related issue or feature (if applicable):** https://discourse.pi-hole.net/t/failure-to-create-gravity-floating-point-error/65705

**Pull request in [docs](https://github.com/pi-hole/docs) with documentation (if applicable):** *to be checked*

---

## Checklist:

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1870,9 +1870,9 @@ get_binary_name() {
             printf "%b  %b Detected AArch64 (64 Bit ARM) architecture\\n" "${OVER}" "${TICK}"
             # set the binary to be used
             l_binary="pihole-FTL-arm64"
-        elif [[ "${cpu_arch}" == "armv6KZ" ]]; then
-            printf "%b  %b Detected ARMv6KZ architecture\\n" "${OVER}" "${TICK}"
-            # set the binary to be used
+        elif [[ "${cpu_arch}" == "armv6"* ]]; then
+            printf "%b  %b Detected ARMv6 architecture\\n" "${OVER}" "${TICK}"
+            # set the binary to be used (e.g., BCM2835 as found in Raspberry Pi Zero and Model 1)
             l_binary="pihole-FTL-armv6"
         else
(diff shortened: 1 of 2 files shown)


yubiuser:
Does this mean, we can also drop the Debian based build image?

DL6ER:
Yes, we can drop https://github.com/pi-hole/docker-base-images/tree/master/ftl-build/debian (and move `ftl-build/alpine/Dockerfile` `ftl-build`) as well as https://github.com/pi-hole/FTL/blob/74432a9476429d83ab9d8d925577950bad596ef8/.github/workflows/build.yml#L66-L71 and a few related things like the ability to specify the `builder` type and simplify the arch_tests. 

But this would be the first step and would need to be merged first before the others can land. It simplifies quite a few things!

yubiuser (approved):

yubiuser:
Maybe we should rename the title as it is not that drastic

> *BREAKING* Drop support for ancient ARMv4 and ARMv5

To something like
"Switch binary for ancient ARMv4 and ARMv5 from debian to alpine builds"

derfasthirnlosenick:
Hm tried it on my Zero W (which should be ArmV6) and seems like the detection is off. At least now install fails with "this architecture is not supported" anymore.

DL6ER:
Thanks for the report! This was due to a glitch and should be fixed by https://github.com/pi-hole/pi-hole/pull/5451

Issue #2800, opened 2019-06-15
Dashboard page cosmetic issues


jdquadrider:
The PiHole dashboard is missing values for the Y-axis.  There are also cosmetic issues with the floating "bubbles".  See screenshot for more details.
![image](https://user-images.githubusercontent.com/51855820/59552447-07f50100-8f55-11e9-98ae-11451a814e2a.png)

PR #5471, opened 2023-11-02
Implement DHCP Subnet Mask Configuration

biodigitalfish:
This PR introduces the capability to configure the DHCP subnet mask directly from the Pi-hole admin settings. The changes include the addition of a new positional parameter in the `EnableDHCP()` function and an update to the `ProcessDHCPSettings()` function to handle the DHCP subnet mask.

I have also updated the parameter positions following the new addition. The subnet mask is now taken into account when generating the `dhcp-range` setting in the Pi-hole DHCP configuration.

The motivation behind this change is to provide users with the flexibility to define their subnet masks as per their network requirements, enhancing the DHCP functionality of Pi-hole. 

Example: pihole on 192.168.1.10 but having a dhcp range in 192.168.4.100 -> .200. Currently the default config applies a /24 scope, which breaks client connectivity as they receive an incorrect subnet. There is no error or message to describe the intended behaviour in the webUI. This means having to go through and edit dhcp config file directly. However this gets reset each time you make a change to DHCP scope settings. So it would make the experience of managing DHCP a lot easier in these types of setups.

**Corresponding Changes**:
This PR is part of a set of changes, and there is a corresponding update in the `pi-hole/web` repository. The changes in the web repository handle the UI elements for this new configuration option. PR [[#2800](https://github.com/pi-hole/web/pull/2800)]

---

By submitting this pull request, I confirm the following:

5. I have squashed any insignificant commits. ([git rebase](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

 

- [x]    I have read the above and my PR is ready for review. Check this box to confirm

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -448,7 +448,7 @@ ProcessDHCPSettings() {
 #            ANY CHANGES MADE TO THIS FILE WILL BE LOST ON CHANGE             #
 ###############################################################################
 dhcp-authoritative
-dhcp-range=${DHCP_START},${DHCP_END},${leasetime}
+dhcp-range=${DHCP_START},${DHCP_END},${DHCP_SUBNET_MASK},${leasetime}
 dhcp-option=option:router,${DHCP_ROUTER}
 dhcp-leasefile=/etc/pihole/dhcp.leases
 #quiet-dhcp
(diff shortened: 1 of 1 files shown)


DL6ER:
Thank you very much for your PR. We are in the middle of [our public Pi-hole v6.0 beta phase](https://pi-hole.net/blog/2023/10/09/pi-hole-v6-beta-testing/) making it rather unlikely that there will be another regular v5.x Pi-hole release.

We nonetheless appreciate your submission and agree with the motivation so we took liberty and created a PR that will add this functionality in v6.0. See https://github.com/pi-hole/FTL/pull/1731

PR #5502, opened 2023-11-26
Do not update package cache on install/upgrade

DL6ER:

Do not update package cache on install/upgrade. It takes quite a long time and we don't do this for `dnf`/`yum` either.

---

---

## Checklist:

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -77,6 +77,7 @@ PI_HOLE_FILES=(list piholeDebug piholeLogFlush setupLCD update version gravity u
 PI_HOLE_INSTALL_DIR="/opt/pihole"
 PI_HOLE_CONFIG_DIR="/etc/pihole"
 PI_HOLE_BIN_DIR="/usr/local/bin"
+PKG_MANAGER_UPDATE=false
 if [ -z "$useUpdateVars" ]; then
     useUpdateVars=false
 fi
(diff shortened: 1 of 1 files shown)


yubiuser (commented):
Do we wan to keep `notify_package_updates_available()`?

DL6ER:
No, I think this is only due to the artificially clean environment of a CI container. No real OS will know nothing about their sources. We neither support installing into naked container (or do we?).

If we do, the logic needs to be changed to use `apt-cache` to check if we have install candidates for packages and, if not, then indeed run `apt update`.

Converting this  PR to draft and doing this later...

yubiuser (changes requested):
This is not a good place to put the update function.  It will update the cache 3 times in a row on new installs as `install_dependent_packages()` is called 3 times during installation and usually not empty.

DL6ER:
@yubiuser I simplified the approach

PromoFaux:
I'm a little confused as to the logic here...

`PKG_MANAGER_UPDATE` is defaulted to false. once `update_package_cache` runs, it is set to true and so subsequent calls of this function will not do the package update. 

What I can't see here is how we get ourselves into a situation where `update_package_cache` is run more than once... Sure, it would run each time one runs `pihole -up`... but each time that is run the value of `PKG_MANAGER_UPDATE` will be false... and therefore the `update_package_cache` function will run in full.

Maybe I'm just being dense... but what is the goal here?

DL6ER:
I think this PR has been changed too often and it was abandoned too often, too, to reproduce its original intention. Closing l.

PR #5542, opened 2024-01-07
Fix failing tests in development-v6 branch

PromoFaux:

A change was introduced way back in June that presumably broke the tests. This got worked around by uploading the v6 binaries to the latest v5 release. No bueno.

The correct fix should have been to ensure that the `devlopment-v6` branch of FTL is selected before trying to download it in tests (as we do with other tests).

We must ensure these tests get changed when `development-v6` becomes `development`

---

---

### DIFF
--- a/test/test_any_automated_install.py
+++ b/test/test_any_automated_install.py
@@ -12,6 +12,8 @@ from .conftest import (
     run_script,
 )
 
+FTL_BRANCH = "development-v6"
+
 
 def test_supported_package_manager(host):
     """
@@ -80,11 +82,7 @@ def test_installPihole_fresh_install_readableFiles(host):
     host.run("command -v dnf > /dev/null && dnf install -y man")
     host.run("command -v yum > /dev/null && yum install -y man")
     # Workaround to get FTLv6 installed until it reaches master branch
-    host.run(
-        """
-    echo "development-v6" > /etc/pihole/ftlbranch
-    """
-    )
+    host.run('echo "' + FTL_BRANCH + '" > /etc/pihole/ftlbranch')
     install = host.run(
         """
     export TERM=xterm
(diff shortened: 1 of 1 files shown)


test/test_any_automated_install.py
```diff
@@ -427,37 +434,13 @@ def test_FTL_detect_unknown_no_errors(host):
...
-    assert "error" not in download_binary.stdout.lower()
-
-
 def test_FTL_development_binary_installed_and_responsive_no_errors(host):
     """
     confirms FTL development binary is copied and functional in installed location
     """
     host.run(
         """
+    echo "development-v6" > /etc/pihole/ftlbranch
```
yubiuser, line 443: Copy& Paste error. This is a duplicate of L464

PromoFaux: I'll remove 464 so they are all consistent in echo placement

yubiuser:
Should we test for all architectures we have FTL builds for?

PromoFaux:
We probably should. Just FP'd over the last commit, in which I have reduced some code duplication for each arch's test.

Easier to maintain, and less places to remember to change `development-v6` ->`development` in future

test/test_any_automated_install.py
```diff
@@ -379,89 +268,28 @@ def test_FTL_detect_armv8a_no_errors(host):
...
-    assert "error" not in download_binary.stdout.lower()
+    if supported:
+        expected_stdout = info_box + " FTL Checks..."
+        assert expected_stdout in detectPlatform.stdout
+        expected_stdout = tick_box + " Detected " + detected_string + " architecture"
+        assert expected_stdout in detectPlatform.stdout
+        expected_stdout = tick_box + " Downloading and Installing FTL"
+        assert expected_stdout in detectPlatform.stdout
+    else:
+        expected_stdout = "Not able to detect architecture (unknown: " + detected_string + ")"
```
rdwebdesign, line 279: After your last FP, the line length was changed and now tests are failing:
- https://github.com/pi-hole/pi-hole/actions/runs/7441027184/job/20242687467?pr=5542

```suggestion
        expected_stdout = (
            "Not able to detect architecture (unknown: " + detected_string + ")"
        )
```

rdwebdesign (changes requested):

yubiuser (approved):

yubiuser:
Wait, we don't check for i386.

And do we really need a check for `arm7l` and `arm7`?

PromoFaux:
> Wait, we don't check for i386.
> 
> And do we really need a check for both `arm7l` and `arm7`?

i386 not as straightforward to test (see the code in the basic install that handles it)

I did look at adding a mock for the dpkg command earlier (uname is still x86_64) , but it didn't work. I can try again if it is important.

Really, aside from adding risc, I just used the same arch's that were already in the test. Can only assume i386 was left out on purpose before, too.

yubiuser:
Maybe it's not important :)

PromoFaux:
I have FP'd away the previous commit that added the i386 tests. On closer inspection, the only OS we support that even has a 386 flavour is debian, and even then - I doubt all that many people are using it. It is likely why the test was left out initially. Adding it to the new tests causes the Fedora/Centos tests to fail - we could code around that with some exceptions, but it's a lot of completixty for what probably is not a lot of gain

PR #5570, opened 2024-02-09
Merge development into development-v6 (with merge conflicts resolved)

PromoFaux:
This needs some more eyes on it than just mine - I don't quite trust myself.


yubiuser:
Do we really need to include all the dependabot commits? We run dependabot on v6 as well, so the dependencies should be up to date already. This causes a lot of noise. 
I think the better approach would be to really look which commits need to be ported to v6 and cherry those.

PromoFaux:
IMO, it's about easy mergability (is that a word? It is now)

If we don't do this we'll have difficulty releasing, I think.

DL6ER (approved):

PR #5591, opened 2024-02-27
Add Custom wildcard DNS records.

DjoSmer:

**Please make sure you**

---

**Wildcard DNS**, for those who don’t know – is a trick to allow any kind of host name – to have a specific same IP as it’s DNS name. For example, if my domain is test.com and my IP is 1.2.3.4, the record full.test.com will also have the same IP: 1.2.3.4.

https://github.com/pi-hole/pi-hole/assets/8194467/6ccfe8de-9ea4-4fa0-9d09-6294e3c85d23

I wrote 3 new admin commands which do this.

---

---

### DIFF
--- a/advanced/Scripts/webpage.sh
+++ b/advanced/Scripts/webpage.sh
@@ -19,6 +19,8 @@ readonly FTLconf="/etc/pihole/pihole-FTL.conf"
 readonly dhcpstaticconfig="/etc/dnsmasq.d/04-pihole-static-dhcp.conf"
 readonly dnscustomfile="/etc/pihole/custom.list"
 readonly dnscustomcnamefile="/etc/dnsmasq.d/05-pihole-custom-cname.conf"
+readonly dnsCustomWildcardDNSFile="/etc/dnsmasq.d/15-pihole-custom-wildcarddns.conf"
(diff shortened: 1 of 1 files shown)


PromoFaux:
Thank you for the submission, however unfortunately we are no longer accepting new features for version 5 of Pi-hole.

I could _potentially_ see this being acceptable as an idea (not nessacerily this implementation) in v6 - but It should be discussed first with @DL6ER, he may have some opinions on why this is not currently a feature...

Closing for now, but thanks again for the submission!

DjoSmer:
> This is already available via regex blocking which is a much more powerful tool when it comes to wildcard matching. You can easily reconstruct your example using something like
> 
> ```
> (^|\.)test\.com;reply=1.2.3.4
> ```
> 
> to get the behavior you are describing above. I don't think it's a wise idea to add "the same thing" twice in a different form as this makes it hard for users to decide when they should use one or the other.

You said this is already available, I agree. It is already exists in dnsmasq. PI-hole has used dnsmasq, I just add some commands in Pi-hole's script.
If this feature exists in dnsmasq, why it can't be used and your solution is not clear.

I really I don't understand this patch has some changes and you didn't get it.

PR #5613, opened 2024-03-25
Wait after restarting FTL before trying to check version

DL6ER:

Prevent errors like seen in https://github.com/pi-hole/pi-hole/issues/5608#issuecomment-2017000214 when Pi-hole uses itself as `nameserver`:

```bash
 Please note that changing branches severely alters your Pi-hole subsystems

  [✓] Branch fix/ede_neterr exists
  [i] Switching to branch: "fix/ede_neterr" from "fix/ede_neterr"
  [✓] Downloading and Installing FTL
  [✓] Enabling pihole-FTL service to start on reboot...
fatal: unable to access 'https://github.com/pi-hole/pi-hole/': Could not resolve host: github.com
fatal: unable to access 'https://github.com/pi-hole/web/': Could not resolve host: github.com
fatal: unable to access 'https://github.com/pi-hole/FTL/': Could not resolve host: github.com

```

---

---

## Checklist:

### DIFF
--- a/advanced/Scripts/piholeCheckout.sh
+++ b/advanced/Scripts/piholeCheckout.sh
@@ -193,6 +193,19 @@ checkout() {
             FTLinstall "${binary}"
             restart_service pihole-FTL
             enable_service pihole-FTL
+            str="Restarting FTL..."
+            echo -ne "  ${INFO} ${str}"
+            # Wait until name resolution is working again after restarting FTL,
+            # so that the updatechecker can run successfully and does not fail
+            # trying to resolve github.com
+            until getent hosts github.com &> /dev/null; do
+                # Append one dot for each second waiting
+                str="${str}."
+                echo -ne "  ${OVER}  ${INFO} ${str}"
+                sleep 1
+            done
+            echo -e "  ${OVER}  ${TICK} Restarted FTL service"
+
             # Update local and remote versions via updatechecker
             /opt/pihole/updatecheck.sh
         else


advanced/Scripts/piholeCheckout.sh
```diff
@@ -170,10 +170,23 @@ checkout() {
...
+            echo -e "  ${INFO} Switching to branch \"${2}\" from \"${oldbranch}\""
             FTLinstall "${binary}"
             restart_service pihole-FTL
             enable_service pihole-FTL
+            str="Restarting FTL..."
+            echo -ne "  ${INFO} ${str}"
+            # Wait until name resolution is working again after restarting FTL,
+            # so that the updatechecker can run successfully and does not fail
+            # trying to resolve github.com
+            until getent hosts github.com &> /dev/null; do
```
rdwebdesign, line 182: Is it possible `enable_service pihole-FTL` fails (or resolution fails) and we ended up in an infinite loop?

PromoFaux: I think it makes sense to loop until github.com is resolvable - if the network is down then there is no harm in waiting for it to be up again.

yubiuser (commented):
I agree with rdwebdesign. We should be not have an infinite loop that depends on outside conditions (network availability). I think we should have a counter and exit the loop after e.g. 10 seconds and ask users to manually run the update command afterwards.

PromoFaux:
I am not sure where my mind was when I first responded to RD's comment. Reading over the code again, actually it probably is a fair shout that we timeout after x seconds (the user could always `ctrl+c`, but that is not overly friendly

DL6ER:
The script is constantly adding `.` to the output to show it's still alive. Waiting "forever" seems fine to me as the updater has no way to finish properly when there in no way to contact GitHub.

The suggested alternative is erroring out with no option for the user to wait longer if 10 seconds isn't enough to restart FTL on their RPi Zero with 100,000 queries to import from the database. Sure, the timeout could be 60 seconds but my opinion is that waiting "forever" and constantly showing that the script is still alive is the proper way to do this

yubiuser:
I know we had this before, but maybe the answer changed in-between: do we have any signal that would signal FTL's readiness to accept DNS queries (e.g. a second pid file)? 
___
This PR is not about FTL only. The cat might have unplugged the cable just at the right moment - FTL might be ready for a long time, but there would be no network connectivity.

DL6ER:
We could `dig pi.hole` or whatever always resolves without being forwarded.

I hear you and you are right, it's about more than FTL alone. However, without connectivity, there is no (simple) way the updater can exit gracefully.

yubiuser:
`dig pi.hole @localhost` could work. But this tells us only that FTL is ready, but has no meaning for systems that don't use FTL for their host's DNS resolution.

DL6ER:
Quite some time has passed and I'd like to re-inquire how we all feel about this PR. After reading up on its motivation, it seems still very relevant on slow devices. The test currently tests for resolvability of GitHub which makes sense because the next step is querying the versions from GitHub.

Hence, I also feel like the question we should be asking is **not** "is FTL ready" but **rather** "can the host resolve github.com" - which is exactly what the code in the PR wants to implement.

I did just rebase this on latest `development-v6`.

rdwebdesign:
CentOS 9 tests are failing.

DL6ER:
Rebased based on https://github.com/pi-hole/pi-hole/pull/5743 to check effect on CI tests

DL6ER:
GitHub forced me to update the branch *again* after merging the other PR. Nothing changed - except, of course, `HEAD^` hash so we get a new commit hash here, too.

yubiuser (approved):
We also have a similar check if updating `core`: this will trigger the installer, which will run `gravity` right before `updatechecker.sh`. `Gravity` contains a check which tries to resolve `raw.githubusercontent.com` 4 times before failing.

See https://github.com/pi-hole/pi-hole/pull/5137 and
https://github.com/pi-hole/pi-hole/blob/a965afa249740e8f4803d999571b1df69d5eaa21/gravity.sh#L296

PR #5620, opened 2024-03-27
More checks when downloading from file:// scheme

orazioedoardo:

**Please make sure you**

---

Current checks when downloading from file:// scheme can be bypassed by creating a symlink to a file that is not world readable and downloading the symlink, which usually has 777 permissions. Checks also don’t stop non-regular files (though might fail when passed to curl later on). Impact is low IMHO since it requires unprivileged shell access to the system in addition of having access to the admin page.

By the way I’m not sure whether the admin page should still show snippets of invalid adlists when downloading from world readable file://. The set of users with admin access does not necessarily intersect with the users having shell access. Also consider that many distributions still set human home directories as 0755.

Test that the file is a regular file, follow symlinks when checking permissions with stat.

---

---

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -567,14 +567,14 @@ gravity_DownloadBlocklistFromUrl() {
   if [[ $url == "file://"* ]]; then
     # Get the file path
     file_path=$(echo "$url" | cut -d'/' -f3-)
-    # Check if the file exists
-    if [[ ! -e $file_path ]]; then
+    # Check if the file exists and is a regular file (i.e. not a socket, fifo, tty, block). Might still be a symlink.
+    if [[ ! -f $file_path ]]; then
       # Output that the file does not exist
       echo -e "${OVER}  ${CROSS} ${file_path} does not exist"
       download=false
     else
-      # Check if the file has a+r permissions
-      permissions=$(stat -c "%a" "$file_path")
+      # Check if the file or a file referenced by the symlink has a+r permissions
+      permissions=$(stat -L -c "%a" "$file_path")
       if [[ $permissions == *4 || $permissions == *5 || $permissions == *6 || $permissions == *7 ]]; then
         # Output that we are using the local file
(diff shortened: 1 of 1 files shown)


PromoFaux:
Are we sure this is a problem? I tried a test with `sudo ln /etc/shadow blah` and adding it as `file:///home/pi/blah` and get the following output:

```
  [i] Target: file:///home/pi/blah
  [✗] Cannot read file (file needs to have a+r permission)
<------  [✗] Status: Retrieval failed / empty list
```

orazioedoardo:
> Are we sure this is a problem?

Works for me.

> sudo ln /etc/shadow blah

Are you sure `blah` is a symlink? ln without `-s` is a hard link.

PromoFaux:
OK, with -s then it works. So much for responsible disclosure, I guess! I think this is probably an edge-case as it would assume that someone has access to the shell to be able to create the symlink in the first place. However having tested your change with a symlink (created with `sudo ln -s /etc/shadow blah`) I can confirm that the symlink is followed and the original file's permissions are used - and that this doesn't break the already in-place functionality to prevent gravity from reading these files.

> Also you don’t need to create a link with sudo.

```
pi@raspberrypi:/etc/pihole $ ln -s /etc/shadow test
ln: failed to create symbolic link 'test': Permission denied
```

PromoFaux (approved):

orazioedoardo:
> ln: failed to create symbolic link 'test': Permission denied

Of course you can’t create a symlink to `/etc/pihole` as `pi` user.

PromoFaux:
Perhaps, but the point was you said one doesn't need `sudo` to create a symlink. If someone has root shell access, you probably have bigger problems!

orazioedoardo:
> If someone has root shell access, you probably have bigger problems!

You just need shell access and a place to write the symlink, which can be the home folder of the user you have access to, doesn’t need to be root or having sudo access.

DL6ER (approved):

PR #5618, opened 2024-03-27
v5.18.1

PromoFaux:
Release PR

PR #5622, opened 2024-03-28
Only use local files (file://) when they have explicit permissions a+r

DL6ER:

This PR ports the most-recent changes from PRs #5615, #5618 amd #5620 to v6.

---

---

## Checklist:

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -598,7 +598,7 @@ gravity_DownloadBlocklistFromUrl() {
   # running the script.
   if [[ $url == "file://"* ]]; then
     # Get the file path
-    file_path=$(echo "$url" | cut -d'/' -f3-)
+    file_path=$(echo "$url" | cut --delimiter='/' --fields=3-)
     # Check if the file exists and is a regular file (or a symlink to one)
     if [[ ! -e $file_path ]]; then
       # Output that the file does not exist
@@ -610,7 +610,7 @@ gravity_DownloadBlocklistFromUrl() {
       download=false
     else
       # Check if the file has a+r permissions
-      permissions=$(stat -Lc "%a" "$file_path")
+      permissions=$(stat --dereference --format="%a" "$file_path")
       if [[ $permissions == *4 || $permissions == *5 || $permissions == *6 || $permissions == *7 ]]; then
         # Output that we are using the local file
         echo -e "${OVER}  ${INFO} Using local file ${file_path}"
@@ -624,7 +624,7 @@ gravity_DownloadBlocklistFromUrl() {
 
   if [[ "${download}" == true ]]; then
     # shellcheck disable=SC2086
-    httpCode=$(curl --connect-timeout ${curl_connect_timeout} -s -L ${compression} ${cmd_ext} ${heisenbergCompensator} -w "%{http_code}" "${url}" -o "${listCurlBuffer}" 2>/dev/null)
+    httpCode=$(curl --connect-timeout ${curl_connect_timeout} --silent --location ${compression} ${cmd_ext} ${heisenbergCompensator} --write-out "%{http_code}" "${url}" --output "${listCurlBuffer}" 2>/dev/null)
   fi
 
   case $url in
--- a/gravity.sh
+++ b/gravity.sh
@@ -488,6 +488,7 @@ compareLists() {
 gravity_DownloadBlocklistFromUrl() {
   local url="${1}" adlistID="${2}" saveLocation="${3}" target="${4}" compression="${5}" gravity_type="${6}" domain="${7}"
   local heisenbergCompensator="" listCurlBuffer str httpCode success="" ip cmd_ext
+  local file_path permissions ip_addr port blocked=false download=true
 
   # Create temp file to store content on disk instead of RAM
   # We don't use '--suffix' here because not all implementations of mktemp support it, e.g. on Alpine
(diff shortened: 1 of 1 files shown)


DL6ER:
image

gravity.sh
```diff
@@ -591,8 +592,40 @@ gravity_DownloadBlocklistFromUrl() {
...
+      # Output that the file does not exist
+      echo -e "${OVER}  ${CROSS} ${file_path} does not exist"
+      download=false
+    elif [[ ! -f $file_path ]]; then
+      # Output that the file is not a regular file
+      echo -e "${OVER}  ${CROSS} ${file_path} is not a regular file"
+      download=false
+    else
+      # Check if the file has a+r permissions
+      permissions=$(stat -Lc "%a" "$file_path")
```
dschaper, line 613: This is more of a style request but can we use the full name of the flags when we call them in scripts? Saves having to track down the man page to know what it's doing.

```suggestion
      permissions=$(stat --dereference --format="%a" "$file_path")
```

gravity.sh
```diff
@@ -591,8 +592,40 @@ gravity_DownloadBlocklistFromUrl() {
...
 
-  # shellcheck disable=SC2086
-  httpCode=$(curl --connect-timeout ${curl_connect_timeout} -s -L ${compression} ${cmd_ext} ${heisenbergCompensator} -w "%{http_code}" "${url}" -o "${listCurlBuffer}" 2>/dev/null)
+  # If we are going to "download" a local file, we first check if the target
+  # file has a+r permission. We explicitly check for all+read because we want
+  # to make sure that the file is readable by everyone and not just the user
+  # running the script.
+  if [[ $url == "file://"* ]]; then
+    # Get the file path
+    file_path=$(echo "$url" | cut -d'/' -f3-)
```
dschaper, line 601:
```suggestion
    file_path=$(echo "$url" | cut --delimiter='/' --fields=3-)
```

gravity.sh
```diff
@@ -591,8 +592,40 @@ gravity_DownloadBlocklistFromUrl() {
...
+        # Output that the file does not have the correct permissions
+        echo -e "${OVER}  ${CROSS} Cannot read file (file needs to have o+r permission)"
+        download=false
+      fi
+    fi
+  fi
+
+  if [[ "${download}" == true ]]; then
+    # shellcheck disable=SC2086
+    httpCode=$(curl --connect-timeout ${curl_connect_timeout} -s -L ${compression} ${cmd_ext} ${heisenbergCompensator} -w "%{http_code}" "${url}" -o "${listCurlBuffer}" 2>/dev/null)
```
dschaper, line 627:
```suggestion
    httpCode=$(curl --connect-timeout ${curl_connect_timeout} --silent --location ${compression} ${cmd_ext} ${heisenbergCompensator} --write-out "%{http_code}" "${url}" --output "${listCurlBuffer}" 2>/dev/null)
```

dschaper (changes requested):

orazioedoardo:
I think this pull request, and the linked #5620 still wouldn’t catch a situation when `file_path` is a symlink to a non-regular file. Probably best would be to first dereference `file_path` with `readlink --canonicalize`, and then check for non-regular and world-readability the destination.

dschaper:
> is a symlink to a non-regular file.

Does the change from `-e` to `-f` at https://github.com/pi-hole/pi-hole/pull/5620/files#diff-1aa8de157651defa7dd9bb1eec010df250d3a336800e2bad8a4a3e297587747cR571 address non-regular files?

orazioedoardo:
> I think symlinks to non-regular fail the -f check as we would like

> help test, -f dereferences before comparison:

Yeah that's right actually

gravity.sh
```diff
@@ -591,8 +592,40 @@ gravity_DownloadBlocklistFromUrl() {
...
+      echo -e "${OVER}  ${CROSS} ${file_path} does not exist"
+      download=false
+    elif [[ ! -f $file_path ]]; then
+      # Output that the file is not a regular file
+      echo -e "${OVER}  ${CROSS} ${file_path} is not a regular file"
+      download=false
+    else
+      # Check if the file has a+r permissions
+      permissions=$(stat -Lc "%a" "$file_path")
+      if [[ $permissions == *4 || $permissions == *5 || $permissions == *6 || $permissions == *7 ]]; then
```
dschaper, line 614: Do we need to check for `5`|`7` since those are directory or exectuable bits? 

Would it be good to expand on the message at 619 to say more about why it's not being used? Something that would say "Executable bit set ${permissions}, ${file_path} not used."

gravity.sh
```diff
@@ -591,8 +592,40 @@ gravity_DownloadBlocklistFromUrl() {
...
+        # Output that we are using the local file
+        echo -e "${OVER}  ${INFO} Using local file ${file_path}"
+      else
+        # Output that the file does not have the correct permissions
+        echo -e "${OVER}  ${CROSS} Cannot read file (file needs to have o+r permission)"
```
dschaper, line 619:
```suggestion
        echo -e "${OVER}  ${CROSS} Cannot read ${file_path} (file needs to have o+r permission, file is set to ${permissions}) "
```

DL6ER: I didn't do this as it seemed odd to have `o+r` in the front but then octal in the back. Also there is no need to re-print the filename here as it is already printed right above and - IMO - just makes the warning line very long without adding any further useful content

gravity.sh
DL6ER: Yes, we need to check. Everything that has 4 is readable alone (`4--`). 5 is 4+1 (`r-x`), 6 is 4+2 (`rw-`), 7 is 4+2+1 (`rwx`). We need to check for all of them to get all cases where the readable bit is set.

I did not exclude the cases where `x` is set as there are legit filesystems lacking the concept of proper file permissions. Take, for instance, FAT32 which does not support file permissions. The permissions for everything on a FAT32 formatted device are determined by how the drive is mounted - the default is `0755` so `x` set for everybody. I don't think we should exclude files on FAT32 (also original FAT and even the newer exFAT) by such a rule.

gravity.sh
dschaper: Okay, I see the octal vs regular situation but I don't see where the filename is printed when the file is not being used. I see it on line 616 but that is for when the file is being used.

dschaper (approved):

PR #5683, opened 2024-06-10
Remove hardcoded name server for os_check

dagobert:

When installing pihole using `basic-install.sh` in a restricted network where firewall rules disallow DNS queries to other than dedicated IPs, the install fails.
Using this hardcoded NS is not necessary since any system configured DNS server will be sufficient.

Removing the hardcoded DNS server in the dig statement.

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -183,7 +183,7 @@ os_check() {
         detected_os=$(grep '^ID=' /etc/os-release | cut -d '=' -f2 | tr -d '"')
         detected_version=$(grep VERSION_ID /etc/os-release | cut -d '=' -f2 | tr -d '"')
 
-        cmdResult="$(dig +short -t txt "${remote_os_domain}" @ns1.pi-hole.net 2>&1; echo $?)"
+        cmdResult="$(dig +short -t txt "${remote_os_domain}" 2>&1; echo $?)"
         # Gets the return code of the previous command (last line)
         digReturnCode="${cmdResult##*$'\n'}"
 


yubiuser:
The usage of the specific nameserver as introduced by https://github.com/pi-hole/pi-hole/pull/3596/commits/0ff32c3629220f386a45c14d8982aaaf128aa47f

Initially, we didn't had this, but observed multiple issue with users running a local AD server which was unable to correctly return the correct query answer. Therefore we introduced it to make sure resolution wouldn't fail for them.

yubiuser:
See https://github.com/pi-hole/pi-hole/issues/3578 and https://github.com/pi-hole/pi-hole/pull/3596#issuecomment-664509808

dagobert:
How about introducing a command line option to disable this or other way around?!

yubiuser:
You can skip the whole OS check by setting `PIHOLE_SKIP_OS_CHECK=true`

dagobert:
Sorry, I might not understand the purpose of the OS check than. If it not important and can be disabled, why do it at all?
My suggestion was only to turn off using the dedicated nameserver and still do the OS check - or turn the whole thing around and offer to use a dedicated nameserver.

Btw. anyone with an AD environment allowing external nameservers through their firewall must be crazy!

yubiuser:
The purpose is to tell users that the use an unsupported OS and we can't provide support for this. We added the option to skip the check with a long disclaimer basically telling users: "If you go that path, you're on your own". We did not intend to block installation completely but can't test and support every OS out there.

We could add an option/env variable to inject another name server, but I don't see much of a use compared to skipping the test at all.

But opinion on this subject may vary (@pi-hole/core-maintainers)

dagobert:
I was thinking back and forth about this. The version check is not very important it seem from the arguments I read. Also, many people should have read beforehand about the OS support.

To make the whole installation process more seamless I would suggest the following:

1. Put the list of known supported OS in the script.
2. First check using dig with hard coded NS
3. If this fails, check without hard coded NS 
4. If this fails check against hard coded list and put a disclaimer that DNS check failed which means that the current os could still work and proceed with the non-os-check flag to force.

This way only in very limited cases there should be a fail.

In total I think the reason behind using DNS as check is a very nice idea but since mitigation seems to be necessary, it leads to unnecessary code that needs to be maintained. And the benefit compared does not seem to worth it.

yubiuser:
Thanks for your suggestion. I think there is some benefit in one of your points:

> If this fails, check without hard coded NS

____
I'm not convinced by the other points: First, your assumption 
> many people should have read beforehand about the OS support

is unfortunately wrong. From offering support for Pi-hole in the last years my impression was only a few read our documentation and the prerequisites.

> Put the list of known supported OS in the script.

We didn't do that on purpose: as soon we add/remove an OS from the supported list, we would need to release another version. We try to avoid this.

> If this fails check against hard coded list and put a disclaimer that DNS check failed which means that the current os could still work and proceed with the non-os-check flag to force.

If the script fails to get the an answer from a hard-coded name server and a non-hard-coded name server something is likely broken. It's fine if it fails then so users can have a look and fix the underlying issue.

___

Would you like to PR a necessary change to the installer to fallback to a non-hardcoded DNS if it failed before? Please target `deveopment-v6` branch. The code there is a bit more complex already, because we now check via IPv4 and IPv6 separately.

dagobert:
> Would you like to PR a necessary change to the installer to fallback to a non-hardcoded DNS if it failed before? Please target `deveopment-v6` branch. The code there is a bit more complex already, because we now check via IPv4 and IPv6 separately.

I will see what I can do. At the moment my time is limited but it will hopefull change soon.

PR #5689, opened 2024-06-20
Finish core v6 implementation

DL6ER:

This PR accompanies PR https://github.com/pi-hole/FTL/pull/1999 - I checked all `pihole`-related commands except `debug`

---

---

## Checklist:


advanced/Scripts/api.sh
```diff
@@ -51,6 +51,12 @@ TestAPIAvailability() {
             API_PORT=""
         else
             # API is available at this URL combination
+
+            if [ "${availabilityResonse}" = 200 ]; then
```
PromoFaux, line 55: /nitpick

```suggestion
            if [ "${availabilityResponse}" = 200 ]; then
```

DL6ER: The variable is not my invention, it's been there already - including the typo. Adding this suggested change would make the code stop working

PromoFaux: Thanks for changing it, though :)

advanced/Scripts/list.sh
```diff
@@ -5,261 +5,187 @@
...
-
-    listname="$(GetListnameFromTypeId "${typeId}")"
-    param="$(GetListParamFromTypeId)"
-
-    echo "Usage: pihole -${param} [options] <domain> <domain2 ...>
-Example: 'pihole -${param} site.com', or 'pihole -${param} site1.com site2.com'
-${listname^} one or more domains
+    echo "Usage: pihole ${abbrv} [options] <domain> <domain2 ...>
+Example: 'pihole ${abbrv} site.com', or 'pihole ${abbrv} site1.com site2.com'
+${typeId^} one or more ${kindId} domains
```
PromoFaux, line 47: Some additional needed here to line up with other changes to function names, current output of `pihole --help`:

```
54af332e58c9:/etc/pihole# pihole --help
Usage: pihole [options]
Example: 'pihole -w -h'
Add '-h' after specific commands for more information on usage

Whitelist/Blacklist Options:
  -w, whitelist       Whitelist domain(s)
  -b, blacklist       Blacklist domain(s)
  --regex, regex      Regex blacklist domains(s)
  --white-regex       Regex whitelist domains(s)
  --wild, wildcard    Wildcard blacklist domain(s)
  --white-wild        Wildcard whitelist domain(s)
                        Add '-h' for more info on whitelist/blacklist usage
```

rdwebdesign: About the wording:

I'm not against the use of `denylist`, but the option is `-b` and the code also accepts `blocklist`. We already use `-d` for other option.

My questions are: 
- should we accept `denylist`, `blocklist` or both?
- if accepting both, shouldn't we show `blocklist` in help too?

DL6ER: Let me first tell you that I added `blocklist` only to have a "justification" to keep the `-b` shortcut. And that only because `-d` is already taken. After having thought about this, I fully agree this wasn't a very good decision so I'm asking if we should rather:

1. Remove `-d` as shortcut for `debug` and add it to `denylist`
2. Remove `-a` for allowed and add semi-short keys like
   - `pihole -deny stuff.com`
   - `pihole -allow other.net`

   so we can keep `-d` for `debug`

Maybe no. 2 without `-` even, like
```
pihole allow example.org
```
which is basically natural language computing like.

PromoFaux: > Maybe no. 2 without `-` even, like

I like this

DL6ER: Thanks, I went a bit further and allowed `not` for removals as well (`-d` and `--delmode` are still there) so you can undo
```
pihole allow example .com
```
by
```
pihole allow not example.com
```

I know we could have done something like `disallow` or even "`unallow`" but I didn't want to get too fancy because this would have needed more code changes.

rdwebdesign: I like the solution.

PromoFaux: I'm not entirely convinced by `not`, but I'll not die on that hill

PromoFaux (approved):

advanced/Scripts/list.sh
DL6ER: We're still here and open for better alternatives

rdwebdesign: Reading the comments again, I think the "allow not" and "deny not" will be confusing.

My suggestions is to use one of "remove", "delete" or "-d" or "-rm" to follow the natural language computing idea.

Examples:
```
pihole allow example.com             ->  add "example.com" to allowlist
pihole allowlist example.com         ->  add "example.com" to allowlist

pihole allowlist remove example.com  ->  remove "example.com" from allowlist
pihole allowlist delete example.com  ->  remove "example.com" from allowlist
pihole allowlist -d example.com      ->  remove "example.com" from allowlist
```

What do you think?

DL6ER: Yes, I like it and committed this change right now.

PromoFaux (dismissed):

yubiuser:
We could also re-write `piholeEnable()` from `pihole` to use the API and then remove `pihole-reenable.sh`

advanced/Scripts/piholeARPTable.sh
```diff
@@ -54,6 +56,20 @@ flushARP(){
         return 1
     fi
 
+    # Flush ARP cache of the host
```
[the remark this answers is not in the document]
rdwebdesign: Good catch.

I really don't know why it was removed and reintroduced, but if we really don't needed to flush the ARP cache, maybe we should change the function name from `flushARP` to `truncateARPtable` or something similar.

advanced/Scripts/api.sh
```diff
@@ -74,26 +80,41 @@ TestAPIAvailability() {
     fi
 }
 
-Authentication() {
-    # Try to authenticate
-    LoginAPI
+LoginAPI() {
+    # If the API URL is not set, test the availability
+    if [ -z "${API_URL}" ]; then
```
yubiuser, line 85: The check for `API_URL` is superfluous, as `API_URL` is not used outside of `api.sh`

advanced/Scripts/query.sh
```diff
@@ -128,22 +134,16 @@ Main() {
     # Test if the authentication endpoint is available
     TestAPIAvailability
```
yubiuser, line 135: This can be removed as `TestAPIAvailability` is now called from within `LoginAPI`

advanced/Scripts/piholeARPTable.sh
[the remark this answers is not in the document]
DL6ER: Thanks for digging this out. I will be honest right away: I don't remember what triggered me to remove the ARP flushing before. However, I think it is very useful because if there is anything wrong in the hosts's ARP cache, then it will stay there and FTL will immediately re-import it from there even after the Pi-hole-internal ARP table has been flushed.

I suppose the bad consequences we have talked about in this commit were that the server shortly becomes quite disoriented inside the local network. It may take up to a few seconds until it rediscovered its neighbors in the network and can continue to use the network as usual. I just tried this on my local machine (having Wireshark opened at the same time to watch ARP traffic) and - for me - it took about 8 milliseconds until my server rediscovered the local gateway and was again able to talk to the Internet. Everything in between failed to be sent.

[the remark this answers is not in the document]
yubiuser: The issue was with the mock devices of the network table.
https://discourse.pi-hole.net/t/request-ability-to-add-clients-to-known-client-list-from-network-page/28601/86?u=yubiuser

advanced/Scripts/api.sh
DL6ER: It is used for efficiency reasons. If LoginAPI is ever used twice, we have already derived how to reach the API. We can then simply reuse this knowledge without having to inquiry again.

DL6ER:
I took liberty and generalize the option to temporarily disable your Pi-hole so you could also use it to temporarily *enable* it.

PromoFaux (approved):

PR #5734, opened 2024-08-23
Make the help text of "pihole checkout [what] [branch]" more colorful

DL6ER:

See title.

image
---

---

## Checklist:

### DIFF
--- a/pihole
+++ b/pihole
@@ -411,6 +411,7 @@ piholeCheckoutFunc() {
       echo "Switch Pi-hole subsystems to a different GitHub branch
     Usage: ${COL_GREEN}pihole checkout${COL_NC} ${COL_YELLOW}shortcut${COL_NC}
         or ${COL_GREEN}pihole checkout${COL_NC} ${COL_PURPLE}repo${COL_NC} ${COL_CYAN}branch${COL_NC}
+
   Example: ${COL_GREEN}pihole checkout${COL_NC} ${COL_YELLOW}master${COL_NC}
        or  ${COL_GREEN}pihole checkout${COL_NC} ${COL_PURPLE}ftl ${COL_CYAN}development${COL_NC}
 
--- a/pihole
+++ b/pihole
@@ -408,19 +408,21 @@ piholeCheckoutFunc() {
     unsupportedFunc
   else
     if [[ "$2" == "-h" ]] || [[ "$2" == "--help" ]]; then
-      echo "Usage: pihole checkout [repo] [branch]
-  Example: 'pihole checkout master' or 'pihole checkout core dev'
-  Switch Pi-hole subsystems to a different GitHub branch
+      echo "Switch Pi-hole subsystems to a different GitHub branch
+    Usage: ${COL_GREEN}pihole checkout${COL_NC} ${COL_YELLOW}shortcut${COL_NC}
+        or ${COL_GREEN}pihole checkout${COL_NC} ${COL_PURPLE}repo${COL_NC} ${COL_CYAN}branch${COL_NC}
+  Example: ${COL_GREEN}pihole checkout${COL_NC} ${COL_YELLOW}master${COL_NC}
+       or  ${COL_GREEN}pihole checkout${COL_NC} ${COL_PURPLE}ftl ${COL_CYAN}development${COL_NC}
 
-  Repositories:
-    core [branch]       Change the branch of Pi-hole's core subsystem
-    web [branch]        Change the branch of Web Interface subsystem
-    ftl [branch]        Change the branch of Pi-hole's FTL subsystem
+  Shortcuts:
+    ${COL_YELLOW}master${COL_NC}            Update all subsystems to the latest stable release
+    ${COL_YELLOW}dev${COL_NC}               Update all subsystems to the latest development release
 
-  Branches:
-    master              Update subsystems to the latest stable release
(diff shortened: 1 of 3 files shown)


pihole
```diff
@@ -408,19 +408,21 @@ piholeCheckoutFunc() {
...
-
-  Branches:
-    master              Update subsystems to the latest stable release
-    dev                 Update subsystems to the latest development release
-    branchname          Update subsystems to the specified branchname"
+      echo "Switch Pi-hole subsystems to a different GitHub branch
+    Usage: ${COL_GREEN}pihole checkout${COL_NC} ${COL_YELLOW}shortcut${COL_NC}
+        or ${COL_GREEN}pihole checkout${COL_NC} ${COL_PURPLE}repo${COL_NC} ${COL_CYAN}branch${COL_NC}
+  Example: ${COL_GREEN}pihole checkout${COL_NC} ${COL_YELLOW}master${COL_NC}
+       or  ${COL_GREEN}pihole checkout${COL_NC} ${COL_PURPLE}ftl ${COL_CYAN}development${COL_NC}
```
rdwebdesign, line 415: The "Usage/Example" block is a little confusing.

Maybe adding a blank line and changing the indentation will look beter:

```suggestion
  Usage:
    ${COL_GREEN}pihole checkout${COL_NC} ${COL_YELLOW}shortcut${COL_NC}
    ${COL_GREEN}pihole checkout${COL_NC} ${COL_PURPLE}repo${COL_NC} ${COL_CYAN}branch${COL_NC}

  Example:
    ${COL_GREEN}pihole checkout${COL_NC} ${COL_YELLOW}master${COL_NC}
    ${COL_GREEN}pihole checkout${COL_NC} ${COL_PURPLE}ftl ${COL_CYAN}development${COL_NC}
```

image

rdwebdesign (changes requested):

yubiuser (changes requested):
Please do not combine these two unrelated changes:
1) Changing the color -> fine
2) adding a new `Sourcing` variable for `basic-install.sh`

Both changes are not related at all. Additionally, we already have a variable indicating "sourcing" of the installer script: `SKIP_INSTALL` ;-)

DL6ER:
It is not unrelated at all, please see the comment I added above the new `if` in `basic-install.sh`. `SKIP_INSTALL` does not prevent `set -e` and extending `SKIP_INSTALL` to do `set -e` may have unwanted side-effects in other scripts.

The point is that `pihole checkout ftl non/existing_branch` is currently entirely broken. As soon as the script detects that a branch is not available, it does a `return 1` internally and `set -e` kills the script **instead** of the script continuing and showing a list of available branches as it ought to be.

Hence, we indeed need both changes in exactly this combination or you cannot review the changes properly. As the variable is only added in one place and does not exist anywhere else, there are no side-effects.

yubiuser:
I see what you mean, but don't think this is the best solution. The issue I see here is that the whole check for available FTL binaries/branches is mixed and reverted compared to `core` and `web`. In the latter you first get the list of available branches and then check if the desired branch is part of the array. For FTL you check if the binary server has the directory and if not, you check the branches on github. Those two should match in theory, but this might not always be the case. 

For the current situation I see a very simple solution wich does not require a new variable (I don't like the entanglement of all those bash scripts) and a more complex solution.
Simple is to add `set +e` in the checkout script after `basic-install`has been sourced. 
The complex is getting a list of directories from `ftl.pihole.net` and check if the desired path is within this array. Something like 

```
curl --silent 'https://ftl.pi-hole.net/?'  --data-raw '{"action":"get","items":{"href":"/","what":1}}' | jq --raw-output .items[].href
```

DL6ER:
I think this is a well-justified difference between the repos. You can really only checkout what exists as a branch for the two repos but FTL does not really work on a branch-basis (of course, but it does so in a rather indirect way)

I don't really like the proposed idea of getting a list of downloadable stuff as it'd require at least maintaining a filter list to excluding possible library mirror paths we have on there, plus what might be added in the future for similar reasons.

First querying GitHub for the availability of branches needs another step and, hence, was probably not considered for speed/efficiency/need to actually do it this way. It just doesn't matter if the branch exists or not as long as the binary is there. 

Thinking about it, it's nonetheless probably better to first check for the branch availability as it will prevent users finding old topics having some `pihole checkout ftl ...` command for something that has long been merged (= branch deleted) but where the corresponding binaries may still exist due to their significantly longer retention times.

I undid the new variable change to `basic-install.sh` and fixed it by other means. Currently, `check_download_exists()` (and how it is used) is broken in `development-v6`.

rdwebdesign (approved):
From the formatting/visual point of view it is approved.

yubiuser (approved):
```
  [i] Checking for availability of branch fix/teleporter_v5_cnames on GitHub
  [✓] Branch fix/teleporter_v5_cnames exists on GitHub
  [i] Checking for pihole-FTL-arm64 binary on https://ftl.pi-hole.net
  [✗] Binary for requested branch is not available, please try again later.
      If the issue persists, please contact Pi-hole Support and ask them to re-generate the binary.

```
```
  [i] Checking for availability of branch asdad on GitHub
  [✗] Requested branch is not available

  [i] Available branches are:
      - ci/macos14
      - development
      - development-v6
      - fix/cli_pw_restart
      - fix/ede_neterr
...
```

PR #5751, opened 2024-09-03
Add fallback option for OS check without hard-coded nameserver

yubiuser:

Inspired by https://github.com/pi-hole/pi-hole/pull/5683 this PR adds a fallback to the OS check. Currently we check for supported OS via IPv4 and IPv6 via hard-coded name server (ns1.pi-hole.net). If, for any reason (DNS server down, firewall blocking certain nam servers), this fails the installer will exit despite the host might have a working DNS resolution. So I added a check to `versions.pi-hole.net` without any hard-coded nameserver.

Re-use the code from the IPv6 test but remove the hard-coded nameserver

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -173,7 +173,7 @@ os_check_dig(){
     local nameserver="$3"
     local response
 
-    response="$(dig "${protocol}" +short -t txt "${domain}" "${nameserver}" 2>&1
+    response="$(dig -"${protocol}" +short -t txt "${domain}" "${nameserver}" 2>&1
     echo $?
     )"
     echo "${response}"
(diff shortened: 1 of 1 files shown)


DL6ER (commented):
To me, it seems we have duplicated the exact same code three times here. With only the `dig` command being *slightly* different each time - this extreme code duplication screams for outsourcing into a dedicated function/subroutine, right? ;-)

automated install/basic-install.sh
```diff
@@ -167,63 +167,108 @@ is_command() {
     command -v "${check_command}" >/dev/null 2>&1
 }
 
+os_check_dig(){
+    local protocol="$1"
+    local domain="$2"
+    local nameserver="$3"
+    local response
+
+    response="$(dig "${protocol}" +short -t txt "${domain}" "${nameserver}" 2>&1
```
rdwebdesign, line 176: Did you forgot the dash before the protocol argument (`-4`, `-6`)?

```suggestion
    response="$(dig -"${protocol}" +short -t txt "${domain}" "${nameserver}" 2>&1
```

rdwebdesign (dismissed):

automated install/basic-install.sh
yubiuser: Thanks for spotting this.

DL6ER (approved):

PR #5766, opened 2024-09-12
Return early during v6 migration if migration dir exists

yubiuser:

Base decision if we are on v6 not on `pihole.toml` but on the existence of the migration directory. As the next step is the creation of the dir, it's better to check if the dir exist. Otherwise all users running the `development` or `dev-v6` branch will never get this directory. But not having this dir breaks v5 teleporter archiv importing. See https://github.com/pi-hole/FTL/pull/2057#issuecomment-2347079935

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -2146,8 +2146,8 @@ migrate_dnsmasq_configs() {
     # avoid conflicts with other services on this system
 
     # Exit early if this is already Pi-hole v6.0
-    # We decide this on the presence of the file /etc/pihole/pihole.toml
-    if [[ -f /etc/pihole/pihole.toml ]]; then
+    # We decide this on the presence of the migration dir
+    if [[ -d "${V6_CONF_MIGRATION_DIR}" ]]; then
         return 0
     fi
 


DL6ER (approved):
Yes, in this case, it is indeed more meaningful to check for the directory as this is everything the migration function is caring about.

yubiuser:
I think it's enough. 

I see no reason why the dir should exist but the toml file not. If it's v6 FTL binary, it will write the toml at the first start. But the dir is handled by the bash script.

PromoFaux:
Wait wait....

What if the user, post-migration, wants to delete the directory? After all, it is no use once the migration has happened.

As it is, the directory will just keep getting recreated every time this function is run...

DL6ER:
FTL may be permission-limited. If, e.g., mounting `/etc/pihole` we don't actually know if `pihole-FTL` can create files/directories. But this would very likely already have shown other issues as it wouldn't be possible to create `pihole.toml` and friends either. It won't hurt to have FTL check and create the directory if it isn't there. If it fails, we can still `ERROR` log this so users should be able to find out about this rather easily: https://github.com/pi-hole/FTL/pull/2061

PromoFaux:
> If, e.g., mounting `/etc/pihole`

Before we start FTL on the container - we also run the pre-start script from the core repo, which should chown -R the entire /etc/pihole directory? 

https://github.com/pi-hole/pi-hole/blob/development/advanced/Templates/pihole-FTL-prestart.sh#L15C1-L15C51

yubiuser:
Reverting PR https://github.com/pi-hole/pi-hole/pull/5768

PR #5773, opened 2024-09-17
Add database optimization and gravity timing

DL6ER:

This PR adds two features for `pihole -g`:

1. Database optimization
  The SQLIte3 `ANALYZE` command gathers statistics about tables and indices and stores the collected information in internal tables of the database where the query optimizer can access the information and use it to help make better query planning choices
2. Timing feature
  A new optional flag `pihole -g -t` is added which can be used to easily get timing details about the individual steps gravity is running while creating the database. This is added so it will be simple to test the impact of the added database optimization (no. 1 in this PR) on various hardware during review.

---

---

4. I have commented my proposed changes within the code.
5. I am willing to help maintain this change if there are issues with it later.
6. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
7. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

## Checklist:

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -837,6 +837,24 @@ database_recovery() {
   echo ""
 }
 
+gravity_optimize() {
+    # The ANALYZE command gathers statistics about tables and indices and stores
+    # the collected information in internal tables of the database where the
+    # query optimizer can access the information and use it to help make better
+    # query planning choices
+    local str="Optimizing database"
+    echo -ne "  ${INFO} ${str}..."
+    output=$( { pihole-FTL sqlite3 -ni "${gravityTEMPfile}" "PRAGMA analysis_limit=0; ANALYZE" 2>&1; } 2>&1 )
+    status="$?"
+
+    if [[ "${status}" -ne 0 ]]; then
+        echo -e "\\n  ${CROSS} Unable to optimize database ${gravityTEMPfile}\\n  ${output}"
+        gravity_Cleanup "error"
+    else
+        echo -e "${OVER}  ${TICK} ${str}"
+    fi
+}
+
 helpFunc() {
   echo "Usage: pihole -g
 Update domains from blocklists specified in adlists.list
@@ -948,6 +966,9 @@ gravity_build_tree
 # Compute numbers to be displayed (do this after building the tree to get the
 # numbers quickly from the tree instead of having to scan the whole database)
 gravity_ShowCount
+
+# Optimize the database
+gravity_optimize
 
 # Migrate rest of the data from old to new database
 # IMPORTANT: Swapping the databases must be the last step before the cleanup
--- a/gravity.sh
+++ b/gravity.sh
@@ -428,7 +428,7 @@ gravity_DownloadBlocklists() {
     if [[ "${check_url}" =~ ${regex} ]]; then
       echo -e "  ${CROSS} Invalid Target"
     else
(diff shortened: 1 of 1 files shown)


yubiuser (commented):
We once had an automatic gravity optimization PR but droped it (https://github.com/pi-hole/FTL/pull/1385). 

Do you see real improvements after optimizing the database?

DL6ER:
The other PR was trying to optimize the `pihole-FTL.db` database but was not delivering performance benefits for the reasons therein. Here, the situation is different. First of: The `ANALYZE` command's documentation is not very good. It is really important to go to `sqlite3.c` and check what is going on. Even though it is no real fun to read this file.

There are many places where the `sqlite_stat1` table which is generated by `ANALYZE` is used. One particular example relevant for `gravity` is
image
(Github does not support links to specific rows to such large source files)
which causes the database to run `ANALYZE` *when* `OPTIMIZE`  is run next time (which we never do).

As `gravity` is a non-unique index (incidental domains can included multiple times from different adlists), `OPTIMIZE` would run `ANALYZE` which seems to make it a useful thing to do this ourselves (due to the missing `OPTIMIZE`). What `ANALYZE` then does is scanning the table and listing the exact number of lines and also recording that the index `idx_gravity` can be used to *exactly* identify matches even tough there is this non-uniqueness, see this line of the `sqlite_stat1` table: `gravity|idx_gravity|233983 1 1`

It means that, in my case, there are 233983 rows in the table (which is good to know as `sqlite3` does not have to open the index and do the (quick) counting itself or use the somewhat imprecise "row estimator"), and that queries using the index will match 1:1 both for `domain=?` and `domain = ? AND adlist_id = ?` lookups. This makes a few paths in the query planner easier as SQLite3 is more confident that the index will return precise results even though it is non-unique.

Will it make a real world measurable difference? I don't know because this is really hard to be measure. However, looking into the SQLite3 source code, it seems to make a difference in a few places in the query plan optimizer. If it turns out to need several seconds on low-end devices, it is *probably* not worth it. If it is quick enough, it may still be useful.

yubiuser:
On low-end NanoPi Neo Plus 2

```
  [✓] Building tree
  --> took 0.291 seconds
  [i] Number of gravity domains: 65980 (63478 unique domains)
  [i] Number of regex denied filters: 1
  [i] Number of exact allowed domains: 4
  [i] Number of regex allowed filters: 6
  --> took 0.173 seconds
  [✓] Optimizing database
  --> took 0.089 seconds
  [✓] Swapping databases
  --> took 0.032 seconds
  [✓] Cleaning up stray matter
  --> took 0.026 seconds

```

Optimization step is neglectable compare to ~0,7 sec download/parsing time per adlist

gravity.sh
```diff
@@ -837,12 +837,76 @@ database_recovery() {
...
+  fi
+
+  # Capture the end time
+  end_time=$(date +%s%3N)
+
+  # Calculate the elapsed time
+  elapsed_time=$((end_time - start_time))
+
+  # Display the elapsed time
+  printf "  --> took %d.%03d seconds\n" $((elapsed_time / 1000)) $((elapsed_time % 1000))
```
rdwebdesign, line 898: We could add a little bit of color here (blue or maybe gray):
```suggestion
  printf "  %b--> took %d.%03d seconds%b\n" ${COL_BLUE} $((elapsed_time / 1000)) $((elapsed_time % 1000)) ${COL_NC}
```

image

rdwebdesign (dismissed):

yubiuser (approved):

PR #5779, opened 2024-09-23
Remove no longer needed `restartDNS` function from the `pihole` script

PromoFaux:

Something we discussed doing a while back when we first started exploring the changes necessary for v6, but we ultimately did not get around to doing.

All settings changed in the `pihole.toml` file, or via the API or CLI which require a restart of FTL will trigger this automagically, so this function is no longer needed

---

---

### DIFF
--- a/advanced/bash-completion/pihole
+++ b/advanced/bash-completion/pihole
@@ -7,7 +7,7 @@ _pihole() {
 
     case "${prev}" in
         "pihole")
-            opts="allow allow-regex allow-wild deny checkout debug disable enable flush help logging query reconfigure regex restartdns status tail uninstall updateGravity updatePihole version wildcard arpflush api"
+            opts="allow allow-regex allow-wild deny checkout debug disable enable flush help logging query reconfigure regex status tail uninstall updateGravity updatePihole version wildcard arpflush api"
             COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
         ;;
         "allow"|"deny"|"wildcard"|"regex"|"allow-regex"|"allow-wild")
--- a/gravity.sh
+++ b/gravity.sh
@@ -777,12 +777,6 @@ gravity_Cleanup() {
 
   echo -e "${OVER}  ${TICK} ${str}"
 
-  # # Only restart DNS service if offline
-  # if ! pgrep pihole-FTL &> /dev/null; then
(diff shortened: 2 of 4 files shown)


rdwebdesign (approved):

DL6ER (approved):

yubiuser:
This requires change here: https://docs.pi-hole.net/core/pihole-command/

Bucking-Horn:
PromoFaux wrote:
> All settings changed in the pihole.toml file, or via the API or CLI which require a restart of FTL will trigger this automagically, so this function is no longer needed

What about `pihole restartdns reload` and `pihole restartdns reload-lists`?
Would changes in custom dnsmasq configuration files also be detected?

PromoFaux:
Good question... @DL6ER ?

(I personally have never used either of those commands, but I don't have much more than a set and forget instance)

PromoFaux:
Though thinking about it, for bare metal, could we not just direct users to:

`systemctl restart pihole-FTL` / `service pihole-FTL restart` ?

(And for docker just restart the container?)

PromoFaux:
> This requires change here: [docs.pi-hole.net/core/pihole-command](https://docs.pi-hole.net/core/pihole-command/)

https://github.com/pi-hole/docs/pull/1099

PromoFaux:
> If I remember correctly, it would be advised to run those e.g. when updating the database via custom SQL commands.

Can the same not be achieved by using the system to restart FTL?

> `systemctl restart pihole-FTL` / `service pihole-FTL restart` ?

PromoFaux:
Thinking out loud.

Maybe we drop the `restartdns` option, but move the `reload` and `reload-lists` up a level (i.e `pihole reload` / `pihole reload-lists`).. Let me look into an alternative PR

PromoFaux:
> but move the `reload` and `reload-lists` up a level

image

Looks like this has already been done for just `reload`, so I think it makes sense

Bucking-Horn:
> Maybe we drop the `restartdns` option, but move the `reload` and `reload-lists` up a level

No objections from me, that sounds rather intuitive. 

That would then just leave custom dnsmasq configurations to be considered?

PromoFaux:
> That would then just leave custom dnsmasq configurations to be considered?

I think that's where we then advise to use the system service commands to restart the FTL daemon

Issue #5772, opened 2024-09-17
Duplicate output when updating FTL


compulsions:
### Versions
Core
    Version is v5.18.3-490-g0cf0b2ab (Latest: null)
    Branch is development
    Hash is 0cf0b2ab (Latest: 0cf0b2ab)
Web
    Version is v5.21-947-g05c286df (Latest: null)
    Branch is development
    Hash is 05c286df (Latest: 05c286df)
FTL
    Version is vDev-72bfc98 (Latest: null)
    Branch is development
    Hash is 72bfc983 (Latest: 72bfc983)

### Platform
- OS and version:  Debian 12
- Platform:  Raspberry Pi W Zero 2

### Expected behavior
Not have repeatable output when doing pihole -up

### Actual behavior / bug
  [i] Checking for updates...
  [i] Pi-hole Core:     up to date
  [i] Web Interface:    up to date
  [i] FTL binary already installed, verifying integrity...
  [✗] Remote binary is different, downloading...
  [i] FTL:              update available

  [✓] Detected AArch64 (64 Bit ARM) architecture
  [i] FTL binary already installed, verifying integrity...
  [✗] Remote binary is different, downloading...
  [✓] Downloading and Installing FTL
  [✓] Installing scripts from /etc/.pihole

### Steps to reproduce
Perform a pihole -up when there is an update available.

## Debug Token
- URL:  https://tricorder.pi-hole.net/lafcwxOC/

### Screenshots
![image](https://github.com/user-attachments/assets/b0e38867-dd2a-4682-8c7a-bea1ae9bffdd)

### Additional context
Is it possible to remove the following output
  [i] FTL binary already installed, verifying integrity...
  [✗] Remote binary is different, downloading...
   at the top when checking if the components are up to date and keep the other one?

yubiuser:
Should be fixed by https://github.com/pi-hole/pi-hole/pull/5800

PR #5800, opened 2024-10-18
Do not print FTL update check details on pihole -up

yubiuser:

Should fix https://github.com/pi-hole/pi-hole/issues/5772.

Currently we source `FTLcheckUpdate()` from the install script to check for FTL updates. This prints a few lines during the update check. If a FTL update exists, the same function is executed again as part of the installer run later during the update. 

While only being cosmetic, I can understand that a duplicated output is not desired. . As we already have code in place to deal with a non-zero return code I think it is fine to suppress the detailed output during the check.

---

---

### DIFF
--- a/advanced/Scripts/update.sh
+++ b/advanced/Scripts/update.sh
@@ -144,7 +144,7 @@ main() {
     local binary
     binary="pihole-FTL${funcOutput##*pihole-FTL}" #binary name will be the last line of the output of get_binary_name (it always begins with pihole-FTL)
 
-    if FTLcheckUpdate "${binary}"; then
+    if FTLcheckUpdate "${binary}" &>/dev/null; then
         FTL_update=true
         echo -e "  ${INFO} FTL:\\t\\t${COL_YELLOW}update available${COL_NC}"
     else


advanced/Scripts/update.sh
```diff
@@ -144,27 +144,29 @@ main() {
...
-                ;;
-            *)
-                echo -e "  ${INFO} FTL:\\t\\t${COL_LIGHT_RED}Something has gone wrong, contact support${COL_NC}"
-                exit 1
-        esac
-        FTL_update=false
-    fi
+    local ftl_needs_update
+    ftl_needs_update=$(FTLcheckUpdate "${binary}") #check if FTL needs an update
+    case "${ftl_needs_update}" in
```
DL6ER, line 149: How does this work? `ftl_needs_update` will contain the returned `output` of the function and the `case` uses this. However, the function has been designed to return these numbers as *exit codes*, so something like `$?` is required here.

yubiuser: You're correct. I pushed an update.

yubiuser:
```
nanopi@nanopi:/etc/.pihole$ sudo pihole -up
  [✓] Update local cache of available packages
  [✓] Checking for iproute2

  [i] Checking for updates...
  [i] Pi-hole Core:	up to date
  [i] Web Interface:	up to date
  [i] FTL:		update available
  [i] Warning: You are using FTL from a custom branch (new/lookup_table) and might be missing future releases.

  [i] FTL out of date, it will be updated by the installer.

  [✓] Root user check

....

  [i] FTL Checks...

  [✓] Detected AArch64 (64 Bit ARM) architecture
  [i] FTL binary already installed, verifying integrity...
  [✗] Remote binary is different, downloading...
  [✓] Downloading and Installing FTL
  [✓] Installing scripts from /etc/.pihole


...
  [✓] Done.

  [i] The install log is located at: /etc/pihole/install.log
  [✓] Update complete! 

Core
    Version is v5.18.3-517-g1ad135a (Latest: null)
    Branch is fix/no_duplicate
    Hash is 1ad135ad (Latest: 1ad135ad)
Web
    Version is v5.21-988-g8ad3397d (Latest: null)
    Branch is dependabot/npm_and_yarn/development/chart.js-4.4.5
    Hash is 8ad3397d (Latest: 8ad3397d)
FTL
    Version is vDev-cf0f8e8 (Latest: null)
    Branch is new/lookup_table
    Hash is cf0f8e8d (Latest: cf0f8e8d)
```

PR #5818, opened 2024-11-16
Gravity database resilience

DL6ER:

When something bad (e.g. file corruption) happens to the `gravity.db` file, users do not only loose their current `gravity` domains but also their possibly laboriously composed subscribed lists, allowed/denied domains and regex filters.

Gravity runs trying to use a corrupted database return something like:
``` bash
$ sudo pihole -g
  [✓] DNS resolution is available

  [i] Neutrino emissions detected...
  [✓] Pulling blocklist source list into range
  [i] No source list found, or it is empty


  [✓] Preparing new gravity database
  [i] Creating new gravity databases...
  [✗] Unable to copy data from /etc/pihole/gravity.db to /etc/pihole/gravity.db_temp
  Runtime error near line 3: file is not a database (26)
Parse error near line 11: no such table: OLD.group
Parse error near line 12: no such table: OLD.domain_audit
Parse error near line 14: no such table: OLD.domainlist
Parse error near line 15: no such table: OLD.domainlist_by_group
Parse error near line 16: no such table: OLD.domainlist_by_group
Parse error near line 18: no such table: OLD.adlist
Parse error near line 19: no such table: OLD.adlist_by_group
Parse error near line 20: no such table: OLD.adlist_by_group
Parse error near line 22: no such table: OLD.client
Parse error near line 23: no such table: OLD.client_by_group
Parse error near line 24: no such table: OLD.client_by_group
   [✗] Unable to create gravity database. Please try again later. If the problem persists, please contact support.
```

This PR creates up to ten backups of the current `gravity.db` **after each *successful* `pihole -g` run**. This ensures that only valid databases are backed up. These database copies are actually stripped-down variants (tables `gravity` and `antigravity` are dropped, then the database is vacuumed to shrink in size) to avoid huge backups lying around.
This cleaning procedure reduces the file size of `gravity.db` from dozens of MB to typically 64 KB, i.e., up to 10 backups means up to less than 1 MB of space used for these new backups.

If the database file is getting corrupted, this new branch is able to recover *if there was at least one successful `pihole -g` run on this branch before*, e.g.:
``` bash
# Just a safety backup and then intentional file corruption
cp /etc/pihole/gravity.db ~/gravity.db
echo "abc" > /etc/pihole/gravity.db 

$ sudo pihole -g
  [✓] DNS resolution is available

  [i] Neutrino emissions detected...

  [✓] Preparing new gravity database
  [i] Creating new gravity databases...
  [✗] Unable to copy data from /etc/pihole/gravity.db to /etc/pihole/gravity.db_temp
  Runtime error near line 3: file is not a database (26)
Parse error near line 11: no such table: OLD.group
Parse error near line 13: no such table: OLD.domainlist
Parse error near line 14: no such table: OLD.domainlist_by_group
Parse error near line 15: no such table: OLD.domainlist_by_group
Parse error near line 17: no such table: OLD.adlist
Parse error near line 18: no such table: OLD.adlist_by_group
Parse error near line 19: no such table: OLD.adlist_by_group
Parse error near line 21: no such table: OLD.client
Parse error near line 22: no such table: OLD.client_by_group
Parse error near line 23: no such table: OLD.client_by_group
  [i] Attempting to restore previous database from backup                                    <------------
  [✓] Successfully restored from backup (/etc/pihole/gravity_backup.db.1)                    <------------
  [✓] Creating new gravity databases
  [✓] Pulling blocklist source list into range
  [i] Using libz compression

  [✓] Parsed 121556 exact domains and 0 ABP-style domains (blocking, ignored 0 non-domain entries)
  [i] List stayed unchanged

  [✓] Building tree
  [i] Number of gravity domains: 121556 (121556 unique domains)
  [i] Number of regex denied filters: 0
  [i] Number of exact allowed domains: 4
  [i] Number of regex allowed filters: 0
  [✓] Optimizing database
  [✓] Swapping databases
  [✓] Cleaning up stray matter

  [✓] Done.
```

---

**Related issue or feature (if applicable):** actually quite a few over the past weeks

---

## Checklist:


gravity.sh
```diff
@@ -325,32 +391,6 @@ gravity_DownloadBlocklists() {
...
-
-  local str="Pulling blocklist source list into range"
-  echo -e "${OVER}  ${TICK} ${str}"
-
-  if [[ -z "${sources[*]}" ]] || [[ -z "${sourceDomains[*]}" ]]; then
-    echo -e "  ${INFO} No source list found, or it is empty"
-    echo ""
-    unset sources
-  fi
-
```
DL6ER, line 353: This block has been moved without changing below after a possibly happening database restoration

gravity.sh
```diff
@@ -317,6 +344,45 @@ gravity_CheckDNSResolutionAvailable() {
...
+#              restoration is successful, it returns 0. Otherwise, it returns 1.
+# Returns:
+#   0 - If the backup is successfully restored.
+#   1 - If no backup is available or if the restoration fails.
+try_restore_backup () {
+  local num
+  num=$1
+  # Check if a backup exists
+  if [ -f "${gravityBCKfile}.${num}" ]; then
+    echo -e "  ${INFO} Attempting to restore previous database from backup"
```
rdwebdesign, line 360: Maybe we should move this message to line 428 (before the for loop) to avoid repetition, or change the message to include the backup file number:

```suggestion
    echo -e "  ${INFO} Attempting to restore previous database from Backup no. ${num}"
```

rdwebdesign (changes requested):

gravity.sh
DL6ER: Any specific reason for writing `Backup` starting with a capital letter? Not being a native speaker, my feeling is still that nothing is to be written with a capital letter in English mid-sentence - except for names.

rdwebdesign: No reason at all... I just copied/pasted the text from line 382 and didn't notice the uppercase letter.

I just fixed my suggestion.

yubiuser (commented):
I'm not a huge fan of making `gravity` even more complicated. Shouldn't we not encourage to backup their settings and gravity via `teleporter` instead?
If we backup and restore by default, most users won't notice if there are any issues, because gravity will run silently during Sun night. Most likely the file will be corrupted due to file system issues - this should not be hidden to the users, but instead fail hard to make them aware.

DL6ER:
I disagree on "fail hard" as, currently, it goes together with **(unrecoverable) data loss** which can really upset users and make them migrate away from Pi-hole. Assume what you might be doing when you are only a user, maybe invested a lot of time and effort into your multi-million blocked domains collection and also spent nights collecting regular expressions online in some shady forums. I don't think gravity should ever loose its content but current it can - and watching/reading Discourse this happens at least on the order of once per week. And these are only the users that do the extra lengths of creating an account and asking for help.

We *could* remove the automated recovery mode and only offer it through manual `pihole -g --repair` or something but I don't really see the benefit and - very probably - it would take even a few lines code in addition. This change comes at about 50 additional lines if you strip existing code that is just moved around and comments - and I don't think it adds any complicated logic (like crazy `awk` stuff, etc.) to the file.

yubiuser:
> I disagree on "fail hard" as, currently, it goes together with (unrecoverable) data loss which can really upset users and make them migrate away from Pi-hole. Assume what you might be doing when you are only a user, maybe invested a lot of time and effort into your multi-million blocked domains collection and also spent nights collecting regular expressions online in some shady forums. I don't think gravity should ever loose its content but current it can - and watching/reading Discourse this happens at least on the order of once per week. And these are only the users that do the extra lengths of creating an account and asking for help.


I get your point, but still think we should let users when know if there were any issues that needed a backup replacement. Can we add warning to the `message` table at least?

DL6ER:
image

together with https://github.com/pi-hole/FTL/pull/2122

PromoFaux:
So, this only kicks in when `gravity.db` is corrupted beyond help, right? We're not likely to overwrite a known-good `gravity.db` with an older one?

DL6ER:
Exactly

dschaper:
10? Can we cap it at 3 or so? We only need one valid backup to pull from and if 9 of the backups are corrupted then there are likely other issues to address that should not be quietly 'fixed' by pulling a backup that is 10 weeks old. (Assuming the default of running gravity once on Sunday).

dschaper:
How resource intensive is the vacuum process? Will this impact low memory, slow single core setups?

DL6ER:
> 10? Can we cap it at 3 or so?

We could but this offers a simple way of "going" back in case users have deleted stuff accidentally. A backup gravity file wights around 50 KB even with a few hundred manual lists and the vacuum process is needed only once.

> How resource intensive is the vacuum process? Will this impact low memory, slow single core setups?

It is a very easy task requiring less than 150 KB of memory. On my small `x86_64` microserver, the process takes less than a millisecond. As the possibly multi-million entries table `gravity` is *not* included in the backup, I expect no performance differences even for users running possibly dozens of millions of blocked domains.

yubiuser:
This needs to be rebased on `development`

darkexplosiveqwx:
Hello,
I would suggest adding a timestamp in some sort to the backup,
so the user better knows what configurations have been lost and to help finding the cause of the corruption.
This timestamp could also be shown in the gravity message.
I know that the files are timestamped in a way with the numbers, but it might not be clear from when that particular number is.

DL6ER:
> I would suggest adding a timestamp in some sort to the backup,

There is already a timestamp - even if only indirectly:
```
$ ls -lh /etc/pihole/gravity_backup.db.*
-rw-rw---- 1 pihole pihole 88K Jan 12 03:12 /etc/pihole/gravity_backup.db.1
-rw-rw---- 1 pihole pihole 96K Jan  5 03:12 /etc/pihole/gravity_backup.db.2
-rw-rw---- 1 pihole pihole 96K Nov 29 03:12 /etc/pihole/gravity_backup.db.3
-rw-rw---- 1 pihole pihole 96K Nov 22 03:12 /etc/pihole/gravity_backup.db.4
```

Furthermore, we have the last updated timestamp inside the databases themselves, e.g.,
```
$ pihole-FTL sqlite3 -h /etc/pihole/gravity_backup.db.1 "SELECT * FROM info;"
property       value     
-------------  ----------
version        19        
updated        1736647923   <------------- Sunday, 12. January 2025 03:12:03 CET
gravity_count  169279
```

yubiuser (commented):
I'd like to keep `/etc/pihole` clean a bit. Can me have a dedicated folder for this? Like we have it for config_backup, migration_backup, migration_backup_v6...

darkexplosiveqwx:
> > I would suggest adding a timestamp in some sort to the backup,
> 
> There is already a timestamp - even if only indirectly:
> 
> ```
> $ ls -lh /etc/pihole/gravity_backup.db.*
> -rw-rw---- 1 pihole pihole 88K Jan 12 03:12 /etc/pihole/gravity_backup.db.1
> -rw-rw---- 1 pihole pihole 96K Jan  5 03:12 /etc/pihole/gravity_backup.db.2
> -rw-rw---- 1 pihole pihole 96K Nov 29 03:12 /etc/pihole/gravity_backup.db.3
> -rw-rw---- 1 pihole pihole 96K Nov 22 03:12 /etc/pihole/gravity_backup.db.4
> ```
> 
> Furthermore, we have the last updated timestamp inside the databases themselves, e.g.,
> 
> ```
> $ pihole-FTL sqlite3 -h /etc/pihole/gravity_backup.db.1 "SELECT * FROM info;"
> property       value     
> -------------  ----------
> version        19        
> updated        1736647923   <------------- Sunday, 12. January 2025 03:12:03 CET
> gravity_count  169279
> ```

I would still suggest showing the time in the Gravity Message, because it it probably not clear to the user that Backups are being made every Sunday(I suppose corruption is not happening every other week).

rdwebdesign:
> I would suggest adding a timestamp in some sort to the backup,
> so the user better knows what configurations have been lost

This backup is only saving Gravity information (adlists, domains, clients and groups). 
This is not a full backup. No configurations will be stored.

Maybe we could add a message here, saying the old database was moved to the backup directory:
https://github.com/pi-hole/pi-hole/blob/cc25ee940e00d211c94cdbad1575c19eb634a35b/gravity.sh#L145-L146

(see my code suggestion)

gravity.sh
```diff
@@ -111,10 +122,32 @@ gravity_swap_databases() {
...
+    if [ ! -d "${gravityBCKdir}" ]; then
+      mkdir -p "${gravityBCKdir}"
+    fi
+
+    # If multiple gravityBCKfile's are present (appended with a number), rotate them
+    # We keep at most 10 backups
+    rotate_gravity_backup
+
+    # Move the old database to the backup location
+    mv "${gravityDBfile}" "${gravityBCKfile}.1"
```
rdwebdesign, line 146: I'm not sure if this is necessary, but maybe we could inform the old database was moved to the backup location:

```suggestion
    mv "${gravityDBfile}" "${gravityBCKfile}.1"
    echo -e "  ${INFO} Previous database moved to ${gravityBCKfile}.1"
```

DL6ER: I don't think this adds much value here but makes the output unnecessarily longer

DL6ER:
@darkexplosiveqwx Ah, yes. Thanks for rephrasing it again in other words. I wholehearted agree that showing the timestamp in the message is a lot more user-friendly than showing some (arbitrarily looking) filename. Change already pushed to this PR.

rdwebdesign (approved):

DL6ER:
A message will be shown in the Pi-hole diagnosis system once https://github.com/pi-hole/FTL/pull/2122 has been merged

gravity.sh
```diff
@@ -390,10 +445,53 @@ gravity_DownloadBlocklists() {
...
+    if [[ -d "${gravityBCKdir}" ]]; then
+      for i in {1..10}; do
+        if try_restore_backup "${i}"; then
+          break
+        fi
+      done
+    fi
+
+    # If none of the attempts worked, return 1
+    if [[ "${i}" -eq 10 ]]; then
```
yubiuser, line 459: Doesn't this falsely signal backup failed it restoration worked on the 10th backup?

yubiuser (commented):
Do we still need the full-size gravity backup added by https://github.com/pi-hole/pi-hole/pull/4202 now that we have up to 10 small size backups?

DL6ER:
> Do we still need the full-size gravity backup added by #4202 now that we have up to 10 small size backups?

Checking the related [Discourse discussion](https://discourse.pi-hole.net/t/gravity-database/46182) it seems this was never meant to be a backup but rather:

> ... this to allow users to determine which domains have been added / removed since the last run ...

for which we still need this database as it is.

gravity.sh
DL6ER: https://github.com/pi-hole/pi-hole/pull/5868

PR #5827, opened 2024-12-14
Fix rare case when apt and rpm package managers are found

yubiuser:

Fixes an edge case when `apt` and `rpm` both are present and the meta package is build/installed

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -484,9 +484,8 @@ build_dependency_package(){
 
         # Move back into the directory the user started in
         popd &> /dev/null || return 1
-    fi
 
-    if is_command rpm; then
+    elif is_command rpm; then
         # move into the tmp directory
         pushd /tmp &>/dev/null || return 1
 
@@ -517,6 +516,13 @@ build_dependency_package(){
 
         # Move back into the directory the user started in
         popd &> /dev/null || return 1
+
+    # If neither apt-get or yum/dnf package managers were found
+    else
+        # we cannot build required packages
+        printf "  %b No supported package manager found\\n" "${CROSS}"
+        # so exit the installer
+        exit 1
     fi
 
     # Remove the build directory
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1517,11 +1517,9 @@ install_dependent_packages() {
             printf "  %b Error: Unable to find Pi-hole dependency meta package.\\n" "${COL_LIGHT_RED}"
             return 1
         fi
-    fi
-
     # Install Fedora/CentOS packages
-    if is_command rpm; then
-            if [ -f /tmp/pihole-meta.rpm ]; then
+    elif is_command rpm; then
+        if [ -f /tmp/pihole-meta.rpm ]; then
             eval "${PKG_INSTALL}" "/tmp/pihole-meta.rpm"
             rm /tmp/pihole-meta.rpm
         else
@@ -1529,7 +1527,14 @@ install_dependent_packages() {
             return 1
         fi
 
+    # If neither apt-get or yum/dnf package managers were found
+    else
+        # we cannot install the dependency package
+        printf "  %b No supported package manager found\\n" "${CROSS}"
+        # so exit the installer
+        exit 1
     fi
+
     printf "\\n"
     return 0
 }


DL6ER (approved):
If both `apt-get` and `rpm` are installed/available, this gives precedence to `apt-get`. This solution will be sufficient for the current issue and, as it replicates what is in use elsewhere in the installer, is logical. However, can we maybe (for a future PR) determine what is the preferred package manager by the system?

```
Ubuntu :~$ apt-cache show dnf
Package: dnf
Architecture: all
Version: 4.14.0-4.1ubuntu1
Multi-Arch: foreign
Priority: optional
Section: universe/admin
Origin: Ubuntu
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: Frédéric Pierret <frederic.pierret@qubes-os.org>
Bugs: https://bugs.launchpad.net/ubuntu/+filebug
Installed-Size: 67
Depends: python3:any, dnf-data (= 4.14.0-4.1ubuntu1), python3-dnf (= 4.14.0-4.1ubuntu1), libmodulemd2 (>= 2.9.4-1~), python3-dbus, sqlite3
Breaks: dnf-plugins-core (<< 4.0.6-1~)
Filename: pool/universe/d/dnf/dnf_4.14.0-4.1ubuntu1_all.deb
Size: 37658
MD5sum: 35e99bb064d28cba5d31113f6d3677d6
SHA1: 19e486fc75b4dcfbc241af712cc594554d0cb64f
SHA256: 71794a344bb091436aebfba82ea627081fbdb8d1abd341c1533966f744a0c2c2
SHA512: 28e60ce1bed4bdad55ca86ce552bd3a5eb99e733efb684d78962d7db24f6af93707d9c1617258bbf7690e711264b6e91aa30a693fc43758af11c4d79e4dafc05
Homepage: https://github.com/rpm-software-management/dnf
Description-en: Dandified Yum package manager
 Package manager forked from Yum, using libsolv as a dependency resolver.
 Dnf requires a working rpm installation and is meant to be used in chroots,
 see README.Debian for more information.
 .
 This package provides the dnf program.
Description-md5: 35ecd037027703f5e8e5541dfb39e073
```
```
Fedora:~ # dnf search apt
Updating and loading repositories:
Repositories loaded.
Matched fields: name (exact)
 apt.x86_64: Command-line package manager for Debian packages
```

The only reason I could see why users are doing this is when they want to install an `rpm` package for something where a `deb` package isn't available (commonly seen for some proprietary printer drivers).

rdwebdesign:
> The only reason I could see why users are doing this is when they want to install an `rpm` package for something where a `deb` package isn't available (commonly seen for some proprietary printer drivers).

I think this is one reason for having `dnf` in Debian: 

> From [Debian packages page](https://packages.debian.org/bullseye/dnf):
> 
> ### Package: dnf (4.5.2-6)
> #### Dandified Yum package manager
> Package manager forked from Yum, using libsolv as a dependency resolver. 
> Dnf requires a working rpm installation and ***is meant to be used in chroots***, see README.Debian for more information.

Apparently `apt` can be installed in Fedora/RH, but it can't be used to install packages:
- https://docs.fedoraproject.org/en-US/quick-docs/dnf-vs-apt/#_why_is_apt_in_the_fedora_repositories

> APT can not be used to install packages on Fedora, you have to use DNF instead. 

> The apt command on Fedora used to — until [Fedora 32](https://fedoraproject.org/wiki/Changes/Move_apt_package_from_RPM_to_DPKG_backend) — actually be APT-RPM, which basically mapped normal `apt` commands so that they worked with Fedora’s RPM package management system.
> However, APT-RPM is unmaintained, broken, and insecure, and so was dropped in favour of shipping the actual Debian APT software. Since APT exclusively deals with `.deb` packages, the `apt` command can no longer be used to manage Fedora packages. Its purpose is now purely as a tool for people building packages for Debian-based distributions on a Fedora system.

yubiuser:
> However, can we maybe (for a future PR) determine what is the preferred package manager by the system?

I'm not sure there is an easy way to do this. The only idea I have right now is to deduct this from the distro used.

dschaper:
Perhaps we could see if both package managers are installed and then check the `/etc/os-release` in that condition? Prefer the manager that matches the release. I don't remember why we didn't look at the `os-release` file when making the installer, or if we did, why we are not using it. My guess would be that we didn't care what the release was since we were looking for the active and usable package manager.

DL6ER:
My assumption is that the current implementation "just works". It rarely happens that both tools are installed, it is interesting that we even encountered a single setup like this. But then the past seems to indicate that using `apt` when it is available always succeeded. Checking `os-release` is like the one-out-of-a-trillion type of solution.

rdwebdesign:
> It rarely happens that both tools are installed

I agree.
`dnf` on Debian is very rare. I think `apt` on Fedora is just as (if not more) rare.

Pi-hole v5 uses the same method (`is_command apt` and `is_command rpm`) to identify the package manager and I don't remember this ever happening before.

PR #5790, opened 2024-10-02
Use pihole.toml to decide if installer runs on an update

yubiuser:
**What does this PR aim to accomplish?:**

Currently, a fresh v6 installation is not possible. The installer always assumes  it is an update. Reason is the removal of `setupVars.conf` by https://github.com/pi-hole/pi-hole/commit/7cbe713873d38eb4b4952ea7e24a59f4c0084ed7

This PR adds a check for an existing installation by checking the existence of `pihole.toml`

---
**By submitting this pull request, I confirm the following:**

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against.
2. I have commented my proposed changes within the code and I have tested my changes.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
6. I have checked that another pull request for this purpose does not exist.
7. I have considered, and confirmed that this submission will be valuable to others.
8. I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
9. I give this submission freely, and claim no ownership to its content.

---
- [x] I have read the above and my PR is ready for review. *Check this box to confirm*

PR #5830, opened 2024-12-21
Explicitly migrate from v5 to v6

DL6ER:
# What does this implement/fix?

**Explicitly** migrate config from v5 to v6 instead of :crossed_fingers: that this is done automagically on the first (re)start of `pihole-FTL`. We have seen at least [one report](https://discourse.pi-hole.net/t/broken-installation/74638/24) where a race-collision was able to destroy that hope.

---

**Related issue or feature (if applicable):** N/A

**Pull request in [docs](https://github.com/pi-hole/docs) with documentation (if applicable):** N/A

---
**By submitting this pull request, I confirm the following:** 

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against. 
2. I have commented my proposed changes within the code.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

## Checklist:

- [x] The code change is tested and works locally.
- [x] I based my code and PRs against the repositories `developmental` branch.
- [x] I [signed off](https://docs.pi-hole.net/guides/github/how-to-signoff/) all commits. Pi-hole enforces the [DCO](https://docs.pi-hole.net/guides/github/dco/) for all contributions
- [x] I [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) all my commits. Pi-hole requires signatures to verify authorship
- [x] I have read the above and my PR is ready for review.

PR #5832, opened 2024-12-22
Fix v5 -> v6 update

yubiuser:

Should fix v5 -> v6 updated where `pihole.toml` is not present yet.
Background: https://github.com/pi-hole/pi-hole/pull/5790 fixed fresh v6 installations, but forgot about v5->v6 updates.

Re-add `setupVars.conf` at the right place. If present, this is an update.

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -2401,8 +2401,8 @@ main() {
         exit 1
     fi
 
-    # in case of an update
-    if [[ -f "${PI_HOLE_V6_CONFIG}" ]]; then
+    # in case of an update (can be a v5 -> v6 or v6 -> v6 update)
+    if [[ -f "${PI_HOLE_V6_CONFIG}" ]] || [[ -f "/etc/pihole/setupVars.conf" ]]; then
         # if it's running unattended,
         if [[ "${runUnattended}" == true ]]; then
             printf "  %b Performing unattended setup, no dialogs will be displayed\\n" "${INFO}"


PromoFaux (commented):
So this basically puts the logic back to how it was (or is, in the current `master` branch) so that the update dialogs run (or not) when we detect an existing install of v5, with the addition of checking for the v6 file. Right? 

I think I've understood that correctly!

yubiuser:
Yes this is how I meant it. We can't get rid of `setupVars.conf` completely until everyone is on v6 and above. We *could* add a step where we delete this file once we did create the toml file. Or move it out of the way

yubiuser:
I would do such a migration step after this: https://github.com/pi-hole/pi-hole/pull/5830

yubiuser:
pinging  @DL6ER where FTL would expect `setupVars.conf` during the update? Still in `/etc/pihole` or already in `/etc/pihole/migration_backup_v6`

DL6ER:
> Or move it out of the way

FTL moves it away for us:

grafik

PromoFaux:
This seems to work as intended, I don't get the dialogs when migrating from v5->v6 with this branch.

However this bring up something else... In the test environment I did not first uninstall `lighttpd`, so when FTL starts up it has failed over to using port 8080 - but there is nothing in the output to suggest this. Open to discussion elsewhere, but should we add a check here, or just document this occurrence very well in any release notes?

yubiuser:
>  just document this occurrence very well in any release notes?

I think this is the way to go. No more migration code please.

PromoFaux (approved):

DL6ER (approved):
Approving this after testing this in both a Ubuntu and a Debian VMs

DL6ER:
I'm afraid I have two more PRs, at least one directly related to migration (but you already know it: #5830 )

PR #5835, opened 2024-12-23
Disable lighttpd if found

DL6ER:

Lighttpd is not needed anymore, so let's disable it to avoid conflicts concerning (at least) port 80. We keep all the configuration files in place, so the user can re-enable it if needed and can then also decide how they want to assign ports. Pi-hole v6.0 is a major release and such a "soft" breaking change is fine. In the end, the vast majority of Pi-hole users will only be using `lighttpd` for the Pi-hole web interface which makes it a useless service for them with v6.0+.

---

---

## Checklist:

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -2305,8 +2305,8 @@ disableLighttpd() {
     if [[ -t 0 ]]; then
         # The terminal is interactive
         dialog --no-shadow --keep-tite \
-            --title "Pi-hole v6.0 does no longer need lighttpd" \
-            --yesno "Pi-hole v6.0 has its own embedded web server so lighttpd is no longer needed *unless* you have custom configurations. In this case, you can opt-out of disabling lighttpd and pihole-FTL will try to bind to an alternative port such as 8080.\\n\\nDo you want to disable lighttpd (recommended)?" "${r}" "${c}"
+            --title "Pi-hole v6.0 no longer uses lighttpd" \
+           --yesno "\\n\\nPi-hole v6.0 has its own embedded web server so lighttpd is no longer needed *unless* you have custom configurations.\\n\\nIn this case, you can opt-out of disabling lighttpd and pihole-FTL will try to bind to an alternative port such as 8080.\\n\\nDo you want to disable lighttpd (recommended)?" "${r}" "${c}"
         response=$?
     else
         # The terminal is non-interactive, assume yes. Lighttpd will be stopped
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -2301,13 +2301,13 @@ copy_to_install_log() {
 }
 
 disableLighttpd() {
+    local response
     # Detect if the terminal is interactive
     if [[ -t 0 ]]; then
         # The terminal is interactive
         dialog --no-shadow --keep-tite \
             --title "Pi-hole v6.0 no longer uses lighttpd" \
-           --yesno "\\n\\nPi-hole v6.0 has its own embedded web server so lighttpd is no longer needed *unless* you have custom configurations.\\n\\nIn this case, you can opt-out of disabling lighttpd and pihole-FTL will try to bind to an alternative port such as 8080.\\n\\nDo you want to disable lighttpd (recommended)?" "${r}" "${c}"
-        response=$?
+           --yesno "\\n\\nPi-hole v6.0 has its own embedded web server so lighttpd is no longer needed *unless* you have custom configurations.\\n\\nIn this case, you can opt-out of disabling lighttpd and pihole-FTL will try to bind to an alternative port such as 8080.\\n\\nDo you want to disable lighttpd (recommended)?" "${r}" "${c}" && response=0 || response="$?"
     else
         # The terminal is non-interactive, assume yes. Lighttpd will be stopped
         # but keeps being installed and can easily be re-enabled by the user
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -2301,6 +2301,24 @@ copy_to_install_log() {
 }
 
 disableLighttpd() {
+    # Detect if the terminal is interactive
+    if [[ -t 0 ]]; then
+        # The terminal is interactive
+        dialog --no-shadow --keep-tite \
+            --title "Pi-hole v6.0 does no longer need lighttpd" \
+            --yesno "Pi-hole v6.0 has its own embedded web server so lighttpd is no longer needed *unless* you have custom configurations. In this case, you can opt-out of disabling lighttpd and pihole-FTL will try to bind to an alternative port such as 8080.\\n\\nDo you want to disable lighttpd (recommended)?" "${r}" "${c}"
+        response=$?
+    else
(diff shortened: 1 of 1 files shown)


PromoFaux:
I have mixed feelings about this - somehow I think it should be provided as an option to the user via a dialog or some other interactive check, while still defaulting to this behaviour. 

I agree with your assessment that the majority of users are _probably_ only using `lighttpd` for Pi-hole, but at this stage there could be untold numbers of people who have gone "Well, Pi-hole has installed `lighttpd`, and I also want a webserver for another project - so I will use that" - those will be the people that make the most noise

DL6ER:
I do get your point but the upgrade process is meant to run "unattended", i.e., it won't show any dialogs. We can change that but make sure to show the dialog *very* early, actually before doing anything. I don't think we can actually show it before pulling the repos. My point here is the high likeliness that users run an update cronjob which would faild hard at any kind of dialog. Unless there is a reliable way to detect this non-interactivity and then assume a Yes on this dialog? What do you think?

DL6ER:
grafik

automated install/basic-install.sh
```diff
@@ -2300,6 +2300,39 @@ copy_to_install_log() {
     chown pihole:pihole "${installLogLoc}"
 }
 
+disableLighttpd() {
+    # Detect if the terminal is interactive
+    if [[ -t 0 ]]; then
+        # The terminal is interactive
+        dialog --no-shadow --keep-tite \
+            --title "Pi-hole v6.0 does no longer need lighttpd" \
```
PromoFaux, line 2308:
```suggestion
            --title "Pi-hole v6.0 no longer uses lighttpd" \
```

Just to un-german the grammar

PromoFaux (changes requested):

automated install/basic-install.sh
```diff
@@ -2300,6 +2300,39 @@ copy_to_install_log() {
...
+            --yesno "Pi-hole v6.0 has its own embedded web server so lighttpd is no longer needed *unless* you have custom configurations. In this case, you can opt-out of disabling lighttpd and pihole-FTL will try to bind to an alternative port such as 8080.\\n\\nDo you want to disable lighttpd (recommended)?" "${r}" "${c}"
```
rdwebdesign, line 2309: Also, I suggest some line breaks to add a little space and increase readability:

 ```suggestion
            --yesno "\\n\\nPi-hole v6.0 has its own embedded web server so lighttpd is no longer needed *unless* you have custom configurations.\\n\\nIn this case, you can opt-out of disabling lighttpd and pihole-FTL will try to bind to an alternative port such as 8080.\\n\\nDo you want to disable lighttpd (recommended)?" "${r}" "${c}"
```
image

**Review by rdwebdesign — DISMISSED** [review]

**Comment by AyanokojiK1yotaka** [comment]
It would be helpful to provide a mechanism for users to uninstall Lighttpd, as the majority of users may not need this service. The only exception would be those who use Lighttpd for specific purposes, such as acting as a reverse proxy or load balancing.

**Comment by PromoFaux** [comment]
Testing this with `pihole checkout`...

Selecting yes works as expected - `lighttpd` is disabled and `pihole-FTL` binds to port 80
Selecting no throws the following:

image

Relevant `bash -x` output:

```txt
+ systemctl is-enabled systemd-resolved
+ printf '  %b %b Testing if systemd-resolved DNSStub-Listener is active' '\r' '[i]'
 [i] Testing if systemd-resolved DNSStub-Listener is active+ grep -E '#?DNSStubListener=yes' /etc/systemd/resolved.conf
+ printf '%b  %b Systemd-resolved does not need to be restarted\n' '\r' '[i]'
  [i] Systemd-resolved does not need to be restarted
+ disableLighttpd
+ [[ -t 0 ]]
+ dialog --no-shadow --keep-tite --title 'Pi-hole v6.0 no longer uses lighttpd' --yesno '\n\nPi-hole v6.0 has its own embedded web server so lighttpd is no longer needed *unless* you have custom configurations.\n\nIn this case, you can opt-out of disabling lighttpd and pihole-FTL will try to bind to an alternative port such as 8080.\n\nDo you want to disable lighttpd (recommended)?' 20 70
```

DL6ER:
Ah, sure, `set -e` again. Should be fixed.

PromoFaux (approved):
Now working as expected when selecting no.

```
root@adam-test-droplet:~# netstat -tulpn | grep pihole-FTL
tcp        0      0 0.0.0.0:8080            0.0.0.0:*               LISTEN      13696/pihole-FTL    
tcp        0      0 0.0.0.0:443             0.0.0.0:*               LISTEN      13696/pihole-FTL    
tcp6       0      0 :::8080                 :::*                    LISTEN      13696/pihole-FTL    
tcp6       0      0 :::443                  :::*                    LISTEN      13696/pihole-FTL    
udp        0      0 0.0.0.0:53              0.0.0.0:*                           13696/pihole-FTL    
udp        0      0 0.0.0.0:123             0.0.0.0:*                           13696/pihole-FTL    
udp6       0      0 :::123                  :::*                                13696/pihole-FTL    
root@adam-test-droplet:~# netstat -tulpn | grep lighttpd
tcp        0      0 0.0.0.0:80              0.0.0.0:*               LISTEN      12435/lighttpd      
tcp6       0      0 :::80                   :::*                    LISTEN      12435/lighttpd 
```

And as expected with yes:

```
root@adam-test-droplet:~# netstat -tulpn | grep pihole-FTL
tcp        0      0 0.0.0.0:443             0.0.0.0:*               LISTEN      13736/pihole-FTL    
tcp        0      0 0.0.0.0:80              0.0.0.0:*               LISTEN      13736/pihole-FTL    
tcp6       0      0 :::443                  :::*                    LISTEN      13736/pihole-FTL    
tcp6       0      0 :::80                   :::*                    LISTEN      13736/pihole-FTL    
udp        0      0 0.0.0.0:53              0.0.0.0:*                           13736/pihole-FTL    
udp        0      0 0.0.0.0:123             0.0.0.0:*                           13736/pihole-FTL    
udp6       0      0 :::123                  :::*                                13736/pihole-FTL    
root@adam-test-droplet:~# netstat -tulpn | grep lighttpd
root@adam-test-droplet:~# 
```

DL6ER:
@AyanokojiK1yotaka completely removing lighttpd automatically seems too much IMO given that this makes it (a bit) more complicated to get it going again if needed. New v6 installations won't have it from the beginning

automated install/basic-install.sh
```diff
@@ -2489,6 +2522,9 @@ main() {
     # but before starting or resttarting the ftl service
     disable_resolved_stublistener
 
+    # Disable lighttpd server
+    disableLighttpd
```
yubiuser, line 2526: This is placed here unconditionally. It will trigger on each update/checkout. 
We should put it into the migration part (where it is guarded by `if [[ -f "${PI_HOLE_V6_CONFIG}" ]]; then` and additionally check for the presence and enable/disable state of `lighttpd` at all.

DL6ER: https://github.com/pi-hole/pi-hole/pull/5849

PR #5846, opened 2024-12-31
Check before install

yubiuser:

Do the OS check early, esp. before building and installing the dependency package. The `oc_check()` needs two external commands (`dig` and `grep`) which users need to install manually if they are not found.

Additionally, removes the unused function `test_dpkg_lock`.

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -218,6 +218,21 @@ is_command() {
     command -v "${check_command}" >/dev/null 2>&1
 }
 
+check_os_check_dependencies() {
+    # Check if the required dependencies for the OS check are installed
+    if ! is_command grep; then
+        printf "  %b %bThe Pi-hole installer needs 'grep' but it was not found. Please install it. %b\\n" "${INFO}" "${COL_LIGHT_RED}" "${COL_NC}"
+        exit 1
+    fi
+
+    if ! is_command dig; then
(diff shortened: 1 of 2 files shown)


PromoFaux:
I'm not sold on this one tbh.

The command checks are separate, a user could run the install script once, find that `grep` is not installed, so install it, and then run the installer again only to find out they don't have `dig` installed either.

Secondly, it (very slightly) makes the install process seem more complicated to the end user. The beauty of it as-is is that they (mostly) run one command and away they go. Does the new dependency package not go in stages like the old method?

yubiuser:
> Does the new dependency package not go in stages like the old method?

No it does not. We would need 3 meta packages to achieve this.

I'm fine if we drop the PR, it was mainly intended to prevent user complaints as "you installed all dependencies and only told me afterwards that my OS is unsupported"

yubiuser:
Failing test and review commends were addressed. 

But this still needs discussion if we want it.

DL6ER:
> The command checks are separate, a user could run the install script once, find that grep is not installed, so install it, and then run the installer again only to find out they don't have dig installed either.

No, this won't happen. If any of the two is missing, the user is informed. If both are missing, the user is told that as well during the first run. So. yeah, maybe the user needs to re-run the installer once after taking care of the manual dependencies installation. But I do agree to the feeling that this is making the installation somewhat more complex than necessary. Mind that many users may still be first-time-Linux and first-time-CLI users. They may not have a clue how to install the packages or what this `apt` or `rpm` is about.

yubiuser:
As the doubt regarding this change is bigger than the seen benefit, I'm going to close the PR

PR #5849, opened 2025-01-01
Improve lighttpd disabling

DL6ER:

Show `lighttpd` disabling dialog only once during v6 migration and when `lighttpd` is actually installed

---

## Checklist:

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -2301,6 +2301,11 @@ copy_to_install_log() {
 }
 
 disableLighttpd() {
+    # Return early when lighttpd is not installed
+    if [[ ! -f /etc/lighttpd/lighttpd.conf ]]; then
+        return
+    fi
+
     local response
     # Detect if the terminal is interactive
     if [[ -t 0 ]]; then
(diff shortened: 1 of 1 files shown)


automated install/basic-install.sh
```diff
@@ -2301,6 +2301,11 @@ copy_to_install_log() {
 }
 
 disableLighttpd() {
+    # Return early when lighttpd is not installed
+    if [[ ! -f /etc/lighttpd/lighttpd.conf ]]; then
```
yubiuser, line 2305: We are not touching any lighttpd config file when disabling/stopping it. I think we should instead use something like `check_service_active()` (which we already have) and/or 

```
nanopi@nanopi:~$ systemctl is-enabled lighttpd.service 
lighttpd.service is not a native service, redirecting to systemd-sysv-install.
Executing: /lib/systemd/systemd-sysv-install is-enabled lighttpd
enabled
nanopi@nanopi:~$ systemctl is-active lighttpd.service 
inactive

```

DL6ER: The routine calling this is only run once during 5 -> 6 migration, this code you are commenting on here will only be running once and my check was for if `lighttpd` is installed at all. I'm not really seeing how the config file being there may not be a good sign of it being installed. I don't think anyone would have installed and then removed it again leaving the config in place. I guess a Pi-hole repair/upgrade would have then reinstalled it.

Anyway, `check_service_active` is fine for me, too. I will push another commit

yubiuser (approved):

PR #5867, opened 2025-01-29
Gravity: Use ETags

DL6ER:

Use HTTP ETags in addition to the `If-Modified-Since` header (if supported by `curl`) to avoid downloading from servers not supporting using the currently used method method. The most relevant example is https://raw.githubcontent.com

Before this PR, `sudo pihole -g` **always** downloads https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts. After this PR, the cached file is used as the [Heisenberg compensator](https://memory-alpha.fandom.com/wiki/Heisenberg_compensator) tries to achieve.

---

---

## Checklist:

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -507,17 +507,12 @@ gravity_DownloadBlocklists() {
     echo -e "  ${INFO} Libz compression not available\n"
   fi
 
-  # Check if etag is supported by the locally available version of curl by
-  # comparing the version string being >= 7.68.0 (released Jan 2020)
-  # https://github.com/curl/curl/pull/4543 followed by
+  # Check if etag is supported by the locally available version of curl
+  # (available as of curl 7.68.0, released Jan 2020)
+  # https://github.com/curl/curl/pull/4543 +
   # https://github.com/curl/curl/pull/4678
-  if echo "${curlVersion}" | grep -q "curl 7\.[6-9][8-9]"; then
+  if curl --help all | grep -q "etag-save"; then
     etag_support=true
-  else
-    # Check if the version is >= 8
-    if echo "${curlVersion}" | grep -q "curl 8"; then
-      etag_support=true
-    fi
   fi
 
   # Loop through $sources and download each one
--- a/gravity.sh
+++ b/gravity.sh
@@ -414,7 +414,7 @@ gravity_DownloadBlocklists() {
     echo -e "  ${INFO} Storing gravity database in ${COL_BOLD}${gravityDBfile}${COL_NC}"
   fi
 
-  local url domain str target compression adlist_type directory curlVersion success
+  local url domain str target compression adlist_type directory success
   echo ""
 
   # Prepare new gravity database
@@ -498,8 +498,7 @@ gravity_DownloadBlocklists() {
   # Use compression to reduce the amount of data that is transferred
   # between the Pi-hole and the ad list provider. Use this feature
   # only if it is supported by the locally available version of curl
-  curlVersion=$(curl -V)
-  if echo "${curlVersion}" | grep -q "Features:.* libz"; then
+  if curl -V | grep -q "Features:.* libz"; then
     compression="--compressed"
     echo -e "  ${INFO} Using libz compression\n"
   else
--- a/gravity.sh
+++ b/gravity.sh
@@ -534,7 +534,7 @@ gravity_DownloadBlocklists() {
 
     # Save the file as list.#.domain
     saveLocation="${piholeDir}/list.${id}.${domain}.${domainsExtension}"
-    activeDomains[$i]="${saveLocation}"
+    activeDomains[i]="${saveLocation}"
 
     # Check if we can write to the save location file without actually creating
     # it (in case it doesn't exist)
@@ -1055,7 +1055,7 @@ timeit(){
   elapsed_time=$((end_time - start_time))
 
   # Display the elapsed time
-  printf "  %b--> took %d.%03d seconds%b\n" ${COL_BLUE} $((elapsed_time / 1000)) $((elapsed_time % 1000)) ${COL_NC}
+  printf "  %b--> took %d.%03d seconds%b\n" "${COL_BLUE}" $((elapsed_time / 1000)) $((elapsed_time % 1000)) "${COL_NC}"
 
   return $ret
 }
--- a/gravity.sh
+++ b/gravity.sh
@@ -414,7 +414,7 @@ gravity_DownloadBlocklists() {
     echo -e "  ${INFO} Storing gravity database in ${COL_BOLD}${gravityDBfile}${COL_NC}"
   fi
 
-  local url domain str target compression adlist_type directory
+  local url domain str target compression adlist_type directory curlVersion
   echo ""
 
   # Prepare new gravity database
@@ -496,7 +496,8 @@ gravity_DownloadBlocklists() {
   # Use compression to reduce the amount of data that is transferred
   # between the Pi-hole and the ad list provider. Use this feature
   # only if it is supported by the locally available version of curl
-  if curl -V | grep -q "Features:.* libz"; then
+  curlVersion=$(curl -V)
+  if echo "${curlVersion}" | grep -q "Features:.* libz"; then
     compression="--compressed"
     echo -e "  ${INFO} Using libz compression\n"
   else
@@ -508,11 +509,11 @@ gravity_DownloadBlocklists() {
   # comparing the version string being >= 7.68.0 (released Jan 2020)
   # https://github.com/curl/curl/pull/4543 followed by
   # https://github.com/curl/curl/pull/4678
-  if curl -V | grep -q "curl 7\.[6-9][8-9]"; then
+  if echo "${curlVersion}" | grep -q "curl 7\.[6-9][8-9]"; then
     etag_support=true
   else
     # Check if the version is >= 8
-    if curl -V | grep -q "curl 8"; then
+    if echo "${curlVersion}" | grep -q "curl 8"; then
       etag_support=true
     fi
   fi
--- a/gravity.sh
+++ b/gravity.sh
@@ -601,7 +601,7 @@ compareLists() {
 # Download specified URL and perform checks on HTTP status and file content
 gravity_DownloadBlocklistFromUrl() {
   local url="${1}" adlistID="${2}" saveLocation="${3}" target="${4}" compression="${5}" gravity_type="${6}" domain="${7}"
-  local heisenbergCompensator="" listCurlBuffer str httpCode success="" ip cmd_ext
+  local modifiedOptions="" listCurlBuffer str httpCode success="" ip cmd_ext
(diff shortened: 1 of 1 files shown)


DL6ER:
Rebased on latest `development` which wasn't up-to-date on my machine.

rdwebdesign:
Just a question:

Are you sure `--etag-save` option works in every supported OS?

I think it does, but I remember a previous attempt to implement this option failed (`unknown option`) because a few supported OS were using older `curl` versions. We need to check.

mwoolweaver:
https://daniel.haxx.se/blog/2019/12/06/curl-speaks-etag/

> Starting in version 7.68.0 (due to ship on January 8th, 2020)

gravity.sh
```diff
@@ -502,6 +503,20 @@ gravity_DownloadBlocklists() {
...
+  # https://github.com/curl/curl/pull/4543 followed by
+  # https://github.com/curl/curl/pull/4678
+  if curl -V | grep -q "curl 7\.[6-9][8-9]"; then
+    etag_support=true
+  else
+    # Check if the version is >= 8
+    if curl -V | grep -q "curl 8"; then
+      etag_support=true
+    fi
+  fi
```
dschaper, line 518: Could you save the output of `curl -V` in to a temp variable and then parse the temp variable instead of potentially calling `curl` twice in the case of `curl` version 8 or a `curl` that is less than 7.68? If most of the installs are not capable of etags then most of the installs will have two unnecessary `curl` calls. Doing a temp variable will limit that to just one `curl` call.

gravity.sh
```diff
@@ -595,12 +610,37 @@ gravity_DownloadBlocklistFromUrl() {
...
+  # For all remote files, we try to determine if the file has changed to skip
+  # downloading them whenever possible.
+  if [[ $url != "file"* ]]; then
+    # Use the HTTP ETag header to determine if the file has changed if supported
+    # by curl. Using ETags is supported by raw.githubusercontent.com URLs.
+    if [[ "${etag_support}" == true ]]; then
+      # Save HTTP ETag to the specified file. An ETag is a caching related header,
+      # usually returned in a response. If no ETag is sent by the server, an empty
+      # file is created and can later be used consistently.
+      heisenbergCompensator="--etag-save ${saveLocation}.etag"
```
dschaper, line 622: We really need to get rid of the 'fun' names of functions and call the functions what they are/what they do.

dschaper (changes requested):

darkexplosiveqwx:
The `--etag-save` flag was introduced in version [7.68.0](https://curl.se/ch/7.68.0.html)

I found Debian 11(Bullseye) has [7.74.0](https://packages.debian.org/search?keywords=curl), Fedora 40 has [8.6.0](https://packages.fedoraproject.org/pkgs/curl/curl/), Ubuntu 20.04(The Focal Fossa) has [7.68.0](https://launchpad.net/ubuntu/+source/curl)
and CentOS 9 has
[7.76.1](https://centos.pkgs.org/9-stream/centos-baseos-x86_64/curl-7.76.1-26.el9.x86_64.rpm.html).

DL6ER:
> We need to check.

We do

gravity.sh
DL6ER: Done

gravity.sh
```diff
@@ -497,13 +498,28 @@ gravity_DownloadBlocklists() {
...
   else
     compression=""
     echo -e "  ${INFO} Libz compression not available\n"
   fi
+
+  # Check if etag is supported by the locally available version of curl by
+  # comparing the version string being >= 7.68.0 (released Jan 2020)
+  # https://github.com/curl/curl/pull/4543 followed by
+  # https://github.com/curl/curl/pull/4678
+  if echo "${curlVersion}" | grep -q "curl 7\.[6-9][8-9]"; then
```
yubiuser, line 514: Maybe this is a nicer way to compare
https://github.com/pi-hole/PADD/blob/9b94e4b7925b18e0cd8c1abc376ed1849676d90d/padd.sh#L1327-L1331

yubiuser (commented):
Thanks for starting to rename the gravity functions to something users can understand. 
As we are starting to create additional `*.etag` files, this might be a good moment to move all the downloaded files for the adlists into a separate directory to clean `/etc/pihole` a bit

```
-rw-rw----  1 pihole pihole 891K 29. Jan 20:59 list.19.v.firebog.net.domains
-rw-rw----  1 pihole pihole   84 29. Jan 20:59 list.19.v.firebog.net.domains.sha1
-rw-rw----  1 pihole pihole  69K 29. Jan 20:59 list.20.v.firebog.net.domains
-rw-rw----  1 pihole pihole   84 29. Jan 20:59 list.20.v.firebog.net.domains.sha1
-rw-rw----  1 pihole pihole  49K 29. Jan 20:59 list.21.raw.githubusercontent.com.domains
-rw-rw----  1 pihole pihole   96 29. Jan 20:59 list.21.raw.githubusercontent.com.domains.sha1
-rw-rw----  1 pihole pihole 9,2K 29. Jan 20:59 list.22.raw.githubusercontent.com.domains
-rw-rw----  1 pihole pihole   96 29. Jan 20:59 list.22.raw.githubusercontent.com.domains.sha1
-rw-rw----  1 pihole pihole 403K 29. Jan 20:59 list.7.sysctl.org.domains
-rw-rw----  1 pihole pihole   80 29. Jan 20:59 list.7.sysctl.org.domains.sha1
-rw-rw----  1 pihole pihole 916K 29. Jan 20:59 list.9.small.oisd.nl.domains
-rw-rw----  1 pihole pihole   83 29. Jan 20:59 list.9.small.oisd.nl.domains.sha1
-rw-r--r--  1 root   root    358 28. Aug 20:50 logrotate

```

yubiuser:
What's about your idea to store the etag in the gravity database itself?
https://github.com/pi-hole/pi-hole/pull/4766#issuecomment-1193163372

gravity.sh
dschaper: Yeah, `awk` might be a better tool here instead of `grep`.

dschaper:
> What's about your idea to store the etag in the gravity database itself? [#4766 (comment)](https://github.com/pi-hole/pi-hole/pull/4766#issuecomment-1193163372)

If we only check to see if the `curl` supports `etag`s then storing this in the database might be a better idea. Once we know that `curl` supports `etag`s then it's not going to change and drop that support. A flag in the database that says "Only need to check if the current support is `false`" and once that status is `true` then no need to check again.

DL6ER:
> As we are starting to create additional *.etag files, this might be a good moment to move all the downloaded files for the adlists into a separate directory to clean /etc/pihole a bit

Yes, I will do this.

> What's about your idea to store the etag in the gravity database itself?

This was before `curl` had this in-built feature. Unfortunately, it can only work with files, so we'd need to write/read tmp files from/to the database if we want to do this. It would be cleaner conceptually but it really doesn't seem worth the extra maybe 20 lines of code, especially when we have the new directory `/etc/gravity/`.

> If we only check to see if the curl supports etags then storing this in the database might be a better idea. Once we know that curl supports etags then it's not going to change and drop that support.

Yes and No. Previously, we've lately talked a lot about the code complexity of `gravity.sh`. Checking `curl`'s version is really fast and done in a few lines (even less than right now when using the PADD variant). Storing it in the database, however, requires additional `sqlite3` calls and we cannot even be 100% that this is true (e.g. `gravity.db` copied from one device to another - users are very creative). I'd very much like to stay with the lightweight "live" `curl` feature testing.

DL6ER:
@yubiuser https://github.com/pi-hole/pi-hole/pull/5869

yubiuser (approved):

yubiuser:
Independently of this PR, we have a but in the `bash` > `FTL` interaction while parsing `gravity`.



Even when the list's content stayed unchanged, `FTL` will always update the `date_updated` field in the gravity database. T

DL6ER:
@yubiuser @dschaper The last commit adds what was discussed in #5873 but it felt easier to add it right here than to mess around with the history of the other PR.

gravity.sh
```diff
@@ -597,12 +608,37 @@ gravity_DownloadBlocklistFromUrl() {
...
+      modifiedOptions="--etag-save ${saveLocation}.etag"
+
+      if [[ -f "${saveLocation}.etag" ]]; then
+        # This option makes a conditional HTTP request for the specific ETag read
+        # from the given file by sending a custom If-None-Match header using the
+        # stored ETag. This way, the server will only send the file if it has
+        # changed since the last request.
+        modifiedOptions="${modifiedOptions} --etag-compare ${saveLocation}.etag"
+      fi
+    fi
```
yubiuser, line 629: Maybe we should use an `else` here. I was looking for how `etag` and `If-Modified-Since` relate to each other, but only found examples using one or the other:

https://curl.se/mail/archive-2022-03/0010.html
https://everything.curl.dev/http/modify/conditionals.html
https://daniel.haxx.se/blog/2019/12/06/curl-speaks-etag/

(and within the `-z` we probably also want `--remote-time`)

gravity.sh
```diff
@@ -497,13 +498,23 @@ gravity_DownloadBlocklists() {
   # Use compression to reduce the amount of data that is transferred
   # between the Pi-hole and the ad list provider. Use this feature
   # only if it is supported by the locally available version of curl
-  if curl -V | grep -q "Features:.* libz"; then
+  curlVersion=$(curl -V)
```
yubiuser, line 501: `curlVersion` is not needed anymore, as we use a different command now to check for `etag` support.

gravity.sh
DL6ER: I checked the code in `curl` and we should definitely use both at the same time. `-z` sets the time in the `If-Modified-Since` header and the others control the ETags. By setting both, we assure the best compatibility with any servers users may be downloading stuff from.

And I also don't think we need `--remote-time`. If we download today (Sunday), we can ask next time "only download if never than Sunday". It doesn't matter if the remote file has last been actually modified weeks ago. For some protocols (like FTL) this will require an extra step to get the timestamp, also I'd find is quite artificial to have `curl` create a file with a timestamp that is maybe years ago. But I also have no hard feeling against if you still think this is a good idea after reading my comment.

DL6ER: Actually, after having read on the links you provided, I dare to differ with the statement

> Specifying both is rarely useful - there are corner cases where it may help.

most exactly because of what they say two sentences earlier

> Etag only applies to http. At the server, Etags take precedence over last-modified.

We support more protocols than HTTP alone.

yubiuser: My concern was only that the time checking algorithm could fail due to different time zones when we use the local time stamp.

DL6ER: ETags should always be preferred by the server when it supports them. `-z` is really just a legacy thing. We've been using it for long and I don't recall it ever causing any trouble for us. We're overall improving with also bringing in ETags.

yubiuser (approved):

dschaper (approved):

PR #5869, opened 2025-01-30
Move gravity list cache into dedicated directory

DL6ER:

Migrate list's files (the lists themselves, their SHA1 sum and, after PR #5867, their ETags) into a dedicated directory `/etc/pihole/listsCache` in an attempt to clean up `/etc/pihole`

---

---

## Checklist:

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -30,6 +30,9 @@ PIHOLE_COMMAND="/usr/local/bin/${basename}"
 
 piholeDir="/etc/${basename}"
 
+# Gravity aux files directory
+listsCacheDir="${piholeDir}/listsCache"
+
 # Legacy (pre v5.0) list file locations
 whitelistFile="${piholeDir}/whitelist.txt"
 blacklistFile="${piholeDir}/blacklist.txt"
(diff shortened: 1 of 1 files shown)


yubiuser (commented):
I'm not sure if we really need a migration step here. If you think so, we might but it into the installer script when we already move all other v6 stuff. This makes it easier to remove the migration code when releasing v8 or so

DL6ER:
The migration step ensure we can continue using the cached lists. Without it, all lists would need to be re-downloaded which would cause trouble if users are using abandoned lists which are not online any longer. I'd like to keep it here because it is small and not really a v6 migration, this isn't a real breaking change and could have happened any time.

yubiuser:
Ok. Will wait with testing until https://github.com/pi-hole/pi-hole/pull/5867 has been merged.

DL6ER:
It is actually entirely independent. It uses `list.*` so any files, may their extension be `.sha1` or `.etag` or whatever may be added in the future will automatically be covered by this PR.

yubiuser:
Sure, but once I switch branches back and forth and the dir already exists, it won't move the files again.

DL6ER:
Yes, is this a problem? You can delete any of the `list.*` files at any time and gravity will recreate them (in whatever directory it sees fit at that time).

yubiuser (approved):

PR #5873, opened 2025-01-31
Use versionConverter instead of two grep calls

DL6ER:

See https://github.com/pi-hole/pi-hole/pull/5867

This PR is an amendment to PR implementing the alternative version comparison scheme borrowed from PADD. IMO it is making things a lot more complicated but I still propose it as alternative for discussion.

However, since I am not convinced this is the better method, I open this change as another PR into the former one instead of simply adding this commit into it.

---

---

## Checklist:

### DIFF
--- a/advanced/Scripts/utils.sh
+++ b/advanced/Scripts/utils.sh
@@ -93,3 +93,15 @@ setFTLConfigValue(){
     exit 5
   fi
 }
+
+#######################
+# converts a given version string e.g. v3.7.1 to 3007001000 to allow for easier
+# comparison of multi digit version numbers
+#
+# Credits https://apple.stackexchange.com/a/123408
+#
+# Example usage: versionConverter "v3.7.1"
+#######################
+versionConverter() {
+  echo "$@" | tr -d '[:alpha:]' | awk -F. '{ printf("%d%03d%03d%03d\n", $1,$2,$3,$4); }';
+}
--- a/gravity.sh
+++ b/gravity.sh
@@ -414,7 +414,7 @@ gravity_DownloadBlocklists() {
     echo -e "  ${INFO} Storing gravity database in ${COL_BOLD}${gravityDBfile}${COL_NC}"
   fi
 
-  local url domain str target compression adlist_type directory curlVersion success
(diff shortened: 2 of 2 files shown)


yubiuser:
Let's rephrase the issue we want to solve here: instead of checking version numbers, we're actually looking for supported option.

Why not something like 

```
curl --help all | grep etag
     --etag-compare <file> Pass an ETag from a file as a custom header
     --etag-save <file> Parse ETag from a request and save it to a file
```

dschaper:
What happens when you try to use `curl` with a feature that is unsupported? Is it an option to do something like

```
curl --etag-save <URL> || curl <URL>
```

That would keep from having to store the etag support status anywhere.

mwoolweaver:
> What happens when you try to use `curl` with a feature that is unsupported?
```
curl: option --tls-earlydata: is unknown
curl: try 'curl --help' or 'curl --manual' for more information
```

checking for emptiness would be easier than removing the `--help` output when the `--etag-save` option is missing, right?

```
if [[ -z "$(curl --help all | grep etag)" ]];  then
//no Etag support    
fi
```

DL6ER:
Yes, doing `curl --etag-save ...` and seeing if this fails may be doable as would be `--help all | grep`. I think I'd actually prefer the latter as the former may fail because of other things (like the URL we specified cannot be resolved, etc.). And we have learned from history to never grep for natural language as `curl` and friends may be speaking other languages on other systems.

However, do we know for sure that `--help all` is always available? Or can it be that the `help` feature is omittable, e.g., for using on systems with limited storage such as routers or tiny docker containers where this feature is unlikely to ever be used?

Bottom line is: I still think we can just use what I proposed with checking the version with two `grep` calls in the original PR.

dschaper:
```powershell
C:\Users\dan> docker container run -it --rm alpine sh
/ # apk -q add curl
/ # curl --help all | grep etag
     --etag-compare <file>                         Load ETag from file
     --etag-save <file>                            Parse incoming ETag and save to a file
/ #
```

DL6ER:
Obsolete as of https://github.com/pi-hole/pi-hole/commit/37718c65e9efd348824948aec3dc29483b49ec55

Issue #5874, opened 2025-02-01
[v6] `pihole -g` domain (regex_block, regex_allow, exact_block, exact_allow) count does not match what is shown in the WebUI @ https://pi.hole/admin/groups/domains


mwoolweaver:
### Versions

- Core:  
       - Version is v5.18.4-641-ge682f69a (Latest: null)
       - Branch is development
       - Hash is e682f69a (Latest: e682f69a)

- Web:  
       - Version is v5.21-1082-gfb368323 (Latest: null)
       - Branch is development
       - Hash is fb368323 (Latest: fb368323)

- FTL:  
       - Version is vDev-80ef20e (Latest: null)
       - Branch is development
       - Hash is 80ef20ed (Latest: 80ef20ed)

### Platform

- OS and version: Ubuntu  24.04.1

- Platform:  Raspberry Pi 

### Expected behavior

`pihole -g` domain count should match [WebUI](http://pi.hole/admin/groups/domains) domain count

### Actual behavior / bug

I am unsure of the exact math but it seems `pihole -g` is not counting these domains correctly due to the number of groups each individual domain is in. 

### Steps to reproduce

1. Go to https://pi.hole/admin/groups/domains
2. Note the number of each type of domain listed there (regex_block, regex_allow, exact_block, exact_allow) shown here and how many groups each domain is in
3. Go to  https://pi.hole/admin/gravity
4. Update Gravity
5. Note the number of each type of domain listed there (regex_block, regex_allow, exact_block, exact_allow)
6. Notice that these pages do not show the same information

Output after `sudo pihole -g`

```
  [✓] Building tree
  [i] Number of gravity domains: 302693 (302681 unique domains)
  [i] Number of exact denied domains: 3
  [i] Number of regex denied filters: 48
  [i] Number of exact allowed domains: 23
  [i] Number of regex allowed filters: 2
  [✓] Optimizing database
  [✓] Swapping databases
  [✓] The old database remains available
  [✓] Cleaning up stray matter

```

note the number of regex blocked shown in the screenshot below

screenshot from WebUI 
![Image](https://github.com/user-attachments/assets/b820e62c-7ae7-40d5-96f9-8b248f72b4e7)

## Debug Token

- URL:  https://tricorder.pi-hole.net/P9s3zlrH/

mwoolweaver:
it seems `pihole -g` is just reading the view provided by the database.

[...]

PR #5881, opened 2025-02-01
Fix counting of domains in the gravity summary

DL6ER:

Fix counting of domains at the end of `pihole -g`. The existing scheme used the SQL `VIEW`s for simplicity, however, they have the unintended side-effect of - when being queried without a `GROUP BY` clause - returning entries multiple times if they are assigned to several groups

**Related issue or feature (if applicable):** Fixes https://github.com/pi-hole/pi-hole/issues/5874.

---

## Checklist:

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -864,11 +864,11 @@ gravity_Table_Count() {
 gravity_ShowCount() {
   # Here we use the table "gravity" instead of the view "vw_gravity" for speed.
   # It's safe to replace it here, because right after a gravity run both will show the exactly same number of domains.
-  gravity_Table_Count "gravity" "gravity domains" ""
-  gravity_Table_Count "vw_blacklist" "exact denied domains"
-  gravity_Table_Count "vw_regex_blacklist" "regex denied filters"
-  gravity_Table_Count "vw_whitelist" "exact allowed domains"
-  gravity_Table_Count "vw_regex_whitelist" "regex allowed filters"
+  gravity_Table_Count "gravity" "gravity domains"
+  gravity_Table_Count "domainlist WHERE type = 1" "exact denied domains"
+  gravity_Table_Count "domainlist WHERE type = 3" "regex denied filters"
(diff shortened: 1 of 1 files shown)


mwoolweaver:
also worth noting It appears this issue might exist in current v5 release as well 

https://github.com/pi-hole/pi-hole/blob/2cf046d559f2430253fde4e3bf1a89ef4d8f1587/gravity.sh#L705

rdwebdesign (approved):
As I said, I suspected using "WHERE type = ..." would be a lot faster than using "DISTINCT".

rdwebdesign:
> this issue might exist in current v5 release as well

Yes it does (tested and confirmed), but apparently no one noticed until now (!!!), so we probably won't waste time fixing this small issue in v5.

yubiuser:
One aspect should be discussed: using the existing `view` had the benefit that it only counted `enabled` domains`. But the web UI and this PR count all 'configured' domains. I'm fine with this change, but this is a design decision which we should be aware of....

DL6ER:
Yes, I have been thinking about this, too, but then it says
```
  [i] Number of exact denied domains: 2
  [i] Number of regex denied filters: 4
  [i] Number of exact allowed domains: 1
  [i] Number of regex allowed filters: 1
```
and not "Number of enabled ..." which is why I left out an `enabled = 1` filter clause.

yubiuser:
I' mot 100% sure, but I think we could even drop all the views. I checked the `core` repo and they are only used for displaying the count. I don't think the web or FTL are using the views

DL6ER:
FTL is indeed using them
grafik

yubiuser:
And this now creates a mismatch between the gravity output and the FTL api (`info/api`) and the web UI if a domain is configured but not enabled.

Example: one disabled deny domain

mwoolweaver:
> Yes it does (tested and confirmed), but apparently no one noticed until now (!!!), so we probably won't waste time fixing this small issue in v5.

Cosmetic issues can easily overlooked, especially when the are in a place most people aren't looking

PR #5885, opened 2025-02-03
installer: use a drop-in to disable systemd-resolved stub listener

enr0n:
systemd-resolved supports drop-in config snippets, e.g. by placing .conf files in /etc/systemd/resolved.conf.d. During install, use a drop-in config to set DNSStubListener=no, instead of modifying the main config.

This is generally better practice, and also prevents conflicts when distribution packages are upgraded, which may include new versions of /etc/systemd/resolved.conf.

---

Improve the way systemd-resolved's stub resolver is disabled during install.

Use `/etc/systemd/resolved.conf.d/pi-hole-disable-stub-listener.conf` instead of modifying `/etc/systemd/resolved.conf` directly.

n/a

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1436,20 +1436,15 @@ disable_resolved_stublistener() {
     printf "  %b Testing if systemd-resolved is enabled\\n" "${INFO}"
     # Check if Systemd-resolved's DNSStubListener is enabled and active on port 53
     if check_service_active "systemd-resolved"; then
-        # Check if DNSStubListener is enabled
-        printf "  %b %b Testing if systemd-resolved DNSStub-Listener is active" "${OVER}" "${INFO}"
-        if (grep -E '#?DNSStubListener=yes' /etc/systemd/resolved.conf &>/dev/null); then
-            # Disable the DNSStubListener to unbind it from port 53
-            # Note that this breaks dns functionality on host until ftl are up and running
-            printf "%b  %b Disabling systemd-resolved DNSStubListener" "${OVER}" "${TICK}"
-            # Make a backup of the original /etc/systemd/resolved.conf
-            # (This will need to be restored on uninstallation)
-            sed -r -i.orig 's/#?DNSStubListener=yes/DNSStubListener=no/g' /etc/systemd/resolved.conf
-            printf " and restarting systemd-resolved\\n"
-            systemctl reload-or-restart systemd-resolved
-        else
-            printf "%b  %b Systemd-resolved does not need to be restarted\\n" "${OVER}" "${INFO}"
-        fi
+        # Disable the DNSStubListener to unbind it from port 53
+        # Note that this breaks dns functionality on host until FTL is up and running
+        printf "%b  %b Disabling systemd-resolved DNSStubListener\\n" "${OVER}" "${TICK}"
+        mkdir -p /etc/systemd/resolved.conf.d
+        cat > /etc/systemd/resolved.conf.d/90-pi-hole-disable-stub-listener.conf << EOF
+[Resolve]
+DNSStubListener=no
+EOF
+        systemctl reload-or-restart systemd-resolved
     else
         printf "%b  %b Systemd-resolved is not enabled\\n" "${OVER}" "${INFO}"
     fi
--- a/automated install/uninstall.sh
+++ b/automated install/uninstall.sh
@@ -94,8 +94,9 @@ removePiholeFiles() {
     echo -e "  ${TICK} Removed config files"
 
     # Restore Resolved
-    if [[ -e /etc/systemd/resolved.conf.orig ]]; then
-        ${SUDO} cp -p /etc/systemd/resolved.conf.orig /etc/systemd/resolved.conf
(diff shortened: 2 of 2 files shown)


yubiuser (changes requested):
Thanks for the PR. Please note, that all pull requests should be based on and target `development` branch

enr0n:
> Thanks for the PR. Please note, that all pull requests should be based on and target `development` branch

Rebased.

automated install/uninstall.sh
```diff
@@ -98,6 +98,10 @@ removePiholeFiles() {
         ${SUDO} cp -p /etc/systemd/resolved.conf.orig /etc/systemd/resolved.conf
         systemctl reload-or-restart systemd-resolved
     fi
+    if [[ -e /etc/systemd/resolved.conf.d/pi-hole-disable-stub-listener.conf ]]; then
+        ${SUDO} rm -f /etc/systemd/resolved.conf.orig /etc/systemd/resolved.conf
```
yubiuser, line 102: I think this is a copy&past error

enr0n: Whoops. Fixed.

automated install/basic-install.sh
```diff
@@ -1436,20 +1436,15 @@ disable_resolved_stublistener() {
...
-            # Make a backup of the original /etc/systemd/resolved.conf
-            # (This will need to be restored on uninstallation)
-            sed -r -i.orig 's/#?DNSStubListener=yes/DNSStubListener=no/g' /etc/systemd/resolved.conf
-            printf " and restarting systemd-resolved\\n"
-            systemctl reload-or-restart systemd-resolved
-        else
-            printf "%b  %b Systemd-resolved does not need to be restarted\\n" "${OVER}" "${INFO}"
-        fi
+        # Disable the DNSStubListener to unbind it from port 53
+        # Note that this breaks dns functionality on host until dnsmasq/ftl are up and running
```
PromoFaux, line 1440:
```suggestion
        # Note that this breaks dns functionality on host until FTL is up and running
```

Just a pedantic change request. Appreciate the grammar wasn't quite right before this PR

enr0n: Done :+1:

yubiuser (commented):
Is there a reason why you disabled the check

`if (grep -E '#?DNSStubListener=yes' /etc/systemd/resolved.conf &>/dev/null); then`
?

___

I think we should keep the check and extend it also check for the presence of `/etc/systemd/resolved.conf.d/pi-hole-disable-stub-listener.conf`

enr0n:
> Is there a reason why you disabled the check
> 
> `if (grep -E '#?DNSStubListener=yes' /etc/systemd/resolved.conf &>/dev/null); then` ?

Because it's unnecessary. If the stub resolver is enabled (the default) then we disable it. If it's already disabled, the drop-in has no effect. But, does that really matter? The drop-in is cheap and it documents in a way that pi-hole wants the stub-listener disabled.

automated install/basic-install.sh
```diff
@@ -1436,20 +1436,15 @@ disable_resolved_stublistener() {
...
+        printf "%b  %b Disabling systemd-resolved DNSStubListener" "${OVER}" "${TICK}"
```
yubiuser, line 1441:
```suggestion
        printf "%b  %b Disabling systemd-resolved DNSStubListener\\n" "${OVER}" "${TICK}"
```

yubiuser (commented):
You're probably right.

enr0n:
> You're probably right.

I added a `90-` prefix as well to make sure the drop-in has high priority among potential others.

automated install/uninstall.sh
```diff
@@ -98,6 +98,10 @@ removePiholeFiles() {
         ${SUDO} cp -p /etc/systemd/resolved.conf.orig /etc/systemd/resolved.conf
         systemctl reload-or-restart systemd-resolved
     fi
+    if [[ -e /etc/systemd/resolved.conf.d/90-pi-hole-disable-stub-listener.conf ]]; then
+        ${SUDO} rm -f /etc/systemd/resolved.conf.d/90-pi-hole-disable-stub-listener.conf
+        systemctl reload-or-restart systemd-resolved
```
yubiuser, line 103: Can you combine this and the check above to prevent `systemd-resolved` to restart twice in a row?

rdwebdesign: Or use a variable to decide if the service needs to be restarted:

```sh
    # Restore Resolved
    local restart_service="false"
    if [[ -e /etc/systemd/resolved.conf.orig ]]; then
        ${SUDO} cp -p /etc/systemd/resolved.conf.orig /etc/systemd/resolved.conf
        restart_service="true"
    fi
    if [[ -e /etc/systemd/resolved.conf.d/90-pi-hole-disable-stub-listener.conf ]]; then
        ${SUDO} rm -f /etc/systemd/resolved.conf.d/90-pi-hole-disable-stub-listener.conf
        restart_service="true"
    fi

    if [[ "${restart_service}" == "true" ]]; then
        systemctl reload-or-restart systemd-resolved
    fi
```

yubiuser (approved):

rdwebdesign (approved):

PR #5887, opened 2025-02-06
Remove 'reconfigure' option

yubiuser:

Removes the `reconfigure` option. 
This option was useful in the past, when the listening interface needed to be set or users wanted to add the web interface which they did not choose during the initial installation. Also settings were scattered around in different config files (`setupVars.conf`, `pihole-FTL.conf`)

With v6 the option is mostly useless: the web interface is installed by default, we only offer one default blocklist to choose from and all settings are nicely combined in one place (`pihole.toml`) where it's easy to change them (edit toml file, use `pihole-FTL --config`)

This PR removes this option, but retains the option to **repair** the installation (which is basically a repo reset).

The only place the `reconfigure` option was used internally was during an update. I changed the code in a way that the update should work even without this flag.

---

---


yubiuser:
The changes to `clone_or_update_repos()` look huge, but aren't. I only swapped the code blocks for `if` and `else` - that way an update will trigger the former `reconfigure` (now the `else` block`) and a `repair` need to be requested now explicitly.

yubiuser:
The main change is the removal of the `reconfigure` dialog choice in the former `update_dialogs()` which was only reachable when an existing installation was detected but the installer was **not** called with the `unattended` flag. 
The option set `useUpdateVars=false` which would trigger all the dialog sequences in the installer afterwards.

yubiuser:
Ready for testing and review

PromoFaux:
Not currently able to test this - but the code looks right.

automated install/basic-install.sh
```diff
@@ -1787,21 +1787,12 @@ displayFinalMessage() {
...
-        strAdd="You will remain on the same version"
-    else
-        # Otherwise, set some variables with different values
-        opt1a="Update"
-        opt1b="This will retain existing settings."
-        strAdd="You will be updated to the latest version."
-    fi
-    opt2a="Reconfigure"
-    opt2b="Resets Pi-hole and allows re-selecting settings."
+repair_dialog() {
```
PromoFaux, line 1790: Correct me if I'm wrong, but does this now mean that `pihole -r` will still display a dialog box, but the dialog box will only have one "option" to select?

If so, does that maybe render the dialog box superfluous?

yubiuser: I just left it to give users the chance to cancel the process. If you think it's not necessary, I can also just remove it.

dschaper: I don't think repairing is a destructive process that needs to have a cancel option. My vote would be to remove the dialog.

yubiuser: Removed it.

PromoFaux (approved):

dschaper (approved):

Issue #5980, opened 2025-02-23
Updates do not honour webserver.paths set in pihole.toml


nexusgoblin:
### Versions
- Pi-hole:  v6.0.4
- AdminLTE:  v6.0.1
- FTL:  v6.0.2

### Platform
- OS and version:  Debian 12.9 aarch64
- Platform:  Raspberry Pi 4

### Expected behavior

TL;DR

Make `update.sh`, `updatecheck.sh` and `uninstall.sh` do not honour changes to `webserver.paths.webroot` and `webserver.paths.webhome` in `pihole.toml`.

Details

Pi-hole web interface, Settings, All settings, Webserver and API allows setting custom values for `webserver.paths.webroot` and `webserver.paths.webhome`. If the `admin` directory is moved to the new location, the web UI works. However, running `sudo pihole -up` will complain that the web UI is not installed.

Standard (basic) install followed by configuration changes for custom configuration. In this instance, the host is a utility box that runs Nginx and multiple virtual hosts. Having the Pi-hole `admin` directory in `/var/www/html` isn't great when using virtual hosts; relocating it to somewhere like `/var/www/pihole.domain/html` and changing the `webserver.paths` settings works nicely... until you try to update.

### Actual behavior / bug

```BASH
$ sudo pihole -up
  [✓] Supported OS detected
  [✓] Update local cache of available packages
  [✓] Building dependency package pihole-meta.deb
  [✓] Installing Pi-hole dependency package

  [i] Checking for updates...
  [i] Pi-hole Core:	up to date

  Error: Web Admin repo is missing from system!
  Please re-run install script from https://pi-hole.net
```

### Steps to reproduce

1. Install Pi-hole
2. Copy web UI `admin` directory to new location
3. Use the web UI to reconfigure `webserver.paths.webroot` to a location other than `/var/www/html` (wherever `admin` has been copied to)
4. `sudo systemctl stop pihole-FTL`
5. Remove original `admin` directory (`sudo rm -rf /var/www/html/admin`)
6. `sudo systemctl start pihole-FTL`
7. Verify that UI is still working
8. `sudo pihole -up` see error as shown above.

## Debug Token- URL:  

Can be provided if necessary

### Additional context

This might be a bit of an edge case, however, the setting are there, they're just not being used/honoured in scripts.

PR #5981, opened 2025-02-23
Update update.sh, updatecheck.sh and uninstall.sh to honour pihole.to…

nexusgoblin:
…ml settings for webserver.paths.webroot and webserver.paths.webhome

Attempt to address #5980

Update `advanced/Scripts/update.sh` to source `utils.sh` and build the `ADMIN_INTERFACE_DIR` variable from configuration in `pihole.toml` via the `getFTLConfigValue()` function. Also change a hard coded reference of `/var/www/html/admin` in output string to use the `ADMIN_INTERFACE_DIR` variable.

Update `advanced/Scripts/updatecheck.sh` to build an `ADMIN_INTERFACE_DIR` variable from configuration in `pihole.toml` via the `getFTLConfigValue()` function and change hard coded references to `/var/www/html/admin` to use the added `ADMIN_INTERFACE_DIR` variable.

Update `automated install/uninstall.sh` to source `utils.sh` and to build an `ADMIN_INTERFACE_DIR` variable from configuration in `pihole.toml` via the `getFTLConfigValue()` function. Add a check that if web UI path is not the default of `/var/www/html/admin` do not automatically delete the custom path, but tell user to delete manually. This is a bit of a work around to deal with potentially "unsafe" paths getting pulled from `pihole.toml` (and its a custom config, so possibly a fair ask).

---

---

### DIFF
--- a/automated install/uninstall.sh
+++ b/automated install/uninstall.sh
@@ -9,6 +9,9 @@
 # Please see LICENSE file for your rights under this license.
 
 source "/opt/pihole/COL_TABLE"
+source "/opt/pihole/utils.sh"
+
+readonly ADMIN_INTERFACE_DIR=$(getFTLConfigValue "webserver.paths.webroot")$(getFTLConfigValue "webserver.paths.webhome")
 
 while true; do
     read -rp "  ${QST} Are you sure you would like to remove ${COL_WHITE}Pi-hole${COL_NC}? [y/N] " answer
@@ -53,17 +56,9 @@ removeMetaPackage() {
 }
 
 removePiholeFiles() {
-    # Only web directories/files that are created by Pi-hole should be removed
+    # Remove the web interface of Pi-hole
     echo -ne "  ${INFO} Removing Web Interface..."
-    ${SUDO} rm -rf /var/www/html/admin &> /dev/null
-
-
-    # If the web directory is empty after removing these files, then the parent html directory can be removed.
-    if [ -d "/var/www/html" ]; then
-        if [[ ! "$(ls -A /var/www/html)" ]]; then
-            ${SUDO} rm -rf /var/www/html &> /dev/null
-        fi
-    fi
+    ${SUDO} rm -rf "${ADMIN_INTERFACE_DIR}" &> /dev/null
     echo -e "${OVER}  ${TICK} Removed Web Interface"
 
     # Attempt to preserve backwards compatibility with older versions
--- a/advanced/Scripts/update.sh
+++ b/advanced/Scripts/update.sh
@@ -12,7 +12,6 @@
 
 # Variables
 readonly ADMIN_INTERFACE_GIT_URL="https://github.com/pi-hole/web.git"
-readonly ADMIN_INTERFACE_DIR="/var/www/html/admin"
 readonly PI_HOLE_GIT_URL="https://github.com/pi-hole/pi-hole.git"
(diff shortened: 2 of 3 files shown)


automated install/uninstall.sh
```diff
@@ -55,16 +58,23 @@ removeMetaPackage() {
...
+        # If the web directory is empty after removing these files, then the parent html directory can be removed.
+        if [ -d "/var/www/html" ]; then
+            if [[ ! "$(ls -A /var/www/html)" ]]; then
+                ${SUDO} rm -rf /var/www/html &> /dev/null
+            fi
         fi
+        echo -e "${OVER}  ${TICK} Removed Web Interface"
+    else
+        # The web directory has been customised, tell the user to clean up manually.
+        echo -e "${OVER}  ${CROSS} Please remove Web Interface at '${ADMIN_INTERFACE_DIR}' manually"
```
PromoFaux, line 76: Why not `rm -rf ${ADMIN_INTERFACE_DIR}` here?

nexusgoblin: Good point. My thinking was: if someone is customising the install, then:

- They may have changed other things within the directory structure, so just deleting might be unexpected;
- They've customised, so they shouldn't be adverse to manual clean up.

Though I can see the argument that this is _**Pi-Hole's**_ directory and should be directly maintainable by Pi-Hole. In which case, `rm -rf ${ADMIN_INTERFACE_DIR}` would be, and should be, fine.

I'll make the change.

nexusgoblin: I also missed quoting some variable usages. Will fixed that too.

automated install/uninstall.sh
```diff
@@ -55,14 +58,21 @@ removeMetaPackage() {
...
+    if [[ "${ADMIN_INTERFACE_DIR}" == "/var/www/html/admin/" ]]; then
+        ${SUDO} rm -rf /var/www/html/admin &> /dev/null
 
 
-    # If the web directory is empty after removing these files, then the parent html directory can be removed.
-    if [ -d "/var/www/html" ]; then
-        if [[ ! "$(ls -A /var/www/html)" ]]; then
-            ${SUDO} rm -rf /var/www/html &> /dev/null
+        # If the web directory is empty after removing these files, then the parent html directory can be removed.
+        if [ -d "/var/www/html" ]; then
```
PromoFaux, line 68: I know this isn't added by your PR - but I'm tempted to say that we needn't be running this check in any case....

Leaving an empty directory here for the user is not really going to cause any harm - and it's one less complexity check in the uninstall script if we only delete "wherever the admin page is"...

paging @pi-hole/core-maintainers  for thoughts...

dschaper: If we can't know that we created the parent directory then we shouldn't delete the parent directory.

automated install/uninstall.sh
yubiuser: I agree with PromoFaux here. I think even if the location was changed, we should remove it. Only changing the location does not imply users also changed the content.

automated install/uninstall.sh
yubiuser: I agree. Leave `/var/www/html` untouched in any case, we should just remove the whole check here.

nexusgoblin: Removed the web interface parent directory empty check and possible removal. Simplified to just removing the `"${ADMIN_INTERFACE}"` location.

yubiuser (approved):

Issue #5987, opened 2025-02-24
After upgrade, incorrect permissions on file /etc/pihole/versions


robcmo:
### Versions
/usr/local/bin/pihole: line 34: /etc/pihole/versions: Permission denied
/opt/pihole/version.sh: 20: .: cannot open /etc/pihole/versions: Permission denied

Core version is v6.0.4 (Latest: v6.0.4)
Web version is v6.0.1 (Latest: v6.0.1)
FTL version is v6.0.2 (Latest: v6.0.2)

### Platform
- OS and version:  Raspbian GNU/Linux 11 (bullseye)
- Platform:  Raspberry Pi

### Expected behavior

Version checks should provide version numbers for console user without sudo.

### Actual behavior / bug

The file /etc/pihole/versions owned by pihole:pihole mode 660 so 'other' users cannot read version file.

### Steps to reproduce

1. $ pihole -v
2. $ sudo ls -l /etc/pihole/versions
-rw-rw---- 1 pihole pihole 323 Feb 23 14:46 /etc/pihole/versions

### Resolution

$ sudo chmod 664 /etc/pihole/versions
$ pihole -v
Core version is v6.0.4 (Latest: v6.0.4)
Web version is v6.0.1 (Latest: v6.0.1)
FTL version is v6.0.2 (Latest: v6.0.2)

yubiuser:
> The file /etc/pihole/versions owned by pihole:pihole mode 660 so 'other' users cannot read version file.

This should not be like this. Correct permissions are
```

nanopi@nanopi:~$ ls -lha /etc/pihole/versions 
-rw-r--r-- 1 pihole pihole 329 24. Feb 14:01 /etc/pihole/versions
```

On every update check, we reset the permissions.
https://github.com/pi-hole/pi-hole/blob/567bb724b18bf174f033d0b8b289cc44a5e9c8eb/advanced/Scripts/updatecheck.sh#L50-L53

yubiuser:
The issue is here: https://github.com/pi-hole/pi-hole/blob/567bb724b18bf174f033d0b8b289cc44a5e9c8eb/advanced/Templates/pihole-FTL-prestart.sh#L17

robcmo:
Great work @yubiuser, that makes sense. Thanks for putting in the PR!

yubiuser:
In general (and as a workaround), I suggest you put your local user in the `pihole` group. Then you would also be able to use things like `pihole -q` from the CLI without need to authentication

[...]

PR #6002, opened 2025-02-24
versions file should readable by others (make pihole -v work for non-root users)

yubiuser:

We allow all users to run `pihole -v`, so they should have read access to `/etc/pihole/version`.
We do set those permissions in 

https://github.com/pi-hole/pi-hole/blob/567bb724b18bf174f033d0b8b289cc44a5e9c8eb/advanced/Scripts/updatecheck.sh#L50-L53

However, if we (re)start FTL after the updatescript run, it will reset the permissions to `0660` until the next updatecheck due to this line

https://github.com/pi-hole/pi-hole/blob/567bb724b18bf174f033d0b8b289cc44a5e9c8eb/advanced/Templates/pihole-FTL-prestart.sh#L17

Add an exception for `/etc/pihole/version` by setting its permissions to `0644` also at FTL start.

Fixes https://github.com/pi-hole/pi-hole/issues/5987

---

---

### DIFF
--- a/advanced/Scripts/updatecheck.sh
+++ b/advanced/Scripts/updatecheck.sh
@@ -52,7 +52,7 @@ rm -f "/etc/pihole/localversions"
 # Create new versions file if it does not exist
 VERSION_FILE="/etc/pihole/versions"
 touch "${VERSION_FILE}"
-chmod 644 "${VERSION_FILE}"
+chmod 640 "${VERSION_FILE}"
 
 # if /pihole.docker.tag file exists, we will use it's value later in this script
 DOCKER_TAG=$(cat /pihole.docker.tag 2>/dev/null)
--- a/advanced/Scripts/updatecheck.sh
+++ b/advanced/Scripts/updatecheck.sh
@@ -52,7 +52,7 @@ rm -f "/etc/pihole/localversions"
 # Create new versions file if it does not exist
 VERSION_FILE="/etc/pihole/versions"
 touch "${VERSION_FILE}"
-chmod 640 "${VERSION_FILE}"
+chmod 644 "${VERSION_FILE}"
 
 # if /pihole.docker.tag file exists, we will use it's value later in this script
 DOCKER_TAG=$(cat /pihole.docker.tag 2>/dev/null)
--- a/advanced/Templates/pihole-FTL-prestart.sh
+++ b/advanced/Templates/pihole-FTL-prestart.sh
@@ -12,8 +12,9 @@ FTL_PID_FILE="$(getFTLConfigValue files.pid)"
 # Ensure that permissions are set so that pihole-FTL can edit all necessary files
 mkdir -p /var/log/pihole
 chown -R pihole:pihole /etc/pihole/ /var/log/pihole/
-# allow all users read version file
-chmod 0664 /etc/pihole/versions
+
+# allow all users read version file (and use pihole -v)
+chmod 0644 /etc/pihole/versions
 
 # allow pihole to access subdirs in /etc/pihole (sets execution bit on dirs)
 find /etc/pihole/ /var/log/pihole/ -type d -exec chmod 0755 {} +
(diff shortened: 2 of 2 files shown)


yubiuser:
Alternative would be to change the updatescript to also set `660` and tell users to put their local user into `pihole` group.

(Which they might need to do anyway to access `cli_pw`)

yubiuser:
I changed the updatecheck script accordingly

PromoFaux:
@yubiuser your commit message says `660`, but you've written `640` in the code 🫣

yubiuser:
Upps... but looking at https://github.com/pi-hole/pi-hole/pull/6002 **640** will be likely be the new default for almost everything in `/etc/pihole`.

So it's the commit message that is wrong here ;-)

DL6ER:
What happens if users without access to this file run `pihole -v` after this PR?

yubiuser:
> /usr/local/bin/pihole: line 34: /etc/pihole/versions: Permission denied
/opt/pihole/version.sh: 20: .: cannot open /etc/pihole/versions: Permission denied

See fist post of https://github.com/pi-hole/pi-hole/issues/5987

DL6ER:
This doesn't seem desirable...

rdwebdesign:
>  I know apt list has some flags that run without authentication but I think they only show what version would be installed or is available?

I think `apt list -i` shows every installed package, with versions.

dschaper:
> This doesn't seem desirable...

Can you clarify what you mean? I'm not sure if you mean that it's not desirable to prevent users from seeing the versions or if the error message format/output is not desirable.

orazioedoardo:
To me this is security theater, an attacker with local access to the system can easily determine the version of Pi-hole by looking at how it behaves, taking the hash of pihole-FTL, the scripts or the cloned git repo, etc.

mwoolweaver:
https://github.com/pi-hole/pi-hole/issues/5987#issuecomment-2680522852
> In general (and as a workaround), I suggest you put your local user in the pihole group. Then you would also be able to use things like pihole -q from the CLI without need to authentication

would it be too much to ask the user for the name of the `local user` and then add this user to the pihole group during install?

yubiuser:
> would it be too much to ask the user for the name of the local user and then add this user to the pihole group during install?

Asking in the sense of letting them type it in might be to error prone. But we could offer to add the user with UID 1000 (usually the first real user) to the group.

mwoolweaver:
> Asking in the sense of letting them type it in might be to error prone.

Agreed.



> But we could offer to add the user with UID 1000 (usually the first real user) to the group.

I assume in most cases this would be the user most people are using.

PromoFaux:
Thinking about this, 644 is fine for the versions file,  I don't think there is any need to make the versions of Pi-hole unreadable.

`apt list -i` will rattle of a list of all installed packages and their versions, and I can also get the version number of an application using it's `--version` switch:

```
pi@raspberrypi:~ $ nano --version
 GNU nano, version 5.4
 (C) 1999-2011, 2013-2020 Free Software Foundation, Inc.
 (C) 2014-2020 the contributors to nano
 Compiled options: --disable-libmagic --enable-utf8
pi@raspberrypi:~ $ apt --version
apt 2.2.4 (arm64)
pi@raspberrypi:~ $ service --version
service ver. 1.60
pi@raspberrypi:~ $ sysctl --version
sysctl from procps-ng 3.3.17
pi@raspberrypi:~ $ rm --version
rm (GNU coreutils) 8.32
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Paul Rubin, David MacKenzie, Richard M. Stallman,
and Jim Meyering.
```

mwoolweaver:
even sudo gives version with out elevated privileges 

```sh
vpn@pi-hole:~$ sudo --version
Sudo version 1.9.15p5
Sudoers policy plugin version 1.9.15p5
Sudoers file grammar version 50
Sudoers I/O plugin version 1.9.15p5
Sudoers audit plugin version 1.9.15p5
```

dschaper (approved):

PR #6008, opened 2025-02-25
Show an warning dialoge before performing an update

yubiuser:

Shows an warning and confirmation dialog before users can proceed with available updates. Defaults to 'Exit'

---

---


yubiuser:
Probably not, but the current update broke some users system and they asked for an opportunity to stop the update process. I partly agree, because once they run `pihole -up` there is nothing to stop the update and it won't tell them if this is a hotfix release or a major upgrade

yubiuser:
> This also means that we establish a contract that we make a blog post for every update. For minor things i've not previously bothered, but that's probably lazyness on my part

I'm not seeing us writing blog posts for every patch/minor update

yubiuser:
It could look like this

dschaper:
I don't know, feels a little too "Clippy" to me.

> Looks like you want to update Pi-hole, do you need some help??

PromoFaux:
> I'm not seeing us writing blog posts for every patch/minor update

Heh, yeah this was kind of my point

yubiuser:
Putting this on hold after internal discussion.

DL6ER:
> IMPORTANT:\\Zn Make a (teleporter) backup of your system!

Note that we have `pihole-FTL --teleporter` which can automatically create Teleporter archives for them.

DL6ER:
The more I think about this, the more I believe we should add this. And only if to break the cron-job based auto-updaters.

One particular example is that we've correctly suggested to users (and users account each other, too) to add `no-0x20-encode` to their configs. Simon Kelley is looking to make this a default-off feature which will *remove* this option. The problem going forward from here is that all users who have added this setting will not be able to start `pihole-FTL`. I want to have a tool where I can say them:

> Wait, look, think what you are doing.

If you have `no-0x20-encode`, remove it **before proceeding with this update***. The only other options are (a) believing they will read the change log / blog post (good joke, I know) or (b) and handholding code that searches all the possible places where they may have added this, and remove it safely from there for them. If it is in a `root`-owned file in `/etc/dnsmasq.d` then `pihole-FTL` couldn't even do it is it wanted (it we first have to mess with permissions). Or, maybe (c) patch `dnsmasq` that said option is valid for us and a harmless no-op.

I'd rather like to have a "stop, think for a second what you are going to do, please confirm here you've looked into the change log " dialog.

aWZHY0yQH81uOYvH:
I'm one of the users guilty of the cron auto-updater that got my stuff broken on v6 upgrade.

The reason I have an auto-updater is that I often will not log into the admin console to see the flashing "update available" for weeks. For a relatively security-critical application like Pihole, I would like to keep it updated. For all of my time using v5, these auto-updates have never been a problem since there have been no breaking changes (at least with my configs, which are nonstandard).

While in an ideal world all users would read the changelogs for each version, I would argue that it's probably unnecessary when updating from e.g. 5.4.1 to 5.4.2. Minor version changes by definition should be minor changes that likely won't break things.

Adding this big warning thing I think would just lead to [alarm fatigue](https://en.wikipedia.org/wiki/Alarm_fatigue) with users just bypassing it because usually it doesn't matter. Just as people might regularly run `pihole -up` without thinking.

In my opinion, it would be best to have a warning like this between major version changes i.e. v5 to v6 where users *really should read the damn patch notes*, make backups, or at least be there for the process of things going up in flames. Alternatively, maybe just have `pihole -up` only perform minor updates and notify the user of an available major update to be installed with `pihole -major-up`. Similar to how `apt upgrade` usually just installs minor package updates, while upgrading Ubuntu 22.04 to 24.04 requires a different command.

rdwebdesign:
> For a relatively security-critical application like Pihole, I would like to keep it updated. 

The same reason can be used to explain the need to avoid auto-updates:

"For a relatively critical application like Pihole, _I would avoid auto-updating to keep it running when I'm away from my computer._"

> The reason I have an auto-updater is that I often will not log into the admin console to see the flashing "update available" for weeks.

I understand, but maybe you could change your approach and create a script to run `pihole -v` and send an e-mail with the contents... or even check if the versions are the same.

dschaper:
I am for automatically creating a teleporter backup when `pihole -up` is called. I'm also for adding a flag that lets a user run `pihole -up --no-backup` (`pihole -up --I-know-what-im-doing` if you want to be really onerous). 

Users are going to run `pihole` how they want to run `pihole` and that means cronjobs for updates. I can't get behind intentionally breaking cronjobs (mostly because the pitchforks will come out, warranted or not).  

> `aWZHY0yQH81uOYvH` For all of my time using v5, these auto-updates have never been a problem since there have been no breaking changes (at least with my configs, which are nonstandard).

Past performance is not an indication of future returns. We really limited ourselves and created a lot of _uneeded_ work to keep things from breaking. I don't want to be limited like that in the future.


> Adding this big warning thing I think would just lead to [alarm fatigue](https://en.wikipedia.org/wiki/Alarm_fatigue) with users just bypassing it because usually it doesn't matter.

Then it's entirely on the user for shooting themselves in the foot. 

So, a fair middle ground is to allow users to footgun but have a box of bandages (teleporter backup) to try to stem the bleeding.

PromoFaux:
Closing as I don't think this is necessary - feel free to reopen / re discuss, but I think it was more of a reaction to some of the issues that were seen after v6 was released

PR #6022, opened 2025-02-27
Allow uses to skip binary check and installing FTL in case the use a self-compiled binary

yubiuser:

We have some users running Pi-hole on (old) unsupported architectures, most famously PogoPlugs. We are not building them because we use Alpine base image to compile FTL, but there is no ARMv5 Alpine image. 
Currently, users fail to install Pi-hole using the installer, because it will error when detecting the arch and tries to download the FTL binary. 
(Cross)-Compiling FTL on Debian for Armv5 works fine and users could run Pi-hole if the installer would allow them.

This PR adds a simple switch to skip the arch detection and FTL download (`PIHOLE_SKIP_FTL_BINARY_CHECK`). Users would need to install the FTL binary **before** running the script at the right place, because the installer will try to enable/start the binary. 

See this discussion for a related issue: https://discourse.pi-hole.net/t/building-pi-hole-v6-0-on-pogoplug-v2/74448

---

---
- [ ] I have read the above and my PR is ready for review. *Check this box to confirm*


davygravy:
Thank you, @yubiuser ... 

I didn't send this to you via git, but here is my thought...

I'm guessing you wanted  `PIHOLE_SKIP_FTL_CHECK` throughout, and not  the occurrence in line 145: `PIHOLE_SKIP_FTL_BINARY_CHECK`.

That is:
```
root@trixie-debian-armel:/etc/.pihole/automated install# cat basic-install.sh | grep -e 'FTL_CHECK\|FTL_BINARY_CHECK'
PIHOLE_SKIP_FTL_BINARY_CHECK=false
    if [ ! "${PIHOLE_SKIP_FTL_CHECK}" = true ]; then
    if [ ! "${PIHOLE_SKIP_FTL_CHECK}" = true ]; then
```

If the OS check is `PIHOLE_SKIP_OS_CHECK` (4 identifiers) , then the FTL check (w/ 4 identifiers, also) could be `PIHOLE_SKIP_FTL_CHECK`.

```
root@trixie-debian-armel:/etc/.pihole# git diff
diff --git a/automated install/basic-install.sh b/automated install/basic-install.sh
index e3d5a6ae..61e9662f 100755
--- a/automated install/basic-install.sh        
+++ b/automated install/basic-install.sh        
@@ -145,7 +145,7 @@ EOM
 # The runUnattended flag is one example of this
 reconfigure=false
 runUnattended=false
-PIHOLE_SKIP_FTL_BINARY_CHECK=false
+PIHOLE_SKIP_FTL_CHECK=false
 # Check arguments for the undocumented flags
 for var in "$@"; do
     case "$var" in
```

yubiuser:
This is embarrassing. Thank for noticing. I'll FP away the old commit.

davygravy:
If I had a Euro/Dollar/BitCoin for every typo I committed, I’d be…  rich. 🤑 Sent from my iPhoneOn Feb 28, 2025, at 9:46 AM, yubiuser ***@***.***> wrote:﻿
  This is embarrassing. Thank for noticing. I'll FP away the old commit.—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you commented.Message ID: ***@***.***>



yubiuser left a comment (pi-hole/pi-hole#6022)
This is embarrassing. Thank for noticing. I'll FP away the old commit.

automated install/basic-install.sh
```diff
@@ -145,6 +145,7 @@ EOM
 # The runUnattended flag is one example of this
 reconfigure=false
 runUnattended=false
+PIHOLE_SKIP_FTL_CHECK=false
```
DL6ER, line 148: How can it be set from the outside when it is overwritten here?

davygravy:
@yubiuser , I gave this a try - looks usable w/ a few tweaks ... left feedback and observations here -  

https://github.com/pi-hole/pi-hole/commit/56ef68eaa11131545a5d456fded70c36a8196edb#commitcomment-153180218

automated install/basic-install.sh
yubiuser: A check was added

yubiuser:
Has been confirmed working on Discourse: https://discourse.pi-hole.net/t/unsupported-binary-branch-1-proposal-and-1-question/77651/6?u=yubiuser

yubiuser:
Doc PR: https://github.com/pi-hole/docs/pull/1237

Samuel1698:
Can't comment on discourse anymore but it doesn't seem to work for me:
{B59ADABA-C97F-40EE-9B17-C921596F67CD}

yubiuser:
@Samuel1698 

I have rebased on latest development and force pushed. Please try again.

Samuel1698:
@yubiuser It did not correctly parse the environmental variables I was giving it again. Same output as before, and I've given up

I'm tired of fighting this device and hitting a roadblock every which way. From convincing it to let me ssh into it, convincing it to use a different kernel, convincing it to let me run a modern version of Debian, to finally convincing it to install pihole. I did that two years ago and no longer have physical access to the device, so I forgot I wasn't supposed to run the update command. Big mistake, now the web gui broke, tried running `pihole -r` to repair and I hit the same snag you tried to solve with FTL. 

Tried using the same FTL that was working last week, too old. Tried updating, device not supported. Tried manually compiling FTL. Won't let me because cmake is too old on this machine. Tried updating that and it gets stuck with no stacktrace to troubleshoot. I'm tired man. 

These pogoplugs I have laying around have become e-waste and it's not because they don't work or are not capable of it, it's because collectively everyone decided to drop support for amel CPUs.

yubiuser:
Currently failing due to upstream issue. The repo's mirror is empty
```

chris@T14Gen5:~$ docker run --rm -it 02ae49ff109e bash
[root@61cf55964677 /]# yum install curl git initscripts
CentOS Stream 10 - BaseOS                                                                                                                             123 kB/s | 6.7 MB     00:55    
CentOS Stream 10 - AppStream                                                                                                                          129  B/s | 3.9 kB     00:31    
Errors during downloading metadata for repository 'appstream':
  - Status code: 404 for http://mirror.alwyzon.net/centos-stream/10-stream/AppStream/x86_64/os/repodata/repomd.xml (IP: 46.102.157.26)
  - Downloading successful, but checksum doesn't match. Calculated: 0bd449559db10554aa78ff24e6f61ff8339871373caf912d3385434efc0966e3743dd3c1e2803c0bdb291b48d2f40df647a377c4f10a052333a5c026602700cb(sha512)  Expected: e993415a71e0b71009ef92b44155f1e174881b1bc1d5651d1bdf3060ba37f0d1471020177823c76e32d82d5c0b808341ee9eeb3b31a132f86ad7d4c54e1a31f2(sha512) 
  - Status code: 404 for https://mirror.alwyzon.net/centos-stream/10-stream/AppStream/x86_64/os/repodata/repomd.xml (IP: 46.102.157.26)
Error: Failed to download metadata for repo 'appstream': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
```

yubiuser:
Tests pass now

dschaper:
I don't like the idea of copying in a binary we know nothing about and installing it in our pihole binary directory under the `pihole-FTL` name. It's way to easy to install a malicious binary or script and cause it to appear as an FTL binary.

yubiuser:
> I don't like the idea of copying in a binary we know nothing about and installing it in our pihole binary directory under the pihole-FTL name. 

We are not copying any binary - it's the user that needs to do that manually. The only thing this PR does is to allow them to skipp downloading FTL from our servers (which they already know they can't run). I don't see a problem with that option. They need to manually set the flag - it's like the 'SKIP_OS_CHECK' flag we had which would signal them that they do advances stuff and we can't take care of them.

dschaper:
Okay, thanks for the clarification. I think that process is fine then. I'm trying to think of the ways users can break things with the proposed process. And #6316 may affect things as well. Can you rebase your changes on the development branch that has 6316 merged in?

yubiuser:
Rebased and force-pushed. 
From a glance https://github.com/pi-hole/pi-hole/pull/6316 should not interfere with this PR and vice-versa, as all this PR is actually doing (the diff rendering on GH is not very good) is wrapping the FTL check logic in an if statement.

Ahh.. there was another merge just now.... rebasing again.

yubiuser:
> the diff rendering on GH is not very good

Hiding whitespace diffs actually shows the real difference way better
https://github.com/pi-hole/pi-hole/pull/6022/files?w=1

dschaper:
Okay, I think this would go in a point release, which would go with #6319. Not in the next bugfix/hotfix release.

automated install/basic-install.sh
```diff
@@ -156,6 +156,9 @@ EOM
 # The runUnattended flag is one example of this
 repair=false
 runUnattended=false
+if [ -z "$PIHOLE_SKIP_FTL_CHECK" ]; then
+    PIHOLE_SKIP_FTL_CHECK=false
+fi
```
dschaper, line 161:
```suggestion
skipFTL=false
```

Can we keep the scheme the same for all of the undocumented flags? I'm not a fan of two different styles. My preference is to copy the existing style, so use a `--skipftl` instead of environment variables. That will also need to include an additional case at line 167 below `"--unattended"`.

dschaper (changes requested):

yubiuser:
I've change the code from checking of the env var `PIHOLE_SKIP_FTL_CHECK` to checking for the presence of the `--skipFTL` flag.
This makes the code in `update.sh` a bit more complicated.

yubiuser:
I also needed to adapt `pihole -r` and `pihole checkout` because those also calling the installer script and need to pass the new flag. I'm not sure if this change is gain compared to then env variable which doesn't require so many checks and passing because because it's just there when needed.

dschaper:
I don't think this is all worth it. More code to maintain and more things to account for, for a very small number of users, most of whom aren't even using their devices anymore.

This would be better handled as a PogoPlug fork from those that see value in it.

dschaper:
I want to close this PR unless there's still a need and an argument is support of it.

PromoFaux:
No argument from me, I agree with your sentiment that it's a lot of complexity for a very small subset of users

dschaper:
Thanks, I've set it back to draft status so it doesn't get merged but then we also don't lose the work in case it's deemed important enough to include.

Issue #5060, opened 2022-12-12
install via --unattended option does not install lighthttp server


wolfebrian2120:
I'm trying to automate the install of pihole into a Vagrant based virtualbox server. So I have a custom script which runs your install script. I place the following setupVars.conf in the appropriate default location and then run the script with the --unattended option.

I believe it is a bug that the script does not install lighthttpd in unattended mode, since there is an option flag that one could append to the script to disable the install of the webserver component.

setupVars.conf
```
PIHOLE_INTERFACE=eth1
PIHOLE_DNS_1=8.8.8.8
PIHOLE_DNS_2=8.8.4.4
QUERY_LOGGING=true
INSTALL_WEB_SERVER=true
INSTALL_WEB_INTERFACE=true
LIGHTTPD_ENABLED=true
CACHE_SIZE=10000
DNS_FQDN_REQUIRED=true
DNS_BOGUS_PRIV=true
DNSMASQ_LISTENING=local
WEBPASSWORD=1addd3920db07c4b09aff3ece5efcdc142f712a05d589fdd88ddfdaf82e4c4a6
BLOCKING_ENABLED=true
```

Script
```
echo "Beginning pihole setup"
mkdir -p /etc/pihole/
cp /vagrant/setupVars.conf /etc/pihole/setupVars.conf
curl -sSL https://install.pi-hole.net > ~/basic-install.sh
chmod +x ~/basic-install.sh
~/basic-install.sh --unattended
```

### Versions

- Pi-hole:  
- AdminLTE:  
- FTL:  

### Platform
[vagrant@pihole ~]$ cat /etc/redhat-release
CentOS Stream release 8

[vagrant@pihole ~]$ uname -a
Linux pihole 4.18.0-408.el8.x86_64 #1 SMP Mon Jul 18 17:42:52 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

[vagrant@pihole ~]$ pihole -v
  Pi-hole version is v5.14.2 (Latest: v5.14.2)
  FTL version is v5.19.2 (Latest: v5.19.2)

- Platform:  
Vagrant/Virtual Box Guest VM running on intel x86_64

yubiuser:
It should install `lighttpd` the way you run the setup. Please check if EPEL is installed before/after you run the setup

https://github.com/pi-hole/pi-hole/blob/0034538794a3754ab532a53480da39d3fd13d7ed/automated%20install/basic-install.sh#L370-L383

[...]

PR #6034, opened 2025-02-28
Improve update/fresh install detection

PromoFaux:

This PR takes the opportunity to:
 - Fix random password being generated on every update if user had specifically set no password. We do this by now _only_ doing the random password routine when we are certain it is a fresh install
 - Replaces the variable `userUpdateVars` with `fresh_install` and flips the logic around. `useUpdateVars` used to make sense as a name when a user could seed a setupVars.conf file, but this is no longer the case and the naming is not so clear.
 - Moves some code into more logical places, and makes it clear whether or not it runs on a fresh install or not.
 - Remove a dedicated function that didn't do a lot - Could possibly further reduce some duplication in the final stages of `main` by building some strings and printing them both to stdout _and_ the dialog (or just drop the dialog)

Tested with:
 - `pihole checkout` on an up to date v6.0.4 install. Blank password is retained though the checkout and out the other side. Subsequent runs of `pihole -r` also retain a blank password.
 - Simulated fresh install by cloning this repo into `/etc/.pihole`, checking out this branch, and then running the install script directly from there.
 - Simulated Fresh install with --unattended flag and pre-seeded `/etc/pihole/pihole.toml` All scripts are installed and FTL is configured directly. 
 - Simulated update by using @MichaIng's fancy hackery from [his comment here](https://github.com/pi-hole/pi-hole/pull/5968#issuecomment-2676409143) (TL;DR, reset repo to previous tag, replace command to run local `basic-install.sh` with a `curl` command to this branch, run `pihole -up`)

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -81,9 +81,7 @@ PI_HOLE_INSTALL_DIR="/opt/pihole"
 PI_HOLE_CONFIG_DIR="/etc/pihole"
 PI_HOLE_BIN_DIR="/usr/local/bin"
 PI_HOLE_V6_CONFIG="${PI_HOLE_CONFIG_DIR}/pihole.toml"
-if [ -z "$useUpdateVars" ]; then
-    useUpdateVars=false
-fi
+fresh_install=true
 
 adlistFile="/etc/pihole/adlists.list"
 # Pi-hole needs an IP address; to begin, these variables are empty since we don't know what the IP is until this script can run
@@ -93,7 +91,6 @@ IPV6_ADDRESS=${IPV6_ADDRESS}
 QUERY_LOGGING=
 WEBPORT=
 PRIVACY_LEVEL=
-v5_to_v6_update=false
 
 # Where old configs go to if a v6 migration is performed
 V6_CONF_MIGRATION_DIR="/etc/pihole/migration_backup_v6"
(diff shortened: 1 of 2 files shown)


DL6ER (approved):
Code looks good from manual inspection. I also tried one of the methods you described (Simulated fresh install by cloning this repo into `/etc/.pihole`, checking out this branch, and then running the install script directly from there) and everything seems to have worked well.

automated install/basic-install.sh
```diff
@@ -81,19 +81,15 @@ PI_HOLE_INSTALL_DIR="/opt/pihole"
 PI_HOLE_CONFIG_DIR="/etc/pihole"
 PI_HOLE_BIN_DIR="/usr/local/bin"
 PI_HOLE_V6_CONFIG="${PI_HOLE_CONFIG_DIR}/pihole.toml"
-if [ -z "$useUpdateVars" ]; then
-    useUpdateVars=false
-fi
+fresh_install=true
```
yubiuser, line 84: I wonder if this should only be set if it's empty. We have users who run the installer in some kind of automatic setup using the `--unattended` flag and passing a config file (`setupVars.conf`) or now `pihole.toml`.

See here: https://github.com/pi-hole/pi-hole/issues/5060


Maybe this is not necessary, I did not deeply review nor test, but it just came to my mind.

PromoFaux: One of the paths I tested was running the installer with `--unattended` set and a preconfigured `/etc/pihole/pihole.toml`. Seemed to work OK!

yubiuser: Perfect!

Issue #6047, opened 2025-03-05
When calling `pihole -g` a `stat: cannot stat` message always appears when loading newly added lists.


LizenzFass78851:
### Versions

- Pi-hole:  v6.0.5
- AdminLTE:  v6.0.2
- FTL:  v6.0.4

### Platform

- OS and version:  Ubuntu 20.04 (over GitHub Codespace as Test)

- Platform:  Docker

### Expected behavior

I added lists to pihole and ran the command "pihole -g" in the docker container and a "stat:" message is displayed when loading new lists.

### Actual behavior / bug

When you call pihole -g you get a message like this:
`stat: cannot stat '/etc/pihole/listsCache/list.149.raw.githubusercontent.com.domains': No such file or directory`

### Steps to reproduce

1. I started pihole as docker compose in a github codespace.
2. I called https://example.invalid/admin/groups/lists and added all of @RPiList's lists to the pihole.
3. I called the pihole -g command in the Docker image (docker exec -it pihole bash).
4. then I saw one of these "stat:" messages.

## Debug Token

- URL: https://tricorder.pi-hole.net/MzbMUbjy/

### Screenshots

![Image](https://github.com/user-attachments/assets/f58acbfe-8f11-4214-b7ef-9b7ac932e28f)

### Additional context

This only occurs if the relevant lists have not been added previously.

yubiuser:
This looks like a side-effect of https://github.com/pi-hole/pi-hole/pull/6030

mwoolweaver:
it seems `-w` flag has been an issue for a few years now with no obvious fixes

https://github.com/alpinelinux/docker-alpine/issues/156

mwoolweaver:
Fixed by https://github.com/pi-hole/pi-hole/pull/6086

PR #6030, opened 2025-02-28
Fix gravity.sh -- Alpine / Docker specific issue

karolkania:
fix the `if statement` that doesn't seem to work for neither of alpine's ash / bash - applying some workaround with the `stat` command

## Thank you for your contribution to the Pi-hole Community!

Please read the comments below to help us consider your Pull Request.

We are all volunteers and completing the process outlined will help us review your commits quicker.

**Please make sure you**

 1. Base your code and PRs against the repositories developmental branch.
 2. [Sign Off](https://docs.pi-hole.net/guides/github/how-to-signoff/) all commits as we enforce the [DCO](https://docs.pi-hole.net/guides/github/dco/) for all contributions
 3. [Sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) all your commits as they must have verified signatures
 4. File a pull request for any change that requires changes to [our documentation](https://docs.pi-hole.net/) at our [documentation repo](https://github.com/pi-hole/docs) 

---

**What does this PR aim to accomplish?:**

**How does this PR accomplish the above?:**

**Link documentation PRs if any are needed to support this PR:**

---
**By submitting this pull request, I confirm the following:**

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against.
2. I have commented my proposed changes within the code and I have tested my changes.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
6. I have checked that another pull request for this purpose does not exist.
7. I have considered, and confirmed that this submission will be valuable to others.
8. I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
9. I give this submission freely, and claim no ownership to its content.

---
- [x] I have read the above and my PR is ready for review. *Check this box to confirm*

PR #6053, opened 2025-03-06
fix: show "stat" message when run "pihole -g"

LizenzFass78851:

Fixes: #6047

this hides the standard error output of the "stat" command and redirects it to /dev/null

N/A

---

---


gravity.sh
```diff
@@ -541,15 +541,15 @@ gravity_DownloadBlocklists() {
     # it (in case it doesn't exist)
     # First, check if the directory is writable
     directory="$(dirname -- "${saveLocation}")"
-    directory_permissions=$(stat -c %a ${directory})
+    directory_permissions=$(stat -c %a ${directory} 2>/dev/null)
     if [ $directory_permissions -lt 700 ]; then
```
dschaper, line 545: I'm not sure why there's a variable created for a single use. 

```suggestion
    if [ $(stat -c %a ${directory} -lt 700) ]; then
```

yubiuser (commented):
This change does not fix the underlying issue.

`stat` does not error at the directory level, but errors because the file does not exist yet (Line 553). 

Before we had
> if [ -e "${saveLocation}" ] && [ ! -w "${saveLocation}" ]; then

and it seems that `[ -w]` is fine checking if a potential file **could** be written, whereas `stat -c` checks permissions for an already existing file/dir.

I think, what we need now is first the check if the file exists and only then check if it's writable

LizenzFass78851:
> I think, what we need now is first the check if the file exists and only then check if it's writable 

I'm not sure what that means? 
Should the logic with the stat command remain and just hide the error message or should this stat logic be replaced? 

I have also added a check to the first line that contains this stat command that checks whether the directory exists.

yubiuser:
> Should the logic with the stat command remain and just hide the error message or should this stat logic be replaced?

The logic needs to be amended. Currently, when the file does not exist, `stats` will error and the `${saveLocation_permissions}` variable is empty. 

```
stat: cannot statx '/etc/pihole/listsCache/list.1.raw.githubusercontent.com.domains': No such file or directory
saveLocation_permissions is 
```

___

 I think there is no need to check for the existence of the directory, as it will be `listsCacheDir` which is created right at the start of gravity in case it does not exist (in `listsCacheDir()`)

LizenzFass78851:
@yubiuser 
I kept the version with the one-liner in the first line that contains the stat command. This means that the standard error output is not visible (the redirection to /dev/null is therefore not necessary)
https://github.com/pi-hole/pi-hole/blob/19d5e5702129434e947bbe76aa8d196bc63eaae0/gravity.sh#L544

In the second area where the stat command can be found, I changed the if phase so that it first checks whether the "${saveLocation}" exists and then the authorization is checked with the stat command.
https://github.com/pi-hole/pi-hole/blob/19d5e5702129434e947bbe76aa8d196bc63eaae0/gravity.sh#L551-L558

I did not want to replace the stat command with the version that existed before it was introduced by https://github.com/pi-hole/pi-hole/pull/6030. This is because it is possible that the alpine-specific problem (regarding ash/sh/bash) will reoccur.

PR #6147, opened 2025-04-04
Shellcheck Tweaks

PromoFaux:

Bump severity level of shellcheck GHA to "warning", and fix all the warnings it complains about.

Have broken it down to one commit per file, with the shellcheck output that prompted the fix as the message - this should make it easier to review.

---

---

### DIFF
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -1,0 +1,1 @@
+disable=SC1090,SC1091 # Ignore warnings about being unable to follow sourced files
--- a/advanced/Scripts/piholeCheckout.sh
+++ b/advanced/Scripts/piholeCheckout.sh
@@ -109,7 +109,7 @@ checkout() {
             echo -e "${OVER}  ${CROSS} $str"
             exit 1
         fi
-        corebranches=($(get_available_branches "${PI_HOLE_FILES_DIR}"))
+        mapfile -t corebranches < <(get_available_branches "${PI_HOLE_FILES_DIR}")
 
         if [[ "${corebranches[*]}" == *"master"* ]]; then
             echo -e "${OVER}  ${TICK} $str"
@@ -136,7 +136,7 @@ checkout() {
             echo -e "${OVER}  ${CROSS} $str"
             exit 1
         fi
-        webbranches=($(get_available_branches "${webInterfaceDir}"))
+        mapfile -t webbranches < <(get_available_branches "${webInterfaceDir}")
 
         if [[ "${webbranches[*]}" == *"master"* ]]; then
             echo -e "${OVER}  ${TICK} $str"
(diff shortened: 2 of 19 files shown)


automated install/basic-install.sh
```diff
@@ -2067,13 +2068,12 @@ FTLcheckUpdate() {
         path="${ftlBranch}/${binary}"
 
         # Check whether or not the binary for this FTL branch actually exists. If not, then there is no update!
-        # shellcheck disable=SC1090
         if ! check_download_exists "$path"; then
-            if [ $? -eq 1 ]; then
+            status=$?
```
yubiuser, line 2072: `status` is quite generic. Maybe add `local status` at L 2064?

advanced/Scripts/piholeCheckout.sh
```diff
@@ -209,13 +212,14 @@ checkout() {
             # Update local and remote versions via updatechecker
             /opt/pihole/updatecheck.sh
         else
-            if [ $? -eq 1 ]; then
+            result=$?
```
yubiuser, line 215: `result` should be `local result` somewhere.

yubiuser (commented):
I'm not sure if it's wise to disable SC2154 and SC2034 in some files globally by putting them right on top of the script. I think I would prefer to use them right at the place where needed.

.shellcheckrc
```diff
@@ -0,0 +1 @@
+disable=SC1090,SC1090 # Ignore warnings about being unable to follow sourced files
```
yubiuser, line 1:
```suggestion
disable=SC1090 # Ignore warnings about being unable to follow sourced files
```

Is there a reason to put it in twice?

PromoFaux: Ah, no - good catch. Actually the second instance should be `SC1091`

PromoFaux: FP'd to fix

.shellcheckrc
```diff
@@ -0,0 +1 @@
+disable=SC1090,SC1091 # Ignore warnings about being unable to follow sourced files
```
yubiuser, line 1: When you set 
```suggestion
external-sources=true
```

here, `shellcheck` should follow external sources (might need `# shellcheck source=somefile` in some places).
See the 'Rationale' section here: https://www.shellcheck.net/wiki/SC1091

PromoFaux: I'm fine with ignoring these two warnings entirely - we always have done, this just makes it a more global change without having to put the directive in any file that needs it.

None of the paths it complains about are relative to the repo from a development point of view - applying the `source=` directive will not help in any of these cases

yubiuser: Ok, let's go on with this, I might need to tweak in a follow-up PR

yubiuser (approved):

PR #6152, opened 2025-04-05
Offer to add local user to pihole group on fresh installations

yubiuser:

Certain CLI functions e.g. `pihole -v` or `pihole api` require access to files in `/etc/pihole` which belong to `pihole:pihole`. This PR adds a dialog to offer to put the user with UID 1000 (usually the first local user) into the `pihole` group during installation.

To be discussed.

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1617,6 +1617,60 @@ create_pihole_user() {
     fi
 }
 
+# offer to add the first local user (UID 1000) to the pihole group
+local_user_to_pihole_group(){
+    local username
+    local str="Checking for user with UID 1000"
+    printf "  %b %s..." "${INFO}" "${str}"
+
+    username=$(getent passwd 1000 | cut -d: -f1)
+
+    # No user with UID 1000 found
+    if [[ -z "${username}" ]]; then
+        str="No user with UID 1000 found"
+        printf "%b  %b %s\\n" "${OVER}" "${CROSS}" "${str}"
+        return
+    fi
+    # User with UID 1000 already in pihole group
+    if id -nG "${username}" | grep -q pihole; then
+        str="User ${username} already in pihole group"
+        printf "%b  %b %s\\n" "${OVER}" "${TICK}" "${str}"
+        return
+    fi
+
+    # User with UID 1000 not in pihole group
+    # Offer dialog to add user to pihole group
+    dialog --no-shadow --keep-tite \
+        --backtitle "Pi-hole Installation" \
+        --title "Add user to pihole group" \
+        --yesno "\\nFor privacy reasons certain Pi-hole CLI functions are only allowed to members of the newly created the local group 'pihole'.\
+\\n\\nYour first local user is '${username}' \
+\\n\\nWould you like to add '${username}' to the pihole group? \
+\\n\\n\\nP.S. Any user can be added to the group later manually. "\
+        "${r}" "${c}" && result=0 || result=$?
+
+    case ${result} in
+    "${DIALOG_OK}")
+        # If they chose yes,
+        printf "  %b Adding user ${username} to pihole group\\n" "${INFO}"
+        if usermod -aG pihole "${username}"; then
+            printf "  %b User ${username} added to pihole group\\n" "${TICK}"
+        else
+            printf "  %b Error adding user ${username} to pihole group\\n" "${CROSS}"
+        fi
+        ;;
+    "${DIALOG_CANCEL}")
+        # If they chose no,
+        printf "  %b Not adding user ${username} to pihole group\\n" "${INFO}"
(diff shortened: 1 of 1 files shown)


yubiuser:
Inspired by https://github.com/pi-hole/pi-hole/pull/6002#issuecomment-2779961094

mwoolweaver:
instead of adding id `1000` could something like what is mentioned [here](https://askubuntu.com/a/1034050)? 

`$SUDO_USER` could be used to get the user that called the script with sudo.

dschaper:
I'm not comfortable with this. I can't think of any other package that offers this, I know Docker (get.docker.com) prints out a message to advise users to add the proper user to the right group(s) but I don't like the idea of doing it programmatically. Too many ways this could go very wrong.

mwoolweaver:
> I can't think of any other package that offers this

I only know of [one](https://wiki.servarr.com/install-script).

yubiuser:
> prints out a message to advise users to add the proper user to the right group(s) 

We could also reduce the scope of this PR to only show an advise to do this.

PromoFaux:
In my opinion we should just set versions file consistently to `644` (so, service wrapper needs updating) and not worry about this. 

There is far too much complexity added in this PR for extremely little gain.

> We could also reduce the scope of this PR to only show an advise to do this.

If anything, then this - but with no checks for the user ID or anything like that. docs.pi-hole.net probably a good target for somewhere to have this information.

dschaper:
We could tag on a note to the last dialog displayed that shows the generated password. Something like https://github.com/docker/docker-install/blob/master/install.sh#L289. It looks like the verbiage has been updated to point to the documentation now instead of suggesting adding the user to the docker group. https://github.com/docker/docker-install/commit/b5567e71b7adca1f42db82b7f9bcd6fb36909bcb

image

yubiuser:
I removed to code for automatic addition and added some info to the final dialog.

automated install/basic-install.sh
```diff
@@ -2489,7 +2489,13 @@ main() {
 \\n\\nIPv4:	${IPV4_ADDRESS%/*}\
 \\nIPv6:	${IPV6_ADDRESS:-"Not Configured"}\
 \\nIf you have not done so already, the above IP should be set to static.\
-\\nView the web interface at http://pi.hole/admin:${WEBPORT} or http://${IPV4_ADDRESS%/*}:${WEBPORT}/admin\\n\\nYour Admin Webpage login password is ${pw}" "${r}" "${c}"
+\\nView the web interface at http://pi.hole/admin:${WEBPORT} or http://${IPV4_ADDRESS%/*}:${WEBPORT}/admin\\n\\nYour Admin Webpage login password is ${pw}\
+\\n
+\\n
+\\nTo allow your user to use all CLI functions without authentication,\
+\\nadd your user to the 'pihole' group with something like:\
```
PromoFaux, line 2496:
```suggestion
\\nadd your user to the 'pihole' group with:\
```

"Something like" is a bit too non-specific.

Also, there is some similar output to the CLI - possibly worth putting it there, too.

yubiuser: Done.

rdwebdesign (approved):

dschaper:
Can we see the dialog with the final changes made please?

dschaper:
Thanks, I can tell you now that we will get support requests saying `sudo usermod -aG pihole $your_user` failed. 

But I'll re-request Promo on it.

yubiuser:
I know... I think you suggested to do it like `docker` which used the phrase above. I added `something like` because it might be a bit different syntax on different OS. But Promo complaint....

We can also get rid of an example command, and only leave 

> add your user to the 'pihole' group.

But this will invoke even more follow-up questions....

PromoFaux:
> We can also get rid of an example command, and only leave

To be honest, I think this is probably best. how about a section on the docs site that shows a different examples for different OS, and provide a link in the dialog rather than an example command.

dschaper:
> But this will invoke even more follow-up questions....

That's exactly how the docker flow is now, 
```bash
	echo
	echo "To run the Docker daemon as a fully privileged service, but granting non-root"
	echo "users access, refer to https://docs.docker.com/go/daemon-access/"
	echo
	echo "WARNING: Access to the remote API on a privileged Docker daemon is equivalent"
	echo "         to root access on the host. Refer to the 'Docker daemon attack surface'"
	echo "         documentation for details: https://docs.docker.com/go/attack-surface/"
	echo
	echo "================================================================================"
	echo
```

yubiuser:
I changed the dialog/cli to refer to our post-installation documentation and squashed the commits. 

The doc PR is at https://github.com/pi-hole/docs/pull/1244

PromoFaux (approved):

Issue #6185, opened 2025-03-08
Root-owned files in pihole/etc/pihole


causalmask:
My backup jobs have been failing since I pulled the most recent version of the container.  It seems to be due to the presence of root-owned files in pihole directories that I have mapped to local directories.  I don't run my backup jobs as root, so the job fails when attempting to read files that are only readable by root. 

The reason I bring this up is because pihole is currently the only container of the tens of containers I am running where this is an issue.  I'm guessing it has to do with the recent addition of logrotate to the container, and I'm not sure of an easy workaround.  For now, I'm excluding certain pihole files from my backup job, but would prefer to be able to backup the entire config directory as I do with my other containers.  

Any workarounds or changes to the container to improve this situation would be greatly appreciated.

yubiuser:
This is a bug not limited to docker. The issue is the gravity script. It will set some owners to root, which will get fixed on the next `pihole-FTL` start.  I'll prepare a fix.
Before gravity
```
nanopi@nanopi:~$ ls -lha /etc/pihole/
insgesamt 461M
drwxrwxr-x  8 pihole pihole 4,0K 20. Apr 23:03 .
drwxr-xr-x 93 root   root   4,0K  6. Apr 20:52 ..
-rw-r-----  1 pihole pihole   65  1. Aug 2023  adlists.list
-rw-r-----  1 pihole pihole   44 20. Apr 23:02 cli_pw
drwxr-xr-x  2 pihole pihole 4,0K 20. Apr 21:30 config_backups
-rw-r-----  1 pihole pihole   47 31. Mär 13:36 dhcp.leases
-rw-r-----  1 pihole pihole 6,4K  5. Apr 21:18 dnsmasq.conf
-rw-r-----  1 pihole pihole   12 20. Apr 23:00 ftlbranch
drwxr-xr-x  2 pihole pihole 4,0K 20. Apr 23:03 gravity_backups
-rw-rw-r--  1 pihole pihole 167M 20. Apr 23:02 gravity.db
-rw-r-----  1 pihole pihole 167M 20. Apr 23:03 gravity_old.db
drwxr-xr-x  2 pihole pihole 4,0K 20. Apr 21:19 hosts
-rw-r-----  1 pihole pihole  408 20. Apr 23:02 install.log
drwxrwxr-x  2 pihole pihole 4,0K 20. Apr 23:02 listsCache
-rw-r-----  1 root   root    358 28. Aug 2024  logrotate
[...]

PR #6186, opened 2025-04-21
All gravity related files and dirs should be owned by pihole:pihole

yubiuser:

`gravity` runs as root, so all new created dirs or copied files will be owned by `root` by default. This PR fixes 3 locations where we need to set `pihole:pihole` explicitly. 
Fixes https://github.com/pi-hole/pi-hole/issues/6185

P.S. On FTL restart, we execute `prestart.sh` which would also fix file/folder permissions.

---

---

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -126,7 +126,7 @@ gravity_swap_databases() {
   oldAvail=false
   if [ "${availableBlocks}" -gt "$((gravityBlocks * 2))" ] && [ -f "${gravityDBfile}" ]; then
     oldAvail=true
-    cp "${gravityDBfile}" "${gravityOLDfile}"
+    cp -p "${gravityDBfile}" "${gravityOLDfile}"
   fi
 
   # Drop the gravity and antigravity tables + subsequent VACUUM the current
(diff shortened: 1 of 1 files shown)


gravity.sh
```diff
@@ -140,6 +140,7 @@ gravity_swap_databases() {
     # Check if the backup directory exists
     if [ ! -d "${gravityBCKdir}" ]; then
       mkdir -p "${gravityBCKdir}"
+      chown pihole:pihole "${gravityBCKdir}"
```
dschaper, line 143:
```suggestion
      mkdir -p "${gravityBCKdir}" && chown pihole:pihole "${gravityBCKdir}"
```

Adding in that `&&` so the `chown` is only called if the directory is created.

gravity.sh
```diff
@@ -1016,6 +1017,7 @@ migrate_to_listsCache_dir() {
   local str="Migrating the list's cache directory to new location"
   echo -ne "  ${INFO} ${str}..."
   mkdir -p "${listsCacheDir}"
+  chown pihole:pihole "${listsCacheDir}"
```
dschaper, line 1020:
```suggestion
  mkdir -p "${listsCacheDir}" && chown pihole:pihole "${listsCacheDir}"
```

Addin in that `&&` so the `chown` is only called if the directory is created.

dschaper (changes requested):

gravity.sh
yubiuser: You mean as a safeguard in case `mkdir` fails?

dschaper: Yes, and because it's impossible to `chown` a directory that doesn't exist. The `&&` is cleaner than doing a "If directory exists then chown directory".

dschaper (approved):

PR #6193, opened 2025-04-24
Fix API logic in api.sh

yubiuser:

Fix the logic in `api.sh`. I noticed we had the leftover `API_PORT` which was intended to signal if we found a responding api. The check for emptiness war wrong in the first place (using `-n` instead of `-z`). 

This PR fixes the logic by better naming the variable and explicitly setting it to true/false.

---

---

### DIFF
--- a/advanced/Scripts/api.sh
+++ b/advanced/Scripts/api.sh
@@ -57,34 +57,33 @@ TestAPIAvailability() {
         authData=$(printf %s "${authResponse%???}")
 
         # Test if http status code was 200 (OK) or 401 (authentication required)
-        if [ ! "${authStatus}" = 200 ] && [ ! "${authStatus}" = 401 ]; then
-            # API is not available at this port/protocol combination
-            apiAvailable=false
-        else
-            # API is available at this URL combination
+        if [ "${authStatus}" = 200 ]; then
+            # API is available without authentication
+            apiAvailable=true
+            needAuth=false
+            break
 
-            if [ "${authStatus}" = 200 ]; then
-                # API is available without authentication
-                needAuth=false
-            fi
-
+        elif [ "${authStatus}" = 401 ]; then
+            # API is available with authentication
+            apiAvailable=true
+            needAuth=true
             # Check if 2FA is required
             needTOTP=$(echo "${authData}"| jq --raw-output .session.totp 2>/dev/null)
-
-            apiAvailable=true
-
             break
-        fi
 
-        # Remove the first URL from the list
-        local last_api_list
-        last_api_list="${chaos_api_list}"
-        chaos_api_list="${chaos_api_list#* }"
+        else
+            # API is not available at this port/protocol combination
+            apiAvailable=false
+            # Remove the first URL from the list
+            local last_api_list
+            last_api_list="${chaos_api_list}"
+            chaos_api_list="${chaos_api_list#* }"
 
-        # If the list did not change, we are at the last element
-        if [ "${last_api_list}" = "${chaos_api_list}" ]; then
-            # Remove the last element
-            chaos_api_list=""
+            # If the list did not change, we are at the last element
+            if [ "${last_api_list}" = "${chaos_api_list}" ]; then
+                # Remove the last element
+                chaos_api_list=""
+            fi
         fi
     done
 
--- a/advanced/Scripts/api.sh
+++ b/advanced/Scripts/api.sh
@@ -20,7 +20,7 @@
 TestAPIAvailability() {
 
     # as we are running locally, we can get the port value from FTL directly
-    local chaos_api_list authResponse authStatus authData
+    local chaos_api_list authResponse authStatus authData apiAvailable
 
     # Query the API URLs from FTL using CHAOS TXT local.api.ftl
     # The result is a space-separated enumeration of full URLs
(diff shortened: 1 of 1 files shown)


advanced/Scripts/api.sh
```diff
@@ -59,7 +59,7 @@ TestAPIAvailability() {
         # Test if http status code was 200 (OK) or 401 (authentication required)
         if [ ! "${authStatus}" = 200 ] && [ ! "${authStatus}" = 401 ]; then
             # API is not available at this port/protocol combination
-            API_PORT=""
+            apiAvailable=false
```
dschaper, line 62:
```suggestion
        echo "API not available. Please check FTL.log"
        echo "Exiting."
        exit 1
```

advanced/Scripts/api.sh
```diff
@@ -20,7 +20,7 @@
 TestAPIAvailability() {
 
     # as we are running locally, we can get the port value from FTL directly
-    local chaos_api_list authResponse authStatus authData
+    local chaos_api_list authResponse authStatus authData apiAvailable
```
dschaper, line 23:
```suggestion
    local chaos_api_list authResponse authStatus authData
```

advanced/Scripts/api.sh
```diff
@@ -71,6 +71,8 @@ TestAPIAvailability() {
             # Check if 2FA is required
             needTOTP=$(echo "${authData}"| jq --raw-output .session.totp 2>/dev/null)
 
+            apiAvailable=true
+
```
dschaper, line 75:
```suggestion
```

advanced/Scripts/api.sh
```diff
@@ -86,9 +88,9 @@ TestAPIAvailability() {
...
 
-    # if API_PORT is empty, no working API port was found
-    if [ -n "${API_PORT}" ]; then
-        echo "API not available at: ${API_URL}"
+    # if apiAvailable is false, no working API was found
+    if [ "${apiAvailable}" = false ]; then
+        echo "API not available. Please check FTL.log"
         echo "Exiting."
         exit 1
     fi
```
dschaper, line 96:
```suggestion
```

dschaper (changes requested):
If the API isn't available then immediately write to log `exit`, no need to create a local variable to track.

yubiuser:
We need the variable because it is set within the loop. The `chaos_api_list` we iterate over can contain multiple URLs and we need to check if we find at least one working API. Only if we exit the loop because we finished iteration and `apiAvailable` is still `false` we know that none of the URLs returned by `chaos_api_list` did work. However, if we find at least one working API, we also exit the loop prematurely and `apiAvailable` will be true, skipping the exit.

dschaper:
Okay, then the logic is off here.

I don't like having `!` tests for things like this, it's backwards logic and doesn't account for changes in the API response. 

What happens if the API is changed to return a 403? 

Do we currently keep iterating through the list of URLs even after we find an open endpoint?

Do we track which URL was 200, or which URL was 401?

dschaper:
1. Get list of URLs
2. Check for response from URL in URLs.
3. if 200 then use that URL and stop looking
4. else if 401 then check if we can authenticate
5. if we can't then continue
6. if were at the end of the list then we haven't found a valid URL, log and exit.
7.

yubiuser:
> Do we currently keep iterating through the list of URLs even after we find an open endpoint?

No.  

https://github.com/pi-hole/pi-hole/blob/1bea6db50adb3a07ba58808791c7c013bc12a671/advanced/Scripts/api.sh#L74-L76

> Do we track which URL was 200, or which URL was 401?

Yes. As we exit the loop once we found a responding URL, it's the one set at this moment in `API_URL`. We know if we got `200` or `401` by setting

https://github.com/pi-hole/pi-hole/blob/1bea6db50adb3a07ba58808791c7c013bc12a671/advanced/Scripts/api.sh#L66-L68

____

> Get list of URLs
Check for response from URL in URLs.
if 200 then use that URL and stop looking
else if 401 then check if we can authenticate
if we can't then continue
if were at the end of the list then we haven't found a valid URL, log and exit.

This is (almost) exactly what we do.

____

> I don't like having ! tests for things like this, it's backwards logic and doesn't account for changes in the API response.

I see your point but we adhere to the HTTP return codes used everywhere. I can reverse the logic to explicitly check for `200` and `400` instead, but this would change nothing if FTL decides to change the response code.

dschaper:
I like explicit, if we need 200 or 401 then check for 200 or 401. 

Read through 

```plain
        # Test if http status code was 200 (OK) or 401 (authentication required)
        if [ ! "${authStatus}" = 200 ] && [ ! "${authStatus}" = 401 ]; then
```

And tell me what we're looking for, compared to 
```bash
# Test if http status code is 200 (OK)
if [ "${authStatus}" = 200 ]; then
    needAuth=false
# Test if http status is 401 (Auth required)
elif [ "${authStatus}" = 401]; then
    # Check if 2FA is required
    needTOTP=$(echo "${authData}"| jq --raw-output .session.totp 2>/dev/null)
else
    # Remove the first URL from the list
    local last_api_list
    last_api_list="${chaos_api_list}"
    chaos_api_list="${chaos_api_list#* }"

    # If the list did not change, we are at the last element
    if [ "${last_api_list}" = "${chaos_api_list}" ]; then
        # Remove the last element
        chaos_api_list=""
       
        # List of URLs is empty and we have not found a valid URL
        echo "API not available. Please check FTL.log"
        echo "Exiting."
        exit 1

    fi
fi
```

yubiuser:
Logic is now reverted.

dschaper (approved):

PR #6206, opened 2025-04-28
Remove check for supported OS

yubiuser:

Checking for the distribution and version has been introduced 4-5 years ago. It was intended to lower the number of support requests due to outdated OS. Since then we have reduced the number of external dependencies (like `lighttpd`, `php`). Also FTL is now `musl` based statically compiled and does not depend on external OS libraries.

Today, we have much less users issues due to OS level issues, but rather due to bugs/oversights we made ourselves. 
The OS check requires constant updates to the necessary TXT records. 

In our documentation we clearly state, which distributions are supported and
> Pi-hole only supports actively maintained versions of these systems.

The debug script still prints the used OS/version so "unsupported" configurations can still be recognized.

This also decouples the list of "supported OS" from the tests - we can easily keep old/"unsupported" tests around until they are dropped upstream.

See discussion https://github.com/pi-hole/pi-hole/discussions/6204

Doc PR: https://github.com/pi-hole/docs/pull/1238

---

---


dschaper:
Have I told you lately that I love your detailed PRs?

Johnny1982fd:
Why will the OS Check be removed?

dschaper:
I think this all looks great. Is this a change we can put in to a minor release? Or can we consider it to be a bug fix? This is the only thing keeping me from approving and merging in to development, I don't want to prevent other bug fixes going out because this change is in the development branch and we can't push out the bugfixes to a release from the development branch while this code also there.

So, consider it approved from my seat but waiting for the time to merge it to the development branch.

yubiuser:
I think this should be in a minor release, but nothing prevents the next release to be s minor bump (instead of a patch)

dschaper:
Okay, then I'll approve and merge which will preclude any patch release on the current development branch and require a minor version step.

dschaper (approved):

PromoFaux:
Great addition (or subtraction, I guess!)

PR #6224, opened 2025-05-09
fix(get_available_interfaces): only drop the loopback device (‘lo’), don’t filter every “lo” substring

PiotrTyrakowski:
### Problem

The `get_available_interfaces()` helper currently does:

```bash
ip --oneline link show up \
  | grep -v "lo" \
  | awk '{print $2}' | cut -d: -f1 | cut -d@ -f1

it removes wlo1 from the output.
i have updated command to use "loo" instead.

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -705,7 +705,7 @@ find_IPv4_information() {
 # Get available interfaces that are UP
 get_available_interfaces() {
     # There may be more than one so it's all stored in a variable
-    availableInterfaces=$(ip --oneline link show up | grep -v "loo" | awk '{print $2}' | cut -d':' -f1 | cut -d'@' -f1)
+    availableInterfaces=$(ip --oneline link show up | awk '{print $2}' | grep -v "^lo" | cut -d':' -f1 | cut -d'@' -f1)
 }
 
 # A function for displaying the dialogs the user sees when first running the installer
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by dschaper** [comment]
```
ip --oneline link show up | awk '{print $2}'
lo:
eth0:
```

I think filtering after the awk step would be safer:
```
dan@Columbia:~$ ip --oneline link show up | awk '{print $2}' | grep -v "^lo"
eth0:
```

File: automated install/basic-install.sh [hunk 1]
```diff
@@ -705,7 +705,7 @@ find_IPv4_information() {
 # Get available interfaces that are UP
 get_available_interfaces() {
     # There may be more than one so it's all stored in a variable
-    availableInterfaces=$(ip --oneline link show up | grep -v "lo" | awk '{print $2}' | cut -d':' -f1 | cut -d'@' -f1)
+    availableInterfaces=$(ip --oneline link show up | grep -v "loo" | awk '{print $2}' | cut -d':' -f1 | cut -d'@' -f1)
```
dschaper, line 708:
```suggestion
    availableInterfaces=$(ip --oneline link show up | awk '{print $2}' | grep -v "^lo" | cut -d':' -f1 | cut -d'@' -f1)
```

```bash
$ ip --oneline link show up | awk '{print $2}'
lo:
eth0:

ip --oneline link show up | awk '{print $2}' | grep -v "^lo"
eth0:
```

dschaper (changes requested):

automated install/basic-install.sh
PiotrTyrakowski: Yea this also would work.

dschaper (approved):

yubiuser:
@dschaper This was merged to master.

yubiuser:
@PiotrTyrakowski 

Can you please open a new PR based on and targeting `development`?

PR #6092, opened 2025-03-18
feat: add Alpine Linux support

jrittenh:
- Add Alpine Linux package manager detection
- Implement meta package for Alpine dependencies
- Add Alpine-specific paths and service management
- Handle Alpine cleanup in uninstaller

## Thank you for your contribution to the Pi-hole Community!

Please read the comments below to help us consider your Pull Request.

We are all volunteers and completing the process outlined will help us review your commits quicker.

**Please make sure you**

 1. Base your code and PRs against the repositories developmental branch.
 2. [Sign Off](https://docs.pi-hole.net/guides/github/how-to-signoff/) all commits as we enforce the [DCO](https://docs.pi-hole.net/guides/github/dco/) for all contributions
 3. [Sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) all your commits as they must have verified signatures
 4. File a pull request for any change that requires changes to [our documentation](https://docs.pi-hole.net/) at our [documentation repo](https://github.com/pi-hole/docs) 

---

**What does this PR aim to accomplish?:**

This PR adds support for Alpine as a target installation OS.

**How does this PR accomplish the above?:**

Minimal changes following the other OS build processes, with definitions specific to Alpine

**Link documentation PRs if any are needed to support this PR:**

---
**By submitting this pull request, I confirm the following:**

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against.
2. I have commented my proposed changes within the code and I have tested my changes.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
6. I have checked that another pull request for this purpose does not exist.
7. I have considered, and confirmed that this submission will be valuable to others.
8. I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
9. I give this submission freely, and claim no ownership to its content.

---
- [x] I have read the above and my PR is ready for review. *Check this box to confirm*

PR #6226, opened 2025-05-11
Alpine Linux installation and tests

rrobgill:

Add support for Alpine linux to the installer, and adds current Alpine Linux to the Pi-Hole test suite.

Install and uninstall scripts:

- Changes for Alpine by @jrittenh from #6092
  (with minor adjustment to allow installation on bare metal and to pass the Pi-hole automated tests)

Test suite:
- Apk package manager added
- Adjusted to use short flags suiting busybox's builtin su
- New alpine test file added, pulling Alpine 3.21 docker image.

n/a

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -90,6 +90,7 @@ PI_HOLE_INSTALL_DIR="/opt/pihole"
 PI_HOLE_CONFIG_DIR="/etc/pihole"
 PI_HOLE_BIN_DIR="/usr/local/bin"
 PI_HOLE_V6_CONFIG="${PI_HOLE_CONFIG_DIR}/pihole.toml"
+CRON_D_DIR="/etc/cron.d"
 fresh_install=true
 
 adlistFile="/etc/pihole/adlists.list"
@@ -145,6 +146,55 @@ Pi-hole dependency meta package
 %changelog
 * Sun Sep 29 2024 Pi-hole Team - 0.1
 - First version being packaged
+EOM
+)
+
+# Content of Pi-hole's meta package control file on Alpine Linux
+PIHOLE_META_PACKAGE_CONTROL_APK=$(
+    cat <<'EOM'
+# Contributor: Pi-hole team <adblock@pi-hole.net>
+# Maintainer: Pi-hole team <adblock@pi-hole.net>
+pkgname=pihole-meta
+pkgver=0.1
+pkgrel=1
+pkgdesc="Pi-hole dependency meta package"
+url="https://pi-hole.net"
+arch="noarch"
+license="EUPL-1.2"
+depends="bash bind-tools binutils coreutils curl git grep iproute2-ss jq libcap logrotate ncurses nmap-ncat procps-ng psmisc shadow sudo tzdata unzip wget bash-completion dialog git"
+makedepends=""
+source=""
+subpackages=""
+install=""
+options="!check"  # No tests for meta package
+
+prepare() {
+    mkdir -p "${srcdir}"
+}
+
+build() {
+    :
+}
+
+package() {
+    # Create required directories with proper paths
+    mkdir -p "${pkgdir}/usr/share/doc/${pkgname}"
+    mkdir -p "${pkgdir}/etc/pihole"
+
+    # Create and install README file with proper permissions
+    cat > "${pkgdir}/usr/share/doc/${pkgname}/README" << INNEREOF
+Pi-hole Meta Package
+===================
+
+This package provides dependencies required for Pi-hole installation.
+For more information, visit https://pi-hole.net
+INNEREOF
+    chmod 644 "${pkgdir}/usr/share/doc/${pkgname}/README"
+
+    # Create marker file to track Alpine installation
+    touch "${pkgdir}/etc/pihole/alpine_install"
+    chmod 644 "${pkgdir}/etc/pihole/alpine_install"
+}
 EOM
 )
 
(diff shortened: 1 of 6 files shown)


test/_alpine_3.Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN mkdir -p $GITDIR $SCRIPTDIR /etc/pihole
+ADD . $GITDIR
+RUN cp $GITDIR/advanced/Scripts/*.sh $GITDIR/gravity.sh $GITDIR/pihole $GITDIR/automated\ install/*.sh $GITDIR/advanced/Scripts/COL_TABLE $SCRIPTDIR/
+ENV PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$SCRIPTDIR
+
+RUN true && \
+    chmod +x $SCRIPTDIR/*
+
+ENV SKIP_INSTALL=true
+ENV OS_CHECK_DOMAIN_NAME=dev-supportedos.pi-hole.net
```
darkexplosiveqwx, line 17: `ENV OS_CHECK_DOMAIN_NAME=dev-supportedos.pi-hole.net` is no longer needed, since #6206 has been merged and removed the OS check. That line can be safely removed.

darkexplosiveqwx:
Please add the test to `.github/workflows/test.yml`, so it will run via GH action and we can see the status here. (Add `alpine_3` into `distro: [ ]`). I would also recommend renaming your test files to `alpine_3_21` to differentiate the single point alpine versions. Alpine 3 was quite a while ago and alpine 4 also quite some time away. This will make it easier to add alpine 3.22 when it releases.

automated install/basic-install.sh
```diff
@@ -262,7 +312,19 @@ package_manager_detect() {
         PKG_COUNT="${PKG_MANAGER} check-update | grep -E '(.i686|.x86|.noarch|.arm|.src|.riscv64)' | wc -l || true"
         # The command we will use to remove packages (used in the uninstaller)
         PKG_REMOVE="${PKG_MANAGER} remove -y"
-    # If neither apt-get or yum/dnf package managers were found
+
+    # If rpm is not found, check for apk.
+    elif is_command apk; then
+        PKG_MANAGER="apk"
+        PKG_INSTALL="${PKG_MANAGER} add --no-cache --allow-untrusted"
```
darkexplosiveqwx, line 319: I would not recommend using `—allow-untrusted` for installing packages from the repositories.
I would recommend something similar to what I did for openSUSE 
```
# SUSE requires locally built rpm packages to be signed by default
PKG_INSTALL_LOCAL="${PKG_MANAGER} --no-gpg-checks install -y"
PKG_INSTALL="${PKG_MANAGER} install -y"
```
and only allow untrusted for the local metapackage.

automated install/basic-install.sh
```diff
@@ -353,7 +415,135 @@ build_dependency_package(){
...
+
+        # Write the APKBUILD file
+        printf "  %b Creating APKBUILD..." "${INFO}"
+        echo "${PIHOLE_META_PACKAGE_CONTROL_APK}" > /var/lib/builder/pihole-meta/APKBUILD
+        chown -R builder:abuild /var/lib/builder/pihole-meta
+        printf "%b  %b Done\\n" "${OVER}" "${TICK}"
+
+        # Install build dependencies
+        printf "  %b Installing build dependencies..." "${INFO}"
+        if ! apk add --no-cache build-base bind-tools binutils coreutils curl git grep iproute2-ss jq libcap logrotate ncurses nmap-ncat procps-ng psmisc shadow tzdata unzip wget >/dev/null 2>&1; then
```
yubiuser, line 511: Do we really need all those packages as build dependencies?

rrobgill:
> Please add the test to `.github/workflows/test.yml`, so it will run via GH action and we can see the status here.

Recommended changes to tests done. Changes to scripting to follow later in the week.

automated install/basic-install.sh
rrobgill: Thankyou, I have followed your approach to explicitly only use untrusted on the local .apk

automated install/basic-install.sh
rrobgill: I believe so. It seems comparable in number to the other package managers.

rrobgill: Ah yes. I see now.
I will have a look at thinning those out.

rrobgill:
Conflicts resolved, good to go.

mgziminsky:
I also have a custom fork adding alpine support that includes a proper openrc script, and uses a virtual package for dependencies instead of requiring the full `abuild` install and logic. Using a virtual package makes it possible to add support with significantly fewer changes and logic, the only needed adjustments are adding openrc branches to the service calls.

If it's preferred to switch to my approach, I can either make a new PR to supersede this, or feel free to adapt or use anything from my fork into this one. There are a few enhancements that can still be made to my version as well in case you decide to go that route. Let me know if you have any questions.

https://github.com/mgziminsky/pi-hole-alpine
https://github.com/mgziminsky/pi-hole-alpine/compare/pi-hole%3Api-hole%3Amaster...alpine

### Additional Notes:
The other main change in my version is changing the PID file to be managed by the service instead of by `pihole-FTL` directly. Service pidfiles should not be owned by unprivileged users with openrc as it is a security risk. My openrc script uses the same file as FTL does by default, which actually prevents FTL from starting unless its pidfile location is changed. This is probably an FLT bug since it fails even though it is being run in the foreground without a pidfile. To fix the startup issue, the service script should probably update its pidfile to be unique, such as by appending a postfix like so: `pidfile="/run/${RC_SVCNAME}_rc.pid"`. The prestart/poststop scripts can probably also be reverted and made to always pass in the service script.

rrobgill:
> I also have a custom fork adding alpine support that includes a proper openrc script, and uses a virtual package for dependencies instead of requiring the full `abuild` install and logic. Using a virtual package makes it possible to add support with significantly fewer changes and logic, the only needed adjustments are adding openrc branches to the service calls.

It's a discussion for the maintainers but after reviewing your code I certainly think your approach has merit @mgziminsky and is probably more easily maintained in the long-run. 

If you put through your PR, I'm happy to change this back to only implement changes to the tests for alpine, or you can incorporate those within your PR.

dschaper:
Two PR's enter, One PR leaves!

rrobgill:
Drafting this for now, and will close it when #6275 merges

rrobgill:
Replacement PR is well underway. Closing this.

PR #6230, opened 2025-05-12
Use shell parameter expansion to split http_code and payload for api.sh

dschaper:

Use bash shell expansion for `curl` payload splitting in to status code and payload variables.

```bash
root@pihole:~# bash c.sh
This is the payload from curl
{"session":{"valid":false,"totp":false,"sid":null,"validity":-1,"message":"no SID provided"},"took":2.1219253540039062e-05}>>401

This is the http code (remove everything up to and including the >> delim
401

This is the payload (remove the >> delim and everything after
{"session":{"valid":false,"totp":false,"sid":null,"validity":-1,"message":"no SID provided"},"took":2.1219253540039062e-05}
```

```bash
root@pihole:~# cat c.sh
b=$(curl -skS -w ">>%{http_code}" http://localhost/api/auth)

echo "This is the payload from curl"
echo ${b}

echo
echo "This is the http code (remove everything up to and including the >> delim"
echo ${b#*>>}

echo
echo "This is the payload (remove the >> delim and everything after"
echo ${b%>>*}
```

---

---


yubiuser (commented):
Nice trick. Could you please do the same at 

https://github.com/pi-hole/pi-hole/blob/1bdbc26a0722e74b404a5d9326f56ba46548a1c1/advanced/Scripts/api.sh#L214-L225

dschaper:
Will that affect the "raw" response? I don't know if there is any other code that parses the raw output and would be confused by the additional `>>` delimiter.

rdwebdesign:
> Will that affect the "raw" response?

Apparently the only place using the "raw" output of this function is line 315:
https://github.com/pi-hole/pi-hole/blob/de31858950a61a5eb90a0a1c59635b32c1dfcdc1/advanced/Scripts/api.sh#L314-L321

Lines 319 and 321 will need to be changed to understand the `>>` delimiter.

dschaper:
This is what I see searching the repo for calls to that function

image

But looking that over makes me want to change the functions. A function called `GetFTLData` should get and return the data and not data or the status. And functions consuming the output from `GetFTLData()` shouldn't also check for the http_code. 

So I'm going to address that in another PR and leave this PR as it is without expanding that to line 214 like @yubiuser has asked.

yubiuser (approved):
> So I'm going to address that in another PR and leave this PR as it is without expanding that to line 214 like @yubiuser has asked.

If you do so, then this is good to go.

PR #6236, opened 2025-05-14
Update `get_available_interfaces()` to correctly filter loopback device (`lo`)

PiotrTyrakowski:
The get_available_interfaces() helper currently does:

ip --oneline link show up | grep -v "lo" | awk '{print $2}'

it removes wlo1 from the output.
i have updated command to use 

ip --oneline link show up | awk '{print $2}' | grep -v "^lo"


rdwebdesign (approved):
Approving, since this is just a redo of https://github.com/pi-hole/pi-hole/pull/6224

dschaper (approved):

Issue #6071, opened 2025-03-08
`basic-install.sh`  :  `if is_command systemctl; then` condition misidentifies boot process - `systemd` vs `sysvinit`


davygravy:
### Versions
Various versions... present in current  master  branch

```
Core
    Version is v6.0.4-40-ge3ad058e (Latest: null)
    Branch is unsupported/binary
    Hash is e3ad058e (Latest: e3ad058e)
Web
    Version is v6.0.2 (Latest: v6.0.2)
    Branch is master
    Hash is 25441178 (Latest: 25441178)
FTL
    Version is v6.0.4 (Latest: v6.0.4)
    Branch is master
    Hash is b7eb53bf (Latest: b7eb53bf)
```

### Platform

- Debian 13 (Trixie Testing), Debian 12 (Bookworm Stable)
- this bug/shortcoming would affect any system that is using sysvinit, yet has the `systemctl` binary present on the rootfs (and this is not a far-fetched possibility)
- this does *not* have to be on an unsupported system or an unsupported architecture

### Expected behavior

- The test for boot method should correctly identify it.  
- A more accurate test method is already used in one location in the installer code near line 1283   :
```    # Install pihole-FTL systemd or init.d service, based on whether systemd is the init system or not
       # Follow debhelper logic, which checks for /run/systemd/system to derive whether systemd is the init system
               if [[ -d '/run/systemd/system' ]]; then
```
### Actual behavior / bug

The installer misidentifies the boot process :  `if [[ -d '/run/systemd/system' ]]; then`  vs `if is_command systemctl; then`  ,  error visible in the last line of output below.

```
root@debian-armel-trixie:/etc/.pihole# PIHOLE_SKIP_OS_CHECK=true  PIHOLE_SKIP_FTL_CHECK=true automated\ install/basic-install.sh 

  [✓] Root user check

        .;;,.
        .ccccc:,.
         :cccclll:.      ..,,
          :ccccclll.   ;ooodc
           'ccll:;ll .oooodc
             .;cll.;;looo:.
                 .. ','.
                .',,,,,,'.
              .',,,,,,,,,,.
            .',,,,,,,,,,,,....
          ....''',,,,,,,'.......
        .........  ....  .........
        ..........      ..........
        ..........      ..........
        .........  ....  .........
          ........,,,,,,,'......
            ....',,,,,,,,,,,,.
               .',,,,,,,,,'.
                .',,,,,,'.
                  ..'''.

  [i] SELinux not detected
[...]

PR #6245, opened 2025-05-27
Use PID1 to determine which command to use when toggeling services

yubiuser:

This is a follow up PR of https://github.com/pi-hole/pi-hole/pull/6043.

It uses the same logic to determine which process has PID1 (hence was the init system) to use the right command to toggle services. This should fix rare cases when users have more than one init system installed.

It also extends the use of service wrapper functions to all scripts we use.

Should fix https://github.com/pi-hole/pi-hole/issues/6071

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -116,11 +116,11 @@ c=70
 PIHOLE_META_PACKAGE_CONTROL_APT=$(
     cat <<EOM
 Package: pihole-meta
-Version: 0.3
+Version: 0.4
 Maintainer: Pi-hole team <adblock@pi-hole.net>
 Architecture: all
 Description: Pi-hole dependency meta package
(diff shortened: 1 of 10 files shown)


yubiuser:
I think it's only the test on the docker images because in a real system Fedora uses `systemd`.
It fails with 

`Enabling pihole-FTL service to start on reboot...", _stderr=b"Redirecting to /bin/systemctl stop pihole-FTL.service\nSystem has not been booted with systemd as init system (PID 1).`

yubiuser:
Tests pass now. It required more changes than I expected, but the good thing is we now test part of the code (mocking `systemd` as init system) we never tested before.

automated install/basic-install.sh
```diff
@@ -228,6 +231,13 @@ is_command() {
     command -v "${check_command}" >/dev/null 2>&1
 }
 
+is_pid1() {
+    # Checks to see if the given command runs as PID 1
+    local is_pid1="$1"
+
+    ps -p 1 -o comm= | grep -q "${is_pid1}"
```
dschaper, line 238:
```suggestion
    # select PID 1, format output to show only CMD column without header
    # quietly grep for a match on the function passed parameter
    ps --pid 1 --format comm= | grep -q "${is_pid1}"
```

dschaper (changes requested):

dschaper (approved):

PR #6255, opened 2025-05-31
Readonly variables should not be changed

rdwebdesign:
### The issue:

The variable was declared readonly, but without a value. A value was set on the next line, causing an error:
```
/usr/local/bin/pihole: line 393: LOGFILE: readonly variable 
```

### How does this PR fixes the above?

Declaring the variable with the value and never changing it after that.

---

---

### DIFF
--- a/pihole
+++ b/pihole
@@ -389,8 +389,7 @@ tailFunc() {
   echo -e "  ${INFO} Press Ctrl-C to exit"
 
   # Get logfile path
-  readonly LOGFILE
-  LOGFILE=$(getFTLConfigValue files.log.dnsmasq)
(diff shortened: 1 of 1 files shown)


yubiuser:
Shellcheck does not like it.

rdwebdesign:
Yeah... I think here we need to remove the `readonly`.

There is no way to declare a `readonly` variable and set its value later without throwing the error.

PromoFaux:
Apparently the workaround would be to be

```
foo=$(bar)
readonly foo
```

But I don't like that, so I would opt to just disable the warning on this line, or remove the readonlyness of it (I don't recall why it is readonly, tbh)

https://github.com/koalaman/shellcheck/issues/1336

dschaper:
I have a working PR in #6261

PR #6257, opened 2025-05-31
Source needed files directly instead of using intermediate variables and remove readonly

yubiuser:

Fixes a copy&paste error in https://github.com/pi-hole/pi-hole/pull/6245.
`PI_HOLE_FILES_DIR` was not declared in all scripts using it now. 

Reported on Discourse: https://discourse.pi-hole.net/t/pi-hole-ftl-v6-2-release-sudo-pihole-flush-command-bug/80192/

Additionally, remove some `readonly` lines which cause trouble when we we source different scripts and variables are already defined.
This should fix: https://github.com/pi-hole/pi-hole/issues/6259

---

---

### DIFF
--- a/advanced/Scripts/piholeLogFlush.sh
+++ b/advanced/Scripts/piholeLogFlush.sh
@@ -17,6 +17,7 @@ utilsfile="${PI_HOLE_SCRIPT_DIR}/utils.sh"
 # shellcheck source="./advanced/Scripts/utils.sh"
 source "${utilsfile}"
 
+readonly PI_HOLE_FILES_DIR="/etc/.pihole"
 SKIP_INSTALL="true"
 # shellcheck source="./automated install/basic-install.sh"
 source "${PI_HOLE_FILES_DIR}/automated install/basic-install.sh"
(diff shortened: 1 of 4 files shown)


rdwebdesign (approved):

dschaper:
Or just revert #6245 since it fixes an edge case and that change appears to affect everyone.

advanced/Scripts/piholeLogFlush.sh
```diff
@@ -17,6 +17,8 @@ utilsfile="${PI_HOLE_SCRIPT_DIR}/utils.sh"
 # shellcheck source="./advanced/Scripts/utils.sh"
 source "${utilsfile}"
 
+PI_HOLE_FILES_DIR="/etc/.pihole"
+readonly PI_HOLE_FILES
```
LudovicRousseau, line 21: Maybe it should be `PI_HOLE_FILES_DIR` instead of `PI_HOLE_FILES`?
the `_DIR` part is missing for the `readonly` command.

yubiuser:
Includes a fix for the readonly failure of `pihole -t`reported here: https://discourse.pi-hole.net/t/sudo-pihole-t/80288
Missed this one: https://github.com/pi-hole/pi-hole/pull/6261 will fore push away.

dschaper:
Let's just revert this for now and then approach it again later. The problem I have is that we now have hardcoded paths in there and that's going to cause issues with the proposed Alpine options that will install to other paths.

Issue #6259, opened 2025-05-31
api.sh: line 25: utilsfile: readonly variable


jacklul:
### Versions

(latest develop)
- Pi-hole:  [cad0d0bf](https://github.com/pi-hole/pi-hole/commit/cad0d0bf1fbb825587d3089c273ee1a858bf5570)
- Web:  [abd4d9d3](https://github.com/pi-hole/web/commit/abd4d9d3b9554e4f08ec9bfbf69b1927b92676f3)
- FTL:  [bc24e856](https://github.com/pi-hole/FTL/commit/bc24e856fbb806087b2093ab4a34291f7f41bda4)

### Platform
- OS and version:  N/A
- Platform:  Router `Linux RT-AX58U 4.19.183 #1 SMP PREEMPT Thu Apr 17 12:22:51 CST 2025 armv7l`
- Bash: 5.2.32

### Expected behavior

No error message is shown

### Actual behavior / bug

Executing `pihole api ...` causes this error to appear: `api.sh: line 25: utilsfile: readonly variable`
`utilsfile` variable is already defined in `pihole` script
`pihole` script sources `api.sh` and then calls functions from it, `TestAPIAvailability` function redefines this variable

### Steps to reproduce

N/A

## Debug Token

- URL:  N/A

### Screenshots

N/A

### Additional context

N/A

dschaper:
Or we declare `utilsfile` in `pihole`
https://github.com/pi-hole/pi-hole/blob/4fed49c5e5d54a19400cbb3236f2f565e2944037/pihole#L23

And then try to redeclare it in `api.sh` which is called by `pihole` and thus is already declared and `readonly` will fail.
https://github.com/pi-hole/pi-hole/blob/4fed49c5e5d54a19400cbb3236f2f565e2944037/advanced/Scripts/api.sh#L25

Seems like the `api.sh` should declare the `utilsfile` as a local variable.

jacklul:
I believe `api.sh` can be sourced in user script without sourcing anything else right now, so moving that variable to `pihole` script would break this.

I don't think declaring it as local works:

```sh
TestAPIAvailability() {

    local chaos_api_list authResponse authStatus authData apiAvailable DNSport utilsfile

    # as we are running locally, we can get the port value from FTL directly
    readonly utilsfile="${PI_HOLE_SCRIPT_DIR}/utils.sh"
    # shellcheck source=./advanced/Scripts/utils.sh
    . "${utilsfile}"
```

```bash
[...]

PR #6262, opened 2025-05-31
Revert "Use PID1 to determine which command to use when toggeling services"

dschaper:
Reverts pi-hole/pi-hole#6245

The changes are causing problems with scripts that source other scripts with masked `readonly` variables and attempts to redeclare readonly variables.

https://github.com/pi-hole/pi-hole/issues/6259

A new PR should be done for that with an eye on `local -r <VARIABLE>` in the downstream scripts that source `pihole` where variables are declared as `readonly` and then the same variable name is attempted to be used and assigned in the downstream script.

I think we'll need to change the package version to 0.5 as we can't step back to version 0.3? @yubiuser can you see if that is true?

### DIFF
--- a/advanced/Scripts/piholeARPTable.sh
+++ b/advanced/Scripts/piholeARPTable.sh
@@ -20,13 +20,6 @@ utilsfile="${PI_HOLE_SCRIPT_DIR}/utils.sh"
 # shellcheck source=./advanced/Scripts/utils.sh
 source "${utilsfile}"
 
-readonly PI_HOLE_FILES_DIR="/etc/.pihole"
-SKIP_INSTALL="true"
-# shellcheck source="./automated install/basic-install.sh"
-source "${PI_HOLE_FILES_DIR}/automated install/basic-install.sh"
-# stop_service() is defined in basic-install.sh
-# restart_service() is defined in basic-install.sh
-
 # Determine database location
 DBFILE=$(getFTLConfigValue "files.database")
 if [ -z "$DBFILE" ]; then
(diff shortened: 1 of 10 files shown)


darkexplosiveqwx:
Cant we just keep the changes to the meta package?
The alphabetical ordering makes sense, and isn't `gawk` needed anyways?

dschaper:
Yeah, that sounds logical.

yubiuser:
We cannot decrease the version of the meta-package. The installer will fail.

dschaper:
I assumed that, hence my question of increasing the version from 0.4 to 0.5.

darkexplosiveqwx's suggestion of excluding the meta package changes from the reversion seems fine. I'll push a commit to do that.

yubiuser:
I propose a different fix: https://github.com/pi-hole/pi-hole/pull/6257

We don't need the intermediate variables. We just use each once in the script, we can source directly.

dschaper:
I've left a comment on the proposed PR but to echo them here, that proposed PR adds in hardcoded paths and I'm not sure that will work with the discussion on Alpine direct installs that will use other paths. 

I'd rather just reset things back to where the were in 6.0 and then look again at this edge case without the rush of having to get a bugfix out now.

yubiuser:
> I'm not sure that will work with the discussion on Alpine direct installs that will use other paths.

I did not think about the changes the Alpine package needs to do. 

Ok, revert this for now without the meta-package changes and I'll retry later.

dschaper:
Thanks, I get that the core code is very complex. Throw in `bash` and it's very unique behavior and things that make sense logically don't always end up doing what you meant to do. Plus it's really hard to try to mock init systems with our 10 year old build and test code.

PR #6274, opened 2025-06-02
do not refresh package cache during update script

mwoolweaver:
package cache will be updated when necessary after update script because install script is run after core OR FTL updates

potentially addresses #6272

## Thank you for your contribution to the Pi-hole Community!

Please read the comments below to help us consider your Pull Request.

We are all volunteers and completing the process outlined will help us review your commits quicker.

**Please make sure you**

 1. Base your code and PRs against the repositories developmental branch.
 2. [Sign Off](https://docs.pi-hole.net/guides/github/how-to-signoff/) all commits as we enforce the [DCO](https://docs.pi-hole.net/guides/github/dco/) for all contributions
 3. [Sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) all your commits as they must have verified signatures
 4. File a pull request for any change that requires changes to [our documentation](https://docs.pi-hole.net/) at our [documentation repo](https://github.com/pi-hole/docs) 

---

**What does this PR aim to accomplish?:**

**How does this PR accomplish the above?:**

**Link documentation PRs if any are needed to support this PR:**

---
**By submitting this pull request, I confirm the following:**

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against.
2. I have commented my proposed changes within the code and I have tested my changes.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
6. I have checked that another pull request for this purpose does not exist.
7. I have considered, and confirmed that this submission will be valuable to others.
8. I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
9. I give this submission freely, and claim no ownership to its content.

---
- [✓] I have read the above and my PR is ready for review. *Check this box to confirm*

PR #6282, opened 2025-06-05
Only update the package cache on fresh installations

yubiuser:

Only update the package cache once on fresh installs, not on checkouts/updates. Addresses https://github.com/pi-hole/pi-hole/discussions/6272

---

---

### DIFF
--- a/advanced/Scripts/update.sh
+++ b/advanced/Scripts/update.sh
@@ -114,6 +114,7 @@ main() {
 
 
     # Install packages used by this installation script (necessary if users have removed e.g. git from their systems)
+    check_fresh_install
     package_manager_detect
     build_dependency_package
     install_dependent_packages
--- a/advanced/Scripts/update.sh
+++ b/advanced/Scripts/update.sh
@@ -114,7 +114,6 @@ main() {
 
 
     # Install packages used by this installation script (necessary if users have removed e.g. git from their systems)
-    check_fresh_install
     package_manager_detect
     build_dependency_package
     install_dependent_packages
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -254,10 +254,6 @@ package_manager_detect() {
         PKG_COUNT="${PKG_MANAGER} -s -o Debug::NoLocking=true upgrade | grep -c ^Inst || true"
         # The command we will use to remove packages (used in the uninstaller)
         PKG_REMOVE="${PKG_MANAGER} -y remove --purge"
-        # Update package cache only on fresh installs
-        if [[ "${fresh_install}" == true ]]; then
-            update_package_cache || exit 1
-        fi
 
     # If apt-get is not found, check for rpm.
     elif is_command rpm; then
@@ -2207,6 +2203,11 @@ main() {
 
     # Check for supported package managers so that we may install dependencies
     package_manager_detect
+
+    # Update package cache only on fresh installs and apt based systems
+    if [[ "${fresh_install}" == true ]] && is_command apt-get; then
+            update_package_cache || exit 1
+    fi
 
     # Notify user of package availability
     notify_package_updates_available
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -2204,8 +2204,8 @@ main() {
     # Check for supported package managers so that we may install dependencies
     package_manager_detect
 
-    # Update package cache only on fresh installs and apt based systems
-    if [[ "${fresh_install}" == true ]] && is_command apt-get; then
+    # Update package cache only on apt based systems
+    if is_command apt-get; then
             update_package_cache || exit 1
     fi
 
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -231,6 +231,13 @@ is_command() {
     command -v "${check_command}" >/dev/null 2>&1
 }
 
+check_fresh_install() {
+    # in case of an update (can be a v5 -> v6 or v6 -> v6 update) or repair
+    if [[ -f "${PI_HOLE_V6_CONFIG}" ]] || [[ -f "/etc/pihole/setupVars.conf" ]]; then
+        fresh_install=false
+    fi
+}
+
 # Compatibility
 package_manager_detect() {
 
@@ -247,8 +254,10 @@ package_manager_detect() {
         PKG_COUNT="${PKG_MANAGER} -s -o Debug::NoLocking=true upgrade | grep -c ^Inst || true"
         # The command we will use to remove packages (used in the uninstaller)
         PKG_REMOVE="${PKG_MANAGER} -y remove --purge"
-        # Update package cache
-        update_package_cache || exit 1
+        # Update package cache only on fresh installs
+        if [[ "${fresh_install}" == true ]]; then
+            update_package_cache || exit 1
+        fi
 
     # If apt-get is not found, check for rpm.
     elif is_command rpm; then
@@ -2193,6 +2202,9 @@ main() {
     # Check for availability of either the "service" or "systemctl" commands
     check_service_command
 
+    # Check if this is a fresh install or an update/repair
+    check_fresh_install
+
     # Check for supported package managers so that we may install dependencies
     package_manager_detect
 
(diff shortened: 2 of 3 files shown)


yubiuser:
The seconds commit should fix the tests. Not sure why we had `fresh_install=false` in a test that should test a fresh installation.

darkexplosiveqwx:
I think we should still update the package cache when the meta package is updated, so all new dependencies are up to date.

yubiuser:
This would be possible too. We would only need to signal `package_manager_detect` that we are invoking from the update script or not.

dschaper:
How does this compare to #6274 ?

yubiuser:
In the current implementation or the one suggestion by @darkexplosiveqwx ?

https://github.com/pi-hole/pi-hole/pull/6274 removes also the building and installing of the dependency package before doing an update - which might fails because some 'installer dependency' is not installed.

dschaper:
Ah, okay. Maybe something like `apt cache policy <dependency package> | grep version` and compare that to the `/etc/.pihole/ <package metadata version>` and "do things" if the proposed updated metadata is higher than the installed package version?

PromoFaux:
`fresh_install` was introduced as a replacement for `useUpdateVars`, but the logic was "reversed" to make the names make sense.

> Replaces the variable `useUpdateVars` with `fresh_install` and flips the logic around. `useUpdateVars` used to make sense as a name when a user could seed a setupVars.conf file, but this is no longer the case and the naming is not so clear.

So I just changed them in the tests, too.

mwoolweaver:
> Only update the package cache once on fresh installs, not on checkouts/updates.

Package cache should also be updated when `pihole-meta` package is being upgraded, right?

mwoolweaver:
could you use the following to get version number of new meta package after being it's created. This command would keep from need another variable for the meta package version number. 

```
dpkg-deb -f pihole-meta.deb Version
```

and then compare that to the the installed version found with

```
dpkg -s pihole-meta | grep '^Version:' | cut -d' ' -f2
```

if they are equal then the package cache update can be skipped as there is no way the meta package would be installed if any of its depends aren't installed.

yubiuser:
It was added in the first place with this commit
https://github.com/pi-hole/pi-hole/commit/cedd1a259105dc231744d58562f0bc80d4294c7a

Back then, the test added configuration to setupVars.conf to automatically install the web interace and the web server. This is no longer needed.

https://github.com/pi-hole/pi-hole/blob/cedd1a259105dc231744d58562f0bc80d4294c7a/test/test_automated_install.py#L198-L214

mwoolweaver:
Would ~~#6292~~ #6298 be a more graceful solution?

yubiuser:
I'm not sure if this is the way to go. While it addresses the suggestions from above, it adds 100 lines of code. I'm not sure if this is worth it, considering 'normal' users will only benefit from it during an update which is not happening that often.

___

I would go another way as a middle ground: only update the cache before we install the meta package initiated by the install script, but not from the updater. 

___

**ADD**
I only noticed now, that we do the cache update only on `apt` based systems, reducing the necessarity to do it at all even more from my view.

yubiuser:
The last two changes allow to update the cache when building/installing the meta package from the installer script, but don't do when building/installing from the update script. 
I intentionally did not squash to allow easy reverting to a previous state depending on how we decide the functions should  work.

mwoolweaver:
> While it addresses the suggestions from above, it adds 100 lines of code.

But it saves about 160 lines with each `pihole -up` that contains no meta data changes and saves SD activity caused by installing the same version of the package again. (when no Pi-hole updates available)

If there is an update to core repo with NO meta packages changes or just an FTL update it would save double the lines and double the SD activity. Currently all package management steps are done twice is these two scenarios, once at the beginning of `update.sh` and again during `basic-install.sh`. see [here](https://github.com/pi-hole/pi-hole/pull/6298#issuecomment-2959931810)

and would also ensure that

> we had users who removed e.g. git which were left with a broken update.[1](https://github.com/pi-hole/pi-hole/pull/6274#issuecomment-2943674650)

does not happen as well.

mwoolweaver:
#6292 also has the added benefit of negating the issue mentioned [here](https://github.com/pi-hole/pi-hole/pull/6262#issuecomment-2925435842)

dschaper:
I'd like to go for this PR for now as it improves the situation. The massive change in #6298 needs some very thorough review and testing and that's going to take time. 

Considering this is a QOL fix and not a bug fix, I'm good with what Yubi has written.

dschaper:
@mwoolweaver:

> But it saves about 160 lines with each `pihole -up` that contains no meta data changes

Where do you see these 160 lines?

This is the output that I see with this branch code:

```
root@pihole:/etc/.pihole# pihole -up
  [✓] Building dependency package pihole-meta.deb
  [✓] Installing Pi-hole dependency package

  [i] Checking for updates...
  [i] Pi-hole Core:     up to date
  [i] Web Interface:    up to date
  [i] FTL:              up to date

  [✓] Everything is up to date!
```

I think I like the idea of always creating/updating the dependency package. It gives us a guarantee that the update will function because it will add the packages that we know we need to run the update. This accounts for the case where a user removes a package that we need from the system, and it's really a light process to run. I don't see it generating a lot of writes to any drive or storage media.



Here's a verbose run of the update code:
```bash
+ build_dependency_package
+ rm -rf '/tmp/pihole-meta_*'
+ local tempdir
++ mktemp --directory /tmp/pihole-meta_XXXXX
+ tempdir=/tmp/pihole-meta_JaZox
+ chmod 0755 /tmp/pihole-meta_JaZox
+ is_command apt-get
+ local check_command=apt-get
+ command -v apt-get
+ pushd /tmp
+ rm -f /tmp/pihole-meta.deb
+ mkdir -p /tmp/pihole-meta_JaZox/DEBIAN
+ chmod 0755 /tmp/pihole-meta_JaZox/DEBIAN
+ touch /tmp/pihole-meta_JaZox/DEBIAN/control
+ echo 'Package: pihole-meta
Version: 0.4
Architecture: all
Depends: awk,bash-completion,binutils,ca-certificates,cron|cron-daemon,curl,dialog,dnsutils,dns-root-data,git,grep,iproute2,iputils-ping,jq,libcap2,libcap2-bin,lshw,netcat-openbsd,procps,psmisc,sudo,unzip
Section: contrib/metapackages
Priority: optional'
+ local 'str=Building dependency package pihole-meta.deb'
+ printf '  %b %s...' '[i]' 'Building dependency package pihole-meta.deb'
  [i] Building dependency package pihole-meta.deb...+ dpkg-deb --build --root-owner-group /tmp/pihole-meta_JaZox pihole-meta.deb
+ printf '%b  %b %s\n' '\r' '[✓]' 'Building dependency package pihole-meta.deb'
  [✓] Building dependency package pihole-meta.deb
+ popd
+ rm -rf /tmp/pihole-meta_JaZox
+ install_dependent_packages
+ local 'str=Installing Pi-hole dependency package'
+ printf '  %b %s...' '[i]' 'Installing Pi-hole dependency package'
  [i] Installing Pi-hole dependency package...+ is_command apt-get
+ local check_command=apt-get
+ command -v apt-get
+ '[' -f /tmp/pihole-meta.deb ']'
+ eval 'apt-get -qq --no-install-recommends install' /tmp/pihole-meta.deb
+ printf '%b  %b %s\n' '\r' '[✓]' 'Installing Pi-hole dependency package'
  [✓] Installing Pi-hole dependency package
+ rm /tmp/pihole-meta.deb
+ printf '\n'
```

mwoolweaver:
@dschaper 

> Where do you see these 160 lines?

the update script calls these 3 functions when installing the meta package

https://github.com/pi-hole/pi-hole/blob/cef7fd4b02c9ae4314e7478c0d542379bfce7f65/advanced/Scripts/update.sh#L116-L119

`package_manager_detect()` is ~38 lines

`build_dependency_package()` is ~92 lines

`install_dependent_packages()` is ~46 lines

so total skipped is 176 lines.

This also doesn't count the calls to other programs used to build the package and install it.

see the `time` test I posted [here](https://github.com/pi-hole/pi-hole/pull/6298#issuecomment-2959931810) for example of time (and likely resources) saved.

Please try #6383 (_this is basically #6298 after following suggestions made by @yubiuser_)

-----

> I think I like the idea of always creating/updating the dependency package.

in an earlier comment you mentioned the following

> >  Maybe something like apt cache policy <dependency package> | grep version and compare that to the /etc/.pihole/ <package metadata version> and "do things" if the proposed updated metadata is higher than the installed package version?
>  >
>  > _Originally posted by dschaper in https://github.com/pi-hole/pi-hole/issues/6282#issuecomment-2944867806_

#6383 accomplishes this

Issue #6281, opened 2025-06-04
"pihole -q anything.com" brings up an error message


ksnortum:
### Versions

- Pi-hole:  Core v6.1.1
- Web:  v6.2.1
- FTL:  V6.2.1

### Platform

- OS and version:  Ubuntu 24.04.2
- Platform:  Intel NUC

### Expected behavior

Typing `pihole -q anything.com` should bring up a list query.

### Actual behavior / bug

This message is displayed:

`/opt/pihole/query.sh: 27: .: cannot open /utils.sh: No such file`

### Steps to reproduce

1. At the command line, type `pihole -q anything.com`
2. An error message is displayed

### Additional context

A solution was provided in [this discourse message](https://discourse.pi-hole.net/t/pihole-q-anything-com-produces-error-message/80364/2)

thomasmerz:
Unfortunately this is still not fixed in Docker Tag [2025.06.1](https://github.com/pi-hole/docker-pi-hole/releases/2025.06.1) Core [v6.1.2](https://github.com/pi-hole/pi-hole/releases/v6.1.2) FTL [v6.2.2](https://github.com/pi-hole/FTL/releases/v6.2.2) Web interface [v6.2.1](https://github.com/pi-hole/web/releases/v6.2.1) because https://github.com/pi-hole/pi-hole/pull/6284 is still open…

vsenn:
Got same error on latest Docker 2025.06.2 version
`docker exec pihole pihole -q ads.facebook.com`
`/opt/pihole/query.sh: .: line 27: can't open '/utils.sh': No such file or directory`

thomasmerz:
@yubiuser , can you please close this as this has been already fixed? I can confirm:
```
# pihole -q ads.facebook.com
Found 0 domains exactly matching 'ads.facebook.com'.

Found 4 adlists exactly matching 'ads.facebook.com'.

  - https://big.oisd.nl/ (block)

    - ||ads.facebook.com^


  - https://raw.githubusercontent.com/RPiList/specials/master/Blocklisten/child-protection (block)

    - ||ads.facebook.com^


  - https://raw.githubusercontent.com/RPiList/specials/master/Blocklisten/spam.mails (block)

    - ||ads.facebook.com^


  - https://raw.githubusercontent.com/notracking/hosts-blocklists/master/hostnames.txt (block)

    - ads.facebook.com
    - ads.facebook.com
```

yubiuser:
[...]

PR #6284, opened 2025-06-05
Fix pihole -q by setting PI_HOLE_SCRIPT_DIR in api.sh

yubiuser:

Fixes https://github.com/pi-hole/pi-hole/issues/6281 by setting  `PI_HOLE_SCRIPT_DIR` in api.sh

---

---

### DIFF
--- a/advanced/Scripts/api.sh
+++ b/advanced/Scripts/api.sh
@@ -22,7 +22,8 @@ TestAPIAvailability() {
     local chaos_api_list authResponse authStatus authData apiAvailable DNSport
 
     # as we are running locally, we can get the port value from FTL directly
-    readonly utilsfile="${PI_HOLE_SCRIPT_DIR}/utils.sh"
+    PI_HOLE_SCRIPT_DIR="/opt/pihole"
+    utilsfile="${PI_HOLE_SCRIPT_DIR}/utils.sh"
     # shellcheck source=./advanced/Scripts/utils.sh
     . "${utilsfile}"
 


dschaper:
This does fix the issue but I'd rather go the route of exporting the variable from `pihole`. We never call `api.sh` directly, it's always called from the parent `pihole` command where the variable is already declared. By duplicating the variable here we run the very real case of updating `PI_HOLE_SCRIPT_DIR` in one file and then missing it in other places. 

My preferred solution would be:

1. `export` `PI_HOLE_SCRIPT_DIR` in `pihole`.
2. Add comment in `api.sh` that `PI_HOLE_SCRIPT_DIR` is being used from the `pihole` script's export.

Or if you want to really isolate `api.sh` then just get rid of the variable. There's no need to create `PI_HOLE_SCRIPT_DIR` as a variable to only be used once in the next line. Using [bash shell parameter expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) we can accept the `export`ed variable from `pihole` or set the path to a fallback if that variable does not exist in the environment. 

`utilsfile="${PI_HOLE_SCRIPT_DIR:-/opt/pihole}/utils.sh"`

yubiuser:
We use `api.sh` in `query.sh` and `list.sh` as well. When we start using `export` from, do you want me to change the lines there as well?

https://github.com/pi-hole/pi-hole/blob/ec892ec096cd1adb90b0209d4c0a155953faf813/advanced/Scripts/list.sh#L12-L19

https://github.com/pi-hole/pi-hole/blob/ec892ec096cd1adb90b0209d4c0a155953faf813/advanced/Scripts/query.sh#L12-L25

dschaper:
Yeah, this is a mess, things changed when we started sourcing scripts inside of other scripts instead of just having one step from `pihole` to the other scripts. Combine that with the fact that it's all shellcode and it's a perfect storm.

I guess we can go with this as is to at least fix the issue but the bigger picture needs to put all of the re-used environment variables in one place (`utils.sh`?) and exported. That means that `pihole` lin 12 (PI_HOLE_SCRIPT_DIR) will need to follow the same advice as the comments for line 17 (PI_HOLE_BIN_DIR):

```bash
readonly PI_HOLE_SCRIPT_DIR="/opt/pihole"

# PI_HOLE_BIN_DIR is not readonly here because in some functions (checkout),
# they might get set again when the installer is sourced. This causes an
# error due to modifying a readonly variable.
```

Or use the fact that the file is `bash` and not `sh` like the scripts and set line 12 there to something like `PI_HOLE_SCRIPT_DIR="${PI_HOLE_SCRIPT_DIR:=/opt/pihole}"`

yubiuser:
> I guess we can go with this as is to at least fix the issue but the bigger picture needs to put all of the re-used environment variables in one place (utils.sh?) and exported.

I think this won't work as we set the dirs in the installer also and some scripts source the installer functions. Then we would source them from `utils.sh` and the installer again (and the repair function of the installer will later re-source `utils.sh`). It's a mess...

PR #6292, opened 2025-06-07
Add dependency check function to avoid installing the same version of pihole-meta package repeatedly 

mwoolweaver:

only installs `pihole-meta` package when the version number has changed or when not installed

potentially addresses #6272 and will 

Lessen the resource impact mentioned here

> running a package cache update is a resource intensive process on SD cards.

adds function `check_for_meta_package()` (name can be changed if needed/wanted) that will check if `pihole-meta` package is installed and only runs package management functions if `pihole-meta` package is missing or version number has changed.

also adds `VersionConverter()` from [PADD](https://github.com/pi-hole/PADD/blob/master/padd.sh#L1357) to make version number comparison easier.

---

---

### DIFF
--- a/advanced/Scripts/update.sh
+++ b/advanced/Scripts/update.sh
@@ -27,6 +27,7 @@ source "/opt/pihole/COL_TABLE"
 # shellcheck source="./advanced/Scripts/utils.sh"
 source "${PI_HOLE_INSTALL_DIR}/utils.sh"
 
+# check_for_meta_package() sourced from basic-install.sh
 # is_repo() sourced from basic-install.sh
 # make_repo() sourced from basic-install.sh
 # update_repo() source from basic-install.sh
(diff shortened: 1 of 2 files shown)


advanced/Scripts/update.sh
```diff
@@ -112,7 +112,7 @@ main() {
     web_update=false
     FTL_update=false
 
-
+    IS_UPDATE=true
```
mwoolweaver, line 115: i was attempting to use the `SKIP_INSTALL` variable but it was causing the CentOS tests to fail so I added `IS_UPDATE`.

the idea for this variable is to use it as signal that we are sourcing this function from the update script, which gives an easy dependency check without requiring the reinstall of the meta package every time `pihole -up` is used. 


if there is an update for the `core` repo OR for `FTL` then `basic-install.sh` is started and the variable has no impact.

mwoolweaver:
closed in favor of #6298

PR #6303, opened 2025-06-16
piholeDebug - Get default route robustly

rrobgill:

Get information about the default route in a more reliable way.

Problem with identifying this information noted on discourse and at bug https://github.com/pi-hole/FTL/issues/2525

Determine address and interface of default route by preceding 'via' and 'dev' fields in ip's json output, instead of field position in the plain text which is not always reliable.

Logs if unable to determine default gateway

N/A

---

---

### DIFF
--- a/advanced/Scripts/piholeDebug.sh
+++ b/advanced/Scripts/piholeDebug.sh
@@ -497,16 +497,25 @@ ping_gateway() {
     ping_ipv4_or_ipv6 "${protocol}"
     # Check if we are using IPv4 or IPv6
     # Find the default gateways using IPv4 or IPv6
-    local gateway gateway_addr gateway_iface
+    local gateway gateway_addr gateway_iface default_route
 
     log_write "${INFO} Default IPv${protocol} gateway(s):"
 
-    while IFS= read -r gateway; do
-        log_write "     $(cut -d ' ' -f 3 <<< "${gateway}")%$(cut -d ' ' -f 5 <<< "${gateway}")"
-    done < <(ip -"${protocol}" route | grep default)
+    while IFS= read -r default_route; do
+        gateway_addr=$(jq -r '.gateway' <<< "${default_route}")
+        gateway_iface=$(jq -r '.dev' <<< "${default_route}")
+        log_write "     ${gateway_addr}%${gateway_iface}"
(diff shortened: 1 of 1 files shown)


yubiuser (commented):
Thanks for your PR.
Regex always give me a headache - whats about a more human-friendly approach using `jp -j`?

```
nanopi@nanopi:~$ ip -4 -j route show default
[{"dst":"default","gateway":"10.0.1.1","dev":"end0","protocol":"static","metric":100,"flags":[]}]

nanopi@nanopi:~$ ip -j -4 route show default | jq -r '.[0].dev'
end0
nanopi@nanopi:~$ ip -j -4 route show default | jq -r '.[0].gateway'
10.0.1.1
```

It's slower than the `sed` but I think this does not matter in the debug script

rrobgill:
> Regex always give me a headache - whats about a more human-friendly approach using `ip -j`?

I must say I like the explicit "null" from jq if there is nothing there.

I'll update with jq instead of sed.

rdwebdesign:
The `jq` solution will fail inside our alpine docker containers.

The `ip` command inside the container is not the same from `iproute2` package, it is a `/bin/busybox` version. This version doesn't have the `-j` option.

Alternatively, we could install `iproute2` in the container and use the `jq` solution, but this is a discussion for docker-pi-hole repo.

rrobgill:
> The `jq` solution will fail inside our alpine docker containers.

That's not ideal.

Will change it back to `sed` for now, at least it is always available and consistent.

dschaper:
The `jq` calls may fail in the docker alpine images but we don't really need to know any of the network related stuff from docker. That is all handled externally by the docker engine. There are some modifications of the debug process for docker and I think just using the output if it's there and if that fails because of alpine then echo something to the output that says "busybox found, docker managed networking". Could probably check if `ip` is a symlink to `busybox` to check if this is a `busybox` based OS.

rrobgill:
`ip` is now provided by iproute2 in development (https://github.com/pi-hole/docker-pi-hole/pull/1854), and `jq` (full version, not busybox) was already in the Dockerfile.

dschaper:
Great!

dschaper (approved):

Issue #6149, opened 2025-04-04
Log Flush not fully completed on Docker


PromoFaux:
Noticed while testing #6148 (can't believe I've never run pihole -f on docker before now...)

This section of code is "problematic" on Docker for a couple of reasons:

https://github.com/pi-hole/pi-hole/blob/0f7803b7753b581ed747eb6398be0c78dbfdc845/advanced/Scripts/piholeLogFlush.sh#L110-L117

1) We have no `service` command
2) If we stopped FTL then the whole container would die.

@DL6ER Is there some way that we can move this functionality inside FTL? Like a message or something to say "Oh hey, stop writing for a sec, and delete a bunch of data first... thanks!")

3ch0less:
Confirmed this bug on Docker. Running `pihole -f` inside a container prints:

```
/opt/pihole/piholeLogFlush.sh: line 107: service: command not found
/opt/pihole/piholeLogFlush.sh: line 113: service: command not found
```

The `service pihole-FTL stop` call fails because there is no `service` command on the Docker image. FTL doesn't actually stop, so the DELETE runs while FTL may still be writing to the DB.

I also tested the API approach from #6350. `PostFTLData "action/flush/logs"` works fine and returned `{"status":"success","took":10.1}`.

@yubiuser mentioned making this universal (not Docker-only) and @dschaper wanted to wait on the api.sh to utils.sh migration. Has there been any progress on that? If not, I can work on a PR that replaces the service stop/restart pattern with the API call across the board.

PR #6319, opened 2025-06-29
Tweak ARP flushing function

yubiuser:
**What does this PR aim to accomplish?:**

Follows discussion from https://github.com/pi-hole/FTL/pull/2541

This PR updates the function flushing the network tables and the ARP table. It separates the two, renaming the whole function to `networkflush` and makes flushing ARP optional by appending `--arp`. Deletion of the network table is now done via FTL's `/action/flush/arp` endpoint.

Documentation: https://github.com/pi-hole/docs/pull/1253

---
**By submitting this pull request, I confirm the following:**

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against.
2. I have commented my proposed changes within the code and I have tested my changes.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
6. I have checked that another pull request for this purpose does not exist.
7. I have considered, and confirmed that this submission will be valuable to others.
8. I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
9. I give this submission freely, and claim no ownership to its content.

---
- [x] I have read the above and my PR is ready for review. *Check this box to confirm*

PR #6350, opened 2025-07-16
pihole -f : use API log flushing when running in docker

rrobgill:

Allow pihole -f to flush the logs when running on docker.

At the moment,as described in #6149, it tries to shut down the pihole-FTL service which is not possible within the docker image, where it is not run as a service.

Tests for the presence of a pihole docker tag, and if found it uses the API to flush the FTL database.

api.sh is added as source for the API functions.

This needs destructive API access and if attempted without will report an error to the user, (also if unable to authenticate via API for any reason)

N/A

---

---

### DIFF
--- a/advanced/Scripts/piholeLogFlush.sh
+++ b/advanced/Scripts/piholeLogFlush.sh
@@ -17,6 +17,10 @@ utilsfile="${PI_HOLE_SCRIPT_DIR}/utils.sh"
 # shellcheck source="./advanced/Scripts/utils.sh"
 source "${utilsfile}"
 
+apifile="${PI_HOLE_SCRIPT_DIR}/api.sh"
+# shellcheck source="./advanced/Scripts/api.sh"
+source "${apifile}"
+
 # In case we're running at the same time as a system logrotate, use a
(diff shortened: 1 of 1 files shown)


yubiuser:
We should wait/combine this PR with https://github.com/pi-hole/pi-hole/pull/6319

yubiuser:
I think we should always use the API to flush the logs, not only on docker. We've started to migrate every `pihole XXX` function to use the API, we should also do it for the flush function.

rrobgill:
I will have a look at that then and reopen when its done.

dschaper:
I concur. I think there should be one approach to a task and making everything go through the API is the sensible way, to me.

I also have been discussing some of the internals with the team and I'd like to get rid of the dedicated `api.sh` file and move its functions over to `utils.sh`. This is so that we don't have situations where you source one file, or both files or the wrong file and the EnvVars get assigned incorrectly. 

I don't know how urgent this fix is needed but if it's a lower priority then it would be good to wait until that migration of `api.sh` over to `utils.sh` is done.

PR #6351, opened 2025-07-16
Use true/false instead of 0/1 to represent bool

yubiuser:

Using `0/1` instead of `true/false` always makes my head hurt. I'll first translate `0/1` to `true/false` and then do the compare. This PR remove one mental step.

---

---

### DIFF
--- a/pihole
+++ b/pihole
@@ -542,7 +542,7 @@ if [[ $# = 0 ]]; then
 fi
 
 # functions that do not require sudo power
-need_root=1
+need_root=
 case "${1}" in
   "-h" | "help" | "--help"        ) helpFunc;;
   "-v" | "version"                ) versionFunc;;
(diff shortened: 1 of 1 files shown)


yubiuser:
I like your idea of checking for being set/unset. Makes it much clearer.

pihole
```diff
@@ -572,8 +572,8 @@ fi
 
 # Check if the current user is not root and if the command
 # requires root. If so, exit with an error message.
-if [[ $EUID -ne 0 && need_root -eq 1 ]];then
-  echo -e "  ${CROSS} The Pi-hole command requires root privileges, try:"
+if [[ $EUID -ne 0 && ${need_root} ]];then
```
dschaper, line 575: This does work but I think we should be explicit with what we are `test`ing for, so `test` for `need_root` to be set/non-empty
```suggestion
if [[ $EUID -ne 0 && -n "${need_root}" ]];then
```

dschaper (changes requested):

dschaper (dismissed):

dschaper:
Looks good. Can you squash the first two commits in to one? Or all three in to one if you'd like.

dschaper (approved):

PromoFaux:
This PR probably needs a new title, as it no longer matches the contents

rdwebdesign:
Maybe it should use the commit text or something like this: _"Invert need_root logic and check if set/unset"_

dschaper:
The title was changed in the merge commit. I think that will be the title that is shown in the release notes?
https://github.com/pi-hole/pi-hole/commit/a9680db21827b964551e3aaaea2ae13f12a4d3e5

dschaper:
image

dschaper:
Or maybe not. I thought changing the merge commit titles would be enough but I'll go back through and change the PR titles to match what the merge commit titles show.

image

Issue #6363, opened 2025-07-25
Output error on Fresh install


PromoFaux:
### Versions
visible in Current master and dev
```
  [✓] Detected x86_64 architecture
  [i] Downloading and Installing FTL...transferred... Failed to stop pihole-FTL.service: Unit pihole-FTL.service not loaded.
  [✓] Downloading and Installing FTL
  [✓] Installing scripts from /etc/.pihole
```

### Expected behavior

No error

### Actual behavior / bug

On a fresh install of Pi-hole the above error is presented during the install. It appears to be cosmetic, as the service is running at the end of the installer.

yubiuser:
Should be fixed with https://github.com/pi-hole/pi-hole/releases/tag/v6.2

PR #6364, opened 2025-07-25
Redirect warning when attempting to stop FTL service on a system where it doesn't yet exist

PromoFaux:

Fixes https://github.com/pi-hole/pi-hole/issues/6363

```
root@ubuntu-s-1vcpu-1gb-lon1-01:/# service somethingthatdoesntexist stop >/dev/null
Failed to stop somethingthatdoesntexist.service: Unit somethingthatdoesntexist.service not loaded.

root@ubuntu-s-1vcpu-1gb-lon1-01:/# service somethingthatdoesntexist stop &>/dev/null
root@ubuntu-s-1vcpu-1gb-lon1-01:/#
```

---

---


yubiuser (commented):
This was explicitly changed in https://github.com/pi-hole/pi-hole/pull/5983

___

I think the proper solution would be to check if the service is running/loaded and only in this case try to stop it.

PromoFaux:
Ah, interesting, I was not aware of that change (probably should have looked at the git blame for that line!)

I've pushed a different solution

automated install/basic-install.sh
```diff
@@ -1794,8 +1794,12 @@ FTLinstall() {
             # Before stopping FTL, we download the macvendor database
             curl -sSL "https://ftl.pi-hole.net/macvendor.db" -o "${PI_HOLE_CONFIG_DIR}/macvendor.db" || true
 
-            # Stop pihole-FTL service if available
-            stop_service pihole-FTL >/dev/null
+
+            # If the binary already exists in /usr/bin, then we need to stop the service
+            # If the binary does not exist, then we can skip this step.
```
yubiuser, line 1799:
```suggestion
            # If the binary does not exist (fresh installs), then we can skip this step.
```

yubiuser (commented):
I think this is an acceptable workaround ;-)

yubiuser (approved):

PR #6367, opened 2025-07-26
Revert "pihole api - use keepalive for curl queries"

PromoFaux:
Reverts pi-hole/pi-hole#6365

Labelled both as `internal` so that the automatic release note generator doesn't pick them up.

Looking deeper into it (after some internal discussion with @DL6ER) adding the `keep-alive` header does not actually do anything.

From [here](https://reqbin.com/req/c-jjwwaevv/curl-command-with-keep-alive-connection-header#:~:text=HTTP%20versions%20support%20persistent%20connections):

> HTTP/1.1 considers all connections persistent unless otherwise specified and does not require additional -H "Connection: keep-alive" headers.

In fact:

```
curl -I http://127.0.0.1/api/auth
HTTP/1.1 404 Not Found
Cache-Control: no-cache, no-store, must-revalidate, private, max-age=0
Expires: 0
Pragma: no-cache
Content-Security-Policy: default-src 'self' 'unsafe-inline';
X-Frame-Options: ALLOW-FROM abc.com
X-XSS-Protection: 0
X-Content-Type-Options: nosniff
Referrer-Policy: strict-origin-when-cross-origin
Access-Control-Allow-Credentials: true
Content-Type: application/json; charset=utf-8
Content-Length: 100
Date: Sat, 26 Jul 2025 10:09:00 GMT
Connection: keep-alive
```

Note that 1.1 is default, and the existence of `Connection: keep-alive` here in the response.

### DIFF
--- a/advanced/Scripts/api.sh
+++ b/advanced/Scripts/api.sh
@@ -55,7 +55,7 @@ TestAPIAvailability() {
         API_URL="${API_URL#\"}"
 
         # Test if the API is available at this URL, include delimiter for ease in splitting payload
-        authResponse=$(curl --connect-timeout 2 -skS -H "Connection: keep-alive" -w ">>%{http_code}" "${API_URL}auth")
+        authResponse=$(curl --connect-timeout 2 -skS -w ">>%{http_code}" "${API_URL}auth")
 
         # authStatus is the response http_code, eg. 200, 401.
         # Shell parameter expansion, remove everything up to and including the >> delim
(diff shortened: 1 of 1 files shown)


DL6ER (approved):
Yes, sending the header is entirely redundant as `curl` uses HTTP/1.1 where it is implicitly enabled just as @PromoFaux quoted.

A different test could, e.g., be
```
$ curl -I -XGET http://pi.hole/api/docs/

HTTP/1.1 200 OK                 <------------------------------------------------------------
Cache-Control: no-cache, no-store, must-revalidate, private, max-age=0
Expires: 0
Pragma: no-cache
Content-Security-Policy: default-src 'self' 'unsafe-inline';
X-Frame-Options: ALLOW-FROM abc.com
X-XSS-Protection: 0
X-Content-Type-Options: nosniff
Referrer-Policy: strict-origin-when-cross-origin
Access-Control-Allow-Credentials: true
Access-Control-Allow-Headers: *
Access-Control-Allow-Methods: *
Content-Type: text/html
Content-Length: 2786
Date: Sat, 26 Jul 2025 17:27:28 GMT
Connection: keep-alive          <------------------------------------------------------------
```

`curl` does HTTP/1.1 by default since the beginning (around 1997) so it is safe to assume *every* available version of `curl` does it.

PR #6368, opened 2025-07-26
Do not call pihole-FTL --config interactively

DL6ER:

Use non-interactive shell in `utils.sh:getFTLConfigValue` to avoid colored output. This became apparent while working on https://github.com/pi-hole/FTL/pull/2584

---

---

## Checklist:

- [x] I based my code and PRs against the repositories `development` branch.

### DIFF
--- a/advanced/Scripts/utils.sh
+++ b/advanced/Scripts/utils.sh
@@ -73,7 +73,9 @@ getFTLPID() {
 # Example getFTLConfigValue dns.piholePTR
 #######################
 getFTLConfigValue(){
-  pihole-FTL --config -q "${1}"
+  # Pipe to cat to avoid pihole-FTL assuming this is an interactive command
+  # returning colored output.
+  pihole-FTL --config -q "${1}" | cat
 }
 
 #######################


yubiuser:
Changing the base was not enough, because this branch was really based on `master`

```
  [✓] Building dependency package pihole-meta.deb
  [✗] Installing Pi-hole dependency package
   Error: Unable to install Pi-hole dependency package.
   Error: Unable to complete update, please contact support
```

(In development the meta package is at 0.5 already, this branch tries to downgrade to 0.4)

yubiuser:
How does FTL determine if the shell is non-interactive. It looks strange to pipe the whole output to `cat`. Most info I could find online suggest to redirect STDIN from `/dev/null` but this does not work for FTL

DL6ER:
FTL uses [`isatty(stdout)`](https://www.man7.org/linux/man-pages/man3/isatty.3.html). Behind the scenes, this is sending `TIOCSWINSZ` to the kernel to get the size of the attached kernel. If this works, `isatty()` returns that this is an interactive terminal. If no size could be obtained, it returns that that's not the case.

This is a pretty standard way of determining whether sending color codes by default is a good idea. IIRC (~~not checked right now~~ I was correct, see [here](https://github.com/NetBSD/src/blob/ffa5d4edf54a323b8166d8ea2d815c3062f2b60f/usr.bin/grep/grep.c#L595)) it's also how `grep` and friends determine if they should show colors.

DL6ER:
Rebased the branch on latest `development`. Turns out it was actually based on `development` but my local `development` was still on f2280eb3300a87928b0b7393e58a02a329b33857 (about one month old). Seems I have not submitted core PRs in a while...

yubiuser:
I'm using

```
nanopi@nanopi:/etc/.pihole$ pihole -v
Core
    Version is v6.1.4-33-g5cb60120 (Latest: null)
    Branch is tweak/non_interactive_config
    Hash is 5cb60120 (Latest: 5cb60120)
Web
    Version is v6.2.1-99-gb778a9a4 (Latest: null)
    Branch is dependabot/npm_and_yarn/development/fortawesome/fontawesome-free-7.0.0
    Hash is b778a9a4 (Latest: b778a9a4)
FTL
    Version is vDev-191aaad (Latest: null)
    Branch is new/cli_config_colors
    Hash is 191aaad7 (Latest: 191aaad7)
```

and still getting the error


____

Switching to any other FTL branch fixes the issue.

____

P.S. Running `sudo pihole -g` from CLI does not show the issue.

yubiuser:
Maybe it's because the web interface sets `FORCE_COLOR` so it will always print the escape sequence for normal color?

**ADD**

I'm quite sure this is the reason. I've added debug code to print the content of the `GRAVITY_TEMPDIR` and switched the web interface to `master` to show the escape codes and voila

DL6ER:
Oh, yes... `FORCE_COLOR` is the issue here. I will suppress being able to enforce colors for the `--config` option. Please update the FTL branch and thanks for testing so thorough. I have, indeed, only tested the CLI not via the web interface.

yubiuser (approved):

PR #6381, opened 2025-08-11
When running with --unattended flag, ensure pre-seeded pihole.toml file exists

### LINKED ISSUE #6380: Installer: `--unattended` silently fails if no config file provided.
# Problem description
The `--unattended` flag is ignored if you run the installer on a new system without a `setupVars.conf` or `pihole.toml` file: it will attempt to show dialogs and thus fail to install pi-hole.

It took me a long time to figure out that my own pre-install script didn't copy my `pihole.toml` file before running the pi-hole installer. This caused the `fresh_install` variable to be set to `true` and have the script follow a code path where it does NOT check if the  `--unattended` flag was provided. See this part of the installer for details:
https://github.com/pi-hole/pi-hole/blob/cef7fd4b02c9ae4314e7478c0d542379bfce7f65/automated install/basic-install.sh#L2238

# Request for improvement
Please improve the installer by exiting with an error message and a non-zero exit code if it is not possible to do an unattended install (e.g. when there's no config file to use).

PromoFaux:

See #6380 

Exits the installer early if no config file is found when passing in the `--unattended` flag. WIthout a config file, FTL cannot be configured.

```
root@adamtest:/etc/.pihole# ./automated\ install/basic-install.sh --unattended
   Error: "/etc/pihole/pihole.toml" not found. Cannot run unattended setup
```

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -167,6 +167,17 @@ for var in "$@"; do
     esac
 done
 
+if [[ "${runUnattended}" == true ]]; then
+    # In order to run an unattended setup, a pre-seeded /etc/pihole/pihole.toml must exist
+    if [[ ! -f "${PI_HOLE_CONFIG_DIR}/pihole.toml" ]]; then
+        printf "  %b Error: \"%s\" not found. Cannot run unattended setup\\n" "${CROSS}" "${PI_HOLE_CONFIG_DIR}/pihole.toml"
+        exit 1
+    fi
+    printf "  %b Performing unattended setup, no dialogs will be displayed\\n" "${INFO}"
+    # also disable debconf-apt-progress dialogs
+    export DEBIAN_FRONTEND="noninteractive"
+fi
+
 # If the color table file exists,
 if [[ -f "${coltable}" ]]; then
     # source it
@@ -2240,15 +2251,6 @@ main() {
     if [[ "${funcOutput}" == "" ]]; then
         printf "  %b Upgrade/install aborted\\n" "${CROSS}" "${DISTRO_NAME}"
         exit 1
-    fi
-
-    if [[ "${fresh_install}" == false ]]; then
-        # if it's running unattended,
-        if [[ "${runUnattended}" == true ]]; then
-            printf "  %b Performing unattended setup, no dialogs will be displayed\\n" "${INFO}"
-            # also disable debconf-apt-progress dialogs
-            export DEBIAN_FRONTEND="noninteractive"
-        fi
     fi
 
     if [[ "${fresh_install}" == true ]]; then


yubiuser (commented):
I **think** this would break v5 > v6 upgrades. `pihole.toml` does not exit yet, but the updater will run with the unattended flag. We need to carry our legacy and also allow for the presence of `setupVars.conf`.

Easiest would be to use `fresh_install` here as well.

PromoFaux:
You're [not wrong](https://github.com/pi-hole/pi-hole/blob/master/advanced/Scripts/update.sh#L225).

OK, let me give it some more thought...

MichaIng:
I was coincidentally just looking for a way to do unattended Pi-hole installations on GitHub Actions runners for testing purpose.

Was this PR aiming to do a step in making this possible? I know it was not intended to make this possible in the past, and the way the code treats this flag currently shows it pretty well:
- The `runUnattended` flag does nothing but `export DEBIAN_FRONTEND="noninteractive"` to skip `debconf` dialogs, which can e.g. pop up on `ca-certificates` installation/update to toggle (new) CAs. On non-APT distros it however has no effect, and `dialog` calls are done even if `runUnattended` is true.
- What has more significant effect is the `fresh_install` flag, which is the condition for all but one `dialog` calls. The exception is one in case of v5 => v6 upgrade whether Lighttpd shall be disabled, and it is shown only if STDIN is a terminal, hence cannot block calls from non-interactive shells, cron jobs, systemd units etc. `debconf` is also skipped automatically in those cases.
- So a reinstall/update/`fresh_install == false` with `--unattended`/`$runUnattended == true` in combination is almost assured to never show any dialog, which is the case if `pihole -up` or `pihole checkout` call the installer.
- Whether `fresh_install` is true or false is decided only based on the existence of either v6 or pre-v6 config file. So pre-seeding an unattended (fresh) installation results in installer treating it as reinstall/update. Did not test the result yet, at least will surely have unexpected/unintended side-effects, if not now then later, when contributors take `fresh_install` for what it says, not expecting it to be false when a fresh install is pre-seeded.

So this PR solves #6380 only indirectly, as the problem is not that a missing config file would cause issues, but that `--unattended` simply has no effect on whether `dialog`s are shown or not. The PR indirectly changes this by enforcing a config file, as the config file triggers the installer to treat it as **no** fresh install. As long as it is not implemented a way that it skips dialogs but assures that all other steps of a fresh install are done, it should be probably made clear in the script that the flag exists for internal update/repair steps only. Probably it would be even better to just remove it and have the two scripts behind `pihole -up` and `pihole checkout` exporting `DEBIAN_FRONTENT`, instead of passing `--unattended` to the installer?

But I would be open to help with and test a true unattended install option. Currently, it is not really possible while pre-creating any config file leads to `$fresh_install == false`.

EDIT: Just saw #6034
> Simulated Fresh install with `--unattended` flag and pre-seeded `/etc/pihole/pihole.toml` All scripts are installed and FTL is configured directly.

Does this mean `$fresh_install == false` in case of an actually fresh install just with pre-created config currently works well? That would be great, but then I suggest to rename the variable from `fresh_install` to `show_dialogs` or `config_exists` (negated) to make clear for contributors that the code section is not assured to run on fresh installs, if a config was pre-created. And then double-check that such an unattended installation really does not miss any expected setup step.

automated install/basic-install.sh
```diff
@@ -167,6 +167,17 @@ for var in "$@"; do
     esac
 done
 
+if [[ "${runUnattended}" == true ]]; then
+    # In order to run an unattended setup, a pre-seeded /etc/pihole/pihole.toml must exist
+    if [[ ! -f "${PI_HOLE_CONFIG_DIR}/pihole.toml" ]]; then
+        printf "  %b Error: \"%s\" not found. Cannot run unattended setup\\n" "${CROSS}" "${PI_HOLE_CONFIG_DIR}/pihole.toml"
+        exit 1
```
rdwebdesign: I know this PR was left in draft mode a few months ago, but the issue with `--unattended` on fresh installs is not solved.

What do you think about using `pihole-FTL create-default-config /etc/pihole/pihole.toml` to create a fresh config file, instead of `exit 1`? 

_https://github.com/pi-hole/FTL/blob/8d1add8d99d8a36ccc6fb97628a33a219cf8abbd/src/args.c#L674-L693_

```suggestion
        pihole-FTL create-default-config "${PI_HOLE_CONFIG_DIR}/pihole.toml"
```

PromoFaux:
I don't _really_ recall where my head was at with this.

> Does this mean $fresh_install == false in case of an actually fresh install just with pre-created config currently works well?

I can only assume so!

PromoFaux:
> Was this PR aiming to do a step in making this possible?

So, to my knowledge this should already be possible, per my (anecdotal) testing in #6034. What _this_ particular PR was aiming for was "Have you tried to run an unattended install _but also haven't pre-seeded a `pihole.toml` file?_ Exit early!" 

But as @yubiuser rightly pointed out, that then breaks a v5->v6 upgrade. Not sure how many of those are left in the wild, mind.

yubiuser:
We should not break the v5 > v6 path yet. (We could probably rip out all pre-v5 stuff, v5 was released 10th of May 2020). 

I think MichaIng has some good points, maybe Promo is up to the challenge to really allow a fresh `--unattended` update install when a pre-seeded `pihole.toml` is present?

MichaIng:
Or split PRs, and just switch to `fresh_install` as condition for now, to keep the intention but restore v5 => v6 updates?

And then we can think about whether/how to merge or rename `--unattended` and/or `$fresh_install` to make it clearer and maybe more explicit for devs and users. As quick pre-liminary thought: The `--unattended` flag could be probably removed: It inconsistently has an effect on `apt` distros only, and AFAIK only `ca-certificates` uses `debconf` among Pi-hole deps, which is almost assured to be installed already. And since the dependency package does not enforce a dependency version, the installer will never upgrade it either. The `$fresh_install` could be renamed to e.g. `$config_present` or `$setup_dialogs_needed` or so, and the only thing needed for Pi-hole to be installed unattended is a pre-seeded config (even an old v5 one would work(?)), no installer flag.

automated install/basic-install.sh
rdwebdesign, line 174: @PromoFaux 

This PR is still a draft, but I think the issue with `--unattended` on fresh installs remains unsolved.

Do you intend to finish it?

If you do, what do you think about trying to create a fresh config file with default settings, instead of `exit 1` immediately? 


```suggestion
        # If the file doesn't exit, we try to create one using the default settings
    
        local createConfig createConfigStatus
        createConfig=$(pihole-FTL create-default-config "${PI_HOLE_CONFIG_DIR}/pihole.toml")
        createConfigStatus=$?
        if [[ "${createConfigStatus}" -eq 0 ]]; then
            printf "  %b Warning: \"%s\" not found. Using default settings.\\n" "${INFO}" "${PI_HOLE_CONFIG_DIR}/pihole.toml"
        else
            # File could not be created
            printf "  %b Error: \"%s\" not found. Cannot run unattended setup\\n" "${CROSS}" "${PI_HOLE_CONFIG_DIR}/pihole.toml"
            exit 1
        fi
```

PR #6298, opened 2025-06-09
changes name of package_manager_detect() to check_for_dependencies() and adds the ability to manage pihole-meta package install/update to prevent installing the same version repeatedly

mwoolweaver:
**What does this PR aim to accomplish?:**

Fix the following points:

 - > running a package cache update is a resource intensive process on SD cards. #6272

 - > Maybe something like `apt cache policy <dependency package> | grep version` and compare that to the `/etc/.pihole/ <package metadata version>` and "do things" if the proposed updated metadata is higher than the installed package version?
   >
   > _Originally posted by dschaper in https://github.com/pi-hole/pi-hole/issues/6282#issuecomment-2944867806_

**How does this PR accomplish the above?:**

change name of `package_manager_detect()` to `check_for_dependencies()` and add the ability to manage `pihole-meta` package install/update to prevent installing the same version repeatedly.

 - only updates package cache for `apt-get` based systems when `pihole-meta` package is installed (_e.g. a fresh install OR if removed inadvertently by user when a dependency is removed_) or if `pihole-meta` package has an upgrade.  

 - prevents the scripts from reinstalling the same version of the package with every call of `pihole -up` and by extension `basic-install.sh` when `core` OR `FTL` have an update with no `pihole-meta` package changes.

Closing #6292 in favor of this.

**Link documentation PRs if any are needed to support this PR:**

N/A

---
**By submitting this pull request, I confirm the following:**

1. I have read and understood the [contributors guide](https://docs.pi-hole.net/guides/github/contributing/), as well as this entire template. I understand which branch to base my commits and Pull Requests against.
2. I have commented my proposed changes within the code and I have tested my changes.
3. I am willing to help maintain this change if there are issues with it later.
4. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
5. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
6. I have checked that another pull request for this purpose does not exist.
7. I have considered, and confirmed that this submission will be valuable to others.
8. I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
9. I give this submission freely, and claim no ownership to its content.

---
- [x] I have read the above and my PR is ready for review. *Check this box to confirm*

PR #6383, opened 2025-08-13
adds function, `check_for_meta_package()`, that helps manage pihole-meta package installation & upgrades

mwoolweaver:

addresses the following points:

 - > running a package cache update is a resource intensive process on SD cards.
   >_Originally posted by dschaper in https://github.com/pi-hole/pi-hole/discussions/6272_

 - > Maybe something like `apt cache policy <dependency package> | grep version` and compare that to the `/etc/.pihole/ <package metadata version>` and "do things" if the proposed updated metadata is higher than the installed package version?
   >
   > _Originally posted by dschaper in https://github.com/pi-hole/pi-hole/issues/6282#issuecomment-2944867806_

adds a function, `check_for_meta_package()`, that adds dependency check that ensures the `pihole-meta` package is always update-to-date.

 - only updates package cache for `apt-get` based systems when `pihole-meta` package is being installed (_e.g. a fresh install OR if removed inadvertently by user_) or if `pihole-meta` package has an upgrade.

 - prevents reinstalling the same version of the meta package with every call of `pihole -up` and by extension `basic-install.sh` when `core` OR `FTL` have an update with no meta package changes.

Closing #6298 in favor of this PR after suggestions by @yubiuser in https://github.com/pi-hole/pi-hole/pull/6298#pullrequestreview-3085872868

note about this suggestion from yubiuser's comment:

 > - Drop `DEPEND_CHECK`. I see where you're coming from, but it only guards two simple functions (getting the installed version and getting the to-be-installed-version - two one-liners) which in my view are not worth the added complexity.

I changed the name of the variable to `META_PACKAGE_CHECK` as it guards the two one-liners mentioned above __AND__ the remainder of the function which includes using `VersionConverter()` and the version number comparison.

N/A

---

---


mwoolweaver:
the biggest impact to be seen is on Fedora/CentOS where the `rpm-build` package is installed __&__ removed before the meta package is installed.

mwoolweaver:
https://github.com/mwoolweaver/pi-hole/tree/QOL/check-for-meta-package-before-install

this branch achieves the same goal with less code duplication

mwoolweaver:
i notice that there have been recent-ish update that add alpine support but i don't have an alpine machine to test with so i won't be able to finish this pr. 

edit: 

this solution would be helpful for scenario's when testing PR's (like #6472) that increase the `pihole-meta` package version and then switching back to the `development`/`master` branch where `pihole-meta`'s version number is lower. 

on debian based systems this won't work without removing the meta package before switch branches

Issue #6390, opened 2025-08-18
Issue Stopping / Starting pihole-FTL on Alpine Linux


lb803:
### Versions

- Pi-hole:  [v6.1.4](https://github.com/pi-hole/pi-hole/releases/v6.1.4)
- Web:  [v6.2.1](https://github.com/pi-hole/web/releases/v6.2.1)
- FTL:  [v6.2.3](https://github.com/pi-hole/FTL/releases/v6.2.3)
- Docker Tag: [2025.08.0](https://github.com/pi-hole/docker-pi-hole/releases/2025.08.0)

### Platform

- OS and version:  Rasberry Pi OS (Trixie)

- Platform:  Docker

### Expected behavior

`/opt/pihole/piholeLogFlush.sh` to be able to stop / start pihole-FTL on Alpine Linux.

### Actual behavior / bug

`/opt/pihole/piholeLogFlush.sh` uses the command `service` to manage services, and this command does not seem to be available on Alpine.

```
user@raspberrypi:~ $ docker exec -it pihole bash
f1eb1d2cb0d5:/# pihole -l off
  [✓] Flushed /var/log/pihole/pihole.log ...
  [✓] Flushed /var/log/pihole/FTL.log ...
  [✓] Flushed /var/log/pihole/webserver.log ...
  [i] Flushing database, DNS resolution temporarily unavailable .../opt/pihole/piholeLogFlush.sh: line 107: service: command not found
/opt/pihole/piholeLogFlush.sh: line 113: service: command not found
  [✓] Deleted 1914 queries from long-term query database
  [i] Disabling logging...
  [✓] Logging has been disabled!
```

An [earlier revision](https://github.com/pi-hole/pi-hole/commit/e1b05028a766c4287e75e5d7d3455010f00467ac) of this script used some useful helper functions [stop_service()](https://github.com/pi-hole/pi-hole/blob/cef7fd4b02c9ae4314e7478c0d542379bfce7f65/automated%20install/basic-install.sh#L1227) and [restart_service()](https://github.com/pi-hole/pi-hole/blob/cef7fd4b02c9ae4314e7478c0d542379bfce7f65/automated%20install/basic-install.sh#L1241) which (I think) might be adapted to work on Alpine as well.

### Steps to reproduce

1. Disable logging with `pihole -l off`
[...]

PR #6391, opened 2025-08-18
Separate log flushing from log rotation and disable log flush on docker container

yubiuser:

Disables `pihole -f` on the docker container. It requires a FTL restart which should not happen inside the container.
See https://github.com/pi-hole/pi-hole/issues/6390

Wrap the function in our docker check.

---

---

### DIFF
--- a/manpages/pihole.8
+++ b/manpages/pihole.8
@@ -100,9 +100,12 @@ Available commands and options:
       -c                Include a Pi-hole database integrity check
 .br
 
-\fB-f, flush\fR
+\fB-f, flush\fR [quite]
 .br
-    Flush the Pi-hole log
+    Flush the Pi-hole log and last 24h from the query database
+.br
+
+      quite           Suppress output
 .br
 
 \fB-r, repair\fR
@@ -240,6 +243,14 @@ Available commands and options:
 .br
 
       verbose           Show authentication and status messages
+.br
+
+\fBlogrotate\fR [quite]
+.br
+    Rotate Pi-hole's log files
+.br
+
+      quite           Suppress output
 .br
 
 .SH "EXAMPLE"
--- a/pihole
+++ b/pihole
@@ -92,6 +92,7 @@ debugFunc() {
 }
 
 flushFunc() {
+    # unsupported in docker because it requires restarting FTL
     if [ -n "${DOCKER_VERSION}" ]; then
         unsupportedFunc
     else
@@ -107,6 +108,11 @@ arpFunc() {
   echo -e "  ${INFO} The 'arpflush' command is deprecated, use 'networkflush' instead"
   "${PI_HOLE_SCRIPT_DIR}"/piholeNetworkFlush.sh "$@"
   exit 0
+}
+
+logrotateFunc() {
+   "${PI_HOLE_SCRIPT_DIR}"/piholeLogRotate.sh "$@"
+    exit 0
 }
 
 networkFlush() {
@@ -506,7 +512,8 @@ Debugging Options:
   -d, debug           Start a debugging session
                         Add '-c' or '--check-database' to include a Pi-hole database integrity check
                         Add '-a' to automatically upload the log to tricorder.pi-hole.net
-  -f, flush           Flush the Pi-hole log
+  -f, flush           Flush the Pi-hole logs and last 24h from the query database
+                        Add 'quiet' to suppress output messages
   -r, repair          Repair Pi-hole subsystems
   -t, tail [arg]      View the live output of the Pi-hole log.
                       Add an optional argument to filter the log
@@ -539,7 +546,9 @@ Options:
   checkout            Switch Pi-hole subsystems to a different GitHub branch
                         Add '-h' for more info on checkout usage
   networkflush        Flush information stored in Pi-hole's network tables
-                        Add '--arp' to additionally flush the ARP table ";
+                        Add '--arp' to additionally flush the ARP table
+  logrotate          Rotate Pi-hole's log files
+                        Add 'quiet' to suppress output messages";
   exit 0
 }
 
@@ -582,6 +591,7 @@ case "${1}" in
   "arpflush"                      ) need_root=true;; # Deprecated, use networkflush instead
   "networkflush"                  ) need_root=true;;
   "-t" | "tail"                   ) need_root=true;;
+  "logrotate"                    ) need_root=true;;
   *                               ) helpFunc;;
 esac
 
@@ -617,5 +627,6 @@ case "${1}" in
   "arpflush"                      ) arpFunc "$@";; # Deprecated, use networkflush instead
   "networkflush"                  ) networkFlush "$@";;
   "-t" | "tail"                   ) tailFunc "$2";;
+  "logrotate"                     ) logrotateFunc "$@";;
   *                               ) helpFunc;;
 esac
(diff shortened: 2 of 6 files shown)


rdwebdesign:
Apparently this will break log rotation inside containers:
- https://github.com/pi-hole/docker-pi-hole/blob/f4aac365653ffe7d8dbe9472193026e78cdfd174/src/crontab.txt#L2C1-L2C75

rdwebdesign:
The more I look at `piholeLogFlush.sh`, the more I think we should have two separate functions to execute the different jobs.

I don't like the completely different jobs:

- `pihole flush`:
  flushes the 3 logs (`pihole.log`, `FTL.log` and `webserver.log`) and delete the last 24 hours from FTL's database.
  No log rotation.

- `pihole flush once`:
  rotate the 3 logs, but it doesn't flush them or touch the database. 
  This should be a separate function. 
  We should change this to something like `pihole rotatelogs`

yubiuser:
I agree. We should separate the 'rotate' function from the 'flushing' function.

yubiuser:
Those PRs depend on this one: https://github.com/pi-hole/pi-hole/pull/6391 and https://github.com/pi-hole/docker-pi-hole/pull/1910

rdwebdesign:
@yubiuser 

I tested the code (core and docker) on a container.

When I run `pihole -f`, the message "Function not supported in Docker images" is printed, as expected.

On the other hand, if I use the web interface "Flush Logs (last 24 hours)" button, the API answers with `status: "success"` and no error message is shown.

Also, this is what I see in `FTL.log`:

```
2025-10-21 17:23:28.450 INFO Received API request to flush the logs
2025-10-21 17:23:28.455 INFO Log has been flushed due to API request
```

This will generate a lot of user confusion, because FTL won't be restarted... the logs won't be completely flushed, but no errors are shown (actaully a success payload is received) and `FTL.log` shows a success message.

We could disable (or hide) the button in containers, but this won't really disable the API endpoint in FTL. Apps or scripts using this endpoint will do the same.

I think the correct solution would be to change FTL [`api_action_flush_logs()` function](https://github.com/pi-hole/FTL/blob/d4af4f7c4e6599b4e7665e41b1881bd464612c26/src/api/action.c#L140-L158) to detect it is running inside a container and disable this API endpoint with a message "Flushing the logs is not possible inside containers - not enough permissions" or something similar.

yubiuser:
The button on the web interface trigger a flush via FTL - it does not use the bash script. From the CLI we need to stop FTL to delete the long- term database history without having FTL writing to it at the same time. FTL internally can handle this better I think so it does not need a restart (maybe only to clean the in-memory database).
But we should block the API call for FTL restart in docker probably (might break some users sync-script?)

rdwebdesign:
> I think so it does not need a restart (maybe only to clean the in-memory database).

That's the part I'm not sure. 

Can FTL API flush the logs without restarting FTL?

If the API can flush the logs without restarting, my PR is not needed and we can leave the web interface button as it is. Also, we could try to change `pihole -f` command to use the API.

> But we should block the API call for FTL restart in docker probably (might break some users sync-script?)

Maybe... We need to check that.

PromoFaux:
> The button on the web interface trigger a flush via FTL - it does not use the bash script. 

Not fully thought this through or looked at the code anywhere, just thinking out loud........

Could (maybe it does already) the bash script not just use the API to do the same thing as the web interface is doing?

manpages/pihole.8
```diff
@@ -100,9 +100,12 @@ Available commands and options:
       -c                Include a Pi-hole database integrity check
 .br
 
-\fB-f, flush\fR
+\fB-f, flush\fR [quite]
```
PromoFaux, line 103:
```suggestion
\fB-f, flush\fR [quiet]
```

manpages/pihole.8
```diff
@@ -100,9 +100,12 @@ Available commands and options:
...
 .br
-    Flush the Pi-hole log
+    Flush the Pi-hole log and last 24h from the query database
+.br
+
+      quite           Suppress output
```
PromoFaux, line 108:
```suggestion
      quiet           Suppress output
```

manpages/pihole.8
```diff
@@ -242,6 +245,14 @@ Available commands and options:
       verbose           Show authentication and status messages
 .br
 
+\fBlogrotate\fR [quite]
```
PromoFaux, line 248:
```suggestion
\fBlogrotate\fR [quiet]
```

manpages/pihole.8
```diff
@@ -242,6 +245,14 @@ Available commands and options:
...
+.br
+    Rotate Pi-hole's log files
+.br
+
+      quite           Suppress output
```
PromoFaux, line 253:
```suggestion
      quiet           Suppress output
```

pihole
```diff
@@ -105,6 +110,11 @@ arpFunc() {
   exit 0
 }
 
+logrotateFunc() {
+   "${PI_HOLE_SCRIPT_DIR}"/piholeLogRotate.sh "$@"
```
PromoFaux, line 114:
```suggestion
    "${PI_HOLE_SCRIPT_DIR}"/piholeLogRotate.sh "$@"
```

pihole
```diff
@@ -578,6 +591,7 @@ case "${1}" in
   "arpflush"                      ) need_root=true;; # Deprecated, use networkflush instead
   "networkflush"                  ) need_root=true;;
   "-t" | "tail"                   ) need_root=true;;
+  "logrotate"                    ) need_root=true;;
```
PromoFaux, line 594:
```suggestion
  "logrotate"                     ) need_root=true;;
```

PromoFaux (commented):
Just some spelling and whitespace issues from me (so far) not tested the code yet!

rdwebdesign:
> Could (maybe it does already) the bash script not just use the API to do the same thing as the web interface is doing?

The currently code executes this sequence:
- the script stops FTL service (to avoid database writes while deleting data);
- then deletes the database entries using sqlite functions and
- finally restarts FTL service.

I think FTL can safely delete recent database entries without the need to stop/restart, but currently there is no API endpoint or command line option to do that.

FTL has a similar function `delete_old_queries_from_db()` (used by GC to remove old queries). We would need a new function (something like `delete_recent_queries_from_db()`) and a new API endpoint.

PromoFaux:
> but currently there is no API endpoint or command line option to do that.

So how does the web interface button work? (Apologies if I'm just being dense and missing something v obvious!)

yubiuser:
It uses https://ftl.pi-hole.net/development/docs/#post-/action/flush/logs

PromoFaux:
Going through old PRs and spotted this one, too https://github.com/pi-hole/pi-hole/pull/6524/changes

It also touches the logrotate functions, so we should make sure they are compatible

rdwebdesign:
> So how does the web interface button work?

Apparently I misunderstood the "old queries" part on `delete_old_queries_from_db()` function.

This function is also [used by `action/flush/logs` API endpoint](https://github.com/pi-hole/FTL/blob/71b6fc62571075740a3c721d5e961db9c3609376/src/log.c#L790-L795) to delete all recent queries from the last 24h:
```c
	// Flush last 24 hours of on-disk database
	if(!delete_old_queries_from_db(false, mintime))
	{
		log_err("Could not flush on-disk database");
		return false;
	}
```

PR #6430, opened 2025-10-22
Loosen requirements for local file access for gravity

yubiuser:

When users add local files with `file:///` as adlists we currently check for `a+r` permissions. However, it's not necessary that everyone has read access to the file, but only the current user executing gravity. (From CLI this is `root`, from web interface via FTL `pihole`).
The current implementation prohibits files like `/etc/pihole/my_list.txt` to be used, because after a FTL restart the file will have `-rw-r-----` permission and not even `sudo pihole -g` will pass the test. 

Only check if the current user has read permission.

---

---

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -729,7 +729,7 @@ gravity_DownloadBlocklistFromUrl() {
     # to match the effective runtime user of FTL; otherwise, check the current user's read access
     # (e.g., in Docker or when invoked by a non-root user). The target must
     # resolve to a regular file and be readable by the evaluated user.
-  if [[ "${url}" == "file://"* ]]; then
+  if [[ "${url}" == "file:/"* ]]; then
     # Get the file path
     file_path=$(echo "${url}" | cut -d'/' -f3-)
     # Check if the file exists and is a regular file (i.e. not a socket, fifo, tty, block). Might still be a symlink.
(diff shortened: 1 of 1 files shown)


rdwebdesign (changes requested):

rdwebdesign:
Is this still needed?

yubiuser:
Yes. The example

> The current implementation prohibits files like /etc/pihole/my_list.txt to be used, because after a FTL restart the file will have -rw-r----- permission and not even sudo pihole -g will pass the test.

will still fail on current development.

rdwebdesign (approved):

gravity.sh
```diff
@@ -724,29 +724,40 @@ gravity_DownloadBlocklistFromUrl() {
...
-  # file has a+r permission. We explicitly check for all+read because we want
-  # to make sure that the file is readable by everyone and not just the user
-  # running the script.
-  if [[ $url == "file://"* ]]; then
+  # If we "download" a local file (file://), verify read access before using it.
+  # When running as root (e.g., via pihole -g), check that the 'pihole' user can read the file
+  # to match the effective runtime user of FTL; otherwise, check the current user's read access
+  # (e.g., in Docker or when invoked by a non-root user). The target must
+  # resolve to a regular file and be readable by the evaluated user.
+  if [[ "${url}" == "file:/"* ]]; then
```
PromoFaux, line 732: Is the change from `file://` to `file:/` intentional here?

yubiuser: Yes. See MM dissucssion.

PromoFaux (approved):

PR #6440, opened 2025-10-26
Create version file, if not existing (on first Container start)

casperklein:

On the first container start `chmod` fails, because `/etc/pihole/versions` does not exist yet:

```
pihole  |   [i] Applying the following caps to pihole-FTL:                                                                                                                                                                 
pihole  |         * CAP_CHOWN
pihole  |         * CAP_NET_BIND_SERVICE
pihole  |         * CAP_NET_RAW
pihole  |         * CAP_NET_ADMIN
pihole  |         * CAP_SYS_NICE
pihole  |         * CAP_SYS_TIME
pihole  |
pihole  | chmod: cannot access '/etc/pihole/versions': No such file or directory
pihole  |   [i] Starting pihole-FTL (no-daemon) as pihole
pihole  |
pihole  |   [i] Version info:
pihole  |       Core
```

This PR fixes that.

Creating the file before the `chmod` execution.

**Additional information**

There is similar code in

https://github.com/pi-hole/pi-hole/blob/66a3c73d6c3bdc0f10f8c081d1f413b3bf19e19b/advanced/Scripts/updatecheck.sh#L54-L56

But that isn't run before `pihole-FTL-prestart.sh`.

---

---

### DIFF
--- a/advanced/Templates/pihole-FTL-prestart.sh
+++ b/advanced/Templates/pihole-FTL-prestart.sh
@@ -14,6 +14,7 @@ mkdir -p /var/log/pihole
 chown -R pihole:pihole /etc/pihole/ /var/log/pihole/
 
 # allow all users read version file (and use pihole -v)
+touch /etc/pihole/versions
 chmod 0644 /etc/pihole/versions
 
 # allow pihole to access subdirs in /etc/pihole (sets execution bit on dirs)


PromoFaux:
Nothing wrong with a bit of belt and braces, of course, but in dockerland we do run pihole updatechecker before we run pihole -v , so the code you linked above should be running, no?

https://github.com/pi-hole/docker-pi-hole/blob/master/src/start.sh#L86

casperklein:
`Pihole -v` is not the problem.

The problem is `pihole-FTL-prestart.sh`,  which expects the file to exist. It runs before the update check:

https://github.com/pi-hole/docker-pi-hole/blob/ad95d1cfdf98c4872f042abb3359dea1fb01f835/src/start.sh#L59

The updatecheck runs in line 86 as you stated.

I think, that `pihole updatechecker` doesn't depend on FTL to be running. If that's true, we could alternatively move the updatecheck before the `pihole-FTL-prestart.sh` execution.

PromoFaux:
> I think, that pihole updatechecker doesn't depend on FTL to be running. 

This is very true. Historically this command was set to run _after_ FTL start because of people setting the container itself to use FTL as it's DNS - which meant that the update checker could not run because it wasn't able to resolve github.com.

We do however explicity advise against using pi-hole as it's own resolver in the readme these days though.

Perhaps the order of things can be changed in docker startup, and perhaps that may even go towards solving [another issue](https://github.com/pi-hole/docker-pi-hole/issues/1918)

casperklein:
I understand. I would go with this PR to avoid any issues that could be introduced with a new order.

PR #6460, opened 2025-11-03
Improve gravity tables presentation on the debug log

rdwebdesign:

1. Add list "type" to the Adlists table. 

    Currently, all lists are shown without differentiation in the log. Only the web interface can differentiate then. If an user mistakenly add all lists as allowlist (antigravity), we can't understand what is wrong without looking at a web interface screenshot or asking for `pihole -g` output.

    image
    
2. Improve tables readability using text columns

1. Adding `type` column to the table:

    image

2. Use text columns instead of number codes on gravity tables:
    - Adlists table (`type` column): replace 0 and 1 with `Block` and `Allow`
    - Domains table (`type` column): replace 0, 1, 2 and 3 with `exact-allow`, `exact-deny`, `regex-allow` and `regex-deny`
    - All tables: use `yes`/`no` for all `enabled` columns.

---

---

### DIFF
--- a/advanced/Scripts/piholeDebug.sh
+++ b/advanced/Scripts/piholeDebug.sh
@@ -1080,15 +1080,15 @@ check_dhcp_servers() {
 }
 
 show_groups() {
-    show_db_entries "Groups" "SELECT id,CASE enabled WHEN '0' THEN '   0' WHEN '1' THEN '      1' ELSE enabled END enabled,name,datetime(date_added,'unixepoch','localtime') date_added,datetime(date_modified,'unixepoch','localtime') date_modified,description FROM \"group\"" "4 7 50 19 19 50"
+    show_db_entries "Groups" "SELECT id,CASE enabled WHEN '0' THEN '  no' WHEN '1' THEN '  yes' ELSE enabled END enabled,name,datetime(date_added,'unixepoch','localtime') date_added,datetime(date_modified,'unixepoch','localtime') date_modified,description FROM \"group\"" "4 7 50 19 19 50"
 }
 
 show_adlists() {
-    show_db_entries "Adlists" "SELECT id,CASE enabled WHEN '0' THEN '   0' WHEN '1' THEN '      1' ELSE enabled END enabled,GROUP_CONCAT(adlist_by_group.group_id) group_ids, CASE type WHEN '0' THEN 'Block' WHEN '1' THEN 'Allow' ELSE type END type, address,datetime(date_added,'unixepoch','localtime') date_added,datetime(date_modified,'unixepoch','localtime') date_modified,comment FROM adlist LEFT JOIN adlist_by_group ON adlist.id = adlist_by_group.adlist_id GROUP BY id;" "5 7 12 5 100 19 19 50"
+    show_db_entries "Adlists" "SELECT id,CASE enabled WHEN '0' THEN '  no' WHEN '1' THEN '  yes' ELSE enabled END enabled,GROUP_CONCAT(adlist_by_group.group_id) group_ids, CASE type WHEN '0' THEN 'Block' WHEN '1' THEN 'Allow' ELSE type END type, address,datetime(date_added,'unixepoch','localtime') date_added,datetime(date_modified,'unixepoch','localtime') date_modified,comment FROM adlist LEFT JOIN adlist_by_group ON adlist.id = adlist_by_group.adlist_id GROUP BY id;" "5 7 12 5 100 19 19 50"
 }
 
 show_domainlist() {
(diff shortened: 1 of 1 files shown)


yubiuser (dismissed):

PromoFaux:
I have another idea here - maybe this fits more under request territory - but I have always disliked the 0 1 2 3 4 in the tables - I find it difficult to parse at a glance, even with the key at top of the table.

Much better would be to have something like `CASE type WHEN '0' THEN 'Block' WHEN '1' THEN 'Allow' ELSE type END` Or something like that. 

It would make it easier to scan down the table. Maybe. Thoughts?

yubiuser:
Would be easier for sure!

rdwebdesign:
I actually tried this on my first attempt, but I didn't like the output. I used "BLOCK" and "ALLOW" (all caps) and the result was confusing to distinguish them. 

I will change to "Allow" and "Block".

PromoFaux:
Can you please provide a screenshot of this?

rdwebdesign:
image

PromoFaux (dismissed):
Approving, but I've also changed the direction of the PR... if we merge this, we should try to follow suit in the other tables if needed

yubiuser:
>  follow suit in the other tables if needed

I think this should happen right away in this PR. Otherwise it will never happen. rd's editor is probably still open ;-)

rdwebdesign:
> if we merge this, we should try to follow suit in the other tables if needed

I think we should keep "disabled" and "enabled" as `0` and `1`.

I'm not sure if replacing the 0, 1, 2 and 3 will improve readability.

image

image

The numbers are on different "columns", making very easy to read and spot the differences between table rows, but their meaning is printed only on the table header.
The texts are more self-explanatory, but the column is larger (11 characters). 

What do you think? Suggestions?

yubiuser:
> I'm not sure if replacing the 0, 1, 2 and 3 will improve readability.

In my opinion: they improve it a lot. I really like the type being written, this saves much mental energy.

> The texts are more self-explanatory, but the column is larger (11 characters).

I have many pixel horizontally ;-)

> I think we should keep "disabled" and "enabled" as 0 and 1.

How does it look like with `yes` and `no`. The header says enabled (?) and yes/no would be the natural answer.

rdwebdesign:
OK. I agree with your comments on using `type` as text.

About the yes/no:

This is the current format, using 0 and 1:

image

This is using yes/no:

image

---

**Note:**

I will use aligned yes/no values.
When I tried using them misaligned (like we had for 0 and 1), the result wasn't very nice and it didn't improve readability: 

image

advanced/Scripts/piholeDebug.sh
```diff
@@ -1080,15 +1080,15 @@ check_dhcp_servers() {
...
+    show_db_entries "Domainlist" "SELECT id,CASE type WHEN '0' THEN 'exact-allow' WHEN '1' THEN 'exact-deny' WHEN '2' THEN 'regex-allow' WHEN '3' THEN 'regex-deny' ELSE type END type,CASE enabled WHEN '0' THEN '  no' WHEN '1' THEN '  yes' ELSE enabled END enabled,GROUP_CONCAT(domainlist_by_group.group_id) group_ids,domain,datetime(date_added,'unixepoch','localtime') date_added,datetime(date_modified,'unixepoch','localtime') date_modified,comment FROM domainlist LEFT JOIN domainlist_by_group ON domainlist.id = domainlist_by_group.domainlist_id GROUP BY id;" "5 11 7 12 100 19 19 50"
```
yubiuser, line 1091:
```suggestion
    show_db_entries "Domainlist" "SELECT id,CASE type WHEN '0' THEN 'exact-allow' WHEN '1' THEN 'exact-deny' WHEN '2' THEN 'regex-allow' WHEN '3' THEN 'regex-deny' ELSE type END type,CASE enabled WHEN '0' THEN '  no' WHEN '1' THEN '  yes' ELSE enabled END enabled,GROUP_CONCAT(domainlist_by_group.group_id) group_ids,domain,datetime(date_added,'unixepoch','localtime') date_added,datetime(date_modified,'unixepoch','localtime') date_modified,comment FROM domainlist LEFT JOIN domainlist_by_group ON domainlist.id = domainlist_by_group.domainlist_id GROUP BY id;" "5 11 7 12 90 19 19 50"
```

I never saw such a long domain in the wild.
(And my screen is not that wide ;-)

rdwebdesign: This is also used for regex entries and I saw some very long examples.

yubiuser (approved):

advanced/Scripts/piholeDebug.sh
yubiuser: Ok

rdwebdesign: Just an example of a huge regex entry:

image

yubiuser: Oh my God! But the is a line break at the end, so the space reserved could easily be a bit shorter.

Issue #6459, opened 2025-11-03
Pi-hole prevents `sudo` shims from being used on Alpine Linux


Sparronator9999:
### Versions

Core version is v6.2.2 (Latest: v6.2.2)
Web version is v6.3 (Latest: v6.3)
FTL version is v6.3.2 (Latest: v6.3.2) 

### Platform

- OS and version: Alpine Linux v3.22
- Platform: Raspberry Pi 5

### Expected behavior

`[doas-sudo-shim](https://github.com/jirutka/doas-sudo-shim)` can be installed in place of `sudo` for privilege escalation.

### Actual behavior / bug

`sudo` is [included in `pihole-meta`](https://github.com/pi-hole/pi-hole/blob/1837b754b945eeab6ad5976443b38b983eb96e6e/automated%20install/basic-install.sh#L181) when installing on Alpine Linux, preventing a shim to an alternative (`doas` on Alpine Linux) from being used.

### Steps to reproduce

1. Login to a shell on a device with Pi-hole installed.
2. Try to install `doas-sudo-shim`: `apk add doas-sudo-shim` (use `doas` or `sudo` if needed to get root permissions).
3. See error due to not being able to select packages (see screenshot below).

## Debug Token

N/A

### Screenshots

<img width="767" height="266" alt="Image" src="https://github.com/user-attachments/assets/1f40ce38-cbea-434f-89ed-1845e4c67701" />

### Additional context

`sudo` has been deprecated for use on Alpine Linux [since v3.15](https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.15.0#Move_from_sudo_to_doas). I was trying to install `doas-sudo-shim` for my own usage/convenience but ran into this bug.

rdwebdesign:
Edit:

I'm not sure if we will be able to modify all functions to work with a `sudo` replacement.

yubiuser:
I think we could make `doas` and `doas-sudo-shim` a dependency of the alpine package. @Sparronator9999 Do you want to PR that change?

Sparronator9999:
> This is not a default Alpine package.

`doas` is installable from the `main` Alpine repository (see [here](https://pkgs.alpinelinux.org/package/v3.22/main/x86_64/doas)), and so should be able to replace `sudo` in the `pihole-meta` package.

> I'm not sure if we will be able to modify all functions to work with a `sudo` replacement.

[...]

Issue #1547, opened 2017-06-22
Pihole script requires root privilege via sudo represents a major security risk


graysky2:
**In raising this issue, I confirm the following (please check boxes, eg [X]) Failure to fill the template will close your issue:**

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the codebase?:**

2

---
**[OTHER]:**
I'm wondering how the project can be changed to avoid the need for `/etc/sudoers.d/pihole` which in my view represents a pretty major security risk.  Can the blocklists, etc. not be under pihole:pihole control perhaps in `/var/lib/pihole` or the like?  I need to familiarize myself more with the code to make a useful suggestion.

Mcat12:
Pi-hole needs sudo in order to restart the `dnsmasq` service. The web interface is only given sudo access to `pihole` commands, so it can not run any code other than what we give it.

graysky2:
I see.  Have you considered using a wrapper for the root privileges?  For example, https://github.com/graysky2/profile-sync-daemon/blob/master/common/psd-overlay-helper

Mcat12:
The wrapper would still need `pihole` sudo access.

jacobsalmela:
Pi-hole is a DNS server, which, by nature is a system-wide service so `sudo` access is to be expected.  I completely understand why some people are concerned with the elevated privileges but the rest of the team can chip in if they want.  So for now I'll label it as a discussion and we can move forward from there.

graysky2:
@jacobsalmela - I was under the impression that pihole was managing a list of domains for dnsmasq to use.  That can occur without root access.  I understand that it needs to start/stop dnsmasq and that the webUI can edit system files, restart the machine, etc.  I'm wondering how much of that could be decoupled from root access.  Clearly:
* start/stop/restart of init system needs root
* editing files in /etc/ owned by root needs root
* what else am I missing that pihole can do that needs root?

[...]

PR #6472, opened 2025-11-12
Eliminate the requirement to specifically use 'sudo' to install pi-hole. 

rrobgill:

Eliminate the requirement to specifically use `sudo` to install and operate pi-hole. 

This allows the user to execute the installer using privilege elevation tools such as `sudo`, `doas` or even `su` as per the default of their distro, or their preference.

This will solve https://github.com/pi-hole/pi-hole/issues/6459 

As basic-install.sh already has a requirement to be executed as root, a simpler and more cross-platform compatible option than sudo/doas is to use [runuser](https://www.man7.org/linux/man-pages/man1/runuser.1.html) to drop to the pihole user to run gravity on installation. This is possible because we are using sudo only to change from root to pihole for this step, and not to increase privileges. Runuser may only be run as root, and therefore is a tool for privilege delegation rather than elevation.

This avoids the need to specify a particular privilege elevation method for installation, and as runuser does not require password confirmation (as it only runs as root), further avoiding the need to handle doas (including the requirement for password confirmation exemptions).

Removes sudo from the dependencies across all platforms, add util-linux (providing runuser) on deb and rpm sysetms, and runuser on apk systems.

Modifies the current saving-provision where the script detects if it has not been run as root (by whatever means) and attempts to re-execute itself by detecting presence of the two most common means of elevation, sudo or doas, and call the script again accordingly. The current message is maintained on systems that have neither of these installed.

N/A 

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -116,11 +116,11 @@ c=70
 PIHOLE_META_PACKAGE_CONTROL_APT=$(
     cat <<EOM
 Package: pihole-meta
-Version: 0.6
+Version: 0.7
 Maintainer: Pi-hole team <adblock@pi-hole.net>
 Architecture: all
 Description: Pi-hole dependency meta package
-Depends: awk,bash-completion,binutils,ca-certificates,cron|cron-daemon,curl,dialog,bind9-dnsutils|dnsutils,dns-root-data,git,grep,iproute2,iputils-ping,jq,libcap2,libcap2-bin,lshw,procps,psmisc,sudo,unzip
+Depends: awk,bash-completion,binutils,ca-certificates,cron|cron-daemon,curl,dialog,bind9-dnsutils|dnsutils,dns-root-data,git,grep,iproute2,iputils-ping,jq,libcap2,libcap2-bin,lshw,procps,psmisc,util-linux,unzip
 Section: contrib/metapackages
 Priority: optional
 EOM
@@ -130,12 +130,12 @@ EOM
 PIHOLE_META_PACKAGE_CONTROL_RPM=$(
     cat <<EOM
 Name: pihole-meta
-Version: 0.3
+Version: 0.4
 Release: 1
 License: EUPL
 BuildArch: noarch
 Summary: Pi-hole dependency meta package
-Requires: bash-completion,bind-utils,binutils,ca-certificates,chkconfig,cronie,curl,dialog,findutils,gawk,git,grep,iproute,jq,libcap,lshw,procps-ng,psmisc,sudo,unzip
+Requires: bash-completion,bind-utils,binutils,ca-certificates,chkconfig,cronie,curl,dialog,findutils,gawk,git,grep,iproute,jq,libcap,lshw,procps-ng,psmisc,util-linux,unzip
 %description
 Pi-hole dependency meta package
 %prep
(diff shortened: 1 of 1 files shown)


dschaper:
This is a really interesting concept. It needs some extensive testing on different distro's and different releases of the distros, mostly because I don't fully trust the test platform to cover everything :)

rrobgill:
Realistically the actual work that makes it possible was back done when FTL stopped using sudo when spawning shell commands (I think for updates and possibly restarts at one point?).

mwoolweaver:
could this also be a potential fix for #1547 as well?

rrobgill:
> could this also be a potential fix for #1547 as well?

The pihole script itself still needs to be called at an elevated privilege to be able to do what it does, whether via sudo, doas, or whatever. But that is done at runtime by the user, not during automated processes internal to Pi-hole itself.

I think the biggest security risk in 1547 was having the pihole script called from within the UI, elevated to root via sudo, and all of the potential badness that could come from that. That hasn't really applied in the same way since v6 (which I would think probably realistically dealt with the actual concerns of 1547).

dschaper:
The automatic installation script re-run with elevated priv's has always felt sketchy to me and I was part of writing it 10 years ago. I'd love to have a way to not need it and I think we can get to that point but with a big change to how the install process happens. 

I think we are at a place where every OS that we can install to has a package manager and that combined with the fact that we don't need to install a whole lot with FTL v6 being the web server and API provider. We've even discussed using the `lua` interpreter in FTL as the shell for the installer, using a small bootstrap shell script that grabs the FTL binary and then the install happens through some `luascsript`ing. That should also address some of thethe controversy of `curl` to `bash`.

But in the end, #1547 comes down to the fact that you either trust Pi-hole or you don't. We are as absolutely transparent as possible and give users every bit of code that will run.

yubiuser (commented):
I like the idea of this PR - the less privs we need, the better. Maybe you could rebase/merge development again?

darkexplosiveqwx:
`sudo` is also used in `gravity.sh` to check whether the `pihole` user can read a local file, however it can be easily be replaced with `runuser`.
quick patch (based on current `development`):
```patch
From 42100137a11d867416d3f79c19b60e39ed5cc01f Mon Sep 17 00:00:00 2001
From: darkexplosiveqwx <101737077+darkexplosiveqwx@users.noreply.github.com>
Date: Thu, 7 May 2026 17:00:57 +0200
Subject: [PATCH] replace sudo with runuser in gravity.sh

Signed-off-by: darkexplosiveqwx <101737077+darkexplosiveqwx@users.noreply.github.com>
---
 gravity.sh | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/gravity.sh b/gravity.sh
index 5ec749af..2f7e6ed8 100755
--- a/gravity.sh
+++ b/gravity.sh
@@ -741,7 +741,7 @@ gravity_DownloadBlocklistFromUrl() {
         if [ "$(id -un)" == "root" ]; then
             # If we are root, we need to check if the pihole user has read permission
             #  otherwise, we might read files that the pihole user should not be able to read
-            if sudo -u pihole test -r "${file_path}"; then
+            if runuser -u pihole -- test -r "${file_path}"; then
                 echo -e "${OVER}  ${INFO} Using local file ${file_path}"
             else
                 echo -e "${OVER}  ${CROSS} Cannot read file (user 'pihole' lacks read permission)"
-- 
2.54.0

```

PR #6503, opened 2025-12-20
Support systemd.timer instead of cron

darkexplosiveqwx:

Use systemd.timer on systemd-based systems instead of cron

Converts the cronjobs to systemd .service and .timer.
These will be installed instead of cron alongside the systemd service for FTL, and uses the same systemd detection (which has caused issued before for systems with multiple init systems installed).

This also makes the weekly gravity update run as `pihole`. Gravity already runs as `pihole` when invoked via FTL so this shouldn't cause any issues

Debian supports multiple init systemd and we shouldn't install systemd when its not present, but it can be accepted as scheduling service.  
Fedora/CentOS only support systemd as init system, so fallback to cron is not needed.
Alpine is very much not using systemd, so this is not affected.

---

---


darkexplosiveqwx:
What is the main functional difference between `pihole flush once quiet` and `logrotate --state /var/lib/logrotate/pihole /etc/pihole/logrotate`? I think it could be justified to add `logrotate` as a dependency to the meta-pkgs and to then unify these two tasks into one. Wouldnt having `logrotate` not installed already cause an error for  `logrotate --state /var/lib/logrotate/pihole /etc/pihole/logrotate`?

darkexplosiveqwx:
I am not that familiar with `tox` so there may be a better way other than removing the checks for `cron.d`.

darkexplosiveqwx:
> What is the main functional difference between `pihole flush once quiet` and `logrotate --state /var/lib/logrotate/pihole /etc/pihole/logrotate`? I think it could be justified to add `logrotate` as a dependency to the meta-pkgs and to then unify these two tasks into one. Wouldnt having `logrotate` not installed already cause an error for `logrotate --state /var/lib/logrotate/pihole /etc/pihole/logrotate`?

I added `logrotate` to RPM and DEB dependencies, it was already present for APK, cleaned up `piholeLogFlush.sh` and combined `pihole-logrotate.systemd` with `pihole-log-flush.systemd`

automated install/basic-install.sh
```diff
@@ -116,11 +116,11 @@ c=70
 PIHOLE_META_PACKAGE_CONTROL_APT=$(
     cat <<EOM
 Package: pihole-meta
-Version: 0.6
+Version: 0.7
 Maintainer: Pi-hole team <adblock@pi-hole.net>
 Architecture: all
 Description: Pi-hole dependency meta package
-Depends: awk,bash-completion,binutils,ca-certificates,cron|cron-daemon,curl,dialog,bind9-dnsutils|dnsutils,dns-root-data,git,grep,iproute2,iputils-ping,jq,libcap2,libcap2-bin,lshw,procps,psmisc,sudo,unzip
+Depends: awk,bash-completion,binutils,ca-certificates,cron|cron-daemon|systemd,curl,dialog,bind9-dnsutils|dnsutils,dns-root-data,git,grep,iproute2,iputils-ping,jq,libcap2,libcap2-bin,logrotate,lshw,procps,psmisc,sudo,unzip
```
rrobgill, line 123: This might cause more problems on debian based distros like MX linux or devuan that don't use systemd but do use apt.

darkexplosiveqwx: If the distribution is really without `systemd` this should be fine.
Some options:
1. `systemd` is installed -> don\`t care if `cron` is installed or not -> uses `systemd.timer`
2. `systemd` is not installed, but `cron`/`cron-daemon` is -> uses `cron.d`
3. None of the three is installed -> install `cron` -> uses `cron.d`
AFAIK for the DEB OR pipe, if none of the packages is currently installed, the first listed package will be installed (as long as it exists in the repos), which is `cron`.

The only edge case I can think of:
`systemd` package is installed, but NOT the active init system AND no process to handle `/etc/cron.d` is available.
This causes `/etc/cron.d/pihole` to be installed, but it will never be read. The fix to tell the users would be to simply install any package reading `/etc/cron.d`.

dschaper:
Why? I don't like potentially breaking something just for the sake of making changes.

DRS1978:
Hello there?

Em sex., 26 de dez. de 2025, 21:59, Dan Schaper ***@***.***>
escreveu:

PR #4489, opened 2021-12-29
Remove pihole-FTL.conf manpage

yubiuser:
- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md), as well as this entire template.
- [x] I have made only one major change in my proposed changes.
- [x] I have commented my proposed changes within the code.
- [x] I have tested my proposed changes, and have included unit tests where possible.
- [x] I am willing to help maintain this change if there are issues with it later.
- [x] I give this submission freely and claim no ownership.
- [x] It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.1)
- [x] I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

Please make sure you [Sign Off](https://docs.pi-hole.net/guides/github/how-to-signoff/) all commits. Pi-hole enforces the [DCO](https://docs.pi-hole.net/guides/github/contributing/).

---
**What does this PR aim to accomplish?:**
Removes the outdated `pihole-FTL.conf` manpage.
FTL's config options change quite often. We document all changes to FTL's config options in `https://https://docs.pi-hole.net/ftldns/configfile/`. Updating them all also in the man pages has been forgotten for over 1 year now and is likely to be forgotten again.  Having them up-to-date online in one place seems sufficient given that Pi-hole is a DNS forwarder.
I added a link to the online resource in the manpage of `pihole-FTL`.

Bildschirmfoto zu 2021-12-29 07-17-20

**How does this PR accomplish the above?:**
Remove the outdated file, update the installer to remove it from existing installations.

PR #6526, opened 2026-02-05
Don't install unused /usr/local/share/man/man5

darkexplosiveqwx:

`/usr/local/share/man/man5` is unused by pihole since `pihole-FTL.conf.5` was removed with https://github.com/pi-hole/pi-hole/pull/4489

`man5`  is for files, but the doc for `pihole.toml` is at docs.pi-hole.net, so I don't think we want to create a man page for it.

For me, removing the directory(if empty) doesn't seem worth it, especially since we don't even clean it up on a failed man install:
```bash
 else
        # Something is wrong with the system's man installation, clean up
        # our files, (leave everything how we found it).
        rm /usr/local/share/man/man8/pihole.8
        printf "%b  %b man page db not updated, man pages not installed\\n" "${OVER}" "${CROSS}"
    fi
```

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1251,10 +1251,6 @@ install_manpage() {
         # if not present, create man8 directory
         install -d -m 755 /usr/local/share/man/man8
     fi
-    if [[ ! -d "/usr/local/share/man/man5" ]]; then
-        # if not present, create man5 directory
-        install -d -m 755 /usr/local/share/man/man5
-    fi
     # Testing complete, copy the files & update the man db
     install -D -m 644 -T ${PI_HOLE_LOCAL_REPO}/manpages/pihole.8 /usr/local/share/man/man8/pihole.8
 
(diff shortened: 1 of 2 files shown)


rdwebdesign:
If you are removing that `install` command, you should also remove the tests for `/usr/local/share/man/man5`:

https://github.com/pi-hole/pi-hole/blob/8c4778025a484fb0e3796759c364f55b6dcf05ef/test/test_any_automated_install.py#L157-L162

darkexplosiveqwx:
> If you are removing that `install` command, you should also remove the tests for `/usr/local/share/man/man5`:
> 
> https://github.com/pi-hole/pi-hole/blob/8c4778025a484fb0e3796759c364f55b6dcf05ef/test/test_any_automated_install.py#L157-L162

Good catch!
After the tests were all green, I simply didn't bother checking them.

Is there something wrong with the tests for not picking this up?

rdwebdesign:
The tests passed because the `/usr/local/share/man/man5` path exists inside the test environment, even when Pi-hole is not using it, but I don't see a reason to test this path if not used.

darkexplosiveqwx:
> The tests passed because the `/usr/local/share/man/man5` path exists inside the test environment, even when Pi-hole is not using it, but I don't see a reason to test this path if not used.

Okay, I was just double-checking in case there was an issue with the tests.
I agree, there is no reason to test for it. My second commit has removed the test.

yubiuser (approved):
Good catch! 
All this was forgotten in https://github.com/pi-hole/pi-hole/pull/4489

PR #6550, opened 2026-02-23
Set versions in /etc/pihole/versions to null if script fails

yubiuser:

When we failed to get the version/branch/hast info in `updatechecker.sh` we returned a mix of `null` or empty vars in `/etc/pihole/versions`.
This PR sets all values to `null` if we fails to get the relevant information.
This is a partly fix raised in https://github.com/pi-hole/web/issues/3728
The second fix is https://github.com/pi-hole/web/pull/3729

---

---

### DIFF
--- a/advanced/Scripts/updatecheck.sh
+++ b/advanced/Scripts/updatecheck.sh
@@ -8,6 +8,8 @@
 # This file is copyright under the latest version of the EUPL.
 # Please see LICENSE file for your rights under this license.
 
+set -o pipefail
+
 function get_local_branch() {
     # Return active branch
     cd "${1}" 2>/dev/null || { echo "null"; return; }
--- a/advanced/Scripts/updatecheck.sh
+++ b/advanced/Scripts/updatecheck.sh
@@ -10,32 +10,32 @@
 
 function get_local_branch() {
     # Return active branch
-    cd "${1}" 2>/dev/null || return 1
-    git rev-parse --abbrev-ref HEAD || return 1
+    cd "${1}" 2>/dev/null || { echo "null"; return; }
+    git rev-parse --abbrev-ref HEAD || echo "null"
 }
 
 function get_local_version() {
     # Return active version
-    cd "${1}" 2>/dev/null || return 1
-    git describe --tags --always 2>/dev/null || return 1
+    cd "${1}" 2>/dev/null || { echo "null"; return; }
+    git describe --tags --always 2>/dev/null || echo "null"
 }
 
 function get_local_hash() {
-    cd "${1}" 2>/dev/null || return 1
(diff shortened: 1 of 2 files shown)


advanced/Scripts/updatecheck.sh
```diff
@@ -10,32 +10,32 @@
...
-        curl -s "https://api.github.com/repos/pi-hole/${1}/releases/latest" 2>/dev/null | jq --raw-output .tag_name || return 1
+        curl -s "https://api.github.com/repos/pi-hole/${1}/releases/latest" 2>/dev/null | jq --raw-output .tag_name || echo "null"
     else
         echo "null"
     fi
 }
 
 function get_remote_hash() {
-    git ls-remote "https://github.com/pi-hole/${1}" --tags "${2}" | awk '{print substr($0, 1,8);}' || return 1
+    git ls-remote "https://github.com/pi-hole/${1}" --tags "${2}" | awk '{print substr($0, 1,8);}' || echo "null"
```
PromoFaux, line 40: Even if the `git ls-remote` command fails (and maybe this is part of the current issue), it's the last command in the pipeline (in this case `awk`) that is evaluated for the exit code.

```sh
#!/usr/bin/env bash
# Demonstrates that || fallback doesn't fire on pipe failure
# when pipefail is not set, because awk exits 0 regardless.

echo "=== Without pipefail (current PR behaviour) ==="

# Simulate get_remote_hash: git fails, awk gets no input, exits 0
result=$(GIT_TERMINAL_PROMPT=0 git ls-remote "https://github.com/pi-hole/DOES-NOT-EXIST" --tags "master" 2>/dev/null | awk '{print substr($0, 1,8);}' || echo "null")
echo "Result: '${result}'"
echo "Expected: 'null', Got: '${result}' (empty — bug!)"

echo ""
echo "=== With pipefail ==="

result=$(set -o pipefail; GIT_TERMINAL_PROMPT=0 git ls-remote "https://github.com/pi-hole/DOES-NOT-EXIST" --tags "master" 2>/dev/null | awk '{print substr($0, 1,8);}' || echo "null")
echo "Result: '${result}'"
echo "Expected: 'null', Got: '${result}'"
```

advanced/Scripts/updatecheck.sh
PromoFaux, line 9:
```suggestion
# Please see LICENSE file for your rights under this license.

set -o pipefail
```

PromoFaux (changes requested):

advanced/Scripts/updatecheck.sh
yubiuser: Good catch!

PromoFaux (approved):

PR #6568, opened 2026-03-17
Replace pytest/tox test suite with BATS

PromoFaux:

May or may not be wanted - we have often talked of replacing the current test suite with something else. I thought I would take a look at BATS, as we use that in the FTL repo already. Got Claude to do the donkey work, but all tested and working well. Personally think it's a little easier to understand/follow/write tests for.

Commit Note from Claude:
The Python-based test infrastructure (pytest, tox, testinfra) is replaced with BATS (Bash Automated Testing System), matching the approach already used in FTL

Changes:
- Add test/run.sh — single entry point replacing all 15 tox.*.ini files; accepts DISTRO env var, builds the test image, installs BATS on demand, and selects test files based on distro family (debian/alpine/rhel)
- Add test/helpers/mocks.bash — bash equivalents of conftest.py's mock_command*, mock_command_2, and mock_command_passthrough helpers; uses base64 transfer to write mock scripts into containers safely
- Add test/test_automated_install.bats — replaces test_any_automated_install.py
- Add test/test_utils.bats — replaces test_any_utils.py
- Add test/test_selinux.bats — replaces test_centos_fedora_common_support.py; only run on CentOS/Fedora (rhel family)
- Remove conftest.py, requirements.txt, setup.py, __init__.py
- Remove all 15 tox.*.ini files
- Remove all three Python test files
- Update .github/workflows/test.yml: drop Python setup, tox invocation, and black formatting check; distro-test job now runs bash test/run.sh
- Update .gitignore: remove Python-specific entries, add test/libs/

---

---


test/run.sh
```diff
@@ -0,0 +1,91 @@
...
+    test_network.bats
+    test_utils.bats
+)
+[[ "$DISTRO_FAMILY" == "rhel" ]] && TEST_FILES+=(test_selinux.bats)
+
+# Use pretty output only when stdout is a real terminal; fall back to TAP in CI.
+# Parallelise across files with --jobs when GNU parallel is available.
+BATS_FLAGS=()
+[[ -t 1 ]] && BATS_FLAGS+=("-p")
+command -v parallel > /dev/null 2>&1 && BATS_FLAGS+=("--jobs" "$(nproc)")
```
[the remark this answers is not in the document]
PromoFaux: Not so worried about this...

.github/workflows/test.yml
```diff
@@ -25,12 +20,13 @@ jobs:
...
         run: |
           IFS=$'\n';
-          for f in $(find . -name '*.sh'); do if [[ ! -x $f ]]; then echo "$f is not executable" && FAIL=1; fi ;done
+          for f in $(find . -name '*.sh' -o -name '*.bats'); do if [[ ! -x $f ]]; then echo "$f is not executable" && FAIL=1; fi ;done
           unset IFS;
           # If FAIL is 1 then we fail.
           [[ $FAIL == 1 ]] && exit 1 || echo "Scripts are executable!"
 
       - name: Differential ShellCheck
+        if: github.event_name == 'pull_request'
```
yubiuser: Why did you add this? Ony 'pull_request` will trigger the workflows, so this seems superfluous

test/helpers/mocks.bash
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env bash
+# Mock command helpers for BATS tests.
+#
+# These are the BATS equivalents of the mock_command* functions in conftest.py.
```
yubiuser:
```suggestion

```

There is not much sense in relating to a file that has just been deleted.

test/run.sh
```diff
@@ -0,0 +1,108 @@
...
+if [[ -z "${DISTRO:-}" ]]; then
+    echo "Error: DISTRO is required."
+    echo "Example: DISTRO=debian_12 bash test/run.sh"
+    echo ""
+    echo "Available distros:"
+    ls _*.Dockerfile | sed 's/^_//;s/\.Dockerfile$//' | sort
+    exit 1
+fi
+
+DOCKERFILE="_${DISTRO}.Dockerfile"
```
yubiuser, line 20:
```suggestion
DOCKERFILE="${DISTRO}.Dockerfile"
```

We could be Cowboys and remove the leading '_' on all these files. I always wondered why they were there at all.

yubiuser (commented):
I think `test/__init__.py‎` can be deleted too

yubiuser (commented):
A general thought: now we run the tests from outside the container, but execute them within. Would it make sense to run the test (as in BATS) from within the container?

test/run.sh
PromoFaux: For one reason, and one reason only... it puts all the dockerfiles in a row when viewed in the file explorer (provided you sort by name!) `_`

image

rdwebdesign: Personally, I think we can keep them.

PromoFaux:
> Would it make sense to run the test (as in BATS) from within the container?

It's a good question, well asked..... Maybe something for a follow up PR. The only negative I can think (maybe) is that it would mean installing BATS in every distro image, might be quite cumbersome

PromoFaux:
Closing in favour of #6598

PR #6581, opened 2026-03-29
Skip apt cache update when pihole-meta is current

PromoFaux:

Stop running `apt-get update` unconditionally on every `pihole -up`. On resource-constrained hardware (SD cards, Raspberry Pi) this is an expensive I/O operation that serves no purpose when the `pihole-meta` dependency package has not changed.

Addresses https://github.com/pi-hole/pi-hole/discussions/6272

---

Adds `PIHOLE_META_VERSION_APT=0.6` as a standalone constant, mirroring the existing `PIHOLE_META_VERSION_APK` pattern already used for Alpine.

The `update_package_cache()` call in `main()` is then gated on whether the installed `pihole-meta` version is absent or older than `PIHOLE_META_VERSION_APT`, using `dpkg-query` and `dpkg --compare-versions`:

```bash
if is_command apt-get; then
    local installed_meta_version
    installed_meta_version=$(dpkg-query -W -f='${Version}' pihole-meta 2>/dev/null)
    if [[ -z "${installed_meta_version}" ]] || \
       dpkg --compare-versions "${installed_meta_version}" lt "${PIHOLE_META_VERSION_APT}"; then
        update_package_cache || exit 1
    fi
fi
```

Behaviour by scenario:

| Scenario | Result |
|---|---|
| Fresh install | `pihole-meta` not present → cache update runs |
| `pihole -up`, meta version unchanged | installed == `PIHOLE_META_VERSION_APT` → skipped |
| `pihole -up`, meta version bumped (new deps added) | installed < `PIHOLE_META_VERSION_APT` → cache update runs |
| Meta package manually removed | not present → cache update runs |

Alpine (`apk`) is unaffected — it was already excluded from this code path by the `is_command apt-get` guard.

---

**Relationship to prior work:**

#6282 separated `update_package_cache()` from `package_manager_detect()` and made the call site explicit, but ultimately kept the unconditional `apt-get update`. The review discussion in that PR raised the concern that skipping the cache update entirely could leave the system unable to resolve new dependencies if the meta package version had changed — a valid point that rules out a simple `fresh_install` guard.

The version-compare approach here directly addresses that concern: the cache is only skipped when the installed `pihole-meta` version already matches what is about to be built, meaning no new dependencies can have been introduced.

The same version-compare idea was also explored in #6383, which was closed due to an Alpine compatibility issue in that implementation. That issue does not apply here.

---

5. I have squashed any insignificant commits.

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -284,8 +284,6 @@ package_manager_detect() {
         UPDATE_PKG_CACHE="${PKG_MANAGER} update"
         # The command we will use to actually install packages
         PKG_INSTALL="${PKG_MANAGER} -qq --no-install-recommends install"
-        # grep -c will return 1 if there are no matches. This is an acceptable condition, so we OR TRUE to prevent set -e exiting the script.
-        PKG_COUNT="${PKG_MANAGER} -s -o Debug::NoLocking=true upgrade | grep -c ^Inst || true"
         # The command we will use to remove packages (used in the uninstaller)
         PKG_REMOVE="${PKG_MANAGER} -y remove --purge"
 
@@ -300,8 +298,6 @@ package_manager_detect() {
 
         # These variable names match the ones for apt-get. See above for an explanation of what they are for.
         PKG_INSTALL="${PKG_MANAGER} install -y"
-        # CentOS package manager returns 100 when there are packages to update so we need to || true to prevent the script from exiting.
-        PKG_COUNT="${PKG_MANAGER} check-update | grep -E '(.i686|.x86|.noarch|.arm|.src|.riscv64)' | wc -l || true"
         # The command we will use to remove packages (used in the uninstaller)
         PKG_REMOVE="${PKG_MANAGER} remove -y"
 
@@ -310,7 +306,6 @@ package_manager_detect() {
         PKG_MANAGER="apk"
         UPDATE_PKG_CACHE="${PKG_MANAGER} update"
         PKG_INSTALL="${PKG_MANAGER} add"
-        PKG_COUNT="${PKG_MANAGER} list --upgradable -q | wc -l"
         PKG_REMOVE="${PKG_MANAGER} del"
 
     else
@@ -1400,23 +1395,6 @@ update_package_cache() {
     fi
 }
 
-# Let user know if they have outdated packages on their system and
-# advise them to run a package update at soonest possible.
-notify_package_updates_available() {
-    # Local, named variables
-    local str="Checking ${PKG_MANAGER} for upgraded packages"
-    printf "\\n  %b %s..." "${INFO}" "${str}"
-    # Store the list of packages in a variable
-    updatesToInstall=$(eval "${PKG_COUNT}")
-
-    if [[ "${updatesToInstall}" -eq 0 ]]; then
-        printf "%b  %b %s... up to date!\\n\\n" "${OVER}" "${TICK}" "${str}"
-    else
-        printf "%b  %b %s... %s updates available\\n" "${OVER}" "${TICK}" "${str}" "${updatesToInstall}"
-        printf "  %b %bIt is recommended to update your OS after installing the Pi-hole!%b\\n\\n" "${INFO}" "${COL_GREEN}" "${COL_NC}"
-    fi
-}
-
 install_dependent_packages() {
     # Install meta dependency package
     local str="Installing Pi-hole dependency package"
@@ -2321,9 +2299,6 @@ main() {
             update_package_cache || exit 1
         fi
     fi
-
-    # Notify user of package availability
-    notify_package_updates_available
 
     # Build dependency package
     build_dependency_package
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -2317,8 +2317,7 @@ main() {
     if is_command apt-get; then
         local installed_meta_version
         installed_meta_version=$(dpkg-query -W -f='${Version}' pihole-meta 2>/dev/null) || true
-        if [[ -z "${installed_meta_version}" ]] || \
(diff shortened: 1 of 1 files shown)


automated install/basic-install.sh
darkexplosiveqwx: Cant we just move `PIHOLE_META_VERSION_APT` above `PIHOLE_META_PACKAGE_CONTROL_APT` and use its value here? This way only one variable need to be bumped
```suggestion
Version: $PIHOLE_META_VERSION_APT
```

yubiuser:
> When bumping pihole-meta to add or remove a dependency, PIHOLE_META_VERSION_APT must be incremented to match the Version: field in PIHOLE_META_PACKAGE_CONTROL_APT. This is what signals to the updater that a fresh cache is needed.

We should have a test for that.

darkexplosiveqwx:
> > When bumping pihole-meta to add or remove a dependency, PIHOLE_META_VERSION_APT must be incremented to match the Version: field in PIHOLE_META_PACKAGE_CONTROL_APT. This is what signals to the updater that a fresh cache is needed.
> 
> We should have a test for that.

This would not be needed with my suggestion.

yubiuser (commented):
Just a thought: should we keep the `notify_package_updates_available()` unconditionally now that we have limited the times when `update_package_cache()` runs? Would it make sens to only run that function when we also updated the package cache? Drawback would be that only `apt`- based systems would ever run it...

automated install/basic-install.sh
```diff
@@ -2305,9 +2309,18 @@ main() {
...
+    # Only update the apt package cache when pihole-meta is absent or outdated.
+    # On a fresh install the package won't exist yet so the cache update always
+    # runs. On upgrades it only runs when the meta version has bumped, meaning
+    # new dependencies may have been added. This saves time and I/O on
+    # resource-constrained hardware (e.g. SD cards).
     if is_command apt-get; then
+        local installed_meta_version
+        installed_meta_version=$(dpkg-query -W -f='${Version}' pihole-meta 2>/dev/null) || true
+        if [[ -z "${installed_meta_version}" ]] || \
+           dpkg --compare-versions "${installed_meta_version}" lt "${PIHOLE_META_VERSION_APT}"; then
```
MichaIng: `dpkg --compare-versions` teats an empty version string as lower, hence it does not need to be checked whether it is empty:
```suggestion
        if dpkg --compare-versions "${installed_meta_version}" lt "${PIHOLE_META_VERSION_APT}"; then
```
_For completeness: the `lt-nl` operator treats an empty version string as **not** lower._

automated install/basic-install.sh
```diff
@@ -2305,9 +2309,18 @@ main() {
...
 
-    # Update package cache only on apt based systems
+    # Only update the apt package cache when pihole-meta is absent or outdated.
+    # On a fresh install the package won't exist yet so the cache update always
+    # runs. On upgrades it only runs when the meta version has bumped, meaning
+    # new dependencies may have been added. This saves time and I/O on
+    # resource-constrained hardware (e.g. SD cards).
     if is_command apt-get; then
+        local installed_meta_version
+        installed_meta_version=$(dpkg-query -W -f='${Version}' pihole-meta 2>/dev/null) || true
```
MichaIng, line 2297: `local` naturally masks the return code of command substitutions, since the `local` command is relevant for the return code of the statement, not the command substitution. Hence, below won't cause a `set -e` exit, unless e.g. the variable is R/O:
```suggestion
        local installed_meta_version=$(dpkg-query -W -f='${Version}' pihole-meta 2>/dev/null)
```
_Sorry for the smart aleck comments on a holiday Friday night 😅._

MichaIng: CI now warns about exactly this 🙈:

> warning: Declare and assign separately to avoid masking return values.

Guess would need to tell it that exactly this return code masking is wanted, or revert 🙁.

MichaIng: Muting this ShellCheck annotation for next line would be:
```sh
        # shellcheck disable=SC2155
        local installed_meta_version=$(dpkg-query -W -f='${Version}' pihole-meta 2>/dev/null)
```
But an extra comment line in exchange for an extra code line ...

An explicit `|| true` or `|| :`, even if redundant, should mute ShellCheck as well.

PromoFaux: Let's just roll it back to where it was

MichaIng:
> Just a thought: should we keep the `notify_package_updates_available()` unconditionally now that we have limited the times when `update_package_cache()` runs? Would it make sens to only run that function when we also updated the package cache? Drawback would be that only `apt`- based systems would ever run it...

It might show "up to date!" even that upgrades are available, due to an outdated package cache 🤔. But at least the notification does not say "run 'apt upgrade'", which can fail if the cache is too old.

Another thought I have is whether the Pi-hole installer needs to inform about available package upgrades at all. Does not hurt, can be valuable for lazy admins, but if additional distro-specific conditionals need to be applied to keep it accurate, it is maybe not worth it, within the scope of the Pi-hole installer?

yubiuser:
I think it was nice to have when we ran `update_package_cache()` every time because we almost had those information ready anyway. Now I think we could also just remove `notify_package_updates_available()`.

___

> It might show "up to date!" even that upgrades are available, due to an outdated package cache

This was exactly what I saw when I tested the branch - it reported up-to-date, but on reverting to `development` it run  `update_package_cache()` and now I had some packages ready to upgrade.

PromoFaux:
> Now I think we could also just remove `update_package_cache()`.

@yubiuser can I confirm this is not a typo? In my mind you mean to remove `notify_package_updates_available`.. I think we still need `update_package_cache` when we have bumped the meta version?

yubiuser:
100% correct. It was a copy& paste error. I edited the comment above to avoid future confusion.

yubiuser (approved):

PR #6598, opened 2026-04-13
Replace pytest/tox with direct in-container BATS

PromoFaux:

An alternative to #6568. WIP

---

---


test/test_automated_install.bats
```diff
@@ -0,0 +1,186 @@
...
+    "
+    assert_success
+
+    local maninstalled=true
+    if [[ "$output" == *"${INFO} man not installed"* ]] || [[ "$output" == *"${INFO} man pages not installed"* ]]; then
+        maninstalled=false
+    fi
+
+    local piholeuser="pihole"
+    _check_perm() { su -s /bin/bash -c "test -${1} ${2}" -p ${piholeuser}; }
```
yubiuser: This could be replaced by https://github.com/bats-core/bats-file

test/helpers/mocks.bash
```diff
@@ -0,0 +1,102 @@
...
+# Write a generated script to /usr/local/bin and clear its log file.
+_write_mock_local() {
+    local script_name="$1" script_content="$2"
+    printf '%s' "$script_content" > "/usr/local/bin/${script_name}"
+    chmod +x "/usr/local/bin/${script_name}"
+    rm -f "/var/log/${script_name}"
+}
+
+# mock_command — matches on $1
+mock_command() {
```
yubiuser: All the mock commands give me a headach. Maybe we can use https://github.com/jasonkarns/bats-mock or https://github.com/capitalone/bash_shell_mock?

yubiuser (commented):
I really like this approach, but would change it a bit. 

1) I would add all the `BATS` installation into the build step (put the cloning code in each Dockerfile, and supply the version as arguments in the build command). 
2) Structure the BATS files in a way that there is no need to 'undo' commands that have modified files within the container. This will require multiple 'docker run' statements as one container is not enough to run all tests, but a few should be enough. E.g. one container could run all tests that only check specific functions, another one could test the install script, ...

test/run.sh
```diff
@@ -0,0 +1,169 @@
...
+# Distro selection
+# ---------------------------------------------------------------------------
+
+if [[ -z "${DISTRO:-}" ]]; then
+    echo "Error: DISTRO is required."
+    echo "Example: DISTRO=debian_12 bash test/run.sh"
+    echo "or:      bash test/run.sh --distro debian_12"
+    echo ""
+    echo "Available distros:"
+    find . -maxdepth 1 -name '_*.Dockerfile' | sed 's|^\./||;s/^_//;s/\.Dockerfile$//' | sort
```
yubiuser, line 55: Can we move this to `usage()`. Otherwise we end up with

```
chris@T14Gen5 ~/S/P/p/test (bats-in-containers)> ./run.sh --distro alpine
Error: Dockerfile not found: _alpine.Dockerfile
```

but no indication which Distros are allowed

test/run.sh
```diff
@@ -0,0 +1,169 @@
...
+    printf '\n=== Suite: %s ===\n' "${label}"
+
+    docker run --rm -t "${IMAGE_TAG}" \
+        bash -euo pipefail -c '
+            cd /etc/.pihole/test
+
+            test_files=("$@")
+
+            # BATS libraries are pre-installed into the image at build time.
+            # Copy them to /tmp/tests so installer tests that mutate /etc/.pihole
```
yubiuser: Do we really need this copy part here? I removed it and it worked fine locally. If you want to play safe, why not clone all the BATS stuff directly to `/tmp/libs` instead of `/etc/.pihole`?

PromoFaux: No - so this was defensive owing to a misunderstanding about the volitility (or lack of) of /etc/.pihole when the installer test runs. We don't need to copy anything at all - and have verified that with a test myself!

test/run.sh
PromoFaux: Yep - simple fix :)

yubiuser (commented):
`/test/README.md` needs adjustment

test/test_fresh_install.bats
```diff
@@ -0,0 +1,124 @@
...
+    command -v yum > /dev/null && yum install -y man || true
+    command -v apk > /dev/null && apk add mandoc man-pages || true
+
+    echo "${FTL_BRANCH}" > /etc/pihole/ftlbranch
+
+    run bash -c "
+        export TERM=xterm
+        export DEBIAN_FRONTEND=noninteractive
+        umask 0027
+        runUnattended=true
```
yubiuser: I still wonder why we have/need `runUnattended=true` twice here`

yubiuser (commented):
I think you force pushed away the BATS parallel execution at some time. If this was on purpose the tests within each BATS file will execute in order, so we wouldn't need `_reset_automated_install_test_state()` except for the first test in `automated_install.bats`

test/test_fresh_install.bats
rdwebdesign: I'm not sure why line 54 is here.

`/opt/pihole/basic-install.sh` is sourced on the next line (55) and [it always sets the variable as `false`](https://github.com/pi-hole/pi-hole/blob/d3bbfea9c00523723dfee788280b69770bbcd8bd/automated%20install/basic-install.sh#L190), before checking if the "--unattended" option is set.

PromoFaux: It's been like this for 5 years :D

yubiuser: But now is a great opportunity to fix it.

yubiuser (dismissed):

yubiuser:
I've created test cases for gravity. Do you wan to get this PR merged first or should I commit what I have to this branch?

PromoFaux:
Sorry - missed the approve and then there were merge conflicts - seeking another approval for merge

yubiuser:
@PromoFaux 

I've created test cases for gravity. Do you wan to get this PR merged first or should I commit what I have to this branch?

PromoFaux:
Lets do them as a separate PR

yubiuser (dismissed):
One last rebase, then ready to merge

yubiuser (approved):

PR #6599, opened 2026-04-13
test: replace pytest/tox with direct in-container BATS (insanity edition)

PromoFaux:
Nothing to see here...


PromoFaux:
Don't take this one seriously - I had an itch that needed scratching.

Expected a lot of them to fail purely based on arch availability for the container images

rdwebdesign:
Pure curiosity:

What is your intention here? What do you want to achieve exactly?

PromoFaux:
So I was looking at the part of the tests that tests the detection of the platform for downloading the correct FTL and thought to myself... "I wonder what it would look like if we just ran that against the correct arch for that distro...."

Obviously there are a lot of gaps because images just don't exist but, I've never been one to let "You can't do that" stop me

yubiuser:
I'll just skip reviewing this...

PromoFaux:
Yeah, this one is too much - I was just curious about the possibility of testing all arch's natively rather than mocking the arch to test the detection code, but it's wayyyy OTT

PR #6605, opened 2026-04-18
Improve gravity error message including curl exit code and errormsg

rdwebdesign:

Improve gravity error messages for `curl`, returning better error messages (when available) based on the exit code and HTTP code.

The current gravity uses `http_code` returned by `curl`, but `curl` answers with `000` for many different reasons, like:

- Timeout: exit_code=28
- DNS Resolution Failure: exit_code=6
- Connection Refused: exit_code=7
- Network Unreachability: exit_code=7 (different error message)
- Untrusted Certificate: exit_code=35 or 60 (or other values... I'm not sure).
- and others: "CA Certificate Issues", "Invalid URL", "Client Disconnection"...

Gravity shows "Connection Refused" for all these cases, but sometimes "Connection Refused" is not the correct error message.

This PR returns the correct exit code and uses HTTP code to show a message returned by `curl` (when available - `curl 7.75.0` and above).

curl_messages

Use `curl` exit code (`$?`) to decide if `curl` command was successful or not.

Then we use the `http_code` returned by `curl` to define what is the correct error message.
Also, when using recent `curl` versions, we use the `errormsg` variable to show the message returned by `curl`.

### Note

This PR replaces #6425, where you can find the related discussion and suggestions.

---

---

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -814,7 +814,7 @@ gravity_DownloadBlocklistFromUrl() {
       case "${httpCode}" in
         "200") echo -e "${OVER}  ${TICK} ${str} Retrieval successful" ;;
         "304") echo -e "${OVER}  ${TICK} ${str} No changes detected"  ;;
-        *) echo -e "${OVER}  ${TICK} ${str} Success (http_code=${COL_RED}${httpCode}${COL_NC})"  ;;
+        *) echo -e "${OVER}  ${TICK} ${str} Success (http_code=${COL_CYAN}${httpCode}${COL_NC})"  ;;
       esac
       success=true
     else
@@ -827,7 +827,7 @@ gravity_DownloadBlocklistFromUrl() {
         "504") echo -e "${OVER}  ${CROSS} ${str} Connection Timed Out (Gateway)" ;;
         "521") echo -e "${OVER}  ${CROSS} ${str} Web Server Is Down (Cloudflare)" ;;
         "522") echo -e "${OVER}  ${CROSS} ${str} Connection Timed Out (Cloudflare)" ;;
-        *) echo -e "${OVER}  ${CROSS} ${str} Retrieval failed (exit_code=${COL_RED}${curlExitCode}${COL_NC} Msg: ${COL_CYAN}${curlErrorMsg}${COL_NC})" ;;
+        *) echo -e "${OVER}  ${CROSS} ${str} Retrieval failed (exit_code=${COL_CYAN}${curlExitCode}${COL_NC} Msg: ${COL_CYAN}${curlErrorMsg}${COL_NC})" ;;
       esac
     fi
     ;;
--- a/gravity.sh
+++ b/gravity.sh
@@ -776,10 +776,14 @@ gravity_DownloadBlocklistFromUrl() {
     # (https://github.com/pi-hole/pi-hole/pull/6605#discussion_r3112153347)
     # First we get the current curl version.
     curlVersion=$(curl --version | awk '{print $2;exit}')
-    # After that, we add "7.75", creating a list with 2 items.
-    # Then we sort the list (ascending order) and return the first item.
-    # If "7.75" is returned, it means that the current version is greater than or equal to "7.75.0".
-    if [[ "$(printf '%s\n' "${curlVersion}" 7.75 | sort -V | head -n1)" == 7.75 ]]; then
+    # After that, we pipe the current version along with the string '7.75' (the minimum version supporting the required option.)
+    # Then we sort the list in natural (version) order and return the first item which will be the lowest version seen.
+    # If it is "7.75" then the current version is greater than or equal to "7.75.0".
+    # Compatibility notes:
+    #   Busybox doesn't support some long flags:
+    #   - "sort -V" is short form of "sort --version-sort"
+    #   - "head -n1" is short form of "head --lines=1"
+    if [[ "$(printf '%s\n' "${curlVersion}" "7.75" | sort -V | head -n1)" == 7.75 ]]; then
         # Use the error message returned by curl
         curlOutputFormat='%{http_code}\n%{errormsg}'
     fi
--- a/gravity.sh
+++ b/gravity.sh
@@ -770,7 +770,7 @@ gravity_DownloadBlocklistFromUrl() {
 
   if [[ "${download}" == true ]]; then
     # Define the generic error message
-    curlOutputFormat='%{http_code}\nNo message available. Non supported curl version.'
+    curlOutputFormat='%{http_code};No message available. Non supported curl version.'
 
     # Check if the installed curl version supports the "-w %{errormsg}" option (available as of curl 7.75.0)
     # (https://github.com/pi-hole/pi-hole/pull/6605#discussion_r3112153347)
@@ -785,7 +785,7 @@ gravity_DownloadBlocklistFromUrl() {
     #   - "head -n1" is short form of "head --lines=1"
     if [[ "$(printf '%s\n' "${curlVersion}" "7.75" | sort -V | head -n1)" == 7.75 ]]; then
         # Use the error message returned by curl
-        curlOutputFormat='%{http_code}\n%{errormsg}'
+        curlOutputFormat='%{http_code};%{errormsg}'
     fi
 
     # This command will output the HTTP code and an error message, if available.
@@ -799,10 +799,7 @@ gravity_DownloadBlocklistFromUrl() {
   fi
 
   # Retrieve http_code and errormsg values, returned by curl command
-  {
-    read -r httpCode;
-    read -r curlErrorMsg;
-  } < <( echo "${curlOutput}" )
+  IFS=";" read -r httpCode curlErrorMsg <<<"$curlOutput"
 
   case $url in
   # Did we "download" a local file?
--- a/gravity.sh
+++ b/gravity.sh
@@ -772,9 +772,13 @@ gravity_DownloadBlocklistFromUrl() {
     # Define the generic error message
     curlOutputFormat='%{http_code}\nNo message available. Non supported curl version.'
 
-    # Check if the installed curl version supports the "-w %{errormsg}" option
-    # (available as of curl 7.75.0)
+    # Check if the installed curl version supports the "-w %{errormsg}" option (available as of curl 7.75.0)
+    # (https://github.com/pi-hole/pi-hole/pull/6605#discussion_r3112153347)
+    # First we get the current curl version.
     curlVersion=$(curl --version | awk '{print $2;exit}')
+    # After that, we add "7.75", creating a list with 2 items.
+    # Then we sort the list (ascending order) and return the first item.
+    # If "7.75" is returned, it means that the current version is greater than or equal to "7.75.0".
     if [[ "$(printf '%s\n' "${curlVersion}" 7.75 | sort -V | head -n1)" == 7.75 ]]; then
         # Use the error message returned by curl
         curlOutputFormat='%{http_code}\n%{errormsg}'
--- a/gravity.sh
+++ b/gravity.sh
@@ -614,7 +614,7 @@ compareLists() {
 # Download specified URL and perform checks on HTTP status and file content
 gravity_DownloadBlocklistFromUrl() {
   local url="${1}" adlistID="${2}" saveLocation="${3}" compression="${4}" gravity_type="${5}" domain="${6}"
-  local listCurlBuffer str curlOutput httpCode curlErrorMsg="" curlExitCode="" curlOutputFormat=""
+  local listCurlBuffer str curlVersion curlOutput httpCode curlErrorMsg="" curlExitCode="" curlOutputFormat=""
   local success="" ip customUpstreamResolver="" file_path permissions ip_addr port blocked=false download=true
   # modifiedOptions is an array to store all the options used to check if the adlist has been changed upstream
   local modifiedOptions=()
@@ -758,10 +758,14 @@ gravity_DownloadBlocklistFromUrl() {
   fi
 
   if [[ "${download}" == true ]]; then
+    # Define the generic error message
+    curlOutputFormat='%{http_code}\nNo message available. Non supported curl version.'
+
     # Check if the installed curl version supports the "-w %{errormsg}" option
     # (available as of curl 7.75.0)
-    curlOutputFormat='%{http_code}\nnomsg'
-    if curl --help --write-out | grep -q "errormsg" ; then
+    curlVersion=$(curl --version | awk '{print $2;exit}')
+    if [[ "$(printf '%s\n' "${curlVersion}" 7.75 | sort -V | head -n1)" == 7.75 ]]; then
+        # Use the error message returned by curl
         curlOutputFormat='%{http_code}\n%{errormsg}'
     fi
 
@@ -781,11 +785,6 @@ gravity_DownloadBlocklistFromUrl() {
     read -r curlErrorMsg;
   } < <( echo "${curlOutput}" )
 
-  # If curl is too old to return errormsg, we provide a generic message here
-  if [[ "${curlErrorMsg}" == "nomsg" ]]; then
-    curlErrorMsg="No message available. Non supported curl version."
-  fi
-
   case $url in
   # Did we "download" a local file?
   "file"*)
@@ -804,6 +803,7 @@ gravity_DownloadBlocklistFromUrl() {
       case "${httpCode}" in
         "200") echo -e "${OVER}  ${TICK} ${str} Retrieval successful" ;;
         "304") echo -e "${OVER}  ${TICK} ${str} No changes detected"  ;;
+        *) echo -e "${OVER}  ${TICK} ${str} Success (http_code=${COL_RED}${httpCode}${COL_NC})"  ;;
       esac
       success=true
     else
--- a/gravity.sh
+++ b/gravity.sh
@@ -614,8 +614,8 @@ compareLists() {
 # Download specified URL and perform checks on HTTP status and file content
 gravity_DownloadBlocklistFromUrl() {
   local url="${1}" adlistID="${2}" saveLocation="${3}" compression="${4}" gravity_type="${5}" domain="${6}"
-  local listCurlBuffer str httpCode success="" ip customUpstreamResolver=""
-  local file_path permissions ip_addr port blocked=false download=true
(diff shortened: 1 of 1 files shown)


gravity.sh
```diff
@@ -758,9 +758,33 @@ gravity_DownloadBlocklistFromUrl() {
...
+        # Use the error message returned by curl
+        curlOutputFormat='%{http_code}\n%{errormsg}'
+    fi
+
+    # This command will output the HTTP code and an error message, if available.
+    # Error messages are suppressed by "-s" option.
+    # By default curl returns exitcode=0 even an HTTP code happens (403, 404, 500, etc). To fix
+    # this, we use the "--fail" option to force curl to return a non-zero exit code.
+    # If curl version is older than 7.75.0, curl can't generate the errormsg output. In this case,
+    # the command will return "nomsg" and this will be replaced later by a generic error message.
```
MichaIng: As the comment describes an old state of the PR:
```suggestion
    # a generic message is returned.
```

MichaIng:
Retested on Debian Bullseye (curl <7.75.0) and Bookworm (curl >7.75.0), and it works now as expected:

#### Bullseye
```
  [i] Target: https://raw.githubusercontent.df
  [✗] Status: Retrieval failed (exit_code=6 Msg: No message available. Non supported curl version.)

  [i] Target: http://
  [✗] Status: Retrieval failed (exit_code=3 Msg: No message available. Non supported curl version.)

  [i] Target: http://10.1.1.2/list
  [✗] Status: Retrieval failed (exit_code=28 Msg: No message available. Non supported curl version.)

  [i] Target: http://localhost:81/list
  [✗] Status: Retrieval failed (exit_code=7 Msg: No message available. Non supported curl version.)

  [✗] Status: Retrieval failed (exit_code=60 Msg: No message available. Non supported curl version.)

  [i] Target: https://dietpi.com/nonexistent
  [✗] Status: Not found

  [i] Target: https://dietpi.com/.well-known/
  [✗] Status: Forbidden
```

#### Bookworm
```
  [i] Target: https://raw.githubusercontent.df
  [✗] Status: Retrieval failed (exit_code=6 Msg: Could not resolve host: raw.githubusercontent.df)

  [i] Target: http://
  [✗] Status: Retrieval failed (exit_code=3 Msg: URL using bad/illegal format or missing URL)

  [i] Target: http://10.1.1.2/list
  [✗] Status: Retrieval failed (exit_code=28 Msg: Failed to connect to 10.1.1.2 port 80 after 10014 ms: Timeout was reached)

  [i] Target: http://localhost:81/list
  [✗] Status: Retrieval failed (exit_code=7 Msg: Failed to connect to localhost port 81 after 0 ms: Couldn't connect to server)

  [✗] Status: Retrieval failed (exit_code=60 Msg: SSL certificate problem: self-signed certificate in certificate chain)

  [i] Target: https://dietpi.com/nonexistent
  [✗] Status: Not found

  [i] Target: https://dietpi.com/.well-known/
  [✗] Status: Forbidden
```

yubiuser (dismissed):

casperklein:
Just a suggestion, withdraw it if you don't like it.

You could use ``;`` as delimiter, which makes it a bit simpler IMO:

```bash

curlOutputFormat='%{http_code};%{errormsg}'

curlOutput=$(curl ..)

IFS=";" read -r httpCode curlErrorMsg <<<"$curlOutput"
```

yubiuser (dismissed):

gravity.sh
```diff
@@ -784,27 +808,28 @@ gravity_DownloadBlocklistFromUrl() {
...
-      ;;
-    "304")
-      echo -e "${OVER}  ${TICK} ${str} No changes detected"
+    # Use the exit code to determine if curl was successful or not.
+    # Use HTTP code only to select the correct error message.
+    if [[ "${curlExitCode}" == "0" ]]; then
+      case "${httpCode}" in
+        "200") echo -e "${OVER}  ${TICK} ${str} Retrieval successful" ;;
+        "304") echo -e "${OVER}  ${TICK} ${str} No changes detected"  ;;
+        *) echo -e "${OVER}  ${TICK} ${str} Success (http_code=${COL_RED}${httpCode}${COL_NC})"  ;;
```
PromoFaux: Can we go with yellow, or no colour at all for the `httpCode` output? Seems confusing to have a success message with red text in it...

gravity.sh
```diff
@@ -769,9 +769,33 @@ gravity_DownloadBlocklistFromUrl() {
...
+        curlOutputFormat='%{http_code}\n%{errormsg}'
+    fi
+
+    # This command will output the HTTP code and an error message, if available.
+    # Error messages are suppressed by "-s" option.
+    # By default curl returns exitcode=0 even an HTTP code happens (403, 404, 500, etc). To fix
+    # this, we use the "--fail" option to force curl to return a non-zero exit code.
+    # If curl version is older than 7.75.0, curl can't generate the errormsg output. In this case,
+    # a generic message is returned.
+    curlOutput=$(curl --connect-timeout ${curl_connect_timeout} -s --fail -L ${compression:+${compression}} ${customUpstreamResolver:+${customUpstreamResolver}} "${modifiedOptions[@]}" -w "${curlOutputFormat}" "${url}" -o "${listCurlBuffer}")
```
PromoFaux, line 793: > Error messages are suppressed by "-s" option

Are we sure? Should probably add the `2>/dev/null` back in?

```
adam@adam-desktop:~$ curl --silent
curl: (2) no URL specified
curl: try 'curl --help' or 'curl --manual' for more information
adam@adam-desktop:~$ echo $?
2
adam@adam-desktop:~$ curl --silent 2>/dev/null
adam@adam-desktop:~$ echo $?
2
```

PromoFaux (changes requested):

gravity.sh
rdwebdesign: This was removed after a [suggestion made on the other PR](https://github.com/pi-hole/pi-hole/pull/6425#discussion_r2443367043).

This is actually hiding the errors we expect to encounter, like using an invalid URL:
```sh
# Without -s option:
rdweb@pi4b:~ $ curl http://
curl: (3) URL rejected: No host part in the URL
rdweb@pi4b:~ $

# With -s option:
rdweb@pi4b:~ $ curl -s http://
rdweb@pi4b:~ $
```

It will only show really unexpected errors, if something is damaged or if we are using a very old `curl` version that doesn't understand some basic options. In this case, I think hiding all errors is bad for diagnosing issues.

gravity.sh
rdwebdesign: Yeah... 

I will use blue (cyan). No need to use a different color.

Blue is the color used for the messages and the color used for "information" in the web interface.

rdwebdesign: I will also change the exit code color when it is different than zero. We already have a red "X", on errors.

gravity.sh
PromoFaux: OK, conceded and point withdrawn!

gravity.sh
```diff
@@ -769,9 +769,33 @@ gravity_DownloadBlocklistFromUrl() {
...
 
   if [[ "${download}" == true ]]; then
-    httpCode=$(curl --connect-timeout ${curl_connect_timeout} -s -L ${compression:+${compression}} ${customUpstreamResolver:+${customUpstreamResolver}} "${modifiedOptions[@]}" -w "%{http_code}" "${url}" -o "${listCurlBuffer}" 2>/dev/null)
+    # Define the generic error message
+    curlOutputFormat='%{http_code}\nNo message available. Non supported curl version.'
+
+    # Check if the installed curl version supports the "-w %{errormsg}" option
+    # (available as of curl 7.75.0)
+    curlVersion=$(curl --version | awk '{print $2;exit}')
+    if [[ "$(printf '%s\n' "${curlVersion}" 7.75 | sort -V | head -n1)" == 7.75 ]]; then
```
dschaper: Seems a bit convoluted but if you want to keep it please add a comment to explain what is happening and why.

dschaper (changes requested):

gravity.sh
MichaIng: The comments explain why the version check is done, but not the code, indeed. For completeness, the related discussion in the old PR: https://github.com/pi-hole/pi-hole/pull/6425#discussion_r3105724017

So the previous approach to know whether `-w %{errormsg}` is supported used `curl --help`, which however changed as well in some later `curl` version, so that a number of versions (notably the one shipped with Debian Bookworm) were not detected correctly to support `-w %{errormsg}`. Hence, the suggestion was to check the actual `curl` version. But that is not that easy in a distro-agnostic way. Printing it along with the breaking point version 7.75 on two lines, and sorting them via `sort -V` (`-V` => `--version-sort`) is the best I could think of:
- `sort` sorts ascending.
- Empty strings (if `curlVersion` failed to be obtained) are treated as lowest version, i.e. show up first.
- The first line (`head -n1`) is hence "7.75" only if a larger `curl` version could be obtained.
- If no `curl` version could be obtained, the first line is an empty string, and if a lower version was obtained, that one will be the first line, hence the condition is false in those two cases.

rdwebdesign: I added a comment explaining the code and pointing to this github comment.

PromoFaux: I'm guessing we can't just easily test directly if `-w %{errormsg}` is supported by "trying it" and detecting an error message?

rdwebdesign: I think code complexity won't be very different.

Other curl functions were tested using [`curl --help all`](https://github.com/pi-hole/pi-hole/blob/16ec32275b3301727f7725763c46a0779cd2307f/gravity.sh#L529-L531), but in some specific packages (like the one used for Bookworm), the `-w %{errormsg}` option doesn't exist on the help text.

We know for sure that curl supports this option starting from `7.75.0`. We just need to compare the version.

PromoFaux (dismissed):

gravity.sh
PromoFaux: Works for me, but I'll let dan resolve the thread himself if he agrees :)

MichaIng: > I'm guessing we can't just easily test directly if `-w %{errormsg}` is supported by "trying it" and detecting an error message?

I thought about the same, like catching the respective error code or message from STDERR, and using that as indicator to throw the "No message available. Non supported curl version." message. But the error code is not specific, and capturing STDERR and STDOUT separately, is as well very complex in shell, even harder to understand: https://github.com/pi-hole/pi-hole/pull/6425#issuecomment-3418463232

gravity.sh
```diff
@@ -769,9 +769,37 @@ gravity_DownloadBlocklistFromUrl() {
...
+    # Define the generic error message
+    curlOutputFormat='%{http_code}\nNo message available. Non supported curl version.'
+
+    # Check if the installed curl version supports the "-w %{errormsg}" option (available as of curl 7.75.0)
+    # (https://github.com/pi-hole/pi-hole/pull/6605#discussion_r3112153347)
+    # First we get the current curl version.
+    curlVersion=$(curl --version | awk '{print $2;exit}')
+    # After that, we add "7.75", creating a list with 2 items.
+    # Then we sort the list (ascending order) and return the first item.
+    # If "7.75" is returned, it means that the current version is greater than or equal to "7.75.0".
```
dschaper:
```suggestion
    # After that, we pipe the current version along with the string '7.75' (the minimum version supporting the required option.)
    # Then we sort the list in natural (version) order and return the first item which will be the lowest version seen. 
    # If it is "7.75" then the current version is greater than or equal to "7.75.0".
```

gravity.sh
```diff
@@ -769,9 +769,37 @@ gravity_DownloadBlocklistFromUrl() {
...
+    if [[ "$(printf '%s\n' "${curlVersion}" 7.75 | sort -V | head -n1)" == 7.75 ]]; then
```
dschaper:
```suggestion
    if [[ "$(printf '%s\n' "${curlVersion}" "7.75" | sort --version-sort | head --lines=1)" == 7.75 ]]; then
```

dschaper (changes requested):

gravity.sh
PromoFaux: This inspired me to go looking at the rest of the scripts for short -> longform switch conversion (will followup in another PR if desired)

Unfortunately these particular changes would prevent gravity running on bare metal alpine - though I agree we should use the longform where we can for readability. 

```
(base) PS C:\Users\micro> docker run --rm busybox sh -lc "cat /etc/hosts | head --lines=1"
head: invalid option -- '-'
```

```
(base) PS C:\Users\micro> docker run --rm busybox sh -lc "cat /etc/hosts | sort --version-sort"
sort: unrecognized option '--version-sort'
```

MichaIng: Weird, Alpine `sort` and `head` support the short form only? Is this BusyBox? Yeah seems so:
- https://manpages.debian.org/trixie/busybox/busybox.1.en.html#head
- https://manpages.debian.org/trixie/busybox/busybox.1.en.html#sort

So BusyBox tools support mostly short form options only 🤔.

PromoFaux: It's a _general_ rule of thumb, except for the exception. e.g `mkdir --parents` IS supported under Busybox (despite the help output saying otherwise)

```
(base) PS C:\Users\micro> docker run --rm alpine sh -lc "mkdir --parents /tmp/etc/test/one/two && echo $?"
True
```

MichaIng: Good to know. So to assure cross-compatibility, short forms are actually better for commands, which are no shell builtins. But less self-explaining.

PromoFaux: Playing with something....

```
adam@adam-desktop:/mnt/e/source/pi-hole-workspace/pi-hole$ bash test/lint-switch-style.sh
gravity.sh:595: sha1sum option style violation: got --check, expected -c
gravity.sh:595: sha1sum option style violation: got --status, expected -s
Found 2 switch style violation(s).
```

PromoFaux: POC here. https://github.com/pi-hole/pi-hole/pull/6608 

Might be something in it, might not.

https://github.com/pi-hole/pi-hole/actions/runs/24693695900/job/72221476949?pr=6608

dschaper:
@rdwebdesign Can you do this as well, I think it's better to use the `;` as the delim instead of a newline because it's possible for the output to contain a newline at some point and that would cause issues. If we declare the IFS to be a certain char then we can be safe to assume that `;` will not appear in the output and can be relied on as the field separator.

> Just a suggestion, withdraw it if you don't like it.
> 
> You could use `;` as delimiter, which makes it a bit simpler IMO:
> 
> ```shell
> curlOutputFormat='%{http_code};%{errormsg}'
> 
> curlOutput=$(curl ..)
> 
> IFS=";" read -r httpCode curlErrorMsg <<<"$curlOutput"
> ```

gravity.sh
dschaper: I was afraid that `busybox` would bite us again. Can you add https://pkgs.alpinelinux.org/package/v3.23/main/x86_64/coreutils to the alpine images? It does add 1MB to the size but it includes the GNU versions of about 100 commands. 

I think the size is acceptable in exchange for being able to use the full flags. I guess the alternative would be to use the short form but add a comments explaining the flags.

```
head -n1 == head --lines=1
```

rdwebdesign:
Maybe I can use just a space as separator.

The `%{http_code}` can never contain spaces or be empty.
It will always be a string with 3 characters.
Even when there is no HTTP code because `curl` couldn't complete the connection, the code will be "000".

The part before the first space will be stored in the first variable.
The rest of the text will be on the second one:

```sh
rdweb@pi4b:~ $ read -r a b <<<"000 Message containing spaces and all kind of characters" && echo "A=$a B=$b"
A=000 B=Message containing spaces and all kind of characters
```

https://onlinegdb.com/h-DYDb16_

dschaper:
You get some odd and unexpected results when you separate based on whitespace. I think it's better to use `;` since bash already uses that for statement separation.

rdwebdesign:
OK.

dschaper:
Thanks! Can you add @casperklein as co-author on that last semicolon commit? 

And then squash out the commits you can and force it up. We still need to solve the alpine issue with @PromoFaux up in https://github.com/pi-hole/pi-hole/pull/6605#discussion_r3113899424 but that's the last part.

gravity.sh
```diff
@@ -776,10 +776,10 @@ gravity_DownloadBlocklistFromUrl() {
...
     # First we get the current curl version.
     curlVersion=$(curl --version | awk '{print $2;exit}')
-    # After that, we add "7.75", creating a list with 2 items.
-    # Then we sort the list (ascending order) and return the first item.
-    # If "7.75" is returned, it means that the current version is greater than or equal to "7.75.0".
-    if [[ "$(printf '%s\n' "${curlVersion}" 7.75 | sort -V | head -n1)" == 7.75 ]]; then
+    # After that, we pipe the current version along with the string '7.75' (the minimum version supporting the required option.)
+    # Then we sort the list in natural (version) order and return the first item which will be the lowest version seen.
+    # If it is "7.75" then the current version is greater than or equal to "7.75.0".
+    if [[ "$(printf '%s\n' "${curlVersion}" "7.75" | sort --version-sort | head --lines=1)" == 7.75 ]]; then
```
PromoFaux: @rdwebdesign - Per the [discussion below the original change request](https://github.com/pi-hole/pi-hole/pull/6605#discussion_r3113899424) here -  these (`-V` -> `--version-sort` and `-n1` -> `--lines=1`) should **not** have been changed from as it will break alpine compatibility in gravity.

Please revert for the time being - we can address long vs shortform of commands in another PR.

Just document the shorthand like this:

https://github.com/pi-hole/pi-hole/blob/d3bbfea9c00523723dfee788280b69770bbcd8bd/gravity.sh#L122-L124

PromoFaux (changes requested):

gravity.sh
rdwebdesign: Done.

rdwebdesign:
@dschaper 

Co-author added.
Also, I fixed the Alpine/Busybox compatibility and squashed this new change into [a7baecb](https://github.com/pi-hole/pi-hole/pull/6605/commits/a7baecb57cb3c97e3c89bb92163738187acf067c) commit.

I'm not sure if I should squash other commits because I want to avoid conflicts (there is a merge commit on the middle).

dschaper:
Thanks again, and if you don't want to squash out any commits and feel they are all needed then we can go with the PR as is.

dschaper (approved):

PromoFaux (approved):

PR #6607, opened 2026-04-19
Prevent double error message output in gravity run with invalid file

PromoFaux:

(View with without whitespace to see we're just moving a `fi` down 30 or so lines!)

Discovered an existing bug while testing #6605, where an adlist with an invalid protocol produces two error lines in gravity output - one correct, one spurious.

When a URL fails the protocol check (lines 752–756), download is set to false and an appropriate error is printed. However, because the `case $url in` block ran unconditionally after the curl guard, the remote URL (*) branch would also execute with an uninitialised `httpCode`, producing a second meaningless error line.

This was verified against the live `pihole/pihole:latest` Docker image with an adlist entry of `invalidproto://example.com/list.txt`

Before (current `master`):
```
[✗] Status: Invalid protocol specified. Ignoring list.
      Ensure your URL starts with a valid protocol like http:// , https:// or file://
[✗] Status: invalidproto://example.com/list.txt ()
```

After this fix:
```
[✗] Status: Invalid protocol specified. Ignoring list.
    Ensure your URL starts with a valid protocol like http:// , https:// or file://
```

Note: this bug also affects the `file://` branches (file not found, bad permissions) in the same way — those checks already print their own error before setting `download=false`, but the case block would still fire with an empty `listCurlBuffer`, printing "Retrieval failed / empty list" as a second line.

---

---

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -796,47 +796,48 @@ gravity_DownloadBlocklistFromUrl() {
     # a generic message is returned.
     curlOutput=$(curl --connect-timeout ${curl_connect_timeout} -s --fail -L ${compression:+${compression}} ${customUpstreamResolver:+${customUpstreamResolver}} "${modifiedOptions[@]}" -w "${curlOutputFormat}" "${url}" -o "${listCurlBuffer}")
     curlExitCode="$?"
+
+
+    # Retrieve http_code and errormsg values, returned by curl command
+    IFS=";" read -r httpCode curlErrorMsg <<<"$curlOutput"
+
+    case $url in
+    # Did we "download" a local file?
+    "file"*)
+      if [[ -s "${listCurlBuffer}" ]]; then
+        echo -e "${OVER}  ${TICK} ${str} Retrieval successful"
+        success=true
+      else
+        echo -e "${OVER}  ${CROSS} ${str} Retrieval failed / empty list"
+      fi
+      ;;
+    # Did we "download" a remote file?
+    *)
+      # Use the exit code to determine if curl was successful or not.
+      # Use HTTP code only to select the correct error message.
+      if [[ "${curlExitCode}" == "0" ]]; then
+        case "${httpCode}" in
+          "200") echo -e "${OVER}  ${TICK} ${str} Retrieval successful" ;;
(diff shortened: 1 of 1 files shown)


rdwebdesign:
Why don't you just include a commit to 6605?

PromoFaux:
Because I only discovered it while testing that PR, but didn't want to confuse matters as it is not a regression introduced by that PR, but an already existing bug.

Just feels like it belongs separately to that PR

rdwebdesign:
Do you want to merge 6605 first and rebase this one? or do you prefer to merge this one first?

gravity.sh
```diff
@@ -764,49 +764,49 @@ gravity_DownloadBlocklistFromUrl() {
   # Check for allowed protocols
   if [[ $url != "http"* && $url != "https"* && $url != "file"* && $url != "ftp"* && $url != "ftps"* && $url != "sftp"* ]]; then
     echo -e "${OVER}  ${CROSS} ${str} Invalid protocol specified. Ignoring list."
-    echo -e "      Ensure your URL starts with a valid protocol like http:// , https:// or file:// ."
+    echo -e "    Ensure your URL starts with a valid protocol like http:// , https:// or file:// ."
```
rdwebdesign, line 767: I don't think we need this change:

```suggestion
    echo -e "      Ensure your URL starts with a valid protocol like http:// , https:// or file:// ."
```


This is how it is shown:

CLI:
Image

Web interface:
image

rdwebdesign (changes requested):

PromoFaux:
I think that's the merge conflict sorted...

yubiuser (approved):
Before

```
  [i] Target: dadfasdfsdafsf.com
  [✗] Status: Invalid protocol specified. Ignoring list.
      Ensure your URL starts with a valid protocol like http:// , https:// or file:// .
  [✗] Status: Retrieval failed (exit_code= Msg: )
```

After

```
  [i] Target: dadfasdfsdafsf.com
  [✗] Status: Invalid protocol specified. Ignoring list.
      Ensure your URL starts with a valid protocol like http:// , https:// or file:// .
```

rdwebdesign (approved):

PR #6609, opened 2026-04-24
fix: honor quiet mode in log flush helpers

### LINKED ISSUE #6289: message output with parameter 'quiet' set
### Versions
- Core version is v6.1.2 (Latest: v6.1.2)
- Web version is v6.2.1 (Latest: v6.2.1)
- FTL version is v6.2.2 (Latest: v6.2.2)

### Platform
- OS and version: Debian Bookworm
- Platform: Bare Metal native x86_64

### Expected behavior
 On call of pihole with parameter 'quiet' I don't expected output of status messages because this sends out emails on the cron job running.

### Actual behavior / bug

On call of pihole for rotating (aka. piholeLogFlush.sh) the logs in the cron job there are output messages even when the parameter 'quiet' is set. It seems that the parameter 'quiet' is not redirected to the function rotate_log() ( and flush_log() ) and therefore not correctly checkable in that functions.

### Steps to reproduce

1. call '/opt/pihole/piholeLogFlush.sh once quiet' (or 'pihole flush once quiet')
2. see output message for correct rotated logfiles
3. (expect output only on error)

kuishou68:
## Summary
- detect quiet mode once at script startup
- reuse the quiet flag inside rotate_log() and flush_log()
- suppress helper output when `pihole flush once quiet` is used

Closes #6289

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -29,12 +29,12 @@ jobs:
     # Initializes the CodeQL tools for scanning.
     -
       name: Initialize CodeQL
-      uses: github/codeql-action/init@95e58e9a2cdfd71adc6e0353d5c52f41a045d225 #v4.35.2
+      uses: github/codeql-action/init@c10b8064de6f491fea524254123dbe5e09572f13 #v4.35.1
       with:
         languages: 'python'
     -
       name: Autobuild
-      uses: github/codeql-action/autobuild@95e58e9a2cdfd71adc6e0353d5c52f41a045d225 #v4.35.2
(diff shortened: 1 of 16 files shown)


darkexplosiveqwx:
Please follow the PR template.

You are missing at least the following: targeting `development` branch and DCO

yubiuser:
Please base on and target development

dschaper:
You'll need to fix the conflicts, your commits are missing the development history as the base so rebase all of your commits on to development. You'll need to sign off your commits as well.

I don't quite see how this fixes anything, it appears to just move the argument/flag detection in to a variable assignment in the main script? There is no functional change, it looks purely stylistic.

Issue #6615, opened 2026-04-26
Error when updating gravity via Web Interface


rdwebdesign:
### Versions
- Core version is v6.4.2 
- Web version is v6.5 
- FTL version is v6.6.1 

### Platform
- Platform:  Docker (`2026.04.1`)

### Expected behavior

Gravity output should look the same, when executed using the web interface or the command line.

### Actual behavior / bug

Gravity shows errors, but only when executed on the web interface.

_(**Note:** the process is completed, but some error messages are shown)_

<img width="1128" height="1026" alt="Image" src="https://github.com/user-attachments/assets/812b2cbd-bbf9-4c69-b3c9-1dcf5f33e8ca" />

rdwebdesign:
My tests indicate that the issue happens when we pipe `sort -V` output into `head -n1`, in the same subshell.
I still don't know why it only happens when gravity update is executed on the Web Interface.

Suggestions to fix:

1. We can simply silence the errors coming from `sort`. The code works even when the message is printed. It seem to be just a cosmetic issue.
```diff
-    if [[ "$(printf '%s\n' "${curlVersion}" "7.75" | sort -V | head -n1)" == 7.75 ]]; then
+    if [[ "$(printf '%s\n' "${curlVersion}" "7.75" | sort -V 2>/dev/null | head -n1)" == 7.75 ]]; then
```

2. We can break the commands into 2 lines. This will work without issues:

```diff
-    if [[ "$(printf '%s\n' "${curlVersion}" "7.75" | sort -V | head -n1)" == 7.75 ]]; then
+    sortedVerList=$(printf '%s\n' "${curlVersion}" "7.75" | sort -V)
+    if [[ $(echo -e "${sortedVerList}" | head -n1) == 7.75 ]]; then
```

2 Replace `head` with `sed`:
```diff
-    if [[ "$(printf '%s\n' "${curlVersion}" "7.75" | sort -V | head -n1)" == 7.75 ]]; then
+    if [[ "$(printf '%s\n' "${curlVersion}" "7.75" | sort -V | sed '1q')" == 7.75 ]]; then
```

rdwebdesign:
@0crap

Can you please check if this happens only on the web interface? I can't replicate the issue on the command line.

```
docker exec -it <pihole_container_name> pihole -g
```

0crap:
I manage my containers with Portainer.

[...]

PR #6620, opened 2026-04-27
Revert "Replace `head` with `sed` to avoid error messages on the web interface"

PromoFaux:
Reverts pi-hole/pi-hole#6616


yubiuser:
I think I missed why this needs to be reverted. 
I think the discussion in https://github.com/pi-hole/pi-hole/issues/6615 is still ongoing.

dschaper (approved):

Issue #5244, opened 2023-04-02
gravity doesn't exit if there is a storage problem.


jpgpi250:
### Versions
  Pi-hole version is v5.16.2 (Latest: v5.16.2)
  AdminLTE version is v5.19 (Latest: v5.19)
  FTL version is update/dnsmasq vDev-879708d (Latest: v5.22)

### Platform
Raspberry pi 3B
Linux raspberrypi 6.1.19-v7+ #1637 SMP Tue Mar 14 11:04:52 GMT 2023 armv7l GNU/Linux

### Expected behavior

exit 1 on error, generate warning message

### Actual behavior / bug

gravity continues, regardless problem

### Steps to reproduce

using GRAVITY_TMPDIR=/home/pi/tmp (setupVars.conf)
fstab entry for the mount point, configured to small (on purpose)
```
tmpfs /home/pi/tmp tmpfs nodev,noauto,nosuid,size=32M 0 0
```
sudo mount /home/pi/tmp
run pihole -g
error message (repeats after every list as soon space runs out, thus multiple messages):
```
sed: couldn't write 16 items to stdout: No space left on device
tee: /home/pi/tmp/tmp.PmGM1s9bMg.ph-non-domains: No space left on device
``` 
final message:
```
  [✗] Unable to fill gravity table in database /etc/pihole/gravity.db_temp
  /home/pi/tmp/tmp.gIRXzLCIDD.gravity:1386090: expected 2 columns but found 1 - filling the rest with NULL
/home/pi/tmp/tmp.gIRXzLCIDD.gravity:1386090: INSERT failed: NOT NULL constraint failed: gravity.adlist_id
```
gravity completes (no exit 1), dashboard says "adlists updated 00:05 (hh:mm) ago, but actually using old database (updated field in info table is modified, regardless of success)
## Debug Token

debug has been created after changing the TMPFS size, and build gravity with success.
https://tricorder.pi-hole.net/FblkfEu9/

### Screenshots

![image](https://user-images.githubusercontent.com/24669386/229344149-4bab705d-cbdb-4355-a0d1-32cfdfed1a6f.png)

![image](https://user-images.githubusercontent.com/24669386/229344070-a4d4657a-d161-4df3-8519-a5ad7de65c10.png)

### Additional context

trying to run gravity.sh on TMPFS, the to save the SD card, works fine, unless available TMPFS storage is to small, current adlists require 96Mb

I think this line:
```
[...]

Issue #5168, opened 2023-02-11
Failed gravity update loses whole DB / configuration


Zocker1999NET:
### Versions
<!-- Please run `pihole -v` and provide the following: -->
- Pi-hole:  v5.15.5
- AdminLTE:  v
- FTL:  v5.21

### Platform
<!-- OS examples: Debian 10, Debian Buster, Ubuntu 18.04 -->
- OS and version:  Armbian Buster (Debian Buster but with custom kernel for Banana Pi)
<!-- Platform examples: Raspberry Pi, Docker, Synology, VM/VPS -->
- Platform:  Banana Pi M2 Berry (comparable to Raspberry Pi 2)

### Expected behavior

When the update of the gravity fails due to a full `/tmp` filesystem, I expect the update to fail.
I expect that Pi-Hole can resume working with the old gravity version.
I also expect that Pi-Hole keeps its configuration (which should be stored in `/etc/pihole`, which had plenty of free storage left, ca. 1.7 GiB).

### Actual behavior / bug

Pi-Hole losts its full configuration (at least groups and configured adlists) and the "blocked domains" counter dropped to 0. Any restart of Pi-Hole did not restore the old database.

### Steps to reproduce


1. Add an example group & add a (preferably small) adlist.
2. Update the gravity successfully (because of small adlist, `/tmp` was sufficient for now)
3. Add a whole bunch of adlists (e.g. all from [here](https://github.com/RPiList/specials/blob/master/Blocklisten.md))
4. Update the gravity again, in my case with `pihole -g` (because of huge adlists, the gravity update "hopefully fails")
5. See that the "blocked domains" counter dropped to 0 and Pi-Hole forgot all adlists (including the working one before) and all groups (including the example one).

## Debug Token

- URL:  

<!-- 
Token generated by running `pihole -d`. https://docs.pi-hole.net/core/pihole-command/#debugger

The token is displayed at the end of the debug process if you allow for uploading the log file. 

[✓] Your debug token is: https://tricorder.pi-hole.net/wim5hft4rq

Debug logs are visible ONLY to developers and support staff. They are not publically accessible and all logs are automatically deleted after 48 hours.
-->

### Screenshots

<details><summary>all screenshots</summary>

#### Before Failed Update

Mind the working list & already added (but not yet downloaded) new adlists:
![image](https://user-images.githubusercontent.com/1645646/218250404-ed414af2-d1d5-4db7-b249-138b8b668d1a.png)
[...]

PR #6630, opened 2026-05-02
fix: check return codes in gravity_build_tree and database_recovery()

jluzzi123:

**Please make sure you**

---

> **Behavior change to flag:** This PR changes `pihole -g` from exit-0-on-silent-failure to exit-non-zero-on-failure for the `database_recovery` error paths. The weekly cron job will now correctly report failure on I/O errors that previously fell through silently. This is the intended behavior, but worth flagging for any orchestration that treats non-zero exit codes as restart signals.

Two unchecked return codes in `gravity.sh` can silently degrade or completely break DNS blocking with no error logged.

`gravity_build_tree` is called twice (lines 1218 and 1221) with no exit status check. If either call fails (disk full, I/O error, database corruption), execution continues and `gravity_swap_databases` installs a database with no search indexes. FTL then performs full table scans on every DNS lookup against a potentially million-entry blocklist, causing DNS timeouts with no error in the logs. This runs on every `pihole -g`, including the weekly cron job.

`database_recovery()` pipes `.recover` output directly into a second `sqlite3` command. Without `pipefail`, bash evaluates the pipe on the RHS exit status only. If `.recover` fails, the second `sqlite3` receives empty input, exits 0, and the if-branch signals success. The empty `.recovered` database is then swapped into production, silently breaking DNS blocking. This is a bug in the error-recovery path itself: it triggers precisely when the database is already corrupt and makes the situation worse without any logged indication.

Neither issue has a specific crash report behind it, but both connect to open issues documenting real failures in the field. The gravity_build_tree gap relates to the class of bug described in #5244 ("gravity doesn't exit if there is a storage problem"). The recovery pipe silent-success path is a contributing factor in failures like those documented in #5168 ("Failed gravity update loses whole DB / configuration"). Both are logic gaps identified during a code review of the gravity update flow.

For `gravity_build_tree`: wrap both calls in `if !` guards and call `exit 1` on failure. No additional error message is needed; `gravity_build_tree` already prints `"Unable to build ${table} tree in ${gravityTEMPfile}"` and the sqlite3 error detail on failure. The `if ! timeit ...; then exit 1; fi` pattern matches lines 1192-1195 and 1234-1237 in the same flow. Confirmed `gravity_build_tree` only returns non-zero on genuine failure (sqlite3 `CREATE INDEX` error), so the hard-exit on failure is appropriate.

For `database_recovery`: stage the `.recover` SQL output to a temp file (`${gravityDBfile}.recovery.sql`) so each command's exit status is checked independently before proceeding. The error paths use `return 1` instead of `exit 1`, which lets a `RETURN` trap handle temp-file cleanup across all paths (success, failure, and interruption) without explicit `rm -f` in each branch. On failure, the function returns 1 with a clear error message; the call site uses `|| exit 1` to terminate the script. This eliminates the silent empty-database scenario entirely.

Considered `set -o pipefail` for `database_recovery`. Chose the temp-file approach to keep the change scoped to the function being fixed; enabling `pipefail` script-wide could surface latent issues elsewhere in `gravity.sh` that are out of scope for this PR. Happy to switch approaches if maintainers prefer.

---

---

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -965,6 +965,8 @@ database_recovery() {
   local result
   local str="Checking integrity of existing gravity database (this can take a while)"
   local option="${1}"
+  local recoverySQL="${gravityDBfile}.recovery.sql"
+  trap 'rm -f "${recoverySQL}"' RETURN
   echo -ne "  ${INFO} ${str}..."
   result="$(pihole-FTL sqlite3 -ni "${gravityDBfile}" "PRAGMA integrity_check" 2>&1)"
 
@@ -993,17 +995,26 @@ database_recovery() {
   echo -ne "  ${INFO} ${str}..."
   # We have to remove any possibly existing recovery database or this will fail
   rm -f "${gravityDBfile}.recovered" >/dev/null 2>&1
-  if result="$(pihole-FTL sqlite3 -ni "${gravityDBfile}" ".recover" | pihole-FTL sqlite3 -ni "${gravityDBfile}.recovered" 2>&1)"; then
-    echo -e "${OVER}  ${TICK} ${str} - success"
-    mv "${gravityDBfile}" "${gravityDBfile}.old"
-    mv "${gravityDBfile}.recovered" "${gravityDBfile}"
-    echo -ne " ${INFO} ${gravityDBfile} has been recovered"
-    echo -ne " ${INFO} The old ${gravityDBfile} has been moved to ${gravityDBfile}.old"
+  # Stage .recover output to a temp file so each command's exit status can
+  # be checked independently. Piping would hide a failing .recover because
+  # bash reports the RHS exit status for the whole pipe.
+  if pihole-FTL sqlite3 -ni "${gravityDBfile}" ".recover" > "${recoverySQL}"; then
+    if result="$(pihole-FTL sqlite3 -ni "${gravityDBfile}.recovered" < "${recoverySQL}" 2>&1)"; then
+      echo -e "${OVER}  ${TICK} ${str} - success"
+      mv "${gravityDBfile}" "${gravityDBfile}.old"
+      mv "${gravityDBfile}.recovered" "${gravityDBfile}"
+      echo -ne " ${INFO} ${gravityDBfile} has been recovered"
+      echo -ne " ${INFO} The old ${gravityDBfile} has been moved to ${gravityDBfile}.old"
+    else
+      echo -e "${OVER}  ${CROSS} ${str} - the following errors happened:"
+      while IFS= read -r line; do echo "  - $line"; done <<<"$result"
+      echo -e "  ${CROSS} Recovery failed. Try \"pihole -g -r recreate\" instead."
+      return 1
+    fi
   else
-    echo -e "${OVER}  ${CROSS} ${str} - the following errors happened:"
-    while IFS= read -r line; do echo "  - $line"; done <<<"$result"
+    echo -e "${OVER}  ${CROSS} ${str} - .recover command failed"
     echo -e "  ${CROSS} Recovery failed. Try \"pihole -g -r recreate\" instead."
-    exit 1
+    return 1
   fi
   echo ""
 }
@@ -1171,7 +1182,7 @@ if [[ "${recreate_database:-}" == true ]]; then
 fi
 
 if [[ "${recover_database:-}" == true ]]; then
-  timeit database_recovery "$4"
+  timeit database_recovery "$4" || exit 1
 fi
 
 # Migrate scattered list files to the new cache directory
@@ -1204,8 +1215,12 @@ update_gravity_timestamp
 fix_owner_permissions "${gravityTEMPfile}"
 
 # Build the tree
-timeit gravity_build_tree gravity
-timeit gravity_build_tree antigravity
+if ! timeit gravity_build_tree gravity; then
+  exit 1
+fi
+if ! timeit gravity_build_tree antigravity; then
+  exit 1
+fi
 
 # Compute numbers to be displayed (do this after building the tree to get the
 # numbers quickly from the tree instead of having to scan the whole database)


jluzzi123:
Tested the logic of both fixes via a bash unit test that mocks the sqlite3 calls:

**Fix 1 (recovery pipe):** Confirmed that with the old code, `false | true` exits 0, so a failing `.recover` is invisible to the surrounding `if`. The temp file approach independently checks each command's exit status and correctly surfaces the failure. Tested all three paths: `.recover` fails, import fails, both succeed.

**Fix 2 (gravity_build_tree):** Confirmed that without the `if !` guards, a `gravity_build_tree` returning 1 is silently ignored and execution reaches `gravity_ShowCount`; the unindexed database would proceed to `gravity_swap_databases`. With the guards, execution stops at the failing call and exits non-zero.

**Happy path:** Both changes confirmed to have no effect when the underlying commands succeed.

**Additional change since initial submission:** Replaced the three explicit `rm -f` calls in `database_recovery()` with a single `trap 'rm -f "${recoverySQL}"' RETURN`. The trap fires on all function exit paths (both `return` paths and normal completion). To make this work, the two `exit 1` calls inside the function were changed to `return 1`, and the call site was updated to `timeit database_recovery "$4" || exit 1` to preserve script termination behavior. Net change: +1 trap line, -3 `rm -f` lines, exit changed to return in two error branches.

Happy to share the test script if useful.

dschaper:
Let us know when you're satisfied with the PR, it's a bit hard to review things when there are force-pushed changes. Throws off any local clones of the branch.

And sure, seeing the test script would be good.

jluzzi123:
Sorry about the force pushed changes. I was still iterating on the database_recovery cleanup and did not think about the impact on local clones once the PR was open. The branch is stable now.

I am satisfied with the PR as it stands. Here is the test script:

```bash
#!/usr/bin/env bash
# Tests for gravity.sh fixes:
#   Fix 1: database_recovery() pipe hides .recover failure
#   Fix 2: gravity_build_tree return code not checked

PASS=0
FAIL=0

pass() { echo "  PASS  $1"; PASS=$((PASS + 1)); }
fail() { echo "  FAIL  $1"; FAIL=$((FAIL + 1)); }
header() { echo ""; echo "=== $1 ==="; }

# ---------------------------------------------------------------
header "Background: bash pipe behavior without pipefail"
# ---------------------------------------------------------------
# This is the root cause of Fix 1. Without pipefail, bash uses
# the RHS exit status for the whole pipe. A failing LHS is silent.

pipe_exit=$(false | true; echo $?)
[[ "$pipe_exit" == "0" ]] \
  && pass "Confirmed: 'false | true' exits 0 (LHS failure invisible without pipefail)" \
  || fail "Unexpected pipe behavior: $pipe_exit"

# ---------------------------------------------------------------
header "Fix 1: OLD code: pipe hides .recover failure"
# ---------------------------------------------------------------
# Simulate the old code: recover_cmd fails, import_cmd succeeds.
# Expected (buggy) result: if-branch signals SUCCESS despite failure.

test_old_code() {
  local recover_exit=$1 import_exit=$2
  recover_cmd() { return "$recover_exit"; }
  import_cmd()  { cat >/dev/null; return "$import_exit"; }
  if result="$(recover_cmd | import_cmd 2>&1)"; then
    echo "success"
  else
    echo "failure"
  fi
}

out=$(test_old_code 1 0)
[[ "$out" == "success" ]] \
  && pass "OLD: .recover fails + import succeeds → falsely reports SUCCESS (bug confirmed)" \
  || fail "OLD unexpected result: $out"

out=$(test_old_code 0 1)
[[ "$out" == "failure" ]] \
  && pass "OLD: import failure is caught correctly (this path was already working)" \
  || fail "OLD unexpected result: $out"

out=$(test_old_code 0 0)
[[ "$out" == "success" ]] \
  && pass "OLD: both succeed → success (correct)" \
  || fail "OLD unexpected result: $out"

# ---------------------------------------------------------------
header "Fix 1: NEW code: temp file detects .recover failure"
# ---------------------------------------------------------------
# New code stages .recover output to a file, checks each exit status.

test_new_code() {
  local recover_exit=$1 import_exit=$2
  recover_cmd() { return "$recover_exit"; }
  import_cmd()  { cat >/dev/null; return "$import_exit"; }

  local tmpfile
  tmpfile=$(mktemp)

  local outcome
  if recover_cmd > "$tmpfile"; then
    if result="$(import_cmd < "$tmpfile" 2>&1)"; then
      outcome="success"
    else
      outcome="import_failed"
    fi
  else
    outcome="recover_failed"
  fi
  rm -f "$tmpfile"
  echo "$outcome"
}

out=$(test_new_code 1 0)
[[ "$out" == "recover_failed" ]] \
  && pass "NEW: .recover fails → correctly reports FAILURE (fix works)" \
  || fail "NEW unexpected result: $out"

out=$(test_new_code 0 1)
[[ "$out" == "import_failed" ]] \
  && pass "NEW: import fails → reports failure (no regression)" \
  || fail "NEW unexpected result: $out"

out=$(test_new_code 0 0)
[[ "$out" == "success" ]] \
  && pass "NEW: both succeed → success (no regression on happy path)" \
  || fail "NEW unexpected result: $out"

# ---------------------------------------------------------------
header "Fix 1: RETURN trap cleans up temp file on all paths"
# ---------------------------------------------------------------
# The actual implementation uses trap 'rm -f "${recoverySQL}"' RETURN so the
# file is removed on all function exit paths including early returns.

test_trap_cleanup() {
  local recover_exit=$1
  local tmpfile=$2
  recover_cmd() { return "$recover_exit"; }
  import_cmd()  { return 0; }

  trap 'rm -f "${tmpfile}"' RETURN

  if recover_cmd > "${tmpfile}"; then
    import_cmd < "${tmpfile}" >/dev/null 2>&1 || return 1
  else
    return 1
  fi
}

tmpfile=$(mktemp)
test_trap_cleanup 0 "$tmpfile"
[[ ! -f "$tmpfile" ]] && pass "Temp file removed on success (RETURN trap fired)"   || fail "Temp file leaked on success"

tmpfile=$(mktemp)
test_trap_cleanup 1 "$tmpfile"
[[ ! -f "$tmpfile" ]] && pass "Temp file removed on failure (RETURN trap fired)"   || fail "Temp file leaked on failure"

# ---------------------------------------------------------------
header "Fix 2: OLD code: gravity_build_tree failure is ignored"
# ---------------------------------------------------------------
# Old code: timeit gravity_build_tree gravity (no if ! check)
# gravity_build_tree returns 1 → execution continues anyway.

timeit() { "$@"; return $?; }

test_old_build_tree() {
  local gravity_exit=$1

  gravity_build_tree() { return "$gravity_exit"; }

  (
    timeit gravity_build_tree gravity        # no error check
    timeit gravity_build_tree antigravity
    echo "reached:show_count"
  )
}

out=$(test_old_build_tree 1)
[[ "$out" == "reached:show_count" ]] \
  && pass "OLD: gravity_build_tree failure ignored, execution reaches gravity_ShowCount (bug confirmed)" \
  || fail "OLD unexpected result: $out"

# ---------------------------------------------------------------
header "Fix 2: NEW code: if ! guards stop execution on failure"
# ---------------------------------------------------------------

test_new_build_tree() {
  local gravity_exit=$1 antigravity_exit=$2

  gravity_build_tree() {
    [[ "$1" == "gravity"     ]] && return "$gravity_exit"
    [[ "$1" == "antigravity" ]] && return "$antigravity_exit"
  }

  (
    if ! timeit gravity_build_tree gravity; then
      echo "stopped:gravity"; exit 1
    fi
    if ! timeit gravity_build_tree antigravity; then
      echo "stopped:antigravity"; exit 1
    fi
    echo "reached:show_count"
  )
}

out=$(test_new_build_tree 1 0)
[[ "$out" == "stopped:gravity" ]] \
  && pass "NEW: gravity fails → stops at gravity, never reaches show_count" \
  || fail "NEW unexpected: $out"

out=$(test_new_build_tree 0 1)
[[ "$out" == "stopped:antigravity" ]] \
  && pass "NEW: antigravity fails → stops at antigravity, never reaches show_count" \
  || fail "NEW unexpected: $out"

out=$(test_new_build_tree 0 0)
[[ "$out" == "reached:show_count" ]] \
  && pass "NEW: both succeed → reaches show_count (happy path, no regression)" \
  || fail "NEW unexpected: $out"

# Verify exit code is non-zero when build fails
(test_new_build_tree 1 0 >/dev/null 2>&1); exit_code=$?
[[ "$exit_code" -ne 0 ]] \
  && pass "NEW: exits non-zero when build fails (gravity_swap_databases will not run)" \
  || fail "NEW: exited 0 despite failure"

# ---------------------------------------------------------------
echo ""
echo "==========================="
echo "Results: ${PASS} passed, ${FAIL} failed"
[[ $FAIL -eq 0 ]] && echo "ALL TESTS PASSED" || echo "SOME TESTS FAILED"
[[ $FAIL -eq 0 ]]
```

The script covers the core scenarios: Fix 1 confirms that the old pipe silently swallowed a failing .recover command while the temp file approach correctly surfaces it. Fix 2 confirms that without the guards a failing gravity_build_tree is ignored and execution reaches gravity_swap_databases, while the guards stop it at the right point. There are also checks for temp file cleanup across all paths and that neither fix breaks the happy path.

PromoFaux:
@jluzzi123 - could you please rebase on development and add some tests to the test suite based on your test script? We recently moved from pytest to BATS, so it should be straightforward

PromoFaux:
Although, it's not a blocker

PromoFaux (approved):
OK - hiding whitespace changes made this a little easier to follow!

PR #6634, opened 2026-05-12
Improve gravity error handling

### LINKED ISSUE #6633: FTL at 100% and unresponsive to DNS requests if Gravity update fails due to short disk space
### Versions
Core version is v6.4.2 (Latest: v6.4.2)
Web version is v6.5 (Latest: v6.5)
FTL version is v6.6.1 (Latest: v6.6.1)

### Platform
- Platform:  Raspberry Pi (2025-05-06, 49e1078c1b07ff4948d304e0149d3bc698e0006a, stage2)

### Expected behavior

If gravity update fails due to disk space limitations a fallback to the old database should happen. PiHole should remain productive.

### Actual behavior / bug

Once gravity update fails due to limited disk space for DB generation, FTL process raises to a CPU load of 100% and the system does not resolve DNS requests anymore. 

### Steps to reproduce

1. Limit diskspace for database generation (e.g. using fallocate)
2. Run gravity update
3. PiHole is now not resolving DNS queries anymore, FTL process is at 100% CPU 

## Debug Token

No debug token available since upload fails due to inoperable PiHole. Debug logfile available.

yubiuser:

Improve gravity error handling. Fixes https://github.com/pi-hole/pi-hole/issues/6633

0) Guard `update_gravity_timestamp` and `gravity_build_tree gravity/antigravity` by checking their return code and exit in case of an error (This is what fixes  https://github.com/pi-hole/pi-hole/issues/6633)
1) Use `gravity_Cleanup error` instead of exit 1 to cleanup in case of an error
2) Format output nicer in case of an error
3) Remove the temporary gravity database in `gravity_Cleanup()` if it still exists
(4) Fix some shellcheck warnings)

---

---

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -163,7 +163,7 @@ gravity_swap_databases() {
   mv "${gravityTEMPfile}" "${gravityDBfile}"
   echo -e "${OVER}  ${TICK} ${str}"
 
-  if $oldAvail; then
+  if ${oldAvail}; then
     echo -e "  ${TICK} The old database remains available"
   fi
 }
@@ -217,7 +217,7 @@ database_table_from_file() {
   # Get MAX(id) from domainlist when INSERTing into this table
   if [[ "${table}" == "domainlist" ]]; then
     rowid="$(pihole-FTL sqlite3 -ni "${gravityDBfile}" "SELECT MAX(id) FROM domainlist;")"
-    if [[ -z "$rowid" ]]; then
+    if [[ -z "${rowid}" ]]; then
       rowid=0
     fi
     rowid+=1
@@ -541,10 +541,10 @@ gravity_DownloadBlocklists() {
 
   # Loop through $sources and download each one
   for ((i = 0; i < "${#sources[@]}"; i++)); do
-    url="${sources[$i]}"
-    domain="${sourceDomains[$i]}"
-    id="${sourceIDs[$i]}"
-    if [[ "${sourceTypes[$i]}" -eq "0" ]]; then
+    url="${sources[${i}]}"
+    domain="${sourceDomains[${i}]}"
+    id="${sourceIDs[${i}]}"
+    if [[ "${sourceTypes[${i}]}" -eq "0" ]]; then
       # Gravity list
       str="blocklist"
       adlist_type="gravity"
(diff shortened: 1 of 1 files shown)


gravity.sh
```diff
@@ -98,7 +98,8 @@ gravity_build_tree() {
   status="$?"
 
   if [[ "${status}" -ne 0 ]]; then
-    echo -e "\\n  ${CROSS} Unable to build ${table} tree in ${gravityTEMPfile}\\n  ${output}"
+    echo -e "\\n  ${CROSS} Unable to build ${table} tree in ${gravityTEMPfile}\\n"
+    echo -e "  ${CROSS} ${output}\\n"
     echo -e "  ${INFO} If you have a large amount of domains, make sure your Pi-hole has enough RAM available\\n"
     return 1
```
rdwebdesign, line 104: Why return "1"?

I think we should replace `return 1` with `gravity_Cleanup "error"` and handle the error here. 
We can include `${CROSS} Unable to create database. Please contact support.` here and avoid the second error checking (lines 1223-1230).

rdwebdesign (changes requested):

gravity.sh
yubiuser: I just followed the style we had for the other functions in the main routine. It should not be changed for one function only, but all would need to be changed.
Personally, I think it's kind of 'cleaner' as we have it now and propagate the error as far up as we can and exit from there. This makes it clear that we thought about error handling. Otherwise one would need to go down through the functions and see if the have proper error handling.

dschaper: This PR doesn't touch that line so any changes would need to be addressed in another PR/issue.

yubiuser: It does not touch it yet, but I think RDs question is legit. If we decide to change the behavior, this PR would be well suited as it modifies gravity error handling already.

rdwebdesign: @yubiuser

I'm not suggesting to change the behavior. 
Actually, my suggestion was to follow the same pattern already used by other functions on this file, like in:

- `database_table_from_file()` - lines 248-252
- `database_adlist_number()` - lines 284-288
- `gravity_DownloadBlocklists()` - lines 450-454
- `gravity_optimize()` - 1033-1037
... and other functions

We print an error message and call `gravity_Cleanup "error"` inside these functions.

To be honest, this is just a suggestion to simplify the code.
Your solution is already solving the issue. If you think this is not desired, we can keep the way it is.

@dschaper 

I'm suggesting exactly to change that line as part of the same fix here.

The issue this PR is fixing is: 
- The code in `master` shows an error message and returns "1", but we don't do anything with this value and continue.

The solution:
- We need to exit on errors.

I see 2 possible fixes:

1. **yubiuser's suggestion _(current PR code)_:**
  we keep showing the original message and returning "1" (as in `master`).
  Then we check the value, show another error message and after that we call `gravity_Cleanup "error"` to exit.
  _(personally I think it is unnecessary to check for error and print messages for the same issue in 2 different places)_

2. **My suggestion:**
  we don't return "1" and handle the error inside the function. 
  The function would show all error messages and call `gravity_Cleanup "error"`.
  No need to return "1" and check for error again.

3ch0less: I agree with rdwebdesign. `gravity_build_tree` is the only function in this file that returns a failure code for the caller to check. Every other function (`database_table_from_file`, `database_adlist_number`, `database_adlist_status`, `gravity_optimize`, etc.) calls `gravity_Cleanup "error"` directly on fatal error. The fact that this bug exists proves return codes are easy to forget.

Changing line 103 from `return 1` to `gravity_Cleanup "error"` (and dropping the `if !` guards at the call sites) also avoids printing two error messages for one failure, which the current approach would do: "Unable to build gravity tree..." plus "Unable to create database..."

rdwebdesign:
@yubiuser 

Are you going to continue this one?

yubiuser:
After force-pushing the rebase, I can't reopen this one. Instead opened https://github.com/pi-hole/pi-hole/pull/6678

PR #6638, opened 2026-05-19
installer: fix custom DNS entry when only one upstream server is provided

### LINKED ISSUE #4783: [Tracking] Installer assumes users always enter two custom upstream DNS server even if they enter only one

Gilmoursa:

Fix the custom DNS entry in the installer so that entering only one upstream DNS server does not incorrectly duplicate it into both slots. Closes #4783.

- Replaced the two numbered variables `PIHOLE_DNS_1` / `PIHOLE_DNS_2` with a single bash array `PIHOLE_DNS=()`. This naturally supports one or more entries without needing special-case logic for the single-server scenario.
- User input is split on commas into the array. Empty entries are skipped. Invalid entries are filtered out while valid ones are retained.
- The confirmation dialog is built by joining array elements, so it accurately reflects exactly what was entered regardless of how many servers were provided.
- Preset DNS selection populates the array from the semicolon-delimited preset string, adding a second element only when one is present.

---

5. I have squashed any insignificant commits.

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -875,19 +875,16 @@ setDNS() {
         until [[ "${DNSSettingsCorrect}" = True ]]; do
             # Signal value, to be used if the user inputs an invalid IP address
             strInvalid="Invalid"
-            if [[ ! "${PIHOLE_DNS_1}" ]]; then
-                if [[ ! "${PIHOLE_DNS_2}" ]]; then
-                    # If the first and second upstream servers do not exist, do not prepopulate an IP address
-                    prePopulate=""
-                else
-                    # Otherwise, prepopulate the dialogue with the appropriate DNS value(s)
-                    prePopulate=", ${PIHOLE_DNS_2}"
-                fi
-            elif [[ "${PIHOLE_DNS_1}" ]] && [[ ! "${PIHOLE_DNS_2}" ]]; then
-                prePopulate="${PIHOLE_DNS_1}"
-            elif [[ "${PIHOLE_DNS_1}" ]] && [[ "${PIHOLE_DNS_2}" ]]; then
-                prePopulate="${PIHOLE_DNS_1}, ${PIHOLE_DNS_2}"
-            fi
+            # Prepopulate the dialog with any previously set DNS values
+            prePopulate=""
+            i=1
+            while true; do
+                varname="PIHOLE_DNS_${i}"
+                [[ -n "${!varname}" ]] || break
+                [[ -n "${prePopulate}" ]] && prePopulate+=", "
+                prePopulate+="${!varname}"
+                ((i++))
+            done
 
             # Prompt the user to enter custom upstream servers
             piholeDNS=$(dialog --no-shadow --keep-tite --output-fd 1 \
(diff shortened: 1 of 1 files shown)


darkexplosiveqwx:
Please follow the PR template.

You are missing at least the following: DCO

Update your description with the proper template from 	https://github.com/pi-hole/.github/blob/master/.github/PULL_REQUEST_TEMPLATE.md

yubiuser:
Thanks for picking this up. I think this is an opportunity to re-work the whole section to be more flexible and react to a unknown number of user provided upstream dns server.

Just image, they would enter 3 servers now...

Gilmoursa:
Thanks for the feedback! I've reworked the custom DNS section to support any number of upstream servers rather than hardcoding for just two.

Here's what I changed:

The _prePopulate_ logic now loops over all set _PIHOLE_DNS_N_ variables instead of checking only _DNS_1_ and _DNS_2_ individually.
Input parsing now uses _IFS=',' read -ra_ to split the user's input into a proper array, which means middle values are no longer silently dropped when 3 or more servers are entered.
Validation, the error dialog, and the confirmation dialog all loop over the full array dynamically.
The _dns.upstreams_ write at install time also loops over all _PIHOLE_DNS_N_ variables so every server makes it into _pihole.toml_.

Let me know if you'd like anything adjusted.

rdwebdesign:
I think we need to decide if we need the `PIHOLE_DNS_N` variables, or if we just need the array.

Gilmoursa:
Good point. Looking at the code, _PIHOLE_DNS_N_ is used in 3 places outside the custom input block: 
- the preset selection branch (lines 1017-1018)
- the "Using upstream DNS" display line
- the config write

My current implementation uses both a local _dns_array_ and the numbered vars, which is redundant.

I think the cleanest path is to replace the numbered vars entirely with a single global array (ex: _PIHOLE_DNS_SERVERS_) populated by both the custom & preset paths, then read from it in the display line and the config write. 

That removes all the indirect variable references too. Would that be better, or keep the numbered vars and just drop the intermediate array?

rdwebdesign:
Personally, I think that using just one variable is cleaner and simplifies the code, but I like to ask for other opinions.

@yubiuser @PromoFaux 

Is there a reason to use the old numbered variables (`PIHOLE_DNS_<n>`) or can we remove them?

PromoFaux:
I think we should go with an array, especially if it means we can take some of the complexity away from the installer. If we're going to rework it then we should make it more resilient 

The numbered vars were only ever an implementation detail, not an API, and I can't see anything that uses them outside of basic-install.sh so we're probably good to go

Gilmoursa:
Thanks for the direction. I've pushed a new commit that replaces all the numbered `PIHOLE_DNS_N` variables with a single `PIHOLE_DNS` bash array throughout `basic-install.sh`. The logic is simpler — no more indirect variable references or manual re-compaction after invalid entries are dropped. Happy to make any further adjustments.

PromoFaux:
Sorry for the delay in re-review! Yep, that's cleaner - and having just tested, confirm that it works as expected

Just some housekeeping now, please squash down to a single commit - the first two just show the evolution, but are not necessary to be included. The other thing is we enforce that commits must have verified signatures. See here for details on how to do that:

https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

PromoFaux:
oh - probably also worht noting that the PR description still describes the two-variable approach, please update that to reflect the final array-based form :)

PromoFaux (changes requested):
Just some housekeeping now, please squash down to a single commit - the first two just show the evolution, but are not necessary to be included. The other thing is we enforce that commits must have verified signatures. See here for details on how to do that:

[docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification)

oh - probably also worht noting that the PR description still describes the two-variable approach, please update that to reflect the final array-based form :)

PromoFaux (approved):

PR #6639, opened 2026-05-21
Add gravity tests

yubiuser:

Adds gravity tests to our new and shiny BATS test suite

---

---

### DIFF
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1791,6 +1791,7 @@ clone_or_reset_repos() {
         printf "  %b Resetting local repos\\n" "${INFO}"
 
         # import getFTLConfigValue from utils.sh
+        # shellcheck source="./advanced/Scripts/utils.sh"
         source "/opt/pihole/utils.sh"
         # Use the configured Web repo location on repair/update
         webInterfaceDir=$(getFTLConfigValue "webserver.paths.webroot")$(getFTLConfigValue "webserver.paths.webhome")
@@ -1802,7 +1803,7 @@ clone_or_reset_repos() {
                 exit 1
             }
         # Reset the Web repo
-        resetRepo ${webInterfaceDir} ||
+        resetRepo "${webInterfaceDir}" ||
             {
                 printf "  %b Unable to reset %s, exiting installer%b\\n" "${COL_RED}" "${webInterfaceDir}" "${COL_NC}"
                 exit 1
--- a/test/run.sh
+++ b/test/run.sh
@@ -98,6 +98,7 @@ SUITE_1=(
     test_installer_ftl.bats
     test_network.bats
     test_utils.bats
+    test_gravity.bats
 )
 [[ "${DISTRO_FAMILY}" == "rhel" ]] && SUITE_1+=(test_selinux.bats)
 
--- a/test/test_gravity.bats
+++ b/test/test_gravity.bats
@@ -191,7 +191,7 @@ teardown() {
     assert_line --partial "${INFO} Migrating content of /etc/pihole/adlists.list into new database"
 
     assert_line --partial "${INFO} Target: http://"
-    assert_line --partial "${CROSS} Status: Retrieval failed (exit_code=3 Msg: URL rejected: No host part in the URL)"
+    assert_line --partial "${CROSS} Status: Retrieval failed (exit_code=3 Msg:"
     assert_line --partial "${CROSS} List download failed: no cached list available"
 
     refute_line --regexp "Parsed [[:digit:]]+ exact domains and [[:digit:]]+ ABP-style domains.*"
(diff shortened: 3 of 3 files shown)


yubiuser:
Interesting, depending on the `curl` version, error output differs...

`[✗] Status: Retrieval failed (exit_code=6 Msg: Could not resolve host: raw.githubusercontent.df (Domain name not found))`

`[✗] Status: Retrieval failed (exit_code=6 Msg: Could not resolve host: raw.githubusercontent.df)`

___

Oh... on docker

```
  [✗] Status: Retrieval failed (exit_code=7 Msg: Failed to connect to localhost port 443 after 0 ms: Could not connect to server)
```

Bare metal

```
  [✗] Status: Retrieval failed (exit_code=60 Msg: SSL certificate problem: self-signed certificate in certificate chain)

```

rdwebdesign:
> depending on the `curl` version, error output differs

Yeah... I noticed that when I was changing the code to retrieve the `exit_code` and `errormsg` values. 

---

> Oh... on docker

I think this was a different issue. I tested using `curl -I https://localhost/list` inside the container and received the SSL error (`exit_code=60`).

rdwebdesign:
Just a question:

Do you think we need all of these tests?

The "StevenBlack" list test will fail if (for some reason) the file is not found, even if Pi-hole is 100% fine.

yubiuser:
Maybe we could remove some of the 'gravity failed' tests - but most test normal functioning and some are security related.

You are right about the external list dependency - but it would then also fail for all fresh installations which is something we should be aware of. This test serves two purposes: 1) check if out external dependency is available and 2) gravity can add the default adlist

yubiuser:
I figured out, why the test currently fail on ubuntu 20. It's an old curl version (curl 7.68.0) which is missing this fix https://github.com/curl/curl/pull/5180 and therefore does not send an etag when `etag-save` and `etag-compare` point to the same file. Therefore it returns 200 instead of 304 and the tests are failing

rdwebdesign:
Do we still need Ubuntu 20 tests?

Ubuntu 20.04 support ended one year ago: https://ubuntu.com/blog/ubuntu-20-04-lts-end-of-life-standard-support-is-coming-to-an-end-heres-how-to-prepare

yubiuser:
I think we once said we support all OS versions that are currently supported, but keep tests even for older one. But sure, we could just drop it

rdwebdesign:
>  but keep tests even for older one.

Yes, but not indefinitely.

dschaper:
We keep the old and unsupported OSes to see things like this exact situation, when new code will not run on older OSes. So we have information and can point users running those OSes to where the changes were that made Pi-hole no longer work for them.

dschaper:
> Ubuntu 20.04 support ended one year ago: [ubuntu.com/blog/ubuntu-20-04-lts-end-of-life-standard-support-is-coming-to-an-end-heres-how-to-prepare](https://ubuntu.com/blog/ubuntu-20-04-lts-end-of-life-standard-support-is-coming-to-an-end-heres-how-to-prepare)

The second is to obtain Expanded Security Maintenance and extend the support period for 20.04 LTS until May 2030 with an Ubuntu Pro subscription. An additional 2 years can be purchased with the Legacy Support add-on, extending the total security maintenance period until May 2032

darkexplosiveqwx:
> > Ubuntu 20.04 support ended one year ago: [ubuntu.com/blog/ubuntu-20-04-lts-end-of-life-standard-support-is-coming-to-an-end-heres-how-to-prepare](https://ubuntu.com/blog/ubuntu-20-04-lts-end-of-life-standard-support-is-coming-to-an-end-heres-how-to-prepare)
> 
> The second is to obtain Expanded Security Maintenance and extend the support period for 20.04 LTS until May 2030 with an Ubuntu Pro subscription. An additional 2 years can be purchased with the Legacy Support add-on, extending the total security maintenance period until May 2032

I think ESM+ Legacy support is a bad metric here, by that metric we would still be supporting Ubuntu 14.04 (Legacy Support until Apr 2029!), even 16.04 had the more standard ESM until April this year. The 10 year ESM is also available for free for individuals.
I think we should support every supported OS at least until the regular EOL(so 5 years for Ubuntu), but still try have reasonable support on a best-effort basis.
From what I made out about the curl error messages and etag quirks is that, while it is annoying and requires a bit of an workaround, it does seem reasonable to keep it around.
If we ever require a newer curl feature where a workaround is not reasonable or just raise the baseline at some point, we should make failure explicit, e.g via a greater than requirement in pihole-meta.

I'll support Dan for keeping it around until we make failure explicit.

rdwebdesign:
> keeping it around until we make failure explicit.

**Just a note:**

Nothing is really failing.

The tests were failing just because, on success, `curl` answers differently on different operating systems, but this is correctly handled by gravity code.

**Gravity is working correctly on every OS**, but since there are 2 different possible "success" outputs (depending on the OS version), the new tests were receiving an unexpected string and fails. 
yubiuser already adjusted the test and now everything is working.

The only question I raised was:
_Should we adjust the test to accept different strings for different versions, or should we remove the old OS?_

There is nothing breaking on this PR or in gravity.
Only 2 different successful results on different operating systems.

We can review this PR.
The decision about the supported OS versions is a different subject and we can decide later.

PromoFaux (approved):

Issue #2960, opened 2019-10-12
Cannot enable/disable with web interface


cars11:
**In raising this issue, I confirm the following:** `{please fill the checkboxes, e.g: [X]}`

- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [x] The issue I am reporting can be *replicated*.
- [x] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the the source code relevant to this issue?:** 1

---
**Expected behaviour:**

clicking enable or disable enables or disables

**Actual behaviour:**

The little loading circle that comes when click turns for a quarter revolution and then stops, and it does not do anything

**Steps to reproduce:**

clicking any of the buttons under disable (even custom 60sec does it)

**Debug token provided by [uploading `pihole -d` log](https://discourse.pi-hole.net/t/the-pihole-command-with-examples/738#debug):**

https://tricorder.pi-hole.net/wx2jxidyx5

**Troubleshooting undertaken, and/or other relevant information:**

`{Steps of what you have done to fix this}`

- tried different browsers (Chrome + edge)  neither work
- sshing to dietpi and running pihole disable works
- restarting the pi didn't help
- this is a fresh install of dietpi + pihole (from yesterday)

dschaper:
The web interface is not set up correctly or you are not using the right credentials / URL. You'll need to ask the Diet-Pi folks about that as they change the URL for the web interface.

```plain

*** [ DIAGNOSING ]: Dashboard and block page
[✗] Block page X-Header: X-Header does not match or could not be retrieved.
HTTP/1.1 200 OK
X-Pi-hole: A black hole for Internet advertisements.
Content-type: text/html; charset=UTF-8
Date: Sat, 12 Oct 2019 15:40:22 GMT
Server: lighttpd/1.4.53


[✗] Web interface X-Header: X-Header does not match or could not be retrieved.
HTTP/1.1 200 OK
Set-Cookie: PHPSESSID=97u4ng422qvurr4sjlrd9qbqrn; path=/
Expires: Thu, 19 Nov 1981 08:52:00 GMT
[...]

PR #6672, opened 2026-07-25
send USR2 after logrotate (v7 Prep)

darkexplosiveqwx:

**Please make sure you**

---

Accompanies https://github.com/pi-hole/FTL/pull/2960
Discussed in https://github.com/pi-hole/FTL/pull/2897#issuecomment-4964470971 2.2

One possible issue to keep in mind:
We intentionally do not replace the user's logrotate file. Installations before v6.4.3 use `copytruncate`, which will continue to work. In v6.4.3 we changed it to so only dnsmasq's log gets a `postrotate` signal action and `FTL` and `webserver` log have neither `postrotate` nor `copytruncate`, due to they being `open()`ed and `closed()`ed per-line. https://github.com/pi-hole/FTL/pull/2960 changes this so all three files utilize a cached file descriptor. This means that systems installed with  v6.4.3 may have issues with logrotate.

I am still unsure on how to proceed and have thought of multiple options:
1. show a package-manager style diff dialog with the old and new file and have the user decide
2. utilize `sed` to insert the `postrotate` hook if it is missing
3. check whether the logrotate file is the unmodified v6.4.3 version and then replace it.

---

4. I have commented my proposed changes within the code and I have tested my changes.
5. I am willing to help maintain this change if there are issues with it later.
6. It is compatible with the [EUPL 1.2 license](https://opensource.org/licenses/EUPL-1.2)
7. I have squashed any insignificant commits. ([`git rebase`](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
8. I have checked that another pull request for this purpose does not exist.
9. I have considered, and confirmed that this submission will be valuable to others.
10. I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
11. I give this submission freely, and claim no ownership to its content.

---

### DIFF
--- a/advanced/Templates/logrotate
+++ b/advanced/Templates/logrotate
@@ -13,7 +13,6 @@
     endscript
 }
 
-# FTL.log and webserver.log are opened and closed for each line, therefore no postrotate is needed
 /var/log/pihole/FTL.log {
     # su #
     weekly
@@ -24,6 +23,9 @@
     notifempty
     nomail
     nodateext
+    postrotate
+        kill -USR2 $(cat /run/pihole-FTL.pid 2>/dev/null) 2>/dev/null || true
+    endscript
 }
 
 /var/log/pihole/webserver.log {
@@ -36,4 +38,7 @@
     notifempty
     nomail
     nodateext
+    postrotate
+        kill -USR2 $(cat /run/pihole-FTL.pid 2>/dev/null) 2>/dev/null || true
+    endscript
 }
--- a/advanced/Templates/logrotate
+++ b/advanced/Templates/logrotate
@@ -13,21 +13,7 @@
     endscript
 }
 
-/var/log/pihole/FTL.log {
-    # su #
-    weekly
-    create 640 pihole pihole
-    rotate 3
-    compress
-    delaycompress
-    notifempty
-    nomail
-    nodateext
-    postrotate
-        kill -USR2 $(cat /run/pihole-FTL.pid 2>/dev/null) 2>/dev/null || true
-    endscript
-}
-
+/var/log/pihole/FTL.log
 /var/log/pihole/webserver.log {
     # su #
     weekly
@@ -38,6 +24,7 @@
     notifempty
     nomail
     nodateext
+    sharedscripts
     postrotate
         kill -USR2 $(cat /run/pihole-FTL.pid 2>/dev/null) 2>/dev/null || true
     endscript
--- a/automated install/basic-install.sh
+++ b/automated install/basic-install.sh
@@ -1614,6 +1614,9 @@ create_pihole_user() {
 installLogrotate() {
     local str="Installing latest logrotate script"
     local target=/etc/logrotate.d/pihole
+    local template="${PI_HOLE_LOCAL_REPO}/advanced/Templates/logrotate"
+    local rendered current rendered_643 current_643
+    local logusergroup
 
     # Remove remnants from older Pi-hole versions: the logrotate config used to
     # live in /etc/pihole/, and logrotate state was tracked in a Pi-hole-specific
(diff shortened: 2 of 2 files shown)


rdwebdesign:
Both logs have exactly the same rotation options (including the same frequency).

I'm not sure if we need to send the signal for both logs independently, since they will be rotated at the same time. Do we need to execute the command twice?

If we don't, we can group them using a single config block and add `sharedscripts` to execute the command only once:

```diff

-/var/log/pihole/FTL.log {
-    # su #
-    weekly
-    create 640 pihole pihole
-    rotate 3
-    compress
-    delaycompress
-    notifempty
-    nomail
-    nodateext
-    postrotate
-        kill -USR2 $(cat /run/pihole-FTL.pid 2>/dev/null) 2>/dev/null || true
-    endscript
-}

+/var/log/pihole/FTL.log
 /var/log/pihole/webserver.log {
     # su #
     weekly
     create 640 pihole pihole
     rotate 3
     compress
     delaycompress
     notifempty
     nomail
     nodateext
+    sharedscripts
     postrotate
         kill -USR2 $(cat /run/pihole-FTL.pid 2>/dev/null) 2>/dev/null || true
     endscript
 }
```

What do you think?

DL6ER:
Your option 3, and I would like all of it in this PR - the `USR2` change alone reaches nobody who already has the file.

Why it is needed now: migration was dropped on purpose in `33feea11` (May), because `/etc/logrotate.d/pihole` was brand new and nobody had a file there yet. Right then, wrong now.

It is also a wider group than your description assumes. The target moved to the new path in v6.4.3, so a v6.4.2 install has nothing there, the `else` branch runs on upgrade, and it gets the v6.4.3 template. Everyone who has updated to v6.4.3 has the file that breaks under https://github.com/pi-hole/FTL/pull/2960, not only fresh installs. Users still on v6.4.2 or older keep their old-path `copytruncate` file and are fine - and stay fine, because when they do land on the new path they will get whatever template is current then, which with this PR already carries the `postrotate`.

There is a blocker in front of option 3. The freshness notice from `ef91b80f` can never report "up to date": it diffs the installed file against the raw template, but the installed copy has had `# su #` rewritten to `su $(stat -c '%U %G' /var/log)` while the template still carries the placeholder, so the three `su` lines always differ. Every user is told their file is outdated on every run, fresh installs included - and it means "is this file unmodified?", the exact question option 3 rests on, currently cannot be answered. Render the template into a temp file with the local `su` line before comparing and both problems go away; the fresh-install path can then install that rendered file directly and drop the post-install `sed`.

For the non-matching case I would keep the "not updating" message but name the missing `postrotate`, so someone who has customized the file knows what to add rather than only being told it is stale.

Scope stays small: only the v6.4.3 rendering needs a reference to compare against, since that is the one shape that is both broken and ours.

The rest is good. The `sharedscripts` merge is right, and I confirmed the stanza parses - `logrotate -d` reports "Handling 2 logs" with both files under the weekly pattern. No sequencing constraint with the FTL side either: `USR2` to an FTL without #2960 reaches dnsmasq's handler and reopens `pihole.log`, which was not rotated, so it is a close and reopen of the same path. This can land whenever it is ready.

darkexplosiveqwx:
I originally read the `/etc/pihole/logrotate` -> `/etc/logrotate.d/pihole` move wrong. You're right -- there is no migration, I later also noticed that in https://discourse.pi-hole.net/t/dnsmasq-log-not-being-rotated/87357/3?u=darkexplosiveqwx


Good catch that ef91b80f5589fcead354414f38be3c15cb12f8c8's out-of-date check was wrong and always reported it as out-of-date.




The tests are failing for some other reason, I suspect something temporary. Can they be re-run or do I need to push a branch update?

DL6ER:
Nothing to do with your changes: `editorconfig-checker` 4.0.0 (Sep 3) renamed its release assets from `ec-*` to `editorconfig-checker-*`, so the pinned action v2.2.0 died with `The binary 'ec-linux-amd64*' not found` on every PR here. The action bump is merged now (#6695).

I closed and reopened this PR to get the checks retriggered - a plain re-run replays the workflow file of the original attempt, only a fresh event picks up the fixed one from `development`. Sorry for the notification noise. `smoke-tests` is green again and the distro tests are running, so no push needed on your side.

DL6ER (approved):
Logrotate template and the v6.4.3 migration check look right now. CI is green apart from the retired `debian_11` job.

PR #6678, opened 2026-08-03
Improve gravity error handling - take II

### LINKED ISSUE #6677: gravity.sh leaves behind tmp.*.phgpb temp files after updateGravity finishes
### Versions
- Pi-hole: v6.4.3
- Web: v6.6
- FTL: v6.7

### Platform
- OS and version: Debian GNU/Linux 13 (trixie)
- Platform: single-board RISC-V computer (bare metal, not Docker), kernel 6.6.63

### Expected behavior

Running `pihole updateGravity` should not leave any temp files behind once
it finishes.

### Actual behavior / bug

`gravity.sh` leaves behind zero-byte temp files named
`tmp.XXXXXXXXXX.phgpb` in `$GRAVITY_TMPDIR` (resolves to `/tmp` on this
system) after a gravity update completes. A single run left about 91 of
these files behind. Since gravity updates run on a schedule (weekly here),
the files keep accumulating run after run unless something else cleans
them up.

The temp file is created around line 623 of gravity.sh:
listCurlBuffer=$(mktemp -p "${GRAVITY_TMPDIR}" --suffix=".phgpb")

It looks like this file (and possibly others created the same way earlier
in the script, e.g. around line 191 with a ".gravity" suffix) is not
removed on every code path, so a run that doesn't hit the specific cleanup
branch leaves its temp file behind.

### Steps to reproduce

1. Run `pihole updateGravity` (or let the scheduled gravity update run).
2. After it finishes, check `$GRAVITY_TMPDIR` (or /tmp if that's where it
   resolves on your system) for leftover `tmp.*.phgpb` files.
3. Files remain despite the run completing normally.

## Debug Token

- URL: https://tricorder.pi-hole.net/Fe8SidaC/

### Screenshots

Not applicable.

### Additional context

Workaround in place on my system: added an `ExecStopPost=` line to the
systemd unit that runs `pihole updateGravity`, which deletes any leftover
`tmp.*.phgpb` files in /tmp after every run, regardless of success or
failure. This is a workaround outside gravity.sh itself, not a fix to the
script. Suggested real fix: a trap (or explicit cleanup at the end of the
relevant function) that removes these temp files on every exit path,
instead of relying on cleanup only in the success path.

Not running Pi-hole in a container, so no compose file applies.

AI disclosure: this report was drafted with the help of Claude (Anthropic),
via Claude Code, which investigated the leak (traced it to gravity.sh:623),
built the local workaround, and drafted this report. I reviewed it before
submitting.

yubiuser:
Continuation of https://github.com/pi-hole/pi-hole/pull/6634

Improve gravity error handling. Fixes https://github.com/pi-hole/pi-hole/issues/6633

0) Guard `update_gravity_timestamp` and `gravity_build_tree gravity/antigravity` by checking their return code and exit in case of an error (This is what fixes  https://github.com/pi-hole/pi-hole/issues/6633)
1) Use `gravity_Cleanup error` instead of exit 1 to cleanup in case of an error
2) Format output nicer in case of an error
3) Remove the temporary gravity database in `gravity_Cleanup()` if it still exists
4) Fix some shellcheck warnings

---

---


yubiuser:
The last commit implements fails-fast as suggested in https://github.com/pi-hole/pi-hole/pull/6634#discussion_r3228188174. 

We now exit within nested functions early instead of propagating the error upwards to the main level. I did not check **every** function but went through by checking each `return 1`.
Two functions still handle their errors outside (`gravity_swap_databases` and `migrate_to_database`): they run a series of (shell) commands for which we don't check all their exit codes and which would be a massive rewrite.

rdwebdesign (approved):

PR #6688, opened 2026-08-27
Use parameter expansion and path canonicalization to safely parse local file URIs

mwoolweaver:

Refactor local file URI parsing logic in `gravity.sh` to be [RFC 8089](https://datatracker.ietf.org/doc/html/rfc8089) compliant

- Replace `$(echo "${url}" | cut -d'/' -f3-)` sub-process pipeline with native Bash parameter expansion (`${url#file://}` and `${file_path#file:}`) to safely handle both standard and minimal local file URIs as shown in the RFC [examples](https://datatracker.ietf.org/doc/html/rfc8089#appendix-B)

- Uses native string substitution (`${file_path//\/\.\//\/}`) to strip a redundant `/./` segment if present. (as seen in original `test/test_gravity.bats`)

- Updates the test suite (`test/test_gravity.bats`) to still give "file does not exist" error test. 

---

---

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -727,8 +727,11 @@ gravity_DownloadBlocklistFromUrl() {
   # (e.g., in Docker or when invoked by a non-root user). The target must
   # resolve to a regular file and be readable by the evaluated user.
   if [[ "${url}" == "file:/"* ]]; then
-    # Get the file path
-    file_path=$(echo "${url}" | cut -d'/' -f3-)
+    # Safely strip both standard (file:///) and minimal (file:/) local file URI schemes
+    file_path="${url#file://}"
+    file_path="${file_path#file:}"
+    # Canonicalize path by stripping any redundant '/./' using native Bash substitution
+    file_path="${file_path//\/\.\//\/}"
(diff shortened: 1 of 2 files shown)


yubiuser:
Please remove the commit changing the `shadow` test. This is a specific test for a security vulnerability that was found some time ago. The file `/etc/shadow` exists on all linux systems and with the path `file:/./etc/shadow` it was possible to read its content. This is not a test simply for a non-existing test, but really specific to prevent regression.

mwoolweaver:
Thank you for the context regarding the `/etc/shadow`check.

I've reverted the test back to `file:/./etc/shadow` and aligned the assertion to check for the proper output from Gravity (`Cannot read file (user 'pihole' lacks read permission)`).

The old `cut` implementation was stripping leading root slashes on minimal URIs like `file:/./etc/shadow`, accidentally turning it into a relative path (`etc/shadow`). With the new expansion and canonicalization logic, it correctly resolves back to the absolute path `/etc/shadow`, keeping the security regression test intact and passing cleanly.

Issue #6685, opened 2026-08-24
Blocklist URL parsing treats text between '@' and '/' as a DNS query


0xpsyduck:
### Versions
- Pi-hole:  v6.4.3
- Web:  v6.6
- FTL:  v6.7

### Platform
- OS and version: Alma 10
- Platform:  Docker

### Summary
When adding a blocklist URL that uses an '@' tag (for example: `https://cdn.jsdelivr.net/gh/hagezi/dns-blocklists@latest/adblock/pro.plus.txt`), Pi‑hole issues a DNS query for the token between '@' and the next '/' (e.g. "latest") instead of treating the entire string as the URL to download. 

### Steps to reproduce
1. Add the URL as a blocklist in WebUI: `https://cdn.jsdelivr.net/gh/hagezi/dns-blocklists@latest/adblock/pro.plus.txt`.
2. Run Gravity update.
3. Monitor live queries, query named `latest`(or any other token between @ and /) would be visible.

yubiuser:
I added the domain and it successfully downloaded the file

```
  [i] Target: https://cdn.jsdelivr.net/gh/hagezi/dns-blocklists@latest/adblock/pro.plus.txt
  [✓] Status: Retrieval successful
  [✓] Parsed 0 exact domains and 248112 ABP-style domains (blocking, ignored 0 non-domain entries)
```

0xpsyduck:
Issue is not successful download of blocklist. It is about url parsing treating token b/w `@` and `/` as a different domain.
Notice in your query log, you would have a query named `latest`.

rdwebdesign:
This issue happens when Pi-hole machine is using Pi-hole as DNS server.

It is caused by a regex that "filters too much". 
The regex was intended to remove only credentials from an URL (`http://`**`user:password@`**`domain.com/list.txt`), but when a "@" is found, like in your URL, the command removes everything up to the "@" and considers the next part the domain.

I already suggested a fix.

PR #6689, opened 2026-08-27
Fix URL parsing in gravity.sh

rdwebdesign:

Fix awk command to return the domain even when there is a "@" character in the middle of the URL: "http://domain.com/path@version/file.txt"

The new command first removes the protocol, then removes the "@" character, but only when it appears in the first part, before the domain: "http://username:password@domain.com/file.txt"

The old command removed everything (including the actual domain) before the "@" character, unconditionally.

Fix https://github.com/pi-hole/pi-hole/issues/6685

Changing the `awk` command to use 2 separate `gsub` calls. One for the protocol and another one to filter credentials. The new command uses a different regex, that removes text only when the "@" character appears before the domain.

---

---

### DIFF
--- a/gravity.sh
+++ b/gravity.sh
@@ -504,8 +504,13 @@ gravity_DownloadBlocklists() {
   mapfile -t sourceDomains <<<"$(
     # Logic: Split by folder/port
     awk -F '[/:]' '{
-      # Remove URL protocol & optional username:password@
-      gsub(/(.*:\/\/|.*:.*@)/, "", $0)
+      # Remove URL protocol
+      gsub(/^.*:\/\//, "", $0)
+
+      # Remove optional credentials "username:password@", but keep other "@" characters,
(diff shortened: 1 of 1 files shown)


rdwebdesign:
**Notes:**

The bad URL parsing is usually invisible and doesn't cause issues for most users, since the lists are correctly downloaded.

The bad URL parsing is visible only when Pi-hole is configured as DNS server for Pi-hole machine itself.
Then, the broken command returns the wrong string as a domain and a DNS query for this domain is visible on the Query Log, like reported in issue 6685.
In this case, this is just a cosmetic issue. Nothing is broken and the lists are still correctly downloaded.

The only case that this bug causes real issues is when the user configured Pi-hole as it own DNS server AND there is something (lists, domains, regex) blocking the "invalid" domain. 
In this specific case, the list won't be downloaded.

